Issue #2, opened 2015-05-01
Can't make guile example work 


ngeiswei:
I'm trying to follow what is described in
```
atomspace/examples/guile/README.md
```
but I'm getting the following error
```
nilg@laptop:~/OpenCog/atomspace/examples/guile$ guile
GNU Guile 2.0.9
Copyright (C) 1995-2013 Free Software Foundation, Inc.

Guile comes with ABSOLUTELY NO WARRANTY; for details type `,show w'.
This program is free software, and you are welcome to redistribute it
under certain conditions; type `,show c' for details.

Enter `,help' for help.
scheme@(guile-user)> (add-to-load-path "/usr/local/share/opencog/scm")
scheme@(guile-user)> (add-to-load-path ".")
scheme@(guile-user)> (use-modules (ice-9 readline))
scheme@(guile-user)> (activate-readline)
scheme@(guile-user)> (use-modules (opencog))
;;; note: auto-compilation is enabled, set GUILE_AUTO_COMPILE=0
;;;       or pass the --no-auto-compile argument to disable.
;;; compiling /usr/local/share/opencog/scm/opencog.scm
;;; opencog.scm:22:0: warning: possibly unbound variable `cog-set-atomspace!'
;;; opencog.scm:22:20: warning: possibly unbound variable `cog-new-atomspace'
;;; compiled /home/nilg/.cache/guile/ccache/2.0-LE-8-2.0/usr/local/share/opencog/scm/opencog.scm.go
While compiling expression:
ERROR: In procedure dynamic-link: file: "libsmob", message: "file not found"
```

linas:
The "file not found" message is misleading. it's failing because of this:

ldd -d libsmob.so
```
libPythonEval.so => not found
```

so why does scheme depend on PythonEval?

linas:
try again, there was a constellation of bugs, mostly install and python init issues, that I fixed over the last week.

ngeiswei:
It is fixed.

Radivarig:
I've built atomspace with nix package manager and when trying to follow the `basic.scm` example am getting `In procedure dynamic-link: file: "libsmob", message: "file not found"` even tho the folder containing `libsmob.so` was set `export LTDL_LIBRARY_PATH="$atomspace_build/opencog/guile"`. Any ideas how to debug this?

linas:
[...]

PR #3, opened 2015-05-01
Remove pln r2l from ure

ngeiswei:
Disentangle PLN and R2L from the unified rule engine.

Perhaps complete but not sure.

I've created a simple example of a crisp deduction system under
```
examples/rule-engine
```
relying on the rule engine. However I was unable to run it from guile due to issue #2 . It does run in the ForwardChainerUTest however.


ngeiswei:
Obviously not complete as the ForwardChainInputMatchCB inherit PLNImplicator. But for I'd need to disentangle PLN from atomspace/query which is too much work for now.

ngeiswei:
Maybe this can be fixed when json is replaced by scheme...

PR #7, opened 2015-05-05
Added new Normalization method according to issue #1512

misgeatgit:
This partly addresses https://github.com/opencog/opencog/issues/1512.I will make another PR from the opencog side once this one is accepted.


linas:
I merged this. But the formula looks completely wrong, to me. Look at what it does: if the AV is greater thatn the boundary, it returns a number between 0.0 and 1.0  But is the AV is small, less than the boundary, it returns a number between 1.0 and infinity.  This is bizarre, I think

PR #12, opened 2015-05-06
Fuzzy matcher

leungmanhin:
This new approach performs multiple searches by using different nodes in the pattern as starting points, captures partial / complete matches on the way and returns the most similar ones based on the number of nodes they have in common & their sizes. This gives more reasonable results than the old one.


linas:
I did not review this code; I was hoping someone else would ...

PR #13, opened 2015-05-06
BackwardChainer: unit test update

williampma:
Added a simple backward chain test.


tests/rule-engine/bc-modus-ponens.scm
```diff
@@ -0,0 +1,56 @@
...
+                    (VariableNode "$B"))
+                (VariableNode "$A"))
+            (ExecutionOutputLink
+                (GroundedSchemaNode "scm: pln-formula-simple-modus-ponens")
+                (ListLink
+                    (VariableNode "$B")
+                    (ImplicationLink
+                        (VariableNode "$A")
+                        (VariableNode "$B")))))))
+
```
ngeiswei: @williampma just in case you didn't know, I put some example of inference system using the URE under

examples/rule-engine

it is crisp, but it could have been re-used instead of redefining rules here.

No problem if you're aware of that and you prefer to redefine test rules here, just wanted to be sure you know about this examples directory.

williampma: Yeah, I was trying to make the tests more self a contained unit.  I didn't do it uniformly enough yet though as other places still refer to the examples folder.

linas: Unit tests should be independent of the examples directory.  If someone decides to "improve" the examples, the unit tests should not start failing.  So William is doing the right thing.

PR #18, opened 2015-05-08
Backward Chainer: implement random target selection

williampma:
Get some sort of target selection going.

Even with this, the criminal example is still not solvable, because there's no backward chaining path that will generate the necessary

```
InheritanceLink
    missile@123
    weapon
```
since
```
InheritanceLink
    $y
    weapon
```
always get immediately satisfied by
```
InheritanceLink
    missile
    weapon
```

To solve this, might need one of the following
 1.  Change the backward chainer so that it doesn't start off with trying to solve the variable, but instead backward chain with a rule immediately... a.k.a changing the "stopping criteria" that determines when a target is done and no longer need to be added back to the potential targets list.
 2.  the mixed forward/backward chainer that will allow the forward chainer to randomly forward chain on the atom.

Anyway...

## TODO ##
* Handle "evaluatable terms" where an "input" of a rule is not an actual input atom, but just a True/False of some EqualLink and stuff...  like the new deduction rule!!!!
 
 Probably need to implement the `evaluate_sentence` callback?  Not sure.  My guess is that `grounding` will not get called when all the inputs are "evaluatable terms"??

### DIFF
--- a/opencog/rule-engine/backwardchainer/BackwardChainer.cc
+++ b/opencog/rule-engine/backwardchainer/BackwardChainer.cc
@@ -60,8 +60,6 @@ void BackwardChainer::set_target(Handle init_target)
 
 	_inference_history.clear();
 
-	_targets_stack = std::stack<Handle>();
-	_targets_stack.push(_init_target);
 	_targets_set = UnorderedHandleSet();
 	_targets_set.insert(_init_target);
 }
@@ -73,10 +71,13 @@ void BackwardChainer::do_full_chain(uint max_steps)
 {
 	uint i = 0;
 
-	while (not _targets_stack.empty())
+	while (not _targets_set.empty())
 	{
 		if (max_steps != 0 && i >= max_steps)
+		{
+			logger().debug("[BackwardChainer] Stopping from reaching max steps");
 			break;
+		}
 
 		do_step();
 		i++;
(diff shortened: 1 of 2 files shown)


linas:
On Thu, May 7, 2015 at 10:49 PM, William Ma <notifications@github.com>
wrote:

> TODO
>
>    -
>
>    Handle "evaluatable terms" where an "input" of a rule is not an actual
>    input atom, but just a True/False of some EqualLink and stuff... like the
>    new deduction rule!!!!
>
>    Probably need to implement the evaluate_sentence callback? Not sure.
>    My guess is that grounding will not get called when all the inputs are
>    "evaluatable terms"??
>
> Not sure; I think it should be. When I did the 'evaluatable terms', I
realized that i was using crisp-logic true/false to implement AndLink,
OrLink NotLink when these get combined with AbsentLink, GreaterThanLink,
EqualLink.

I realized that perhaps one could use other formulas besides T/F so I
turned it into a callback, but then things got confusing when I tried to
work through practical applications.

So, at the moment, its almost a 'solution looking for a problem': where do
you need to combine with AndLink, orLink, NotLink, and use math other than
crisp true/false?   Clearly bayesian conditional probabilities fit the
bill, but I couldn't easily create a compelling demo that had to have
Bayesian math to work correctly.

Of course, PLN could use this too, maybe!?  Just not clear. how, yet.

--linas

Issue #20, opened 2015-05-08
use __path__ to locate additional python locations!


linas:
add the` /usr/local/share/opencog/python` system directory to the `__path__` in the `__init__.py`'s of the "opencog" build packages.

This page looks helpful:
http://stackoverflow.com/questions/2699287/what-is-path-useful-for

I'm guessing that the source/build dir should use `__path__`  to say "hey if you did not find it here, then look at the install location, but let the source/build always have priority over install" 

By contrast, the installed  `__init__` should NOT use ` __path__` since the install location should be "unique" from its own point of view ...

linas:
closing for now, but this needs additional testing.

PR #23, opened 2015-05-08
__path__ code to address "opencog" python module now being in multiple directories

eddiemonroe:
This is a parallel pull request with opencog/opencog#1606
It addresses #20

### DIFF
--- a/opencog/cython/opencog/__init__.py
+++ b/opencog/cython/opencog/__init__.py
@@ -1,0 +1,2 @@
+from pkgutil import extend_path
+__path__ = extend_path(__path__, __name__)


linas:
OK, so ... this `__init__.py` also ends up getting installed in /usr/local/share/opencog/python/opencog/__init__.py .. is that correct? is that the intended behavior?

linas:
And, as I think about it it, isn't this needed in the the opencog repo, not here,  so that the cogserver and agent_finder, etc. can find the rest of the modules they depend on?

eddiemonroe:
Yeah, it's good that the __init__.py ends up getting installed in the install directory--I think it needs to be there (for the case where the install directory comes before the build directories in PYTHONPATH and sys.path). 

And yes, the same __init__.py is needed in the matching directory in opencog, which is done in opencog/opencog#1606. But it is also needed here for the case where the atomspace directories come before the opencog directories in PYTHONPATH and sys.path.   ...assuming I'm understanding things correctly, but things seem to be working now.

PR #24, opened 2015-05-09
rename to be compatible with case-insensitive file systems

eddiemonroe:
Can we change the name of atomspace.h (note all lowercase name) to atomspace_install.h?

Because there is also file AtomSpace.h (note camel case name), this is causing problems on OS X filesystems, which default to being case-insensitive (hmmm.., that surprised me) IOW, both AtomSpace.h and atomspace.h cannot exist on case-insensitive file systems.


linas:
well, in that case, we also nee to change all the users of this file.  I'lll do that now. 

Are you sure OSX didn't fix this bug? Some new update? They talked a lot about this a decade ago, it was considered to be a fatal design flaw ....

linas:
OK, I just pushed an alternative fix its called version.h now

linas:
The other half of the fix was to change lib/FindAtomSpace.cmake to look for the renamed header file.

PR #35, opened 2015-05-14
Fix ure example part i

ngeiswei:
The rule-engine example isn't fixed yet (I'm still having an error, ERROR: In procedure module-lookup: Unbound variable: cog-stv-strength, but it's kind of a pain to debug given guile messing completely up the exception message [I had to hack the C++ code to print that one in order to see the message]).

I'm merging already anyway as we're a bunch working on the URE/PLN as to not drift too much from each others.

Note that now cog-fc takes the JSON file in second argument (beside the source). I know we're gonna get rid of JSON soon but till then that's just makes more sense.

### DIFF
--- a/opencog/rule-engine/InferenceSCM.cc
+++ b/opencog/rule-engine/InferenceSCM.cc
@@ -71,7 +71,8 @@ void InferenceSCM::init_in_module(void* data)
 void InferenceSCM::init(void)
 {
 #ifdef HAVE_GUILE
-    //all commands for invoking the rule engine from scm shell should be declared here
+    // All commands for invoking the rule engine from scm shell should
+    // be declared here
     define_scheme_primitive("cog-fc", &InferenceSCM::do_forward_chaining,
                             this, "rule-engine");
     define_scheme_primitive("cog-bc", &InferenceSCM::do_backward_chaining,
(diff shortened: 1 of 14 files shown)


opencog/rule-engine/InferenceSCM.cc
```diff
@@ -79,16 +80,17 @@ void InferenceSCM::init(void)
 #endif
 }
 
-Handle InferenceSCM::do_forward_chaining(Handle h)
+Handle InferenceSCM::do_forward_chaining(Handle h, const string& conf_path)
```
linas: passing around config paths seems like a terrible idea, to me..

ngeiswei: My feeling is that after we've replaced JSON by Scheme, we can just load the configuration in the AtomSpace, before running cog-fc, and then we can remove the config path from cog-fc. Till then though, in order to get both cog-fc work on different rule systems (PLN, R2L, etc), both the atomspace repo and opencog repo. I don't see another solution.

Or maybe setting the config path in the atomspace conf file, but I don't think there is one on the atomspace repo.

ngeiswei: BTW, it's a just a matter of days before I get rid of JSON (say Monday if I'm lucky).

PR #44, opened 2015-05-18
Fchain

misgeatgit:
The following changes are based on the recent discussion made on how source selection should work. Whether premises should be added to potential source list is not yet decided and will be addressed in upcoming changes.


williampma:
Is the `.clang` on purpose?

misgeatgit:
Ah, its unintentional. I will remove it.

misgeatgit:
Removed.

PR #50, opened 2015-05-20
general reduction in Fold

linas:
Implement reduction for a general fold operator.

This should have been easy, but was annoyingly hard. Doing computer algebra in the atomspace is not going to be easy nor efficient, it seems.  Anyway, this now works as a provisional experiment, but needs re-thinking and redesign.


ngeiswei:
Linas, I'm wondering if it wouldn't save us a lot of work if this could be coded using the pattern matcher and the unified rule engine. I'm not exactly sure how, in particular there would need a way to replace outgoings without destroying a link.

linas:
Yes, I guess so. This was another experiment that didn't go so well.

PR #62, opened 2015-05-27
Fix include from BindLink.h to BindLinkAPI.h

kungfooman:
Fixing this error: http://i.gyazo.com/c65c3fd6d34a874037d59bffab4d8e80.png

     fatal error: opencog/query/BindLink.h: No such file or directory


ngeiswei:
Thanks.

PR #63, opened 2015-05-27
Cython: add __init__.py to load atomspace.pxd from external

kim135797531:
This commit attempt to fix issue https://github.com/opencog/opencog/issues/1614, some part of https://github.com/opencog/opencog/issues/1625, and the problem encountered to YiShan(https://groups.google.com/forum/#!topic/opencog/IxQQJajw2gc).

This fix makes `__init__.py` in ${ATOMSPACE_INCLUDE_DIR}/opencog/cython/opencog/atomspace.pxd during build project.

It would be needed by external project. (e.g. agent_finder module, cogserver module, etc.)

### DIFF
--- a/opencog/cython/opencog/CMakeLists.txt
+++ b/opencog/cython/opencog/CMakeLists.txt
@@ -93,8 +93,9 @@ SET_TARGET_PROPERTIES(atomspace_cython PROPERTIES
 # The cogserver needs the atomspace pxd file to get the Handle, Atom and
 # AtomSpace defintions.
 INSTALL (FILES
+	__init__.py
 	atomspace.pxd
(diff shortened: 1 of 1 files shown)


linas:
Are you sure this is the correct location?  There already is an __init__.py installed in /usr/local/share/opencog/python/opencog/__init__.py  and it is installed at the very bottom of this same file, at line 228

For what I can tell, it is the same __init__.py that you are installing...  I think it is strange that it should be in two different places.

kim135797531:
Refer to mailing list(https://groups.google.com/forum/#!topic/opencog/IxQQJajw2gc), I think both `__init__.py` are necessary.

PR #64, opened 2015-05-28
Fchain

misgeatgit:


williampma:
Why is there both "ForwardChainPatternMatchCB.cc" and "ForwardChainerPMCB.cc"?

misgeatgit:
ForwardChainPatternMatchCB.cc shouldn't be there. I have made another PR that removes it. Thanks.

PR #68, opened 2015-05-29
Fully functional GetLink

linas:
Again, I am ready to declare that I am "code complete" and have no more large changes planned.

The demo to look at here is `/examples/guile/get-put.scm`  which now can do properly nested GetLink and PutLink's ad infinitum.

This means that we could (if we wished) to eliminate BindLink and completely replace it by a Get/Put pair.  This is not a theoretical statement, it should now be "practical" (-ish) Hmm. I guess I should add that to the demo.  I have no plans to do this any time soon; the stuff that is here is sufficient for my other needs. Ergo, I claim (I hope, I pray) I am done.

As Amen has observed, the code is lightly tested, and may prove to be buggy in more complex cases. We shall find out.

### DIFF
--- a/examples/guile/get-put.scm
+++ b/examples/guile/get-put.scm
@@ -54,7 +54,8 @@
 
 ; The cog-satisfying-set function will return the value(s) that
 ; the GetLink finds.  If only one value satsifies the query, then
-; that is returned. Else a SetLink is returned.
+; that is returned. Else a SetLink is returned. Equivalently,
+; the cog-execute! function will do the same thing.
 (cog-satisfying-set get-value)
 (cog-execute! get-value)
 
@@ -79,18 +80,18 @@
 (cog-execute! remove-thing-ab)
 
 ; Look for it.
-(cog-satisfying-set get-value)
+(cog-execute! get-value)
 
 ; Add it back in:
 (cog-execute! to-be-added)
-(cog-satisfying-set get-value)
+(cog-execute! get-value)
 
 ; ... and so on. We can now continue to remove it and add it
 ; back in repeatedly.
 (cog-execute! remove-thing-ab)
-(cog-satisfying-set get-value)
+(cog-execute! get-value)
 (cog-execute! to-be-added)
-(cog-satisfying-set get-value)
+(cog-execute! get-value)
 
 
 ; It is also useful to generically remove any atom matching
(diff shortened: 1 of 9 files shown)


williampma:
Just curious, as I can't find this information anywhere.

How would PutLink works for thing like
```
(PutLink
    (CrazyLink
         (VariableNode "$A")
         (InsaneLink
                 (VariableNode "$B")
                 (VariableNode "$A")
                 (MadnessLink
                        (VariableNode "$A")
                 )
                 (VariableNode "$B")
         )
         (VariableNode "$A")
     )
     (ListLink
          what order of node to specific to replace $A and $B
     )
```

I assume the ordering would replace $A first, then $B?  But what if one of those link is unordered?

ngeiswei:
@williampma Excellent question, this is yet another good reason never to use implicit binding (in spite of its admittedly cool terseness).

ngeiswei:
@linas Yeah I thought of removing BindLink entirely (before starting removing the ImplicationLinks), but BindLink still saves one link as compared to using Get/PutLink, and given how ubiquitous it is, I think this is worth it (especially after removing the ImplicationLink from BindLink, I'm nearly done with that)...

linas:
Multi-part answer:

1) First, let me assume that CrazyLink, InsaneLink are ordered links, and assume that they do NOT bind variables.  Then PutLink will bind them, in the order of occurance, top to bottom, so $A comes first.

2) PutLink  should accept optional variable declarations. I guess I did  not implement this. 

3) If CrazyLink binds the variable $A, then it is not a free variable, and PutLink should not see it.  This is also not implemented, i.e. its a bug.  The way this works is that CrazyLink would be a derived class of ScopeLink (or LambdaLink) and then PutLink could look at the ScopeLink and see that $A is not free.

So 2) and 3) are bugs.  Should probably open bug reports/feature requests for these.  I have no plans to fix this soon, unless they are needed by someone for something.   Yes it would be great if these worked, but I'm running out of time.  Care to fix this?  

Warning: fixing this is both "easy" and "hard". So for example: CrazyLink should inherit from ScopeLink. Well, in that case, BindLink should inherit from ScopeLink.  But right now it doesn't. Why not? Partly historical accident, partly because something was messy and difficult and confusing.  So, first you have to fix this.  Half way through fixing this, you will realize that maybe its a bad idea, or maybe there is a conflict, or maybe there is a better way, or that you are getting sleepy. Or worse: you are almost done when you realize it was all a big mistake. So its conceptually "easy" but the conflicting details and confusions can be deadly.

Issue #54, opened 2015-05-21
Remove ImplicationLink from PM BindLink


ngeiswei:
# Overview
I think we can make BindLink take 3 arguments (or 2 when no variable restrictions) and remove the ImplicationLink without hurting anything.

# Specification
Instead of
```
BindLink
   <Variables> (optional)
   ImplicationLink
      <Clauses>
      <Rewrite>
```
we would have
```
BindLink
   <Variables> (optional)
   <Clauses>
   <Rewrite>
```

# Example
For example, instead of
```
BindLink
   VariableNode X
   ImplicationLink
      InheritanceLink
         VariableNode X
         ConceptNode "Earth"
      VariableNode X
```
we would have
```
BindLink
   VariableNode X
   InheritanceLink
      VariableNode X
      ConceptNode "Earth"
   VariableNode X
```
(which would return a SetLink of all things that inherit "earth").

It is simpler, why not go for it?

# Motivation
1. The PM is such a fundamental part of the AtomSpace, it has to be as perfect as possible, people are gonna use this zillions of times, we cannot afford any cruft on it.
2. No more interferences with PLN rules.

# Yours
If you guys approve I'm happy to take care of it.

linas:
I guess I'm OK w/ it. Maybe @AmeBel or @williampma might see something.

Do you plan to support the old form, for backwards-compat, or do you  plan tear up all the unit tests, and all the other scattered code that uses it, to remove it?

AmeBel:
I see no problem

ngeiswei:
OK, I plan to remove the old form, as this will help me to understand whether it was a good idea to begin with.

ngeiswei:
And also cause it's better to have one way of doing things.

linas:
Some history: back in the day, the original pattern matcher was written and working before the BindLink was invented.  One day, I realized that the bindLink was possible, and would be cool to have, but could not explain it clearly.  Only after putting the ImplicationLink in there did the general concept seem to gel into something that seemed acceptable and understandable by everyone on the mailing list (i.e. understood by Ben, who, after understanding it, lost all objections, but also lost all interest. As Ben is wont to do).

I think that the original idea was that someday, in the distant future, the ImplicationLink would be replaced by more PLN rules, but as you see, that idea didn't really work out. 

Err.. well, in some weird way, what you are proposing is the "generalization" we were going to do in the future ... I guess we didn't understand what that generalization would be.

williampma:
[...]

PR #70, opened 2015-05-29
Remove implication link from bindlink

ngeiswei:
1. Addresses issue #54.
2. 258 lines of code taken away as a result.
3. QuoteUTest fails, I can't understand why, I've decided to merge now anyway, I'll keep investigating, any idea is welcome.


ngeiswei:
Hehe, I think I know why QuoteLink is "failing", that's cause it's not! Since the ImplicationLinks are missing it just generates less junk... I just have to confirm that...

ngeiswei:
Yep, I confirm that the missing outputs are the ImplicationLinks.

linas:
wow!

linas:
Should we find a better name for BindLink  -- maybe RewriteLink ?

PR #72, opened 2015-05-29
Start working on Haskell Bindings.

MarcosPividori:


MarcosPividori:
Hi, I would like, before merge, to know if you agree on this specific commit: 4d28208ab6cd54f6386e15114f7118c9f4fd854f
It adds automatically installation of opencog-atomspace haskell library. Maybe I should remove these lines and make it optional.

examples/haskell/example.hs
```diff
@@ -0,0 +1,16 @@
+
+import OpenCog.AtomSpace.Api
+import OpenCog.AtomSpace.Types
+import Control.Monad.IO.Class
+import Data.Default
+
+main :: IO ()
+main = runOnNewAtomSpace program
+
+program :: AtomSpace ()
```
linas: It is rare, but there are times when multiple atomspaces need to be used at the same time.

opencog/haskell/LICENSE
```diff
@@ -0,0 +1,693 @@
+Unless indicated otherwise (a license reference in file headers, or a 
+LICENSE file in a directory), all OpenCog source files use the AGPL 
+with additional permissions to link OpenCog libraries with non-AGPL
+works; the text of the OpenCog linking excpetion is located at the end
+of this file.
+
+ 
+
+                    GNU AFFERO GENERAL PUBLIC LICENSE
```
linas: Don't include a separate license file. The master license file in the main directory covers everything, already.

opencog/haskell/OpenCog/AtomSpace/Api.hs
```diff
@@ -0,0 +1,58 @@
...
+-- Main Data Type for representing programs working on an AtomSpace.
+-- We have to use the IO monad because of the use of FFI for calling c functions for working
+-- on a mutable instance of the atomspace, so we have side effects.
+type AtomSpace = ReaderT AtomSpaceRef IO
+
+-- Internal functions new and delete, to create and delete C++ instances of the AtomSpace class.
+foreign import ccall "AtomSpace_CWrapper.h AtomSpace_new"
+               c_atomspace_new :: IO AtomSpaceRef
+asNew :: IO AtomSpaceRef
+asNew = c_atomspace_new
```
linas: atomspaces can take other atomspaces as arguments: an atomspace is kind-of like an "environment": it holds everything that is "true" right now. So, if you create atomspace B from A, then everything in A is also in B.  whatever you put  into B will stay in B, and will be deleted when B is deleted, but A will still be there.

opencog/haskell/LICENSE
MarcosPividori: Hi. Ok, I will remove it.
I added it because I tried to develop a complete cabal package that needs to include a license file. But as we are not planning to upload this package anywhere for now (hackage,etc) , I can remove the LICENSE file.

opencog/haskell/OpenCog/AtomSpace/Types.hs
```diff
@@ -0,0 +1,54 @@
+
+module OpenCog.AtomSpace.Types(
+    NodeType (..)
+  , LinkType (..)
+  , TruthValue (..)
+  , Atom (..)
+  , Link (..)
+  , Node (..)
+  , Handle (..)
```
linas: Atoms and Handles should be exactly the same thing. There should be no difference between atoms and handles.

There is a difference in the C++ code, but this is due to a terrible historical accident, and it will someday be fixed.

I guess you could say that a handle is like a pointer to an atom, but it would be best if you had an API that does not use pointers. Neither the scheme nor the python API's have pointers in them: you just work directly with the atoms that you want to work with.

MarcosPividori: Yes! I agree with you. I plan to work on this.

linas:
I made numerous comments. They probably don't need to be fixed before merging.  I won't merge, though, I think this is Nil's job.

Also, if you do prepare fixes, just push them to this same branch, they will automatically become a part of this pull request.

ngeiswei:
I'll take care of merging it. After carefully reviewing it, which I'll do on Monday.

opencog/haskell/LICENSE
MarcosPividori: Done: d99027596b8bfdc69deda49e4cc7f7332c5f834e

MarcosPividori:
Hi thanks for the comments!
You were right about the private header file "AtomSpace_CWrapper.h". I removed it because it is not necessary. (e49960d4af223e1d21976a622cfed0f30e3b0a2f)
About the library name: "atomspace_wrapper" , I think this name is representative of the library functionality. It is a wrapper of the atomspace. If I name it "haskell_cwrapper", I think it will be confusing. It doesnt mention the atomspace...  Maybe I can use: "haskell-atomspace"

linas:
But the atomspace is already written in C; you don't need to wrap C code with C code to use it, that makes no sense. In practice, the only consumer/user of the wrapper functions will be Haskell, as both the scheme and the python bindings already have their own wrappers, and won't be using the ones you create.  

Here's an example of the scheme wrapper for getting the name of a node: https://github.com/opencog/atomspace/blob/master/opencog/guile/SchemeSmobAtom.cc#L51 notice there is a lot of error checking and conversion to/from internal scheme formats.   Here is the equivalent in python: https://github.com/opencog/atomspace/blob/master/opencog/cython/opencog/atom.pyx#L31  -- it is converted into C code by the cython compiler, and then compiled by gcc as normal C code.

MarcosPividori:
Hi Linas, thanks for your comments:
 - The atomspace is already written in C++, but not C. To call foreign functions in haskell these has to be written in C. So, because of that, I use a C wrapper, to offer C functions to be liked with the haskell library. (I am writing a wiki page about these project decisions on: http://wiki.opencog.org/w/Haskell_Bindings_-_GSoC_2015 ). But, I agree with you in that the only consumer of this library is the haskell library, and that it won't be only a wrapper because I am going to include some code for error checking, conversion, etc. If you think this is ok, I can name the library: haskell-atomspace. So in future, when adding Moses bindings we could add a wrapper library: haskell-moses, the same for cogserver integration: haskell-cogserver, etc.
 - Thanks for the Python and Scheme references. I have been reading that code before, but I am not sure if I am understanding properly how things work. So, Scheme bindings are interpreted by Guile inside CogServer. Python bindings can be used to write interpreted code on a CogServer shell and also to build independent executables. Am I right?

examples/haskell/example.hs
MarcosPividori: Hi. Yes, I plan to broaden the api to work with multiples atomspaces. In that case one would need to specify which atomspace wants to use for each action, so the atomspace instance wouldn't be hidden inside a monad.
Now:
   main = runOnNewAtomSpace $ do 
                                                            asAddNode  (Node ....)
                                                            asAddLink  (Link ....)

With multiples atomspaces:
  main = do
                   a1  <-  createAtomSpace
                   a2  <-  createAtomSpace
                   addNode a1 (Node ...)
                   addNode a2 (Node ...)
                   ....
I plan to offer both options. As it is more common to work with only one atomspace, the first approach is far more confortable because you avoid passing the atomspace instance as an argument each time for each action.
I will work on that and let you know!

opencog/haskell/OpenCog/AtomSpace/Api.hs
MarcosPividori: Hi, Ok. I have read the code of the AtomTable class, and I think I have a general idea of how environments work. I found this wiki page: http://wiki.opencog.org/w/Multiple_atom_spaces which explains about this. I will work on this https://github.com/MarcosPividori/atomspace/issues/10!

examples/haskell/example.hs
ngeiswei: @MarcosPividori Just to be sure I understand correctly, if createAtomSpace, addNode, etc binds to the C++ implementation, then those functions would still involve an IO monad in their types, right? Due to using FFI, even if the AtomSpace is passed in argument to addNode, etc, right?

opencog/haskell/AtomSpace_CWrapper.h
```diff
@@ -0,0 +1,20 @@
...
+/*
+    int AtomSpace_getSize(AtomSpace* this_ptr);
+    int AtomSpace_getNumNodes(AtomSpace* this_ptr);
+    int AtomSpace_getNumLinks(AtomSpace* this_ptr);
+    void AtomSpace_addAtom(AtomSpace* this_ptr, AtomPtr atom);
+    Handle AtomSpace_addLink(AtomSpace* this_ptr, Type t, const HandleSeq& outgoing);
+    bool AtomSpace_removeAtom(AtomSpace* this_ptr, Handle h);
+*/
+}
+
```
ngeiswei: I would assume that a C wrapper for the AtomSpace should be moved to its own directory, something like
```
opencog/atomspace/C-wrapper
```
any suggestion, guys?

Anyway for now it's OK to let this under opencog/haskell

opencog/haskell/OpenCog/AtomSpace/Api.hs
```diff
@@ -0,0 +1,55 @@
...
+
+import Foreign
+import Foreign.C.Types
+import Foreign.C.String
+import Control.Exception
+import OpenCog.AtomSpace.Types
+import Control.Monad.IO.Class
+import Control.Monad.Trans.Reader
+import Data.Functor
+
```
ngeiswei: @MarcosPividori I would recommend that you use restrictions on the imports, such that
```haskell
import MyMod (myFun1, myFun2, etc)
```
This minimizes the chances of having ambiguities, but even in the absence thereof, it helps guys like me who are not very familiar with the plethora of Haskell modules to see right away which function belongs to which module.

opencog/haskell/OpenCog/AtomSpace/Api.hs
```diff
@@ -0,0 +1,55 @@
...
+-- Internal AtomSpace reference to a mutable C++ instance of the AtomSpace class.
+newtype AtomSpaceRef = AtomSpaceRef (Ptr AtomSpaceRef)
+
+-- Main Data Type for representing programs working on an AtomSpace.
+-- We have to use the IO monad because of the use of FFI for calling c functions for working
```
ngeiswei: @MarcosPividori we have a coding standard defined on the wiki
http://wiki.opencog.org/w/Development_standards
try you're best to follow it, when it makes sense, for instance here the comment line is more than 80-col long.

opencog/haskell/OpenCog/AtomSpace/Api.hs
```diff
@@ -0,0 +1,55 @@
...
+                    let ntype = fromIntegral $ fromEnum $ nodeType nod
+                        fun = \str -> fromIntegral <$> c_atomspace_addnode asRef ntype str
+                     in liftIO $ withCString (nodeName nod) fun
+
+-- 'runOnNewAtomSpace' creates a new AtomSpace (C++ object), does some computation over it,
+-- and then deletes it.
+-- By using bracket, I ensure properly freeing memory in case of exceptions during the computation.
+runOnNewAtomSpace :: AtomSpace a -> IO a
+runOnNewAtomSpace as = bracket asNew asDelete $ runReaderT as
+
```
ngeiswei: @MarcosPividori this code (in particular the ReaderT part) goes beyond my Haskell skills. I may ask you to walk me through it in details during our next Hangout.

examples/haskell/example.hs
ngeiswei: Never mind, I clearly need to understand more what's going on. We can discuss that during our next Handgout.

opencog/haskell/AtomSpace_CWrapper.h
ngeiswei: Oh I've only just seen @linas and @MarcosPividori discussion about the atomspace C wrapper. I still feel this wrapper would be useful outside of Haskell. But anyway, I guess it doesn't need to be moved outside of the haskell directory just yet.

ngeiswei:
I'm far from understanding how it works but I'm merging.

opencog/haskell/OpenCog/AtomSpace/Api.hs
MarcosPividori: Hi Nil, Yes, I had read this standards before, but I forgot to care about column width! I fixed this in 835fdba7286b2739ee52ea48eb4eff0f3a99e2d5 . I will be more careful.

opencog/haskell/OpenCog/AtomSpace/Api.hs
MarcosPividori: Hi Nil. Ok! I have updated the code with more specific imports (f7a5ef441deb17082ff36f8b2ddce898b4a3ad8e). I hope it is clearer now.

examples/haskell/example.hs
MarcosPividori: @ngeiswei   Yes! They would involve the monad IO. For the example above:
```haskell
createAtomSpace :: IO AtomSpaceInstance
addNode :: AtomSpaceInstance -> Node -> IO Handle
```

linas:
On Sun, May 31, 2015 at 3:52 PM, MarcosPividori <notifications@github.com>
wrote:

> Hi Linas, thanks for your comments:
>
>    - The atomspace is already written in C++, but not C. To call foreign
>    functions in haskell these has to be written in C. So, because of that, I
>    use a C wrapper, to offer C functions to be liked with the haskell library.
>    (I am writing a wiki page about these project decisions on:
>    http://wiki.opencog.org/w/Haskell_Bindings_-_GSoC_2015 ). But, I agree
>    with you in that the only consumer of this library is the haskell library,
>    and that it won't be only a wrapper because I am going to include some code
>    for error checking, conversion, etc. If you think this is ok, I can name
>    the library: haskell-atomspace. So in future, when adding Moses bindings we
>    could add a wrapper library: haskell-moses, the same for cogserver
>    integration: haskell-cogserver, etc.
>
>
yes, that sounds good.


>
>    - Thanks for the Python and Scheme references. I have been reading
>    that code before, but I am not sure if I am understanding properly how
>    things work. So, Scheme bindings are interpreted by Guile inside CogServer.
>    Python bindings can be used to write interpreted code on a CogServer shell
>    and also to build independent executables. Am I right?
>
> Sort-of.   Both scheme and python can be used with and without the
cogserver.   For various technical reasons, we are very very slowly moving
away from the cogserver; or at least, re-thinking it.

Both scheme and python are compiled into a byte-code when possible, and
interpreted when that is not possible.  The byte-codes are very different,
though, and not compatible with each-other nor with e.g. the java JVM.
Guile runs considerably faster than python; guile is multi-threaded, python
is not.   In general, guile is both a lot smaller and a lot more
sophisticated in what it does and can do, as compared to python.

The JVM is not popular because it fails to implement tail recursion, and it
cannot do call-with-continuation. Clojure hacks around this by not
supporting tail recursion or call/cc. I believe Microsoft .net can do both,
and was specifically designed to be a better VM than JVM. Unfortunately, it
never became a standard due to the usual mircrosoft antics, so the world of
VMs remains fragmented.  (Yes, the JVM committee has a stick up their ass,
which is a shame.  Its been like 20 years now, and still nothing!?).
Whatever. In a highly fragmented world, VM's remain a topic of research.

I only mention this because I sometimes daydream of having opencog on a
VM.  Clearly, though, that would be a decade away or more.

To improve performance, the python bindings cache values in the python
code; the opencog guile code does not, but could. Whether it should or not
is not clear:  caching e.g. a Node name (which is immutable, static) in the
guile code, or in haskell local code leads to greater memory usage, which
can lead to a slow-down when traversing large graphs. It all depends on how
the graph is traversed...  Note that the outgoing set of an opencog Link is
immutable was well; the incoming set is not.

One reason that python is slower is that python uses the atom UUID's,
instead of using the atom pointers.  This is due to a design constraint in
cython: cython cannot manage pointer objects.

haskell would be faster if it could access atom pointers directly, instead
of attempting a UUID-to-pointer lookup every time it needs to touch an atom.

-- Linas

PR #79, opened 2015-06-04
Change satisfying_set to return singleton instead of element if one m…

ngeiswei:
…atch

I think it complicates the user's life to silently replace the
singleton by its element when there is only one match. It is also
inconsistent with bindlink behavior.

### DIFF
--- a/opencog/query/Satisfier.cc
+++ b/opencog/query/Satisfier.cc
@@ -99,11 +99,6 @@ Handle opencog::satisfying_set(AtomSpace* as, const Handle& hlink)
 	SatisfyingSet sater(as);
 	bl->satisfy(sater);
 
-	// If there is only one item in the satisfying set, return
-	// that one item. Otherwise, wrap multiple items in a SetLink.
-	if (1 == sater._satisfying_set.size())
-		return sater._satisfying_set[0];
-
 	return as->addLink(SET_LINK, sater._satisfying_set);
 }
 
--- a/tests/query/GetLinkUTest.cxxtest
+++ b/tests/query/GetLinkUTest.cxxtest
@@ -135,9 +135,6 @@ void GetLinkUTest::test_free_bound(void)
 
 	Handle hgnd = eval->eval_h("(cog-satisfying-set is-query)");
 
-	// XXX for better or for worse, satisfying-set skips using the
-	// Setlink if the Setlink only has one element in it.
-	// Handle hans = eval->eval_h("(SetLink (ConceptNode \"human\"))");
 	Handle hans = eval->eval_h("(SetLink (ConceptNode \"human\"))");
 
 	printf("Expected this: %s\n", hans->toString().c_str());
--- a/tests/query/GetLinkUTest.cxxtest
+++ b/tests/query/GetLinkUTest.cxxtest
@@ -138,7 +138,7 @@ void GetLinkUTest::test_free_bound(void)
 	// XXX for better or for worse, satisfying-set skips using the
 	// Setlink if the Setlink only has one element in it.
 	// Handle hans = eval->eval_h("(SetLink (ConceptNode \"human\"))");
-	Handle hans = eval->eval_h("(ConceptNode \"human\")");
+	Handle hans = eval->eval_h("(SetLink (ConceptNode \"human\"))");
 
 	printf("Expected this: %s\n", hans->toString().c_str());
 	printf("Found this answer: %s\n", hgnd->toString().c_str());


ngeiswei:
@linas could have a look at it and merge if you agree? Thanks.

linas:
per other comments, the decision to to this was thought out -- the plusses seemed to outweigh the minuses, at least for the GetLink usage.

How about this: what if GetLink did this, but SatisfactionSet did not? So SatisfactionSet always returned a Setlink, but GetLink would always return just one item if one variable was requested, and one grounding was found?

ngeiswei:
@linas so you mean that
```scheme
(cog-execute! (GetLink <pattern-with-one-match>))
```
would return the element, but
```scheme
(cog-statisfying-set (GetLink <pattern-with-one-match>))
```
would return a singleton, right?

Maybe... but I'm still unsure.

There could be another option, create a GetElementLink, so that
```scheme
(cog-execute! (GetElementLink <pattern>))
```
always return the element, or raise an exception otherwise. Or, instead of introducing a GetElementLink, add the function
```scheme
(cog-get-element (GetLink <pattern>))
```

What do you say?

ngeiswei:
Yeah, when I think about it, I think having the cog-get-element solution is right, because if the user *knows* it's gonna get a singleton, then surely it doesn't want that extra SetLink. But if it doesn't know then silently stripping the SetLink is gonna require the user to consider that case, and make its code a bit longer.

So it seems to me stripping the SetLink by using a dedicated cog-get-element function offers the best of both worlds.

linas:
well, actually, I was thinking of getting rid of  `cog-statisfying-set` and also `cog-bind` and just replacing them with a single generic `cog-execute!` that handles everything that returns atoms.  meanwhile, `cog-evaluate!` handles everything that reeturns TV's.

Its a hassle to have all these different cog-whatever routines: a lot of code to maintain, debug, test, and then you have to remember which is which, and when to use which one. Too complicated.

What I was saying is that thee is a link type called SatisfactionSetLink -- its just the first half of the BindLink, without the implicand. executing it should always yield s set.

Right now SatisfactionSetLink and GetLink are almost the same .. two different concepts that are actually the same or almost the same

AmeBel:
Why are the results wrapped in a SetLink? I mean, if they are not timestamped and associated with the BindLink or thier handles kept in some cache for near term invocation of the pattern matching, what uses do they have in the atomspace?

 Or  could it be for performance, passing a handle arround takes memory space than a list of handles?

ngeiswei:
@linas +1 for reducing the set of functions, -1 for returning stripping the SetLink when there is only one match, don't put this burden on the user, what if that one match is itself a SetLink, then likely the user's code is gonna misbehave.

@AmeBel good question. The fundamental reason for having a SetLink is to move the processing representation in the AtomSpace. But is that a good idea w.r.t. to the pattern matcher call? I don't know. In most (all?) the use I've witnessed so far the results are passed to the caller, and the SetLink could just as well be removed. And I think you are right, unless the results are timestamped and linked to the original call, there is not much use for keeping the result in the atomspace. Or rather I don't see how concurrently operative mind agent would be able to take advantage of that.

@linas @AmeBel so perhaps what this leads to is that we'd benefit of having 2 calls, one wrapping the results in a SetLink, one returning a set in the binding language of choice. Or just one call returning a set, and let the user wrap that in a SetLink if he needs to.

ngeiswei:
But anyway, I still think we should merge that pull request till we progress further ahead with that issue. The way I see it, is that if those SetLinks turn out to be too many in the future we can think of the proper way to deal with them. Meanwhile, having a system that behaves as consistently as possible seems more important to me. Do you guys agree?

linas:
On Fri, Jun 5, 2015 at 1:26 AM, ngeiswei <notifications@github.com> wrote:

> @linas <https://github.com/linas> +1 for reducing the set of functions,
> -1 for returning stripping the SetLink when there is only one match, don't
> put this burden on the user, what if that one match is itself a SetLink,
> then likely the user's code is gonna misbehave.
>

It was added to remove a burden for the user, now you say it adds burden.
Which is it?   The goal of change was to make things easier, not harder.


> @AmeBel <https://github.com/AmeBel> good question. The fundamental reason
> for having a SetLink is to move the processing representation in the
> AtomSpace. But is that a good idea w.r.t. to the pattern matcher call? I
> don't know. In most (all?) the use I've witnessed so far the results are
> passed to the caller, and the SetLink could just as well be removed. And I
> think you are right, unless the results are timestamped and linked to the
> original call, there is not much use for keeping the result in the
> atomspace. Or rather I don't see how concurrently operative mind agent
> would be able to take advantage of that.
>

Huh? You can time-stamp anything you want, yourself.

I picked SetLink instead of ListLink because there is no guaranteed order.
Its not an ordered list, its an unordered collection.

If there is more than one result, it is impossible to "remove" the
SetLink.  That doesn't even make sense.


> @linas <https://github.com/linas> @AmeBel <https://github.com/AmeBel> so
> perhaps what this leads to is that we'd benefit of having 2 calls, one
> wrapping the results in a SetLink, one returning a set in the binding
> language of choice. Or just one call returning a set, and let the user wrap
> that in a SetLink if he needs to.
>
 I don't understand what this is trying to say. Sure, results could be
wrapped in a SetLink or a ListLink or SomeOtherLink .. there is a whole zoo
Links to choose from.  The point is ???

--linas

ngeiswei:
@linas of course it adds a burden. Actually even when you expect a single element you're never 100% sure you're gonna get just one, so you still need to add the code to deal with that exception. I'll get back with some code to show my point if that's necessary...

linas:
There's no "of course" about it. I added this to make things simpler, not to make things harder. In my opinion, it makes things a lot simpler. I'm not even sure what the alternative is, or what you are proposing: maybe a new link type -- say, SingletonLink, and it takes a SetLink and returns the single atom inside the SetLink ? How could that possibly be "easier" in any way? 

The most basic how-to-use-this example is 
```
    PutLink
         blah blah blah
         GetLink
               blah blah blah
```

If GetLink started returning SetLinks for singletons, then this example changes to
```
    PutLink
         blah blah blah
         SingletonLink
              GetLink
                    blah blah blah
```
I don't see how that is "simpler" .. we have to invent a SingletonLink, define what it does ... yuck.

ngeiswei:
Hmm, I think I start to understand even more your motivation.

According to the wiki http://wiki.opencog.org/w/PutLink PutLink works that way
```
PutLink
   <clauses with X, Y, Z>
   ListLink A B C
```
if you replace ListLink A B C by a GetLink that is non singleton (i.e. a SetLink) what happens? do PutLink support that? I would assume so since you've said that
```
PutLink
    <Rewrite>
    GetLink
        <Pattern>
```
is equivalent to
```
BindLink
   <Pattern>
   <Rewrite>
```
More soon...

ngeiswei:
OK, let me illustrate my problem in C++. I want to get all rules from a certain rule-based system, for that I write
```c++
	Handle gl = _as.addLink(GET_LINK,
	                        // MemberLink
	                        //    VariableNode "__URE_RULE__"
	                        //    ConceptNode <RBS>
	                        _as.addLink(MEMBER_LINK, rule_var, rbs));
	Handle rule_names = satisfying_set(&_as, gl);
        for(Handle rn : LinkCast(rule_names)->getOutgoingSet()) {
            // construct the rule, etc.
```
Now if there's only one rule in my rule-based system (it's not so impossible, it happens in my utest actually), this perfectly looking code is gonna crash because rule_names will not contain a SetLink but a rule.

Sure I can add 2 lines of code to deal with the exception of only 1 rule. But, I think you'll agree it's rather inelegant, besides, how that exception should be detected? OK, I would check whether the result is a SetLink or not, in that code that would work because no rule can be a SetLink. In general, I don't know, the check might more complicated. So I think you see how this adds a burden to the user. For that reason I don't use satisfying_set, I use bindlink instead, BindLink is a bit more verbose that GetLink, overall if still makes my code more elegant. Hope you see my point.

linas:
Oh. Heh. Right. Good point. Yes, I think the intent there was that the PutLink, if it sees a set, then in unwraps it and does a Put for each individual member (i.e. behaves like 'zip').  So yes, you are right, -- I guess a SetLink  return value could/should work.

I no longer recall if this is implemented, or if it is tested or if it works...

ngeiswei:
Cool! I should have brought that code snippet right from the start, of course (for adequate use of it).

PR #80, opened 2015-06-05
Replace ure config json by atomese part 2

ngeiswei:


ngeiswei:
URE JSON configuration is gone, it follows precisely what is described in http://wiki.opencog.org/w/URE_Configuration_Format

williampma:
Where are the rules loaded outside of the "example" directory? What about on cogserver?

williampma:
Just tried to write my own loader so I can continue working.  Now that I actually tried it, everything seems a bit verbose and confusing.

Instead of using ConceptNode and InheritanceLink and all that, why not invent a whole new bunch of atom types specific to URE rule loading so that it is easier to see?  I don't think we want PLN to be applied to these atoms right?

ie.  Instead of
```
InheritanceLink
   ConceptNode "PLN"
   ConceptNode "URE"
MemberLink
   ConceptNode "PLN-deduction-rule"
   ConceptNode "PLN"
ExecutionLink
   SchemaNode "Maximum Iterations"
   ConceptNode "URE"
   NumberNode 20
```
which is not clear what is a member or what (both rule-base and rule are ConceptNode), why not use
```
RuleBaseInheritanceLink
   RuleBaseNode "PLN"
   RuleBaseNode "URE"
RuleMemberLink
   RuleNode "PLN-deduction-rule"
   RuleBaseNode "PLN"
RuleSettingLink
   SettingNode "Maximum Iterations"
   ArgumentNode 20
   RuleBaseNode "URE"
```
etc...

AmeBel:
1. why not apply, PLN? Eg. Meta- reasoning
2. How will the new Link types decrease verbosity?

williampma:
Yeah, new link type doesn't decrease verbosity, but it does decrease confusion.  Just throwing out ideas.

So you are saying we might want to reason with the rules themselves??  So as the URE engine applies rules which will then in turn modify the rules themselves, and have the URE engine dynamically reload the modified rules or something...?  Sounds like a mess.

AmeBel:
In the short run no, in the long run yes. For inspiration http://wiki.humanobs.org/_media/public:replicode:replicode.reykjavik.september.2010.videoremoved.ppt slide 4

AmeBel:
@ngeiswei which is the convention Node vs ConceptNode ?

https://github.com/opencog/atomspace/blob/56c5d86db3afa36cd8109643f8f963e73c118a11/examples/rule-engine/rules/crisp-deduction.scm#L74 vs http://wiki.opencog.org/w/URE_Configuration_Format#Rule_Set

williampma:
Node for "rule", ConceptNode for "Rule Base"

ngeiswei:
There is no convention. A ConceptNode has members, defined via using MemberLinks. Can we say that a rule has members, maybe? Maybe not? A rule in my view in a function, I actually thought of being a SchemaNode. Since I wasn't sure I decided it would be a Node. Once Atomese has a proper deep type semantics, we can choose the right type, for now I just wanted a node to refer to a rule.

ngeiswei:
@williampma regarding representing rule-based systems with InheritanceLinks and its possible interference with PLN itself. That's a good question and concern, PLN could indeed be applied and create knew knowledge regarding the URE configuration, but I think this new knowledge should not interfere with the configuration itself, it should be compatible. Note that the URE configuration semantics is compatible with PLN semantics, so I think it should be turn out OK, if not it has to be corrected. Also, normally reasonings will take place in contexts, and attention allocation shouldn't normally make reasoning apply on those URE configuration atoms. We'll see!

Issue #87, opened 2015-06-11
unbound variable on cog-execute! from the cogserver


AmeBel:
Trying to run https://github.com/opencog/atomspace/issues/57
```
Backtrace:
In ice-9/boot-9.scm:
 157: 12 [catch #t #<catch-closure 3339bc0> ...]
In unknown file:
   ?: 11 [apply-smob/1 #<catch-closure 3339bc0>]
In ice-9/boot-9.scm:
 157: 10 [catch #t #<catch-closure 33398e0> ...]
In unknown file:
   ?: 9 [apply-smob/1 #<catch-closure 33398e0>]
   ?: 8 [call-with-input-string "(counter) (counter)(counter)(counter)\n" ...]
In ice-9/boot-9.scm:
2320: 7 [save-module-excursion #<procedure 3119510 at ice-9/eval-string.scm:65:9 ()>]
In ice-9/eval-string.scm:
  44: 6 [read-and-eval #<input: string 1be8dd0> #:lang ...]
  37: 5 [lp (counter)]
In ice-9/eval.scm:
 386: 4 [eval # #]
 393: 3 [eval # #]
In unknown file:
   ?: 2 [memoize-variable-access! #<memoized cog-execute!> #<directory # 18aac60>]
In ice-9/boot-9.scm:
 102: 1 [#<procedure 3118780 at ice-9/boot-9.scm:97:6 (thrown-k . args)> unbound-variable ...]
In unknown file:
   ?: 0 [apply-smob/1 #<catch-closure 3339860> unbound-variable ...]

ERROR: In procedure apply-smob/1:
ERROR: Unbound variable: cog-execute!
ABORT: unbound-variable
```
on the cogserver. .guile has the following

```
; OpenCog paths
(add-to-load-path "/opencog")
(add-to-load-path "/opencog/build")
(add-to-load-path "/opencog/opencog/scm")
(add-to-load-path) "/usr/local/share/opencog")
(add-to-load-path "/usr/local/share/opencog/scm")
(add-to-load-path ".")

; Enable readline features for the REPL
(use-modules (ice-9 readline))
(activate-readline)
```

It used to work without the following lines in .guile
```
(add-to-load-path) "/usr/local/share/opencog")
(add-to-load-path "/usr/local/share/opencog/scm")
```

what am i missing?

AmeBel:
I don't think entries on .guile matter, i get the following
```
guile> %load-path
[...]

PR #88, opened 2015-06-11
include a missing headerfile for cog-execute

AmeBel:
part of fix for https://github.com/opencog/atomspace/issues/87


linas:
this won't fix #87

Issue #97, opened 2015-06-18
Derived AtomSpace's getHandlesByType() bug?


williampma:
Some inconsistency while working with InitiateSearchCB::link_type_search() lead me to this problem:

Empty AtomSpace _as:
```
	AtomSpace temp(_as);

	_as->addNode(CONCEPT_NODE, "lalala");
	temp.addNode(CONCEPT_NODE, "kakaka");

	HandleSeq handle_set;
	temp.getHandlesByType(handle_set, CONCEPT_NODE);
```

I expected `handle_set` will be size 2, but instead got size 1.  Only "kakaka" was found, even though "lalala" should also be considered part of "temp"?

williampma:
I can sort of see the problem.

AtomTable's `getHandlesByType` only look at its `typeIndex` member, but `typeIndex` does not contain information on the parent's `typeIndex`...

Perhaps to fix this I just need to change the function to also copy `_environ->getHandlesByType`...?

linas:
Yes, I think that is the correct semantics and the correct fix.

PR #98, opened 2015-06-19
AtomTable: allow _environ's typeIndex to go thru

williampma:
For https://github.com/opencog/atomspace/issues/97

I did the same thing with nodeIndex, linkIndex, and importanceIndex (specifically getNumNodes, getNumLinks, etc) at https://github.com/williampma/atomspace/tree/ATIndex .  However, apparently the assumption in the scheme version of MultiAtomSpaceUTest (https://github.com/opencog/atomspace/blob/master/tests/scm/MultiAtomSpaceUTest.cxxtest) is different (ie, only the nodes/links in the derived space are counted).  So I ended up not touching them.

It's hard to tell which assumption is correct.

### DIFF
--- a/opencog/atomspace/AtomSpace.cc
+++ b/opencog/atomspace/AtomSpace.cc
@@ -356,7 +356,7 @@ void AtomSpace::clear()
 {
     std::vector<Handle> allAtoms;
 
-    atomTable.getHandlesByType(back_inserter(allAtoms), ATOM, true);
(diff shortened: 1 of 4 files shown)


williampma:
@linas could you review?  Thanks.

williampma:
I am going to merge for now since I need this to fix a bug.

linas:
looks good to me.

PR #99, opened 2015-06-19
URE misc improvements

ngeiswei:
Mainly make the API between the FC and BC more consistent, so they look more like each other. In particular, but not only, have the BC take the rule-base system config atom in constructor argument.

### DIFF
--- a/opencog/rule-engine/backwardchainer/BackwardChainer.cc
+++ b/opencog/rule-engine/backwardchainer/BackwardChainer.cc
@@ -36,20 +36,12 @@
 
 using namespace opencog;
 
-BackwardChainer::BackwardChainer(AtomSpace* as, const std::vector<Rule>& rs)
-	: _as(as), _rules_set(rs)
-{
-	// create a garbage superspace with _as as parent, so codes acting on
-	// _garbage will see stuff in _as, but codes acting on _as will not
-	// see stuff in _garbage
-	_garbage_superspace = new AtomSpace(_as);
-}
-
-BackwardChainer::~BackwardChainer()
-{
-	// this will presumably remove all temp atoms
-	delete _garbage_superspace;
-}
+BackwardChainer::BackwardChainer(AtomSpace& as, Handle rbs)
+	: _as(as), _rbs(rbs), _configReader(as, rbs),
+	  // create a garbage superspace with _as as parent, so codes acting on
+	  // _garbage will see stuff in _as, but codes acting on _as will not
+	  // see stuff in _garbage
+	  _garbage_superspace(&_as) {}
 
 /**
  * Set the initial target for backward chaining.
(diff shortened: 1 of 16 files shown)


williampma:
Oh... I was hoping you would move the config out of ForwardChainer instead of putting it into BackwardChainer. Somehow it just seems like a bad idea to have things so tightly coupled... oh well.

You seem to be quite obsessed with the 80-col limit... It is actually easier to for me to see the code if some of them are in one line, so perhaps you could hold the code cosmetic changes after I am done with the code?

ngeiswei:
Yeah, I wasn't sure about the config being given to the FC and BC. But you do need to pass a structure holding the configuration. Maybe could have a struct with all parameters that UREConfigReader fills, then this is passed to the FC BC constructors. Or to some set_config method, if we don't want to have that in the constructor. What do you say?

williampma:
Are we sure the config FC and BC uses will be the same?  It seems it is possible the parameters they use will be different, especially if the fitness function is in the AtomSpace in the future.  In that case, set_config method probably makes better sense.

ngeiswei:
The way I see it, is that parameters that differ between FC and BC would still be tied to the same rule-based system concept node. And since it doesn't make much sense to use a chainer without specifying a rule-based system why not putting at construction time. Having said that we could still have a set_config method, it comes down to whether we consider useful to change the rule-based system on the same chainer object.

Also I'm not sure whether a chainer should take the rbs node, or a C++ structure with all the parameters fetched from that rbs node.

Anyway I'm willing to let it that way for now, and after gathering more experience.

ngeiswei:
OK, after thinking a bit further I think a chainer should really be tied to an rbs node rather than a C++ structure, and ultimately any modification on the parameters should be reflected in the AtomSpace (unlike now, where set_<PARAMETER> only affects the UREConfigReader object). But again I guess we can let it that way for now.

williampma:
I was thinking maybe the only config that really need to be passed to FC and BC are the rules & fitness functions.  Other configurations (like max steps, stopping criteria) could potential be handled outside by the combined ForwardChainer/BackwardChainer.  ie. the `do_chain` method can be removed and the only thing left for both chainer would be `do_step`.  Additional calling option to the combined caller could make it do only FC or BC.

opencog/rule-engine/backwardchainer/BackwardChainer.cc
```diff
@@ -36,20 +36,12 @@
 
 using namespace opencog;
 
-BackwardChainer::BackwardChainer(AtomSpace* as, const std::vector<Rule>& rs)
-	: _as(as), _rules_set(rs)
```
williampma: Umm... why did you remove the storing the vector of rule objects?  Now _rules_set is empty and no rule is given to the BackwardChainer, break the unit test.

ngeiswei: Oh cool, that was it! I was about to start debugging that, thanks!

ngeiswei:
I understand what you mean @williampma , there is the behavior part (rules + fitness) that does change often, and the control part (max iter) that may change every time you run it. Max iter should ultimately be replaced by a termination criteria, and that starts looking like a fitness, so I'm not completely sure. In fact I'm unsure about many things, once we start using it more extensively we'll surely want to adapt API and the configuration.

PR #106, opened 2015-06-22
Improve Haskell Api.

MarcosPividori:
Hi,
I have been working on Haskell bindings. I want to make this pull request before going too far.
First, I focused on the Atom data type representation. I have gone in depth in this topic and I have written a summary on the project wiki page with the main ideas: [1](http://wiki.opencog.org/w/Haskell_Bindings_-_GSoC_2015#Atoms_Data)
As I far as I understand, we want to have simple bindings, as similar as possible to the Atom notation described on the OpenCog documents. Also we want to take advantage of Haskell type system to impose type restrictions in how atoms relates between them, implemented as an EDSL.
I investigated the main design options available, their advantages/disadvantages, in order to have a clear idea about which is the best. On the wiki page, I explain more on this, the posibility of using template haskell, etc.
On the branch: [data-options](https://github.com/MarcosPividori/atomspace/tree/data-options/opencog/haskell/api_options) I implement 4 different options for the data type representation and I explain their advantage/disadvantages. It looks like the best option is to work with GADTs, so I went in that direction.
On the other hand, I also worked improving the API, with 3 new main functions to interact with the atomspace: insert, remove and get (and a debug function to print the atomspace on stderr). Also, I offer a util function printAtom to show a given atom in opencog notation. I updated the examples with this new interface.
Every comment/ review is welcome!
Thanks,
Marcos

### DIFF
--- a/examples/haskell/README.md
+++ b/examples/haskell/README.md
@@ -1,0 +1,21 @@
+Haskell Examples:
+================
+
+To run these examples you have to previously
+[build](https://github.com/opencog/atomspace#building-atomspace) and
+[install](https://github.com/opencog/atomspace#install) the AtomSpace.
+
+Then you can just compile them with:
+```
+ghc example.hs
+```
+
+If when running an example you get an error: "...cannont open shared object
+file: No such file or directory ...":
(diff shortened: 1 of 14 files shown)


ngeiswei:
@MarcosPividori Excellent, thanks!

PR #111, opened 2015-06-26
add function to refresh non-core types in py

chenesan:
the get_refreshed_types() is for such a situation: We import a cython module which invokes atom_types.inheritance to call classserver to add atom types. Then we want to use atom types in these modules(ex.spacetime/nlp/...). At this time we can use this function to call the type generator again and get new types object.


linas:
thanks!

Issue #115, opened 2015-06-29
convert camel case atomspace API method names to underscore


ngeiswei:
Might seem futile one but frankly it's rather annoying to see this mix of camel case and underscore in the atom and atomspace util libraries.

ngeiswei:
I didn't touch the AtomSpace API per se (it's a lot of work), but at least there isn't this mixture of camel case and underscore in the Atom and AtomSpace utils.

linas:
this is done, right?

ngeiswei:
Not completely, I'm waiting for the next moment I'll be completely brain incapacitated (yet still willing to be productive) to pursue this endeavor.

PR #117, opened 2015-06-30
Convert AtomSpace API method names from camel to underscore

ngeiswei:
Please check and merge if you like.

There would be plenty of other classes to convert like Atom, Node, Link, TV, etc. I won't be taking care of them soon, I just wanted to show the way with the AtomSpace API.


linas:
Why? is it really worth it?  I don't mind a mixture of both; some things are camel case, I know what they are from, some are snake case, I know where they are from.

ngeiswei:
It's definitely not urgent but I think eventually we want something really clean and uniform, at least for all the user API part, the earlier we take care of that the better.

Anyway, that was not that much work, just a proof of concept, we can merge it or leave it. I don't want to decide for the community.

Others?

ferrouswheel:
Consistency makes lives easier for people who are not veterans of the project. :+1: from me if this is the agreed coding style for the rest of the project.

ngeiswei:
It agrees with

http://wiki.opencog.org/w/Development_standards#Multi-word_identifiers

I added that after Curtis complained about the lack of homogeneity in coding style. I don't feel strict about that, but after seeing mixed coding style within the same, and asking myself "what coding style should I use here?", I felt it was time to clean stuff up.

I guess I'll wait for a few days and merge if nobody complains.

ngeiswei:
I'm gonna merge. There are things I'm not sure about like get_TV or get_AV might be better in lower case (get_tv, get_av), that is at least what emacs string-inflection converter has suggested. Most other things like add_node instead of addNode are a no brainer, in terms of following the dev standard, that is. For the rest it can easily be corrected.

linas:
Homogeneity is a usability issue. Usability as multiple components: vision
(glasses), size of font, style of font, indentation, and casing.  When you
make all the code homogeneous, it is harder to read and understand. A
mixture of CamelCase and snake_casse is easier to read, quicker to read,
easier to understand, and has a whole pile of human usability positives.
Converting everything to a single case e.g. snake_case makes things HARDER
to read and understand, not easier.

This is not another case of Linas saying crazy things: this has been
studied by typographers for hundreds of years. There is a reason that the
descenders for the lower case letters p and g go below the line, and that
the ascenders for d and b go up, whole the loops in all these letters are
aligned: the ascenders and descenders make it easier to read.  One could
have placed the letter loops at deifferent levels! and some early fonts
did.   It is also the reason we don't write in capital letters: capital
letters are harder to scan, because different words don't ave different
shapes.

So I think this case-conversion effort is a mistake, and I also think that
anyone who thinks that all code should be the same case is  simply unaware
of the human-usability issues involved.  Until we teach computers to
program, we have to make accomodations for humans, and that means making
things easier for humans.  We should not be beholden to fashion nazi's who
make ridiculous claims about beauty: these people don't have to actually
read and write code for a living, and they don't understand how hard it
is.  A mxiture of snake_case and CamelCase is (much) better than just one
alone.  Arghhhh.

--linas

ferrouswheel:
I think it's less about homogeneity than consistency. If you have a standard and follow/enforce it, you avoid these arguments and can focus on programming. Python has PEP8 to pre-empt this discussion, Go has `go fmt` to enforce a style as part of the compiler. These communities are fun to work in.

> We should not be beholden to fashion nazis who make ridiculous claims about beauty

This is exactly why coding conventions exist!

ngeiswei:
Linas, I understand your argument about human readability, that's exactly what the standard tries to exploit, read it well, for functions and methods snake_case, for class or template types CamelCase, like in Haskell, boost, STL.

Having that some class methods in CamelCase and other in snake_case makes no sense to me. Ultimately you know that this class is in CamelCase and that is in snake_case but the new user doesn't, it has to learn it, and it's completely arbitrary, so yeah, no very beautiful.

I remember when I resumed working on OpenCog (beside MOSES) 6 months ago and I had to first add an atom after something like 5 years without using the AtomSpace API, I actually wrote it in snake case, cause that's what the other parts of OpenCog I knew where mostly using, when I realized it was camel case it felt unnatural.

And again, beauty aside, as I said, what made me trigger that effort is when I saw in the same file a combination of camel and snake, then I had to ask myself what to follow, since the atomspace API was in camel while other components of it are in snake, there was just no clear answer. The advantage of following a standard is that you don't have to ponder about certain things, the standard tells you what to do.

ngeiswei:
Just to point to issue #115

PR #118, opened 2015-06-30
Modify SchemePrimitive for sureal

leungmanhin:
A new int argument is added to control the max number of searches being done by SuRealPMCB

### DIFF
--- a/opencog/guile/SchemePrimitive.h
+++ b/opencog/guile/SchemePrimitive.h
@@ -101,7 +101,7 @@ class SchemePrimitive : public PrimitiveEnviron
 			HandleSeq (T::*q_h)(Handle);
 			HandleSeq (T::*q_hti)(Handle, Type, int);
 			HandleSeq (T::*q_htib)(Handle, Type, int, bool);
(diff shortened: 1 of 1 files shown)


opencog/guile/SchemePrimitive.h
```diff
@@ -139,7 +139,7 @@ class SchemePrimitive : public PrimitiveEnviron
 			Q_H,   // return HandleSeq, take handle
 			Q_HTI, // return HandleSeq, take handle, type, and int
 			Q_HTIB,// return HandleSeq, take handle, type, and bool
-			K_H,   // return HandleSeqSeq, take Handle
+			K_HI,  // return HandleSeqSeq, take Handle, int
```
linas: instead of deleting K_H, why not just add K_HI?  Surely someone needs it ...

leungmanhin: Because sureal is the only one using K_H, so I believe this should not affect the others

leungmanhin: By the way, I've added K_H back to SchemePrimitive

Issue #128, opened 2015-07-06
(use-modules (opencog)) doesn't seem to load scheme files


ngeiswei:
According to opencog/scm/opencog.scm scheme files like utilities.scm should be automatically loaded but aren't.

I've added a bunch of (display "~~~ N ~~~:") in opencog/scm/opencog.scm to see whether those loading lines are executed, they seem to be. I'm really puzzled, see below:

```scheme
nilg@laptop:~/OpenCog/opencog/opencog/reasoning/pln$ guile --no-auto-compile
;;; note: source file /usr/local/share/opencog/scm/opencog.scm
;;;       newer than compiled /home/nilg/.cache/guile/ccache/2.0-LE-8-2.0/usr/local/share/opencog/scm/opencog.scm.go
~~~ 1 ~~~:~~~ 2 ~~~:~~~ 3 ~~~:~~~ 4 ~~~:~~~ 5 ~~~:GNU Guile 2.0.9
Copyright (C) 1995-2013 Free Software Foundation, Inc.

Guile comes with ABSOLUTELY NO WARRANTY; for details type `,show w'.
This program is free software, and you are welcome to redistribute it
under certain conditions; type `,show c' for details.

Enter `,help' for help.
scheme@(guile-user)> (use-modules (opencog))
scheme@(guile-user)> (stv 1 1)
;;; <stdin>:2:0: warning: possibly unbound variable `stv'
<unnamed port>:2:0: In procedure #<procedure 1ba5020 at <current input>:2:0 ()>:
<unnamed port>:2:0: In procedure module-lookup: Unbound variable: stv

Entering a new prompt.  Type `,bt' for a backtrace or `,q' to continue.
scheme@(guile-user) [1]> (load-from-path "utilities.scm")
scheme@(guile-user) [1]> (stv 1 1)
$1 = (stv 1 0.99999982)
```

As you can see (stv 1 1) only runs after re-loading utilities.scm. Obviously (using-modules (opencog)) is present in my .guile config file. Running out of idea, any advice is welcome, thanks!

ngeiswei:
BTW, I've re-added
```scheme
(load-from-path "utilities.scm")
```
at the end of opencog.scm. Same thing.

jswiergo:
Hi. If you use "use-modules" then all variables not exported by the module stay local inside the scope of module. Like a local function variables not visible from outside of the function or private methods in C++.

There are several alternatives:
[...]

PR #133, opened 2015-07-09
Fix scheme opencog modules (partially)

ngeiswei:
I only took care of utilities for opencog (and rule-engine-utils for rule-engine), as that's what I'm using so far.


ngeiswei:
Address partially #128

Issue #137, opened 2015-07-10
The python bindings docs seems outdated


ngeiswei:
Neither

http://wiki.opencog.org/w/Python

nor

https://github.com/opencog/atomspace/tree/master/opencog/python

work for me. They are somewhat contradictory as well. Obviously something is outdated in the python bindings documentations.

ngeiswei:
OK it seems

https://github.com/opencog/atomspace/blob/master/opencog/cython/README.md

is correct. However the other docs should still be fixed. In particular
```
Install the required dependencies:
\```
cd ~/opencog/opencog/python
sudo pip install -r requirements.txt --upgrade
\```
```
in https://github.com/opencog/atomspace/blob/master/opencog/python/README.md doesn't work at all. And I assume this file should point the reader to https://github.com/opencog/atomspace/blob/master/opencog/cython/README.md for further details.

ngeiswei:
I can try to fix it but I'm not totally confident. I'll create a pull request and let the experts merge...

ngeiswei:
I've fixed the wiki to reflect Cython's README.md

http://wiki.opencog.org/w/Python#Setting_up

as well as the building page

http://wiki.opencog.org/w/Building_OpenCog#Building

ngeiswei:
I think it can be closed but since I'm not sure about the requirements.txt thing as described in #138 I'll hand it back to the experts.

linas:
Oh, I see. the requirements are ctually pre-reqs for the python modules, not the cython code.

Many of these prereqs are probably obsolete or no longer needed; they surely depend on the particular python module that someone wants to use.

inflector:
@linas - do you want me to look at this list to narrow it down to the minimum? There may not actually be any requirements for just the atomspace. I can't think of why there would be.

linas:
Are there any outstanding issues? I skimmed the wiki page, and the README's above, and did not see anything obviously flakey.  If its OK with @ngeiswei then maybe this issue can be closed.

inflector:
@ngeiswei Python is working and I've fixed and tested http://wiki.opencog.org/w/Python line by line. I also updated it to reflect the output that one will get by running the examples with the current codebase.

So it's ready to be retested by someone else and closed.

ngeiswei:
@inflector everything was mostly correct (I've fixed some minor discrepancies) till

http://wiki.opencog.org/wikihome/index.php/Python#MindAgents_in_Python

I'm getting the error
```
[...]

PR #138, opened 2015-07-10
Fix python doc

ngeiswei:
Partially addresses issue #137 

I'm think my changes are consistent with the state of the code. However I don't know if the requirements.txt file should be moved from the opencog repo to the atomspace repo (and the pip install instruction put back) instead.

### DIFF
--- a/opencog/python/README.md
+++ b/opencog/python/README.md
@@ -1,19 +1,14 @@
 Using Python with the AtomSpace
 ===============================
 
-Add these lines to the end of your ~/.bashrc file:
-```
-PYTHONPATH="$PYTHONPATH:/usr/local/share/opencog/python"
-export PYTHONPATH
-```
+#### Installation and Usage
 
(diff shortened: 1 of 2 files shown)


opencog/python/README.md
```diff
@@ -1,19 +1,14 @@
...
-```
-PYTHONPATH="$PYTHONPATH:/usr/local/share/opencog/python
-export PYTHONPATH
-```
+#### Installation and Usage
 
-Install the required dependencies:
-```
-cd ~/opencog/opencog/python
-sudo pip install -r requirements.txt --upgrade
```
linas: the `pip install` shoudld be called automatically from `make install`.

I have never ever (manually) done a pip install of our python code. This perhaps explains some bugs?

linas: Never mind.  I just looked at it. The file is mis-named, they're not "requirements" they are "prerequisites".

PR #139, opened 2015-07-10
Make UREConfigReader::fetch_execution_outputs more robust

ngeiswei:
Optionally can ignore outputs of a certain type. That way we can avoid
matching patterns regarding those parameters' ExecutionLink.

### DIFF
--- a/opencog/rule-engine/UREConfigReader.cc
+++ b/opencog/rule-engine/UREConfigReader.cc
@@ -94,20 +94,28 @@ HandleSeq UREConfigReader::fetch_rules(Handle rbs)
 	return LinkCast(rule_names)->getOutgoingSet();
 }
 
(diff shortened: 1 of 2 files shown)


williampma:
I don't really understand what is going on here, but one thing:  why fetch the execution outputs inside the UREConfigReader class and not inside the Rule class?  It seems it should be part of the Rule's functionality.

ngeiswei:
That's cause ure-set-num-parameter https://github.com/opencog/atomspace/blob/master/opencog/scm/rule-engine-utils.scm#L56 creates some cruft of ExecutionLink patterns. Of course I could remove those cruft but I figured they shouldn't be confused as real input/output record, so better ignore them.

Regarding fetching the ExecutionLinks in Rule. Those parameters do not relate to rules anyway. But if they did I still wouldn't be sure. I'm fine with putting all the code pertaining to the URE config format in one place.

Issue #86, opened 2015-06-10
EinsteinUTest taking too long


linas:
Normally, EinsteinUTest runs in 10 seconds, but it is currently running in 300 seconds.  This is due to some pattern matcher bug .. this has happened before, and bisection should reveal what the problem is.

jswiergo:
This is caused by commit fb41305e4f207a87694b3b8305299a5c43af43ee (issue #76). If you revert this change the EinsteinUTest ends in less then 10 seconds but PatternCrashUTest crashes. I am still analyzing how to fix both problems.

Short summary after analyzing the code:
I have found out that after fixing #76 as a side effect you fixed calculation of undead variables in the procedure that select next untried clause. There was a bug before. I printed this undead variables out after and before the change. So this means that #76 change enabled "optimization" that choose a clause with minimum number of ungrounded variables.

Moreover seems like disabling this "optimization" helps much in that case of "elimination-rule" query in EinsteinUTest. (or there is other side effect of this change that helps and I am not aware of, is there?). When I trace the log of clauses and joiners chosen during search I see big difference. For example it looks like it is more lucky to select good joiner e.g. attr=German instead of predicate=Nationality, so it cuts longer branches in the search tree earlier.

Does above makes sense? What are your thought about following optimization strategies? Are they worth to experiment or it is not so important for real situations?

1. I have read the comment about choosing clause with smallest incoming set instead of counting variables. Indeed seems that counting does not help much.

2. In the case of "elimination-rule" that uses "c++:exclusive" predicate it is extremely poor strategy to choose this clause as a last one. What about the strategy that permits "black" clauses when their all variables are already grounded? Maybe that kind of predicates should have some additional "hint" attribute to help pattern matcher distinguish if the predicate is cheap to evaluate or not. Evaluating "c++:exclusive" just after grounding all attributes (attr_*) could prune very large part of the search tree.

linas:
Csesc,

On Sat, Jul 4, 2015 at 3:12 AM, Jacek Świergocki <notifications@github.com>
wrote:
>
> Does above makes sense?
>
Yes!

> What are your thought about following optimization strategies? Are they
> worth to experiment or it is not so important for real situations?
>
 We don't have many "real situations". Both seem like good fixes,
especially the second one.

>
>    1.
>
>    I have read the comment about choosing clause with smallest incoming
>    set instead of counting variables. Indeed seems that counting does not help
>    much.
>    2.
>
[...]

PR #141, opened 2015-07-12
EinsteinUTest optimisation by thinnest clause selection in pattern matcher (#86)

jswiergo:
I have added the first optimization discussed here (#86). Pattern matcher chooses the next joint of clauses as the one with smallest incoming set. If there are many clauses for the same joint then the clause with minimal number of ungrounded variables is selected. This solves the problem with EinsteinUTest running time as well as speeds up other unit tests a little.

I have also found and "fixed" following problems.

1. Weird behaviour of guile 2.0.5, that seemed like concurrent changes in PatternLink/BindLink object constructors. When I changed to guile 2.0.9 it all disappeared. There is also similar comment in the code:
https://github.com/opencog/atomspace/blob/master/opencog/guile/SchemeEval.cc#L239.
So I think it is better to increase required guile versions in CMakeLists to help next persons avoid
such issues.

2. A bug in matching NotLink. This kind of link is unordered but it has only one outgoing atom. There is something wrong in the procedure comparing unordered links. It is randomly missing some search branches. Randomly because incoming sets are ordered by pointers. I have changed comparison of single arity links to work as simple ordered links. I am not sure if it fixes other cases but it solves EinsteinUTest.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -214,8 +214,8 @@ FIND_LIBRARY(GMP_LIBRARY gmp)
 FIND_PATH(GMP_INCLUDE_DIR gmp.h)
 
 # Gnu Guile scheme interpreter
-# Version 2.0.0 is needed for mrs io ports, compilation, etc.
-FIND_PACKAGE(Guile 2.0.0)
+# Version 2.0.9 is needed for mrs io ports, compilation, etc.
+FIND_PACKAGE(Guile 2.0.9)
 IF (GUILE_FOUND AND GMP_LIBRARY AND GMP_INCLUDE_DIR)
 	ADD_DEFINITIONS(-DHAVE_GUILE)
 	SET(HAVE_GUILE 1)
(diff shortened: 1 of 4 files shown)


linas:
Thank you Jacek!

Can you find an example with an unordered link in it, that fails, and write a unit test for it?  I never tested arity-1 unordered links.

It is not the case that every possible ordering needs to be explored: if a certain ordering is impossible, then related orderings will be impossible too. However, the code should find all solutions ...

jswiergo:
>Can you find an example with an unordered link in it, that fails, and write a unit test for it? I never tested arity-1 unordered links.

I have found one case (#148).

Though it seems there is more. I am almost convinced that in the case of shared unordered links there is as well a problem with _perm_state structure that keeps states of permutations being explored.
Now the structure is a map:
 ("sub-pattern handle","grounding handle") => mutation.
But it should consider sharing as well:
 ("sub-pattern handle","position in main pattern","grounding handle") => mutation
It seems it messes up now and cut valid branches.
Trying to catch...

linas:
Thank you!

Yes, that fix does sound right to me.

PR #142, opened 2015-07-13
BackwardChainer: Check for empty VariableList

williampma:
Fix the utest since the PM now requires non-empty VariableList.


jswiergo:
I checked that if this new assertion line is removed then PM works in this case as previously - it returns empty set and this unit test passes. However I do not know what is expected behaviour. It can be easly reverted if needed.

leungmanhin:
Hi @jswiergo, I think it would be nice if that assertion can be removed, in some cases if we want to do a fuzzy match (finding similar but not identical patterns), the input pattern may not contain any variables, this assertion will become a problem...

jswiergo:
Yes, I see it now. Removed: #146

Issue #135, opened 2015-07-10
Haskell build break.


linas:
Something about the haskell bindings breaks the build. I get
```
cabal: Could not resolve dependencies:
trying: opencog-atomspace-0.1.0.0
```

MarcosPividori:
Hi,
I am working on this. I could not reproduce it. It looks like it is a , very common, dependency problem with haskell libraries installed in your computer. I am considering leaving cabal-install to start using stack, which is a new and far more comfortable tool for building and dealing with haskell packages. But I need to go in depth in some details before being sure this is a good option for us.
Thanks,
Marcos

linas:
I have both haskell and cabal installed, the standard installation on ubuntu trusty 14.04

```
cabal -V
cabal-install version 1.16.0.2
using version 1.16.0 of the Cabal library 
$ ghc -v
Glasgow Haskell Compiler, Version 7.6.3, stage 2 booted by GHC version 7.6.3
```

I took a quick look at the CMakefile, it looks like you are maybe trying to install libraries without root access.  The build process needs to be split into two steps: an ordinary build, as a non-root user, and an install step as root.

MarcosPividori:
Hi Linas, Thanks for your response.
The root access is not a problem because I am installing haskell packages in user environment.
I have spent some time reading about the new [Stack](https://github.com/commercialhaskell/stack/wiki) building tool. I think we should definitely stop directly using cabal to start using stack. It solves some very common problems of working with Haskell packages, such as: broken dependencies, different versions of ghc, etc. It ensures we are working with the lts snapshot, a group of haskell libraries that have been tested and work ok, etc. If someone wants to go in deepth about stack, more information on: [[1]](https://www.fpcomplete.com/blog/2015/06/why-is-stack-not-cabal) [[2]](https://www.fpcomplete.com/blog/2015/06/stack-0-1-release).
I have made a pull request #143 , with the main changes.
I think it will work far better in future!

MarcosPividori:
I forgot to mention, I have asked some questions to stack developers before updating the code: [haskell-stack](https://groups.google.com/forum/#!topic/haskell-stack/loqCuEQlvgs) google group.

Issue #83, opened 2015-06-10
haskell errors during install


linas:
`make install` seems to work, but ends with a warning:

```
Warning: The package list for 'hackage.haskell.org' does not exist. Run 'cabal update' to download it.
cabal: Could not resolve dependencies:
trying: opencog-atomspace-0.1.0.0
```

the cmakefiles should not attempt to build haskell unless it is full installed.

MarcosPividori:
Hi,
This happen probably because the user have never executed:
cabal update
So Cabal never downloaded the list of packages available. It needs this list to resolve the dependencies for opencog-atomspace. (opencog-atomspaces needs the packages "transformers" and "data-default" to be installed)
One quick fix would be to run "cabal update" before "cabal install", but I think we should remove all of this from the CMake automatically installation and make this optional to the user that wants to use Haskell Bindings. The same happen if the user wants to install this package globally or locally, etc.
If you agree, we can remove the automatically installation. So, users interested in haskell bindings, go to the haskell folder and run:
 cabal update
 cabal install
If they want other options, for example to install the package globally, they can run:
 cabal install --global

linas:
Well, ideally, it should all be automated and automatic.  Opencog is already too hard to build, install, and asking the user to do more manual steps, such as 'cabal update' etc is just too much.  Recall that most users will not know what haskell is, they just want to be able to say "cmake; make; make install" and have everything "just plain work" without getting errors, and without reading special instructions or how-tos.

MarcosPividori:
Hi Linas, ok! I have modified the code to avoid this problem. I have sent a pull request!

PR #143, opened 2015-07-13
Replace cabal install by stack - Solves Haskell build break. #135

MarcosPividori:


ngeiswei:
@MarcosPividori I've installed the latest haskell platform (after removing all haskell ubuntu packages from my system) + slack. The first time I run cmake .. on your branch I'm getting
```bash
nilg@laptop:~/OpenCog/atomspace-replace-cabal-install-by-stack/build$ cmake ..
-- Build type: Release
-- Boost version: 1.54.0
-- Found the following Boost libraries:
--   date_time
--   filesystem
--   program_options
--   regex
--   serialization
--   system
--   thread
-- Boost version 105400 found.
-- Found CogUtil version 2.0.1 at /usr/local/lib/libcogutil.so
-- CogUtil found.
Using latest snapshot resolver: lts-2.18
Writing global (non-project-specific) config file to: /home/nilg/.stack/global/stack.yaml
Note: You can change the snapshot via the resolver field there.
Downloading lts-2.18 build plan ...
Downloaded lts-2.18 build plan.
Caching build plan
Populating index cache ...
Updating package index Hackage (mirrored at https://github.com/commercialhaskell/all-cabal-hashes.git) ...
Fetching package index ...
Fetched package index.
Populated index cache.
Preparing to download ghc-7.8.4 ...
ghc-7.8.4: download has begun
ghc-7.8.4: 2682746 bytes downloaded...
ghc-7.8.4: 5779322 bytes downloaded...
ghc-7.8.4: 10039162 bytes downloaded...
ghc-7.8.4: 13643642 bytes downloaded...
ghc-7.8.4: 17215354 bytes downloaded...
ghc-7.8.4: 20852602 bytes downloaded...
ghc-7.8.4: 24457082 bytes downloaded...
ghc-7.8.4: 27438970 bytes downloaded...
ghc-7.8.4: 31649658 bytes downloaded...
ghc-7.8.4: 35254138 bytes downloaded...
ghc-7.8.4: 38875002 bytes downloaded...
ghc-7.8.4: 42479482 bytes downloaded...
ghc-7.8.4: 46083962 bytes downloaded...
ghc-7.8.4: 49622906 bytes downloaded...
ghc-7.8.4: 53227386 bytes downloaded...
ghc-7.8.4: 56799098 bytes downloaded...
ghc-7.8.4: 60436026 bytes downloaded...
ghc-7.8.4: 64040506 bytes downloaded...
ghc-7.8.4: 67644986 bytes downloaded...
ghc-7.8.4: 70325112 bytes downloaded...
Downloaded ghc-7.8.4.
Unpacking GHC ...
Configuring GHC ...
Installing GHC ...
Installed GHC.
Would add the following to PATH: /home/nilg/.stack/programs/x86_64-linux/ghc-7.8.4/bin
-- Stack found.
-- GSL found.
-- Guile (2.0.9 >= 2.0.9) was found.
-- HyperTable was not found. Make sure HYPERTABLE_LIBRARY and HYPERTABLE_INCLUDE_DIR are set.
-- Hypertable not found; the experimental distributed persistence DB needs it.
-- Python libs found.
-- Cython ( 0.20.1 >= 0.19.0) found.
-- Python destination dir found: /usr/lib/python2.7/dist-packages
-- Found Intel TBB
-- 
-- UnixODBC was found.
-- Valgrind Prefix: 
-- VALGRIND was found.
-- VALGRIND devel headers found.
-- Found ZeroMQ library: /usr/lib/x86_64-linux-gnu/libzmq.so
-- Detected ZeroMQ version number: 4.0.4
-- ZeroMQ was found.
CMake Warning at tests/persist/sql/CMakeLists.txt:60 (MESSAGE):
  Postgres database not configured for unit tests! See the README!



The following components will be built:
-----------------------------------------------
   AtomSpace         - A weighted and typed hypergraph database.
   Cython bindings   - Cython (python) bindings.
   Doxygen           - Code documentation.
   Haskell bindings  - Haskell bindings.
   Python tests      - Python bindings nose tests.
   Scheme bindings   - Scheme bindings and shell.
   SQL persistance   - Save/Restore of AtomSpace to database.
   TBB               - Intel TBB (Threaded Building Blocks) for multithreading.
   Unit tests        - Unit tests.

The following components WILL NOT be built:
-----------------------------------------------
   Hypertable        - HyperTable for scalable persistance (experimental).

-- Configuring done
-- Generating done
-- Build files have been written to: /home/nilg/OpenCog/atomspace-replace-cabal-install-by-stack/build
```
As you may see it attempts to download ghc 7.8.4, it's probably a weird thing to happen at this point but I guess this is just cabal's initialization behavior. At least didn't do that when I retried on another fresh build dir. Weirdly though I still have ghc 7.8.3, so why cabal went through the effort...?
```bash
nilg@laptop:~/OpenCog/atomspace-replace-cabal-install-by-stack$ ghc --version
The Glorious Glasgow Haskell Compilation System, version 7.8.3
```

```

File: opencog/haskell/README.md [hunk 1]
```diff
@@ -32,20 +29,54 @@ Go through the normal process of
...
-If you want to install only the haskell library, for example a new version,
-move to this directory (/opencog/haskell), build and install the
-opencog-atomspace haskell library with this command:
+If you want to use this library in some project you should include it to the
+package list on the project [stack.yaml](https://github.com/commercialhaskell/stack/wiki/stack.yaml) config file. For example:
+
+```yaml
+...
+packages:
+- ...RelativePathToThisDirectory...
```
ngeiswei: better if you use
```
<ATOMSPACE_ROOT>/opencog/haskell
```
instead of ...RelativePathToThisDirectory...

jswiergo:
Maybe it would be better to move GHC installation to another place? For example here:
https://github.com/opencog/atomspace/blob/master/scripts/ocpkg

I suppose (but not sure) that it is not common practice that cmake scripts install additional packages while configuring. Cmake just check if everything is configured and complain if not. Or if you want to install something you do it explicitly by "sudo make install" or use other tools.

ngeiswei:
@jswiergo agreed. But I suspect that stack only does that at the very first run. I installed the haskell platform + stack then straight away run cmake .. . So stack never got the chance to be run once before. If that's the case then I think it is acceptable.

ngeiswei:
If you don't have stack installed, cmake .. isn't gonna bother anymore. I'm merging...

ngeiswei:
@MarcosPividori one more problem. I'm unable to import the AtomSpace module in ghci
```bash
nilg@laptop:~/OpenCog/atomspace-replace-cabal-install-by-stack/examples/haskell$ ghci
GHCi, version 7.8.3: http://www.haskell.org/ghc/  :? for help
Loading package ghc-prim ... linking ... done.
Loading package integer-gmp ... linking ... done.
Loading package base ... linking ... done.
Prelude> import OpenCog.AtomSpace

<no location info>:
    Could not find module ‘OpenCog.AtomSpace’
    It is not a module in the current program, or in any known package.
Prelude> :q
Leaving GHCi.
nilg@laptop:~/OpenCog/atomspace-replace-cabal-install-by-stack/examples/haskell$ ghc example.hs

example.hs:3:8:
    Could not find module ‘OpenCog.AtomSpace’
    Use -v to see a list of the files searched for.
nilg@laptop:~/OpenCog/atomspace-replace-cabal-install-by-stack/examples/haskell$ stack ghc example.hs
Getting project config file from STACK_YAML environment
[1 of 1] Compiling Main             ( example.hs, example.o )
Linking example ...
```
Never obviously can I compile with ghc only. Any light on that would be welcome.

MarcosPividori:
Hi thanks for your feedback! I will explain the main points:
 - You don't need to install haskell platform anymore. Stacks does all the work for you. If you start from an empty computer, you have to download and install stack and then with: "stack setup" everything is done.
 - In the cmake configuration step I look for the stack executable, if stack is there, I ran "stack setup":
```
stack setup --help
"Get the appropriate ghc for your project"
```
  It looks for the proper version of ghc to your project and automatically download/install it if you doesn't have the proper one. So, yes, it will run **only the first time if needed**. (Proper version according to the LTS packages too, so I think it downloads the version 7.8.4 because that is needed to support LTS packages)
  I agree with you that maybe we could move the setup step to another place... but it is done on user space, not root space.
 - About the compilation with ghc and usage of ghci, you need to use stack to provide the proper environment. I have updated the examples readme file with all this information.
To use ghc:
```
export STACK_YAML=<ATOMSPACE_ROOT>/opencog/haskell/stack.yaml
stack ghc example.hs
```
To use ghci:
```
export STACK_YAML=<ATOMSPACE_ROOT>/opencog/haskell/stack.yaml
stack ghci --ghc-options -lhaskell-atomspace
```
Thanks,

ngeiswei:
Oh, that makes sense now... Yes, as I see stack ghc actually points to ghc 7.8.4. Thanks.

linas:
I very very very very very strongly object to installing GHC automatically. This is absolkutely terrible coding practice, and breaks many serious and fundamental software design rules. Please revert ASAP.

linas:
Guys, I don't know what you are thinking, but this is wrong beyond wrong in entirely new dimensions.  I'm flabergasted that you would even vaguely consider this to be an adequate solution to the problem.

ferrouswheel:
@linas stack installs a ghc inside it's stack environment. Do you dismiss the value of isolated and repeatable builds? Most modern languages embrace them.

If you don't want the haskell bindings uninstall `stack`. Easy!

linas:
Its not OK to have to run any part of the build process as root.

ferrouswheel:
Why/where is root needed?

linas:
because its attempting to install stuff.

bgoertzel:
Well, Marco said " it is done on user space, not root space." ... when let
me to believe login as root is not needed...

Whether there is a better way to organize all this, I'm not sure, though...

On Wed, Jul 15, 2015 at 8:39 PM, Linas Vepštas <notifications@github.com>
wrote:

> because its attempting to install stuff.
>
> —
> <https://github.com/opencog/atomspace/pull/143#issuecomment-121790044>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

"The reasonable man adapts himself to the world: the unreasonable one
persists in trying to adapt the world to himself. Therefore all progress
depends on the unreasonable man." -- George Bernard Shaw

ngeiswei:
I had a good chat with Marcos about that the other day. I think using stack really is a good move. So
the question is, is it right to run `stack setup` during `cmake ..`?

The alternative would be to have cmake just output a message and let the user do it. So assuming that stack is installed on your system you'd have something like

```
The following components will be built:
-----------------------------------------------
   AtomSpace         - A weighted and typed hypergraph database.
   Cython bindings   - Cython (python) bindings.
   Doxygen           - Code documentation.
   Python tests      - Python bindings nose tests.
   Scheme bindings   - Scheme bindings and shell.
   SQL persistance   - Save/Restore of AtomSpace to database.
   TBB               - Intel TBB (Threaded Building Blocks) for multithreading.
   Unit tests        - Unit tests.

The following components WILL NOT be built:
-----------------------------------------------
   Haskell bindings  - Haskell bindings.
   Hypertable        - HyperTable for scalable persistance (experimental).

In order to build Haskell bindings you need to run 'stack setup' and rerun cmake.
```

Then Haskell bindings would moved to the to-be-built components.

Personally I think we can afford to try to run `stack setup` inside cmake, it's simpler, if turns out to be harmful we can always opt for the suggestion above.

What do you say?

pennachin:
How are the other bindings handled? Does cmake install python libs and/or
guile, if they're not present? I believe not...

ngeiswei:
On 07/16/2015 12:00 PM, Cassio Pennachin wrote:
> How are the other bindings handled? Does cmake install python libs and/or
> guile, if they're not present? I believe not...

Well, those libs are expected to be installed on the system. Also, 
Haskell standard evolves faster that Python's, Guile's, C's or C++'s 
thus having a coherent set of libraries + building tools is harder. That 
is why people who use Haskell in real life have come up with Stack.

Stack installs the right Haskell environment for **that** project, on 
user space, as it's been said. So it's a slightly different situation.

I think Stack is probably the equivalent of Leiningen for Clojure 
(according to my limited understanding).

ngeiswei:
Oh, BTW, more than once while compiling projects relaying on CAML (I'm talking about real open source projects, not academic toys), I've seen ./configure (or was it make?) do the CAML installation on the userspace, the first time I said "WTF?", but in the end I was glad it saved me the hassle of finding the right CAML compiler and libraries.

All that to say that it's not an out-of-the-blue practice...

linas:
The basic rules of make and cmake are these:

-- neither make nor cmake should write anything outside of $cwd (current working directory), in this case the build directory.

-- neither make nor cmake should assume any network access.

-- `make install` must not write any files outside of $prefix (default is /usr/local but is user-settable to be anything)

-- `make install` should follow the FHS (File System Hierarchy Standard)

The point of these rules is the need to manage the filesystem, such as access controls; manage the disk space, so that adequate disk space is always available; manage the network, e.g. with containerization, buildbot, automated build tools; and to get repeatable, debuggable builds.   

Writing to the user's home directory is a major no-no: not only does it blow up disk space and threaten stability, but it makes build unrepeatable and thus undebuggable.  To get a repeatable build, it should be enough to delete the build directory and rebuild.

The restriction on root access is and network access is for basic security.  Breaking these rules would break buildbot and other automated build systems.

The restriction on FHS is another disk and system management issue, e.g. when the root file system is remote-mounted, e.g. via NFS, or because it is an overlay FS.  It also helps with debugging: you know which directories contain which files. 

Those old enough to have used MS Windows will recall something called "DLL hell" where different packages installed different, conflicting libraries and commands in different, non-standard  locations.  This is OK for small systems, but sooner or later, usually sooner, when you install enough things, the conflicts would emerge, and the different installed versions would start interfering with each-other.  Two libraries, having the same name, but installed in different locations, would have different symbols defined, because they were different versions.  An app that worked when it linked to one would crash when linked to the other. Each app tried to over-ride the environment variables set by the other, e.g. the file search path, the library path. After about a year of use, the system would become unbootable, and you would have to reinstall windows, and then reinstall the apps that you really liked/needed, and the problem would usually go away.  Sometimes, you would have to give up on using some app you really liked, because it conflicted with e.g. the corporate-mandated firewall.  It was a management nightmare.

In this case, having two different versions of GHC installed just reeks of DLL hell: you go to build something, and sometimes one version gets used, and sometimes the other, depending on some environment variable or file search path being set or not set. If the variable is set, things work, if it gets unset, you end up getting random bug reports from users on the mailing list about how it doesn't build for them.  These are very hard to debug, because its almost impossible to control the user's environment by exchanging emails.    

Linux has mostly avoided DLL hell by carefully hewing to the FHS, and a single package manager (apt, rpm) but the good old bad days have been returning, e.g. with python pip and similar tools.  (If you use python, you would know what I'm talking about. Combining ROS with pip is a debug nightmare, although this is partly catkin's fault, catkin and pip have big disagreements about who is boss).

pennachin:
@ngeiswei If Stack is the Haskell equivalent on Leiningen, then perhaps the Clojure typical practice would be useful. In Clojure, a build tool like cmake would not install lein, but it would call "lein deps", which downloads all necessary Clojure and Java dependencies. Leiningen puts those inside the project tree, though, and that seems to be the only difference.

@linas If Stack is anything close to Leiningen for Clojure, DLL hell is precisely what it avoids by doing project-level dependency tracking and guaranteeing isolated, reproducible builds for the Haskell part of the project.

MarcosPividori:
Hi @linas , thanks for your comments.
*"
The basic rules of make and cmake are these:
-- neither make nor cmake should write anything outside of $cwd (current working directory), in this case the build directory.
-- neither make nor cmake should assume any network access.
-- make install must not write any files outside of $prefix (default is /usr/local but is user-settable to be anything)
-- make install should follow the FHS (File System Hierarchy Standard)
"*

Ok, so I would like to be sure you want to follow this.
I asked you about including "cabal update", to download the list of haskell packages available, on the issue: https://github.com/opencog/atomspace/issues/83#issuecomment-111257562
I suggested to do it manually by the user if interested in haskell bindings, but you pointed out that it would preferably be automatic:

"Well, ideally, it should all be automated and automatic. Opencog is already too hard to build, install, and asking the user to do more manual steps, such as 'cabal update' etc is just too much. Recall that most users will not know what haskell is, they just want to be able to say "cmake; make; make install" and have everything "just plain work" without getting errors, and without reading special instructions or how-tos."

So, I would like to know what would you prefer: automatic installation (this includes downloading the ghc and packages needed when necessary (first time only)) or follow the points you mentioned above which means I should remove all haskell stuff from cmake.

I don't do anything on "make install" step, because I don't need root privilegies. Everything is done on build step.

Again, I need to say that mixing cmake and cabal/stack is not very common, so it needs a bit of tricky code.
If you prefer, again, I can remove automatically haskell bindings installation from cmake, and write some readme steps, which was my initial idea.


Well I am not an expert in stack, it is a new tool, but I think I understand the main functionalities. I will summarize how stack works:

**[LTS Haskell](https://github.com/fpco/lts-haskell#readme)**
"LTS (Long Term Support) Haskell is a curated set of packages which includes non-breaking point releases. It is a companion to Stackage Nightly: whereas Stackage Nightly releases include potentially breaking changes with each new release, LTS Haskell maintains major version stability for a longer period of time."

Ok. LTS is enough for Opencog Haskell bindings right now. Because of that, you can see the file: [stack.yaml](https://github.com/MarcosPividori/atomspace/blob/master/opencog/haskell/stack.yaml) specifies: "resolver: lts-2.18" which means: "I need LTS 2.18 snapshot".


So the installation steps using stack:

**stack setup:** looks for the proper ghc version based on the information on the stack.yaml file of your project. If it is not present in your system, it downloads and installs ghc in: ~/.stack/.... home directory.

**stack install:** automatically looks for packages needed and, if necessary, it downloads and installs them. As mentioned above, all packages that I need are part of the LTS Haskell snapshot.
This packages are installed in: ~/.stack/snapshots/os-version-folder/lts-2.18/7.8.4/lib/
So they can be used by other projects in your computer using the same LTS snapshot.
Then it builds/installs the target package: opencog-atomspace . As it is not part of the LTS, snack build/install it in: ATOMSPACE_ROOT/opencog/haskell/.stack-work/install/os-version-folder/lts-2.18/7.8.4/lib/

So, as resume, stack creates an isolated environment for your project, using a specific ghc version and a specific LTS snapshot.
As ghc and LTS's packages could be used by other projects, it installs them in your ~/.stack....  directory.
On the other hand, project specific stuff, as opencog-haskell bindings package, is installed in your: ATOMSPACE_ROOT/opencog/haskell/.stack-work/ directory, at your own risk.
Then you can work on this environment, setting STACK_YAML to point to your stack.yaml project specific file, and using: stack ghc, stack ghci, etc.

As you can see, "stack setup" and "stack install" both *need network access* and *write to your ~/.stack directory* the first time. So if you think we should follow the rules of make and cmake you mentioned above, I could remove these commands from cmake, and make users do this on their own.

-----------------------------------------------------------------------------------------------

*"
The point of these rules is the need to manage the filesystem, such as access controls; manage the disk space, so that adequate disk space is always available; manage the network, e.g. with containerization, buildbot, automated build tools; and to get repeatable, debuggable builds.
Writing to the user's home directory is a major no-no: not only does it blow up disk space and threaten stability, but it makes build unrepeatable and thus undebuggable. To get a repeatable build, it should be enough to delete the build directory and rebuild.
"*
Stack doesn't write anything in user space except: the initial ghc installation (if needed), and LTS packages. They will be the same for everyone, according to the LTS version mentioned in the stack.yaml file.

-----------------------------------------------------------------------------------------------

*"
Those old enough to have used MS Windows will recall something called "DLL hell" where different packages installed different, conflicting libraries and commands in different, non-standard locations. This is OK for small systems, but sooner or later, usually sooner, when you install enough things, the conflicts would emerge, and the different installed versions would start interfering with each-other. Two libraries, having the same name, but installed in different locations, would have different symbols defined, because they were different versions. An app that worked when it linked to one would crash when linked to the other. Each app tried to over-ride the environment variables set by the other, e.g. the file search path, the library path. After about a year of use, the system would become unbootable, and you would have to reinstall windows, and then reinstall the apps that you really liked/needed, and the problem would usually go away. Sometimes, you would have to give up on using some app you really liked, because it conflicted with e.g. the corporate-mandated firewall. It was a management nightmare.
"*
I agree with you that installing packages/libraries everywhere without a proper management can lead to many problems.
That's exactly a very common problem when using cabal-install, which many people call: [cabal hell](https://wiki.haskell.org/Cabal/Survival#What_is_the_difficulty_caused_by_Cabal-install.3F)
Haskell's Packages management is by far the main problem nowadays when working in a haskell project. As you can see in a recently [Haskell survey](https://www.fpcomplete.com/blog/2015/05/thousand-user-haskell-survey):
"Package management with cabal is the single worst aspect of using Haskell. "

Stack solves some very common problems of working with Haskell packages, such as: broken dependencies, different versions of ghc, etc. It ensures that everyone in a project, is working with the same LTS snapshot, the same compiler version, etc. It ensures everyone BUILDS WITH THE SAME PLAN.
Stack knows exactly which ghc version it needs to use to some specific project.

-----------------------------------------------------------------------------------------------

*"
GHC installed just reeks of DLL hell: you go to build something, and sometimes one version gets used, and sometimes the other, depending on some environment variable or file search path being set or not set. If the variable is set, things work, if it gets unset, you end up getting random bug reports from users on the mailing list about how it doesn't build for them. These are very hard to debug, because its almost impossible to control the user's environment by exchanging emails.
"*

Won't happen! because if you use: stack ghc, stacks automatically figure out the proper ghc version to use based on your stack environment!! :) And everyone will be using the same version, so everyone can reproduce some existing bug.

-----------------------------------------------------------------------------------------------

**Summary:**
In user's home directory, stack only installs:   (on the ~/.stack/... directory)
- GHC with proper version.
- Packages of the LTS snapshot. 
So, they can be reused by another project.
They don't represent any problem in future because stack properly cares of this.

So, we have 2 options:
- Include cmake stuff to check if stack is installed, execute "stack setup" and "stack install" when building.
- Remove this from cmake and include a readme file with these steps, to be manually executed by the user.

Thanks!
Marcos

jswiergo:
Consider option 3) that I suggested at first. Instead of manualy executed steps by the user add this steps to ocpkg or other "install environment packages" administration script but not cmake.

bgoertzel:
From the perspective of a relatively naive user (i.e. I'm pretty damn rusty
about Linux software installation, even if not about AI algorithms), I
would generally have a preference for solutions involving the user to take
fewer steps...

On Thu, Jul 16, 2015 at 12:05 PM, Jacek Świergocki <notifications@github.com
> wrote:

> Consider option 3) that I suggested at first. Instead of manualy executed
> steps by the user add this steps to ocpkg or other "install environment
> packages" administration script but not cmake.
>
> —
> <https://github.com/opencog/atomspace/pull/143#issuecomment-122003868>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

"The reasonable man adapts himself to the world: the unreasonable one
persists in trying to adapt the world to himself. Therefore all progress
depends on the unreasonable man." -- George Bernard Shaw

linas:
Well, installing anything into the user's home directory is really pretty much unacceptable, from the disk-management point of view. It breaks unix conventions (the unix convention being to use the FHS.)  It will certainly break buildbot.   The whole point of having a standard is that ... well, standards exist to solve real problems ...  Having stack break the FHS standard seems like a flawed design choice.

Installing snapshots of haskell also seems like a terrible idea: despite best efforts, development trees are always unstable, with new bugs appearing and disappearing regularly (daily?).  I can't believe that the haskell developers are immune to bugs.   To have stack install a new version of GHC that changes daily or weekly seems like asking for trouble.

I would much rather install a STABLE, released version of GHC, something that won't change except maybe a few times a year (or less often)  Putting the install into ocpkg seems like a good choice.

ferrouswheel:
Try using the STACK_ROOT environment variable, making it use a location in
build tree instead.

bgoertzel:
Hi Linas,

Hmmm....   Well, regarding which Haskell package management tool to use, I
would tend to trust the most knowledgeable Haskell developers among us.
Marcos is definitely the top Haskell developer involved here, and so his
opinion means a lot to me on these matters....

As an amateur Haskell developer (I was pretty good in the mid-90s, but the
Haskell universe was very different then, even though the core language is
the same thing), I have myself experienced cabal-hell and stack seems very
welcome.   A good articulation of the reasons for moving from cabal-install
to stack are given here....

https://www.fpcomplete.com/blog/2015/06/why-is-stack-not-cabal

if you're curious....   If you read that blog post you'll see that your
comment

***
I would much rather install a STABLE, released version of GHC, something
that won't change except maybe a few times a year (or less often) Putting
the install into ocpkg seems like a good choice.
***

is a bit beside the point in the context of practical Haskell development.
GHC itself is not really the issue that's causing so much hassle for
Haskell beginners and intermediates....

Making things easier for developers who are not super-gonzo Haskell experts
seems important in terms of growing the user base for the OpenCog Haskell
bindings....  Many uses of these bindings may be programmers who are
interested in Haskell but not Haskell experts, and who might easily get put
off by dealing with cabal-hell...

...

Of course, issues about the overall make/install process of OpenCog go
beyond Haskell issues, and I trust your expertise on the overall
make/install process generally speaking

...

Marco: I wonder what you think of Jacek's suggestion that

***
Consider option 3) that I suggested at first. Instead of manualy executed
steps by the user add this steps to ocpkg or other "install environment
packages" administration script but not cmake.
***

Naively, this would seem to provide a fully automated solution based on
stack, allowing novice developers to get started w/ minimal hassle, but
without violating the principles Linas cites... (?)

- Ben

On Thu, Jul 16, 2015 at 3:39 PM, Joel <notifications@github.com> wrote:

> Try using the STACK_ROOT environment variable, making it use a location in
> build tree instead.
> On 17 Jul 2015 6:34 am, "Linas Vepštas" <notifications@github.com> wrote:
>
> > Well, installing anything into the user's home directory is really pretty
> > much unacceptable, from the disk-management point of view. It breaks unix
> > conventions (the unix convention being to use the FHS.) It will certainly
> > break buildbot. The whole point of having a standard is that ... well,
> > standards exist to solve real problems ... Having stack break the FHS
> > standard seems like a flawed design choice.
> >
> > Installing snapshots of haskell also seems like a terrible idea: despite
> > best efforts, development trees are always unstable, with new bugs
> > appearing and disappearing regularly (daily?). I can't believe that the
> > haskell developers are immune to bugs. To have stack install a new
> version
> > of GHC that changes daily or weekly seems like asking for trouble.
> >
> > I would much rather install a STABLE, released version of GHC, something
> > that won't change except maybe a few times a year (or less often) Putting
> > the install into ocpkg seems like a good choice.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/opencog/atomspace/pull/143#issuecomment-122043442>.
> >
>
> —
> <https://github.com/opencog/atomspace/pull/143#issuecomment-122064972>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

"The reasonable man adapts himself to the world: the unreasonable one
persists in trying to adapt the world to himself. Therefore all progress
depends on the unreasonable man." -- George Bernard Shaw

MarcosPividori:
Hi,
Thanks for your recommendations!

This is the situation:
I am currently working on haskell bindings. It's a haskell package (opencog-atomspace). It is a library, so I am not building specific executables.
Following stack proposal, instead of installing this package globally, (cabal-install includes all packages in the same database and this generate many problems such as broken dependencies because of mixing different incompatibles packages), stack generates an isolated environment for each project, and installs packages in that environment (unless they are part of the LTS snapshot, which means they are shared).
So, to use the opencog-atomspace package (atomspace haskell bindings), you have to include its path to your stack.yaml file. For example:

- I want to use the opencog-atomspace package in a project.
- I include it in my <PROJECT_ROOT>/stack.yaml file:
```yaml
  ....
  packages:
   - <ATOMSPACE_ROOT>/opencog/haskell
  ...
```
- then, I install it with:
       stack install opencog-atomspace
- Ok. opencog-atomspace package was installed in my isolated environment:
      <PROJECT_ROOT>/.stack-work/.....
- Now I can use this library with: stack ghc, stack ghci, etc.

So, as right now, we don't have a specific project using the library, we only provide the library to the user (we are not building any application using haskell bindings).
The library is built by each project interested in using it in their insolated environment, such as the examples on: /examples/haskell
So, all we have to do is provide this library to the users in a location so they can include it in their projects.
I would like to include this package in: "/usr/local/lib/opencog" ,but it is not possible now because of some stack limitations. I have discussed this with some stack developers on their google group: [haskell-stack](https://groups.google.com/forum/#!topic/haskell-stack/loqCuEQlvgs)

--------------------------------------------------------------------------------------------------------

*"
Well, installing anything into the user's home directory is really pretty much unacceptable, from the disk-management point of view. It breaks unix conventions (the unix convention being to use the FHS.) It will certainly break buildbot. The whole point of having a standard is that ... well, standards exist to solve real problems ... Having stack break the FHS standard seems like a flawed design choice.
Installing snapshots of haskell also seems like a terrible idea: despite best efforts, development trees are always unstable, with new bugs appearing and disappearing regularly (daily?). I can't believe that the haskell developers are immune to bugs. To have stack install a new version of GHC that changes daily or weekly seems like asking for trouble.
I would much rather install a STABLE, released version of GHC, something that won't change except maybe a few times a year (or less often) Putting the install into ocpkg seems like a good choice.
"*

Thanks for your comments @linas .I don't know the main reasons why things work this way. I am not a stack developer. All I can say is that, as a Haskell user, we used to have only one option: cabal-install. By default, it installs haskell packages on ~/.cabal folder so is the same situation.
Now, stack is a new option, definitely more confortable, avoiding many issues in packages management.
So, generally there are only 2 options: cabal-install and stack, and stack looks better.

About ghc installation it is not necessary if you have proper ghc version installed.
If you have ghc-7.8.4 installed, "stack setup" won't install it again.
We could spend some time in modifying the script ocpkg to install ghc-7.8.4 in a proper location (such as /usr/local/bin). Something like:
```
wget https://www.haskell.org/ghc/dist/7.8.4/ghc-7.8.4-x86_64-unknown-linux-centos65.tar.xz
tar xpvf ghc-7.8.4-x86_64-unknown-linux-centos65.tar.xz
cd ghc-7.8.4
./configure
sudo make install
```
Going this way, again, new problems! To install it is necessary to have libgmp.so.3. I don't have it, I have libgmp.so.10, so I need to figure out how to deal with this , etc. For example, it looks like stack developers solved this problem in their software. Also, this ghc installation would only work for ubuntu.
Maybe we should let stack do this automatically, at the price of installing ghc under ~/.stack/.... .
We avoid spending time on ghc installation. Time that we can use to improve haskell bindings! :)
As many people are using stack, probably its ghc installation procedure will be far more robust that the one implemented by us by some script file.

----------------------------------------------------------------------------------------------

"
Marco: I wonder what you think of Jacek's suggestion that

***
Consider option 3) that I suggested at first. Instead of manualy executed
steps by the user add this steps to ocpkg or other "install environment
packages" administration script but not cmake.
***

"

Thanks @jswiergo for this suggestion. Where could I find more information on this? I have being reading/modifying the script, is it updated?. I don't understand why it only executes "make install" when "min" and "demo", and not when "local"... [atomspace/blob/master/scripts/ocpkg](https://github.com/opencog/atomspace/blob/master/scripts/ocpkg#L558)
Could someone give me a little more information on this? :)
I found this wiki page: http://wiki.opencog.org/w/Ocpkg   which refers to the script file on the opencog repo, no the one in the atomspace repo.
What is the common procedure to install the atomspace? executing this script?

-------------------------------------------------------------------------------------------------

*"
Try using the STACK_ROOT environment variable, making it use a location in
build tree instead.
"*

Thanks @ferrouswheel but this is not useful, because it means installing every stack related packages and ghc on build dir, and this doesn't seem to be the right approach. You should be able to remove the build dir after installing.

-------------------------------------------------------------------------------------------------

Thanks!
Marcos

ngeiswei:
@linas you wrote
"To have stack install a new version of GHC that changes daily or weekly seems like asking for trouble." but that is not the case, stack will only install the version of GHC associated to the stack package LTS version specified in OpenCog's stack config file.

So on the stability side, using stack setup should be just fine. The actual remaining problems of having stack setup run inside cmake are
1. Network access
2. User disk space consumption

I have no feeling about how bad that is (I don't have Linas' years of experiences). So I'd be tempted to try it and if it turns out to be problematic we'll think of a solution then. I really **hate** to introduce technical debt, but I'm just not convinced it actually is one.

I don't understand how the buildbot works, does someone who understands could tell us if network access and user disk space consumption is bad? If network access really is a problem, then I suggest to have the haskell environment corresponding to the LTS installed on the system, just for the buildbot. For the lambda user, I still think using stack setup is much better, even inside cmake.

ngeiswei:
Let's be logical here: the buildbot need to fetch the last revision of OpenCog, right? So it necessarily needs network access at some point, right? So surely this shouldn't be such a big problem.

AmeBel:
@ngeiswei there is enough space and bandwidth on the machine running buildbot.

ngeiswei:
Thanks @AmeBel. In that case I suggest @MarcosPividori to resubmit the pull request (after taking care of the minor remarks made earlier about opencog path) and to move on. :)

MarcosPividori:
Hi, Ok! I can ensure you that I have spent some time thinking about this and I couldn't find a better option. So, in summary:
- If you don't have stack installed, cmake won't do any haskell related work.
- If you have stack installed:
    - Probably you are using it so, as it is a common behavior of stack to install ghc on user space, you probably won't be annoyed with this behaviour.
    - "stack setup" will install ghc, only the first time, if needed, but, again, if you don't like this, you can install ghc-7.8.4 before on your own (following this instructions: [ghc-7.8.4](https://www.haskell.org/ghc/download_ghc_7_8_4)), and then, when building with cmake, stack won't attempt to install ghc again. So, if you agree, I can clarify this on the README file.

In future, I could add stack installation and move "stack setup" command to the "ocpkg" script. But I would need some information about "ocpkg" because it doesn't seem to be up to date, as I mentioned before. (previous comment)

Thanks!
Marcos

AmeBel:
@MarcosPividori don't worry about ocpkg just let me know the packages to be installed and I will add it.

AmeBel:
Just updated http://wiki.opencog.org/w/Ocpkg

MarcosPividori:
Hi @AmeBel ! Thanks for your response!
So, what I need to include in ocpkg:

- Stack installation: https://github.com/commercialhaskell/stack/wiki/Downloads#ubuntu
- Run "stack setup" = ghc installation

I was working on that, modifying the ocpkg script in the atomspace repository. I paste the diff, maybe it is useful:

```diff
diff --git a/scripts/ocpkg b/scripts/ocpkg
index a76c065..80d3380 100755
--- a/scripts/ocpkg
+++ b/scripts/ocpkg
@@ -127,6 +127,7 @@ PACKAGES_BUILD="
 		libiberty-dev \
 		libicu-dev \
 		libbz2-dev \
+		stack \
 		cython \
 		python-dev \
 		python-zmq \
@@ -492,8 +493,14 @@ trap quit_trap  INT HUP QUIT TERM
 
 # SECTION 4: Function Definitions
 
+add_stack_repository() {
+wget -q -O- https://s3.amazonaws.com/download.fpcomplete.com/ubuntu/fpco.key | sudo apt-key add -
+echo 'deb http://download.fpcomplete.com/ubuntu/trusty stable main' | sudo tee /etc/apt/sources.list.d/fpco.list
+}
+
 add_repositories() {
 MESSAGE="Adding software repositories..." ; message
+add_stack_repository
 for REPO in $REPOSITORIES ; do 
   sudo apt-add-repository -y $REPO
 done
@@ -557,14 +564,19 @@ fi
 
 case $PACKAGE_TYPE in
   min)	MESSAGE="Installing OpenCog..." ; message
- 	make install; exit 0;;
+ 	make install; build_stack_env; exit 0;;
   demo)	MESSAGE="Installing OpenCog..." ; message
-  	make install; exit 0;;
+ 	make install; build_stack_env; exit 0;;
   dev)	exit 0;;
 esac
 
 }
 
+build_stack_env() {
+export STACK_YAML= $LIVE_SOURCE_BRANCH/opencog/haskell/stack.yaml
+stack setup
+}
+
 locate_live_iso() {
 if [ $INPUT_ISO_NAME ]; then 
   MESSAGE="Using Ubuntu image specified at $INPUT_ISO_NAME" ; message
```
Thanks!
Marcos

MarcosPividori:
So, If you add the "stack setup" step to the ocpkg script. I will remove it from cmake before doing a pull request.

AmeBel:
updated the ocpkg to keep it in sync with that found in opencog. https://github.com/opencog/atomspace/pull/175 @MarcosPividori you can include your changes as part of your pull request, just include build_stack_env @ https://github.com/opencog/atomspace/blob/master/scripts/ocpkg#L868 & https://github.com/opencog/atomspace/blob/master/scripts/ocpkg#L856

ngeiswei:
@MarcosPividori you said '"stack setup" will install ghc, only the first time, if needed, but, again, if you don't like this, you can install ghc-7.8.4 before on your own (following this instructions: ghc-7.8.4), and then, when building with cmake, stack won't attempt to install ghc again. So, if you agree, I can clarify this on the README file.' That's a very good point, so I guess ocpkg should install ghc-7.8.4 **on the system**, but then I don't think you need to have stack setup run by ocpkg again, just let it in one place like cmake, for the user who's not using ocpkg (which includes me).

MarcosPividori:
Hi, Thanks for your comments.
I've made a pull request https://github.com/opencog/atomspace/pull/176
Marcos

PR #144, opened 2015-07-14
Fix ure rule config utest

ngeiswei:
Fixes the UREConfigReaderUTest.

However I fear now BackwardChainerUTest is failing, weird...

# Unit test results
The following tests FAILED:
	 63 - ForwardChainerUTest (OTHER_FAULT)
	 64 - BackwardChainerUTest (Failed)

### DIFF
--- a/opencog/scm/rule-engine-utils.scm
+++ b/opencog/scm/rule-engine-utils.scm
@@ -54,23 +54,24 @@
 ; to be sure there is ever only one value associated to that
 ; parameter. The value is automatically converted into string.
 (define (ure-set-num-parameter rbs name value)
-  ; Delete any previous parameter
-  (cog-bind (BindLink
-               (ExecutionLink
-                  (SchemaNode name)
-                  rbs
-                  (VariableNode "__VALUE__"))
-               (DeleteLink
-                  (ExecutionLink
-                     (SchemaNode name)
-                     rbs
-                     (VariableNode "__VALUE__")))))
+  (define (param-hypergraph value)
+    (ExecutionLink
+       (SchemaNode name)
+       rbs
+       value)
+  )
+  (let ((del-prev-val (BindLink 
+                          (param-hypergraph (VariableNode "__VALUE__"))
+                          (DeleteLink
+                             (param-hypergraph (VariableNode "__VALUE__"))))))
+       ; Delete any previous value for that parameter
+       (cog-bind del-prev-val)
+       ; Delete pattern to not create to much junk in the atomspace
+       (purge-hypergraph del-prev-val)
+  )
 
   ; Set new value for that parameter
-  (ExecutionLink
-     (SchemaNode name)
-     rbs
-     (NumberNode (number->string value)))
+  (param-hypergraph (NumberNode (number->string value)))
 )
 
 ; Set (fuzzy) bool parameters. Given an RBS, a parameter name and its
--- a/opencog/scm/rule-engine-utils.scm
+++ b/opencog/scm/rule-engine-utils.scm
@@ -82,7 +82,7 @@
 ;    rbs
 (define (ure-set-fuzzy-bool-parameter rbs name value)
   (EvaluationLink (stv value 1)
-     (SchemaNode name)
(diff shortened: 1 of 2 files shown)


ngeiswei:
Actually the BackwardChainerUTest passes... really strange I run about 10 times, trying to narrow down which test failed, they all pass. I can only conclude that it randomly occasionally fails.

jswiergo:
I have observed similar behaviour in pattern matcher here (#141). Randomness came from ordering incoming sets by pointers and sometimes it missed some branches. May it be same thing?

williampma:
There's always a small chance the BC utest will fail becasue target/rule selection are still random.  You could in theory not able to reach a solution if you are unlucky.

Ideally, for unit testing, I would set a fixed seed so the random is not so random, but I don't see a good way to do that.  If I could use the cogutil's mt19937ar class then maybe I can set a fixed seed (since it creates a singleton), but it doesn't have the functionality I want.  The internal std::mt19937 object got it, but it seems I need to write the equivalent functionality for when `USE_STL_RANDOM` is not defined...

pennachin:
@williampma Fixing the seed of a RNG is the typical pattern for unit tests, and I guess work in extending the mt19937ar class (or picking some standard RNG if that's simpler) would be useful on other cases. @ngeiswei how do the MOSES unit tests handle randomness? Would that be useful in this case?

ngeiswei:
@williampma @pennachin assuming your code relies solely on the singleton randGen() to generate randomness, you can simply do
```c++
randGen().seed(MY_SEED);
```
that is how MOSES handles randomness.

To the user's end, it should make no difference (in average) whether `USE_STL_RANDOM` is enabled of not, hopefully.

williampma:
No, BC wasn't using randGen(), since cogutil's mt19937ar doesn't support discrete_distribution http://www.cplusplus.com/reference/random/discrete_distribution/

It is possible to support it simply using the internal std::mt19937 object, but not the case when `USE_STL_RANDOM` is disabled.

ngeiswei:
Oh, so the BC wouldn't compile for gcc 4.6 or lower, right? Not a big deal (I don't know who is using gcc 4.6 anymore). Anyway, if randGen() doesn't offer this functionality it should be fixed. Let me look at the code...

ngeiswei:
I see 3 ways to go about that

1. Fix the BC so it uses randGen() + roulette_select (see `<COGUTILS>/opencog/utils/select.h` which offers the same functionality as discrete_distribution)
2. Make randGen compatible with STL random generators, so that it can be used in stl::discrete_distribution, etc.
3. Throw randGen away entirely and rely on STL from now on, which I believe offers the same functionalities as randGen and more. I'm not sure about singleton thing, but surely it must that too, or we can add one. Maybe have randGen() return an STL random generator reference instead of our made up random generator.

Ideally, if we had plenty of time, I'd go with 3, but we have other more important things so... what do you think?

williampma:
I wasn't aware of the `roulette_select`.  Let me look at that and maybe use it.

ngeiswei:
Honestly look at that

http://en.cppreference.com/w/cpp/numeric/random

we really ought to drop our random generator and use STL's instead (or at least make it compatible). I'd be happy to take care of that but I'm too busy, maybe after a couple of weeks. I guess we should create an issue in the cogutils repo.

williampma:
That sounds good too.

It's not that high a priority at the moment, so I am fine with option 3.

ngeiswei:
Do you feel like do it?

But first, before deciding to go with 3, I think you should study the requirements of a STL random generator and see if you can easily turn randGen into one (2). If you can it's gonna save you a lot of work cause you wouldn't have to update the randGen use all over the 4 repos.

ngeiswei:
That is you can both support randGen's and STL's specs. It makes a randGen slightly more bloated but again, saves a lot of edit! And allows future people to take advantage of STL cool stuff. Yeah I think I kinda like it...

ngeiswei:
Oh, you say "it's not that high a priority", I read too quickly and read "I'm not that busy I could take care of it" ;-)

PR #145, opened 2015-07-15
Modify get-similr-sentences to also support SentenceNode as the input

leungmanhin:


linas:
??? 

There should not be any scheme code at all in the /opencog/query directory

There should not be any NLP code in the entire atomspace repository.   

Whatever this is, its just wrong.

AmeBel:
This shoudl be moved to https://github.com/opencog/opencog/tree/master/opencog/nlp/sureal

opencog/query/sentence-matching.scm
```diff
@@ -13,19 +15,30 @@
     ; Generate sentences from each of the R2L-SetLinks
     (define (generate-sentences r2l-setlinks) (if (> (length r2l-setlinks) 0) (map to-sureal r2l-setlinks) '()))
 
+    ; Check if the input-sentence is an actual sentence or a SentenceNode,
+    ; parse and return the corresponding SentenceNode if it is a sentence
+    (define (get-sentence-node)
+        (if (cog-atom? input-sentence)
```
AmeBel: maybe you should rename it to `check-sentence-node` or `sentence-node?`

opencog/query/sentence-matching.scm
```diff
@@ -13,19 +15,30 @@
...
         ; Delete identical sentences from the return set
         (delete-duplicates
             ; Use SuReal to generate sentences from their corresponding SetLinks
             (generate-sentences
-                ; TODO: Filter out non-SetLinks here, or do it in (generate-sentences)
                 ; Search for any similar SetLinks in the atomspace
                 (cog-outgoing-set (cog-fuzzy-match
                     ; Get the R2L SetLink of the input sentence
                     (car (cog-chase-link 'ReferenceLink 'SetLink
                         (car (cog-chase-link 'InterpretationLink 'InterpretationNode
```
AmeBel: If there isn't any function that does this already then this should be an independent function. That way should we change the atomic-representaion we don't have to go hunting for every case that does this. Maybe it is too early to do that, i don't know your call.

williampma:
I think `opencog/nlp/scm` is a better place for it.

AmeBel:
Whichever makes sense

leungmanhin:
Yes sure

PR #151, opened 2015-07-16
Remove sentence-matching.scm

leungmanhin:


linas:
How about this: create a new directory https://github.com/opencog/opencog/nlp/nlu and put this file there?

nlu==natural-language-understanding and infrastructure for the new chatbot can go there.

PR #152, opened 2015-07-17
Fix for loading certain types of atoms from backing store

leungmanhin:
While loading certain types of atoms from the backing store, for example, a NumberNode in this form
```
(ListLink
    (NumberNode "1.000000")
)
```
will return an error `AtomTable - Atom in outgoing set isn't known!`.

To solve this problem, we need to explicitly create these types of atoms.


linas:
wow. Good catch.

PR #153, opened 2015-07-17
Fix for loading a link with empty outgoing set from backing store

leungmanhin:
I got an error `AtomTable - Atom in outgoing set isn't known!` while loading a link with empty outgoing set. Check if the outgoing set is empty before creating the handle would fix this problem.


linas:
thanks!

PR #156, opened 2015-07-20
AtomStorage to support PTV

leungmanhin:
Some of the atoms created via the NLP pipeline are using Probabilistic Truth Value, extend the SQL database to support this.

Utest results:


linas:
well, there is also some remaining work for GenericTV, see issue #195

PR #157, opened 2015-07-20
Fchain

misgeatgit:
- Fixed a bug caused by the URE rule base representation change; i.e the forward-chainer used to assume all rules have a root link of type BIND_LINK which is not the case anymore. Currently BIND_LINKS  could be wrapped in MEMBER_LINK. This fix will avoid relying on either such assumptions or rule representation syntax by working directly with the C++ rule objects.

- Treat Node matching same way as Link matching.This was overlooked in the last PR.

### DIFF
--- a/opencog/rule-engine/forwardchainer/DefaultForwardChainerCB.cc
+++ b/opencog/rule-engine/forwardchainer/DefaultForwardChainerCB.cc
@@ -79,6 +79,7 @@ vector<Rule*> DefaultForwardChainerCB::choose_rules(FCMemory& fcmem)
             DefaultImplicator imp(&rule_atomspace);
             imp.implicand = bl->get_implicand();
             bl->imply(imp);
+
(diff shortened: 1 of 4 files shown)


opencog/rule-engine/forwardchainer/DefaultForwardChainerCB.cc
```diff
@@ -240,3 +195,44 @@ HandleSeq DefaultForwardChainerCB::apply_rule(FCMemory& fcmem)
 
     return product;
 }
+
+HandleSeq DefaultForwardChainerCB::get_implicant_seq(Handle implicant)
```
williampma: Better move this to the Rule class.

misgeatgit: Sure, I'll do that.

Issue #127, opened 2015-07-06
cog-delete-recursive documentation isn't correct


ngeiswei:
cog-delete-recursive's documentation isn't exactly correct here

``` scheme
       ; Verify that the link l is gone:
       guile> l
       Invalid handle

       ; Verify that the node x is gone:
       guile> x
       Invalid handle
```

in reality it shows the atoms referenced by l and x

``` scheme
scheme@(guile-user)> l
$8 = (Link
   (ConceptNode "abc")
   (ConceptNode "def")
)

scheme@(guile-user)> x
$9 = (ConceptNode "abc")
```

However the behavior of cog-delete-recursive is correct, using prt-atomspace as given in http://wiki.opencog.org/w/Scheme#Print_all_atoms_in_the_AtomSpace (EDIT: you can use directly cog-prt-atomspace) I can see that the atoms corresponding to l and x are indeed missing.

I'm not sure whether the documentation should be fixed or what is returned by l and x should be fixed.

ngeiswei:
For those who don't know there exists an already coded cog-prt-atomspace, rather than using prt-atomspace for the wiki.

linas:
Yeah, that's kind of weird.  Here is what is happening:

In the good old days, printing checked the UUID, and if it was -1, it printed "invalid handle".   Now, guile holds on to the atom pointer, and so the atom gets printed even if it is not in any atomspace. 

The correct fix is probably to alter the printing to also print (in a comment?) the atomspace that the atom belongs to.

We should add a new command `(cog-atomspace-of  SOME-ATOM)` that returns the atomspace that SOME-ATOM belongs to.

We probably should change `cog-delete` to `cog-remove`  and likewise for recursive, since it doesn't "delete", is simply removes the atom from the atomspace.

linas:
Closing. Fixed in pull req #1341 Normally its considered polite to let the person who reported the bug to verify and close, but I figure Nil has his hands full, so I'm closing on his behalf.

PR #158, opened 2015-07-20
Fchain

misgeatgit:


williampma:
What if someone want to actually use a SetLink as a single source?

misgeatgit:
I don't have any strong reason actually besides recalling to have read some thread weaks ago where @ngeiswei  was talking about using SetLinks for passing var args. Do u think ListLink should be used instead?

williampma:
Maybe it is better to just take a scheme list?  That way it won't conflict with anything that one might want to use as source.

misgeatgit:
@williampma how do I do that?

AmeBel:
```
(define (par-cog-fc list-of-atoms rule-base-conceptnode)
    (par-map cog-fc list-of-atoms (make-list (length list-of-atoms) rule-base-conceptnode))
)
```

misgeatgit:
@AmeBel my question is how to write the C++ interface that accepts the scheme list and convert it in to SetLink. My scheme knowledge is very limited.

williampma:
Just look at https://github.com/opencog/atomspace/blob/master/opencog/guile/SchemePrimitive.h if you need a new primitive type.  It's pretty straight-forward.

AmeBel:
I don't get why you would want to convert it to a SetLink. Does the present implementaion mean that you are defining the focuse set, i.e. , the forward chainer will only run over the outgoing-set of the SetLink and it wouldn't pattern match over other atoms in the atomspace?

ngeiswei:
I forgot who has the right or not to merge. Please add the result of the unit tests so I can merge.

misgeatgit:
@ngeiswei which unit tests are referring to?

ngeiswei:
Cassio has suggested to paste the results of the unit tests in the pull request message till get some better automated unit test checking on pull request. See my last pull request for instance #127.

ngeiswei:
Ooops, wrong issue see #174

williampma:
This PR is already merged... just saying.

misgeatgit:
OK, will do that in the future PRs.

Issue #46, opened 2015-05-19
"Expecting GroundedPredicateNode!" error


williampma:
Empty atomspace, add the following
```
(EvaluationLink (stv 1.000000 0.000000)
  (PredicateNode "sell") ; [8621]
  (ListLink (stv 1.000000 0.000000)
    (VariableNode "$x") ; [8615]
    (VariableNode "$y") ; [8618]
    (VariableNode "$z") ; [8622]
  ) ; [8623]
) ; [8624]
```
Then do
```
(cog-satisfy (VariableNode "$B"))
```
gives me
```
Backtrace:
In ice-9/boot-9.scm:
 157: 10 [catch #t #<catch-closure 3ec3120> ...]
In unknown file:
   ?: 9 [apply-smob/1 #<catch-closure 3ec3120>]
In ice-9/boot-9.scm:
 157: 8 [catch #t #<catch-closure 38a2fe0> ...]
In unknown file:
   ?: 7 [apply-smob/1 #<catch-closure 38a2fe0>]
   ?: 6 [call-with-input-string "(cog-satisfy (VariableNode \"$B\"))\n" ...]
In ice-9/boot-9.scm:
2320: 5 [save-module-excursion #<procedure 3847600 at ice-9/eval-string.scm:65:9 ()>]
In ice-9/eval-string.scm:
  44: 4 [read-and-eval #<input: string 381c5b0> #:lang ...]
  37: 3 [lp (cog-satisfy (VariableNode "$B"))]
In unknown file:
   ?: 2 [opencog-extension cog-satisfy ((VariableNode "$B")
)]
In ice-9/boot-9.scm:
 102: 1 [#<procedure 38a3940 at ice-9/boot-9.scm:97:6 (thrown-k . args)> C++-EXCEPTION ...]
In unknown file:
   ?: 0 [apply-smob/1 #<catch-closure 38a2fa0> C++-EXCEPTION ...]

ERROR: In procedure apply-smob/1:
ERROR: In procedure cog-satisfy: Expecting GroundedPredicateNode! (/home/william/atomspace/opencog/atoms/execution/EvaluationLink.cc:177)
ABORT: C++-EXCEPTION
```

Why is that?

misgeatgit:
But where is Var node $B ?

williampma:
$B is to be mapped to all possible atoms inside 
```
(EvaluationLink (stv 1.000000 0.000000)
  (PredicateNode "sell") ; [8621]
[...]

PR #161, opened 2015-07-22
Pattern Matcher: exact matching evaluatable in non-logical/non-evaluatable/non-executable

williampma:
Very preliminary, just pushing it out here to be reviewed and see if I am missing anything major.  I haven't tested more complicated clauses yet (probably should... maybe tomorrow).

Basically the code avoid evaluating anything unless at the clause level.

Only this test fail:
```
      Start 76: CythonBindlink
76/78 Test #76: CythonBindlink ...................***Exception: Other  0.29 sec
```
which is because 
```
terminate called after throwing an instance of 'opencog::RuntimeException'
  what():  Trying to re-initialize python interpreter with different
AtomSpace ptr! Current ptr=0x161ab40 New ptr=0x7fffdf707a30
 (/home/william/atomspace/opencog/cython/PythonEval.cc:427)
Aborted (core dumped)
```
Apparently the python interpreter does not allow executing python code in a new temporary atomspace? (and the PM do this extensively with `_temp_aspace`)


jswiergo:
I think the message about python interpreter is misleading and it is result of earlier thrown exception not a cause. I've not dig deep but probably it fails on satisfaction_link and this may be the case you are missing. I would check at least all use cases from API here:
https://github.com/opencog/atomspace/blob/master/opencog/query/BindLinkAPI.h

williampma:
I think it is failing on the stop_go green light & red light test.  And since the new route to evaluatable evaluate stuff inside a temp atomspace, the python interpreter failed (I think).

The same stop_go test in scheme works.

williampma:
Found some annoying stuff.  In PM it is allowed for a clause to be just a (VariableNode $A), and if this VariableNode is matched to some evaluatable, the matched evaluatable for some reason is supposed to be evaluated.

This mean a single VariableNode clause should be considered as an evaluatable clause.  This removes it from "fixed_clause" and then the number of components become 0, so the PM is not even ran...

This special treatment of lone VariableNode clause is kind of annoying...

jswiergo:
@williampma you are right. Python evaluator assumes that there is a singleton instance of atomspace. I am not sure why it is like that. It seems this is too strong assumption and should be changed. 

So now this assumption in python evaluator is totally not consistent with DefaultPatternMatcherCB that invoke it in private _temp_aspace. The problem was hidden because python GroundedPredictateNodes were evaluated by post_link_match() method that uses the main atomspace (created by python interpreter). Seems you have moved this evaluation inside clause_match() that uses PM _temp_aspace.

opencog/query/DefaultPatternMatchCB.cc
```diff
@@ -149,18 +149,7 @@ bool DefaultPatternMatchCB::link_match(const LinkPtr& lpat,
...
-	Handle hp(lpat);
-	if (_dynamic->find(hp) == _dynamic->end()) return true;
-
-	// We will find ourselves here whenever the link contain
-	// a GroundedPredicateNode. In this case, execute the
-	// node, and declare a match, or no match, depending
-	// one how the evaluation turned out.  Its "crisp logic"
-	// because we use a greater-than-half for the TV.
-	// This is the same behavior as used in evaluate_term().
-	TruthValuePtr tv(EvaluationLink::do_evaluate(_as, lgnd->getHandle()));
```
jswiergo: Previously evaluated in the main atomspace created by python interpreter.

opencog/query/DefaultPatternMatchCB.cc
```diff
@@ -179,7 +168,20 @@ bool DefaultPatternMatchCB::post_link_match(const LinkPtr& lpat,
...
-	if (ptrn == grnd) return false;
+	if (ptrn == grnd)
+	{
+		// If the whole clause is evaluatable and has no variable, then
+		// we can evaluate here and now before matching the remaining
+		// clauses, like a pre-condition.
+		if (_have_evaluatables && _dynamic->count(ptrn) != 0)
+		{
+			std::map<Handle, Handle> empty_map;
+			return eval_sentence(grnd, empty_map);
```
jswiergo: This is evaluated in _temp_aspace

williampma:
Yes, the python interpreter calls the pattern matcher, and then pattern matcher calls the python interpreter with a new temporary atomspace.

I think evaluating in _temp_aspace is more consistent, since every other instance of evaluation are within the _temp_aspace.  I simply removed the out-liner.

Though I actually finding the concept of allowing executable to be executed while matching is kind of strange.  The executable can produce its own side-effect even if the whole query ended up matching nothing.  There's even a whole unit test (SequenceUTest) that is matching a query with no variable, and the sole purpose is to let the pattern matcher execute some code in sequence.

Also, it's not clear whether the atoms created by matching-time executable should be in the final atomspace or temp atomspace.  If the executable is storing a value that count how many atoms it created in the atomspace, then if the output atoms are put into a temporary atomspace, the value the executable thinks it has produced will be different from how many actual atoms it has produced in the final atomspace.

bgoertzel:
> Though I actually finding the concept of allowing executable to be
> executed while matching is kind of strange.
>

Strange, perhaps, but definitely intentional....  The PM is now being
viewed as not merely a pattern matcher, but also an interpreter of
"programs" that are implemented as collections of Atoms...


> The executable can produce its own side-effect even if the whole query
> ended up matching nothing. There's even a whole unit test (SequenceUTest)
> that is matching a query with no variable, and the sole purpose is to let
> the pattern matcher execute some code in sequence.
>

True...


> Also, it's not clear whether the atoms created by matching-time executable
> should be in the final atomspace or temp atomspace. If the executable is
> storing a value that count how many atoms it created in the atomspace, then
> if the output atoms are put into a temporary atomspace, the value the
> executable thinks it has produced will be different from how many actual
> atoms it has produced in the final atomspace.
>

In many cases we definitely want those Atoms to end up in the main
Atomspace....  But there may be cases where you'd want to pipe the output
Atoms to a special output Atomspace, I dunno...

-- Ben

williampma:
> Strange, perhaps, but definitely intentional.... The PM is now being viewed as not merely a pattern matcher, but also an interpreter of "programs" that are implemented as collections of Atoms...

I see.  How about the case when a top level clause is a single VariableNode, and it matches to an EvaluationLink with GroundedSchemaNode, that EvaluationLink automatically turned into a matching-time evaluatable?  Like being done for this unit test https://github.com/opencog/atomspace/blob/master/tests/query/eval-var.scm

It seems there will be potential for a EvaluationLink with GroundedSchemaNode to be randomly evaluated by some random unrelated pattern matching that happens to have a VariableNode as a top clause, if the other clauses in the query is not restrictive enough.

williampma:
Updated this a little bit.

1. Removed a leftover code that is incorrectly marking evaluatable in the new design
2. Make sure a lone `(VariableNode "$A")` clause is marked as "evaluatable/virtual" since it could be mapped to a EvaluationLink with GroundedSchemaNode, which PM will want to evaluate during matching-time.
3. As a result, needed to allow the only clause that is virtual to still work
4. To solve the above, adds any unique VariableNode as a fixed clause (so a lone VariableNode clause is now both fixed and virtual)
5. So had to remove one of the OC_ASSERT check

Doesn't solve #165 and #171, becasue of #172.  However, this fixes #46, which is my goal to allow me to continue on the BackwardChainer.

Again, all unit tests passed except for CythonBindLink.

```
99% tests passed, 1 tests failed out of 78

Total Test time (real) =  74.03 sec

The following tests FAILED:
	 76 - CythonBindlink (OTHER_FAULT)
```

bgoertzel:
On Fri, Jul 24, 2015 at 3:09 AM, Ben Goertzel <ben@goertzel.org> wrote:

>
> Though I actually finding the concept of allowing executable to be
>> executed while matching is kind of strange.
>>
>
> Strange, perhaps, but definitely intentional....  The PM is now being
> viewed as not merely a pattern matcher, but also an interpreter of
> "programs" that are implemented as collections of Atoms...
>
>
>> The executable can produce its own side-effect even if the whole query
>> ended up matching nothing. There's even a whole unit test (SequenceUTest)
>> that is matching a query with no variable, and the sole purpose is to let
>> the pattern matcher execute some code in sequence.
>>
>
> True...
>
>
>> Also, it's not clear whether the atoms created by matching-time
>> executable should be in the final atomspace or temp atomspace. If the
>> executable is storing a value that count how many atoms it created in the
>> atomspace, then if the output atoms are put into a temporary atomspace, the
>> value the executable thinks it has produced will be different from how many
>> actual atoms it has produced in the final atomspace.
>>
>
> In many cases we definitely want those Atoms to end up in the main
> Atomspace....  But there may be cases where you'd want to pipe the output
> Atoms to a special output Atomspace, I dunno...
>
>
Yes. All this does sketch the general intent. Yes, certain details are
confusing, and it is sometimes not clear what the "correct semantics" or
what the "best way of doing things" is.  If/when you see a better way of
doing things ... let us know; we got to here by "following our noses" and
doing what seemed like the right thing to do.

opencog/atoms/bind/PatternLink.cc
```diff
@@ -45,13 +45,6 @@ void PatternLink::common_init(void)
 	                 _fixed, _virtual, _pat.black);
 	_num_virts = _virtual.size();
 
-	// unbundle_virtual does not handle connectives. Here, we assume that
-	// we are being run with the DefaultPatternMatchCB, and so we assume
-	// that the logical connectives are AndLink, OrLink and NotLink.
-	// Tweak the evaluatable_holders to reflect this.
-	std::set<Type> connectives({AND_LINK, OR_LINK, NOT_LINK});
-	trace_connectives(connectives, _pat.clauses);
```
linas: Removing this looks wrong.  I made the connectives configurable, because different logic would use different connectives. For example, for probability theory, the connectives would be SET_UNION_LINK, SET_INTERSECTION_LINK, and SET_COMPLEMENT_LINK being the analogs of boolean OR, AND, NOT.  Other logics would use different connectives. 

Because PLN has a "special" interlretation of AndLink, OrLink, NotLink, then a better solution that avoids conflich with PLN would be to change the above to BOOLEAN_AND_LINK, BOOLEAN_OR_LINK, BOOLEAN_NOT_LINK, to make it clear that crsip-logic is being used to combine the results of evaluating terms.

opencog/atoms/bind/PatternLink.cc
```diff
@@ -482,9 +475,58 @@ void PatternLink::unbundle_virtual(const std::set<Handle>& vars,
...
 	for (const Handle& clause: clauses)
 	{
-		bool is_virtual = false;
 		bool is_black = false;
 
+		// Check if all links from top to bottom are either VirtualLink,
+		// logical links, EvaluationLink with GroundedPredicateNode, or
+		// ExecuationOutputLink; cause otherwise none of the evaluatable
+		// can be evaluated and must be matched directly.
+		std::function<bool (const Handle&)> is_virtualable = [&](const Handle& h)
```
linas: well, but this is what "trace_connectives" did; you re-wrote it, but made it less general ... 

Again, different logics, e.g. the kripke logics, or anything based on probability, are going to use different connectives, and we should allow these general cases to be handled.

A little bit off-topic, but treating PLN so that is was more kripke-like might make forward and backward chaining easier to implement: at each time-step, one has a different universe of things that can be "true" (follow from inference).

opencog/atoms/bind/PatternLink.cc
```diff
@@ -506,14 +548,8 @@ void PatternLink::unbundle_virtual(const std::set<Handle>& vars,
 		{
 			_pat.evaluatable_terms.insert(sh);
 			add_to_map(_pat.in_evaluatable, sh, sh);
-			// But they're virtual only if they have two or more
-			// unquoted, bound variables in them. Otherwise, they
-			// can be evaluated on the spot.
```
linas: Removing this comment is wrong, and makig everything virtual, below, is wrong.    Consider, for example, `GreaterThanLink`, and we wish to evaluate whether it is true or not.  There are three cases:

`if 1 < 2` -- this can be evaluated immediately, there are no variables at all.

`if 1 < x` -- this can be evaluated as soon as a grounding for variable x is found.  You do not need any of the complicated virtual-link code to determine if this is true or not; you only need the grounding for x.

`if x < y` -- this is a true "virtual link" that requires the _temp_atomspace and the other complicated machinery.  The problem is that every different possible grounding for x must be compared to every different possible grounding for y.   Terms such as this cannot be evaluated immediately; they must be saved for last, only after each different possible x and y have been determined.  Only this last form are really, truly "virtual", in that they require the complex virtual-code to evaluate them.

opencog/atoms/bind/PatternLink.cc
```diff
@@ -549,50 +582,50 @@ void PatternLink::unbundle_virtual(const std::set<Handle>& vars,
...
-			}
 		}
 		for (const Handle& sh : feol.holders)
 			_pat.executable_holders.insert(sh);
 
-		if (is_virtual)
-			virtual_clauses.push_back(clause);
-		else
-			fixed_clauses.push_back(clause);
+		virtual_clauses.push_back(clause);
```
linas: see comments above; not everything should be virtual.

opencog/atoms/bind/PatternLink.cc
williampma: Removal of this was to allow this https://github.com/opencog/atomspace/issues/46#issuecomment-107873361 to work.

If types like SET_UNION_LINK, SET_INTERSECTION_LINK, and SET_COMPLEMENT_LINK should be used instead of AND_LINK, OR_LINK, etc, then I guess the "is_virtualable" function can be changed to look for them.

opencog/atoms/bind/PatternLink.cc
williampma: Not 100% the same (even if it is less general).  The new method check to make sure only clauses that are truly virtual will be marked as is, so if an evaluatable is under a link that accepts atom instead of TV, then the clause is not virtual.

opencog/atoms/bind/PatternLink.cc
williampma: I don't think you can evaluatable anything on the spot unless it's virtual to the top.  Otherwise it is supposed to be matched as it.

ie.  `1 < 2, 1 < x, x < y` could all potentially be matched as is and should not be evaluated on the spot without looking at what's above it.

opencog/atoms/bind/PatternLink.cc
williampma: Not all are virtual.  Some are marked as "fixed" on line https://github.com/opencog/atomspace/pull/161/files#diff-aafb2f2e1b6e1603ab489430cefcc6fbR526

opencog/atoms/bind/PatternLink.cc
williampma: Doing it this way (ie. not evaluate on the spot) also allow the same atom `x < y` to appear in multiple clauses, so that in one clause it is virtual and should be evaluated, and in the other it should be matched as is.  This is test in the new unit test.

Issue #164, opened 2015-07-23
fail to compile BasicSaveUTest


AmeBel:
```
[ 51%] Building CXX object tests/persist/sql/CMakeFiles/BasicSaveUTest.dir/BasicSaveUTest.cpp.o
In file included from /atomspace/build/tests/persist/sql/BasicSaveUTest.cpp:24:0:
/atomspace/tests/persist/sql/BasicSaveUTest.cxxtest:36:42: fatal error: opencog/nlp/types/atom_types.h: No such file or directory
 #include <opencog/nlp/types/atom_types.h>
                                          ^
compilation terminated.
make[3]: *** [tests/persist/sql/CMakeFiles/BasicSaveUTest.dir/BasicSaveUTest.cpp.o] Error 1
make[2]: *** [tests/persist/sql/CMakeFiles/BasicSaveUTest.dir/all] Error 2
make[1]: *** [CMakeFiles/tests.dir/rule] Error 2
make: *** [tests] Error 2
```
I think this has to be moved to opencog, and add scheme binding test in its place, right ? @linas

AmeBel:
reason I am suggestining moving is because it depends on the cogserver and nlp atom-types.

jswiergo:
This is atom storage test, so it seems to me atomspace repo is good place for it. The dependency on cogserver and nlp-types looks quite easy to cut. I will fix it when my time allows.

AmeBel:
thanks

jswiergo:
done #167

PR #166, opened 2015-07-24
move back scm nlp module to opencog repository

AmeBel:
since there isn't any nlp code here
## Unit Test Result
after disabling BasicSaveUTest and PersistUTest b/c of https://github.com/opencog/atomspace/issues/164#issuecomment-124470493
```
99% tests passed, 1 tests failed out of 77

Total Test time (real) =  49.23 sec

The following tests FAILED:
	 70 - PythonEvalUTest (Failed)
```

### DIFF
--- a/opencog/scm/opencog/nlp/lg-dict.scm
+++ b/opencog/scm/opencog/nlp/lg-dict.scm
@@ -1,7 +1,0 @@
-;
-; OpenCog NLP lg-dict module
-;
-
-(define-module (opencog nlp lg-dict))
-
-(load-extension "liblg-dict" "opencog_nlp_lgdict_init")
--- a/opencog/scm/opencog/nlp/sureal.scm
+++ b/opencog/scm/opencog/nlp/sureal.scm
@@ -1,7 +1,0 @@
-;
-; OpenCog NLP SuReal module
-;
-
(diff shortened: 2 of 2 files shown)


ngeiswei:
As Cassio has recently suggested it would be nice if you include the result of the unit tests. Even if it seems obviously unrelated to the unit test set, cause the obviousness might not be shared across everybody. Thanks!

ngeiswei:
Oh, it's been reverted! I guess it should be closed then.

AmeBel:
I think I have missed what Cassio suggested.  Would you please elaborate ?

ngeiswei:
See the thread on the opencog list entitled "Keeping Master branches stable".

AmeBel:
oh, well the the unit-tests aren't passing 100% for me. Same as buildbot

ngeiswei:
OK, you just have to paste here the utests results (see my last pull request for instance #174). Then it's up to you (or whoever wants to merge) to merge, but at least we know exactly what is the state of the utests.

AmeBel:
@ngeiswei have updated the description

AmeBel:
@linas is there any reason for having the scm module definition here?

ngeiswei:
I've got that Python utest fail as well. I'm merging.

PR #167, opened 2015-07-24
Fix BasicSaveUTest

jswiergo:
This fixes broken BasicSaveUTest that is run if you have sql storage configured. It is partial fix. There is one more sql persistence test (PersistUTest) broken since atomspace repo was separated. It does not compile but only if you have sql persistance enabled.

Test results when PersistUTest is uncommented:
```
...
      Start 21: BasicSaveUTest
21/78 Test #21: BasicSaveUTest ...................   Passed    1.52 sec
...
```
```
```


jswiergo:
Fixed PersistUTest as well. Now all persistence tests pass.
```
```

linas:
thanks

Issue #75, opened 2015-06-02
"SatisfactionLink with no VariableNode" crashes


williampma:
I assume SatisfactionLink is still a thing (kind of lost track with the recent changes).

I tried doing this weird call
```
(cog-satisfy (SatisfactionLink (ConceptNode "ABC")))
```
and cogserver crashed with
```
[2015-06-02 03:52:50:762] [ERROR] Caught signal 11 (Segmentation fault) on thread 140036640466688
	Stack Trace:
	2: basic_string.h:539	~basic_string()
	3: CogServerMain.cc:79	_Z7sighandi()
	4: ??:0	killpg()
	5: PatternMatchEngine.cc:948	  opencog::PatternMatchEngine::explore_link_branches(opencog::Handle const&, opencog::Handle const&, opencog::Handle const&)
	6: InitiateSearchCB.cc:286	  opencog::InitiateSearchCB::neighbor_search(opencog::PatternMatchEngine*)
	7: InitiateSearchCB.cc:481	  opencog::InitiateSearchCB::disjunct_search(opencog::PatternMatchEngine*)
	8: PatternMatch.cc:341	  opencog::PatternLink::satisfy(opencog::PatternMatchCallback&) const
	9: shared_ptr_base.h:779	__shared_ptr()
	10: SchemeModule.cc:41	  opencog::FunctionWrap::prapper(opencog::Handle)
	11: shared_ptr_base.h:545	~__shared_count()
	12: SchemePrimitive.cc:158	  opencog::PrimitiveEnviron::do_call(scm_unused_struct*, scm_unused_struct*)
	13: ??:0	scm_vm_engine()
	14: ??:0	scm_call_1()
	15: ??:0	scm_vm_engine()
	16: ??:0	scm_call_3()
	17: ??:0	scm_vm_engine()
	18: ??:0	scm_call_4()
	19: SchemeEval.cc:606	  opencog::SchemeEval::do_eval(std::string const&)
	20: SchemeEval.cc:535	  opencog::SchemeEval::c_wrap_eval(void*)
	21: ??:0	scm_at_abort()
	22: ??:0	scm_vm_engine()
	23: ??:0	scm_call_4()
	24: ??:0	scm_at_abort()
	25: ??:0	scm_c_with_continuation_barrier()
	26: ??:0	GC_call_with_gc_active()
	27: ??:0	scm_current_processor_count()
	28: ??:0	GC_call_with_stack_base()
	29: ??:0	scm_with_guile()
[...]

PR #168, opened 2015-07-24
Fix + unit test for PM crash when VariableNode is absent (#75)

jswiergo:
```
100% tests passed, 0 tests failed out of 77
```


williampma:
Oh, so `_starter_term` is UNDEFINED and it gets passed down.  But why would `_starter_term` be allowed to be UNDEFINED in the first place...?  Even the dbgprt on https://github.com/opencog/atomspace/blob/master/opencog/query/InitiateSearchCB.cc#L269-L270 is expecting it... weird.

ngeiswei:
Anybody feels confident enough to merge that one? I don't personally.

linas:
thanks!

Issue #148, opened 2015-07-15
Pattern matcher branch exploration problem


jswiergo:
There is a bug in pattern matcher for queries having common atoms shared by different sub-patterns (sub-tree branches). See example:

``` scheme
(define (SharedAtom)
        (AndLink
                (ConceptNode "A")
        )
)

(define (Tree)
        (ListLink
                (SharedAtom)
                (ListLink
                        (SharedAtom)
                        (ConceptNode "B")
                )
        )
)
(Tree)

(define (query1)
        (AndLink
                (VariableNode "$a")
        )
)

(define (query2)
        (VariableNode "$a")
)

(define (tree-query input-query)
        (BindLink
                (VariableList
                        (VariableNode "$a")
                        (VariableNode "$b")
                )
                (AndLink
                        (ListLink
                                (input-query)
                                (ListLink
                                        (input-query)
                                        (VariableNode "$b")
                                )
                        )
                )
                (ListLink
                        (VariableNode "$a")
                        (VariableNode "$b")
                )
        )
)

(cog-bind (tree-query query1))
(cog-bind (tree-query query2))
```

The answer is:

``` scheme
$2 = (SetLink
   (ListLink
      (ConceptNode "A")
      (ConceptNode "B")
   )
   (ListLink
      (ConceptNode "A")
      (ConceptNode "B")
   )
)

$3 = (SetLink
   (ListLink
      (AndLink
         (ConceptNode "A")
      )
      (ConceptNode "B")
   )
)
```

The answer for query1 is duplicated. The right size of satisfying set is 1 for both queries. This is simple example but this may bring weird problems for more complicated settings.

I have checked that it happens when there exists an atom shared by sub-patterns and moreover the main pattern is explored recursively bottom-up. Then all branches are explored.
The difference between this two queries is because the pattern matcher uses "choose thinnest term strategy" and finds as starters different terms. 
For query1 the starter is a bottom atom:

``` scheme
Start term is: (AndLink (stv 1,000000 0,000000)
  (VariableNode "$a") ; [7]
) ; [10]
```

For query2 the starter is a top atom:

``` scheme
Start term is: (ListLink (stv 1,000000 0,000000)
  (VariableNode "$a") ; [7]
  (ListLink (stv 1,000000 0,000000)
    (VariableNode "$a") ; [7]
    (VariableNode "$b") ; [8]
  ) ; [14]
) ; [19]
```

[...]

PR #169, opened 2015-07-26
[WIP] Pattern Matcher: traverse trees of unique terms instead of graph with common atoms

jswiergo:
Just to inform you I am close to finish implementation of solution to the problem of pattern matcher missing some branches when queries contain replicated atoms (#148). I run out of time now and will back later.

The idea of solution is to preprocess pattern query and build internal tree structure of unique terms. Each term corresponds to one atom handle. If the same atom appears many times in the query then it corresponds to many unique terms, one for each occurrence. Having that structures, pattern matcher traverses the pattern tree exactly as before but uses term pointers instead of atom handles to know its exact position. Then it keeps states of unordered links and choice links in the language of term pointers (e.g. permutation of unique pointers instead of permutation of not unique handles).

If you can review present changes or give some other hints you are more then welcome.


linas:
I took a quick look, and it all seems quite plausible.  explore_term_branches is a little confusing. I think I understand why its needed and what it is used for, but additional notes to illustrate it could be good.

Any reason why this cannot be merged yet?

jswiergo:
Thanks for suggestion. Some tests does not pass yet. I should be able to continue in a few days.

jswiergo:
Thanks for info, it's clear. This branch is experimental and the pull request can be dropped. I will be pushing separate changes.

PR #176, opened 2015-07-28
Replace Cabal by Stack and update ocpkg script.

MarcosPividori:
Hi,
I updated the code of the ocpkg script. It will install stack, and also attempt to install proper ghc version globally.
Thanks,
Marcos


ngeiswei:
I can't find where ocpkg is currently being maintained, the following repo https://github.com/opencog/ocpkg seems way to outdated. Any idea?

ngeiswei:
I see, ocpkg is directly in the atomspace repo under the directory scripts.

But the opencog repo also contains suck a script. The wiki points to the version on the opencog repo  for Ubuntu 14.04, and point to the version on the ocpkg repo for Ubuntu 12.04.

Nothing points to the atomspace repo version. This is quite confusing. Why ocpkg hasn't been updated on the ocpkg? It seems the natural place to have it, doesn't it?

ngeiswei:
It seems the ocpkg version on the atomspace repo is not maintained, it was lastly edited by Linas on 2015-04-27 while the opencog version was lastly edited on 2015-06-29 by Amen. So it seems the official currently maintained ocpkg is on the opencog repo. Unless multiple ocpkg exist for each repo, but I don't think that is the case.

So, again, why not using the ocpkg repo, since OpenCog is now split into multiple repo it seems really better?

Any idea?

pennachin:
I'm brainstorming here, but I can see three possibilities for ocpkg: its own repo; the cogutils repo (since everything depends on it); the opencog repo (since it depends on everything). I agree having it sorta duplicated in two repos isn't ideal...

AmeBel:
I had updated ocpkg found here for this pull request. Refer to https://github.com/opencog/atomspace/pull/175

I had been thinking having it in cogutils and cleaning ocpkg and adding some more features, so as to install  it as a tool. That way it will be the single tool for setting workspaces, update installations, setup ci contianers and stuff. Maintaining the seperate repo also works.

ngeiswei:
I think having it in the ocpkg repo makes more sense than having it in the cogutils repo. I don't understand why it's been moved to the opencog repo in the first place (although it was more sensible back then when the opencog repo wasn't split into cogutils/moses/atomspace/opencog).

MarcosPividori:
Hi everyone! Any new on this? I would like to make another pull request on top of this.

ngeiswei:
I'm merging, but the ocpkg change should be done on the right repo (and I don't know which is the right one).

MarcosPividori:
Thanks!

ngeiswei:
@MarcosPividori the ocpkg script on that repo is no longer used (see #186), could you apply your change the one in the ocpkg repository? Thanks.

MarcosPividori:
Hi @ngeiswei ! AmeBel has already introduced my changes to the ocpkg repository, inside this commit: https://github.com/opencog/ocpkg/commit/4b7ea984fc6de91417cbd1859a0c3955d2b74a59
Marcos

AmeBel:
@MarcosPividori The issue raised by https://github.com/opencog/atomspace/pull/143#issuecomment-121154675 is still there. ocpkg doesn't deal with the installation of dependendencies required for the project cmake does. I think  it would be best if ocpkg, similar to what is done for python https://github.com/opencog/ocpkg/blob/master/ocpkg#L563. Just a suggestion, I don't know what the haskalnists way of doing thing :-)

PR #178, opened 2015-07-30
Fix finding starter_term when clause is a Node (#75)

jswiergo:
This is continuation of #168 which @linas merged already just while I was writing this message.

@williampma: The _starter_term should not be passed UNDEFINED to PatternMatchEngine in my opinion. It does not make sense for me. Why do I need to ask PME to match UNDEFINED clause/term when it is obvious that the answer is "it matches nothing and there is nothing to search". That is why I added protection in PME but not assertion.

I have checked your suggestion deeper and found out that the only case when the starter becomes
UNDEFINED is for clauses being simple nodes. InitiateSearchCB::find_starter was missing that case.
I think this was wrong. It should just find the node as starter and pass it down to PME. It is justified
because then PME calls PatternMatchCallback::match_node and this callback has a chance to accept the node or not. So I have added a fix to find_starter method in this commit.

### DIFF
--- a/opencog/query/InitiateSearchCB.cc
+++ b/opencog/query/InitiateSearchCB.cc
@@ -97,9 +97,29 @@ InitiateSearchCB::InitiateSearchCB(AtomSpace* as) :
 // If no constant is found, then the returned value is the undefnied
 // handle.
 //
+
 Handle
 InitiateSearchCB::find_starter(const Handle& h, size_t& depth,
-                               Handle& start, size_t& width)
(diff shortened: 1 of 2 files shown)


jswiergo:
```
100% tests passed, 0 tests failed out of 77
```

williampma:
Ah yes.  I think not allowing it to be UNDEFINED in the first place makes more sense.

linas:
seems I forgot to merge this long ago; sorry.

PR #179, opened 2015-07-31
Rule substitution implementation

misgeatgit:
I have implemented part of rule substitution by source when there is matching rule found. Previously, it
only finds a rule that matches and passes it to the rule selection phase. Now, it also derives a new BindLink/Rule by substituting part of the matched rule with the source. This will boost the efficiency of pattern matching since fewer variables are searched for.  

**Next tasks:**

  -Unit test for substitute_rule_part method.

  -Refactor rule_match method since now its also handling grounding i.e making it more elegant.

  -Execute if implicant is fully grounded and is executable.

  -Store to result list if implicant is fully grounded while rule matching.
   
 **Test report**

Total Test time (real) =  85.48 sec
[100%] Built target test

### DIFF
--- a/opencog/rule-engine/forwardchainer/DefaultForwardChainerCB.cc
+++ b/opencog/rule-engine/forwardchainer/DefaultForwardChainerCB.cc
@@ -33,6 +33,7 @@
 #include <opencog/guile/SchemeSmob.h>
 #include <opencog/rule-engine/URECommons.h>
 
+#include "VarGroundingPMCB.h"
 #include "DefaultForwardChainerCB.h"
 
 using namespace opencog;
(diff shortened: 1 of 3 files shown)


opencog/rule-engine/forwardchainer/DefaultForwardChainerCB.cc
```diff
@@ -56,47 +60,58 @@ DefaultForwardChainerCB::DefaultForwardChainerCB(AtomSpace& as,
...
+                        fv.search_set(it.first);
+
+                        HandleSeq new_rules = substitute_rule_part(
+                                temp_pm_as,
+                                temp_pm_as.add_atom(rule->get_handle()),
+                                fv.varset, gcb.var_groundings);
+                        for (const auto& nr : new_rules)
+                            if (find(new_rules.begin(), new_rules.end(),
+                                     nr) == new_rules.end())
+                                new_rules.push_back(nr);
```
williampma: I am trying to understand what this loop is doing...  

So, for each Handle `nr` in `new_rules`, check if `nr` is **not** in `new_rules` and add `nr` into `new_rules`...? I don't get it.

misgeatgit: Good catch, I think I missed that while refactoring. The local variable new_rules should be named to sth else. There is a global new_rules where new rules are placed.

PR #181, opened 2015-08-03
Assorted updates for the fuzzy pattern matcher

leungmanhin:
- Extended it so that we have more control over the type of atoms being return, and what kind of atoms to exclude
- Added rareness estimation
- Removed some old logic and also updated the unit test

UTest results:


linas:
so when will we hook up the fuzzy matcher to the NLP pipeline?

leungmanhin:
Rodas has hooked up the fuzzy matcher to the chatbot (https://github.com/opencog/opencog/tree/master/opencog/nlp/chatbot) so that it is possible try it on the IRC channel. By the way, it would be nice if you have time to also review and merge this PR (https://github.com/opencog/opencog/pull/1696/files) before trying it :)

Issue #74, opened 2015-06-01
DefineLink full support


ngeiswei:
We're getting close to being able to use Atomese as a real programming language. It seems we still need to add (full) support for DefineLink.

I think the following wiki page

http://wiki.opencog.org/w/Definelink#SchemaNode

correctly defines what we want.

Do we agree on that?

Or do you think PutLink/GetLink can be substituted to DefineLink?

ngeiswei:
## Subtasks
- [X] Generalize DefineLink to handle constants (addressed in PR #185)
- [ ] Add support for definition variables directly in the DefineLink in the case of function/predicate/etc definitions

linas:
The contents of http://wiki.opencog.org/w/Definelink at this moment seem OK.  However ...

The place where DefineLink becomes interesting is in the pattern matcher: what happens when a pattern has define links in it? Should these be searched verbatim?  Should these be understood as defining sub-patterns to be searched for? If a patten is large and complex, it sure would be nice to build it out of smaller, simpler sub-patterns defined in a DefineLink.

Tangled up in the above is the notion of "beta reduction" aka "substitution" aka "composition": all three of these terms refer to the same idea:  given some expression containing x, substitute the value a for all occurrences of x (but do NOT do any further evaluation/execution; just substitute) -- this gets interesting if "a" is a the thing that is defined.

So, for example: the ExecutionOutputLink/SchemaNode example has two steps: a beta-reduction step, where the values 2 and 4 are substituted for X and Y, and then an execution step, where 2+ 4*10 is computed numerically.  It would be nice to have a BetaRedexLink or SubstitutionLink or CompisitionLink (or even a modified PutLink!!??) that only does the beta-reduction, and nothing else.

There are more tangles.  So, the SchemaNode example "works" because we "know" that, whenever we see a SchemaNode, that it was previously "defined" with a DefineLink (and that it is an error if a DefineLink cannot be found).    But what about the PredicateNode "MyPredicate" example?  So, every time that we see a PredicateNode, should we try to see if it is behing a DefineLink?  This would be very CPU-wasteful...  This suggests that the only things that can be "defined" are SchemaNodes, and maybe "DefinedPredicateNode"s and maybe, ummm.  "DefinedPatternNode" (for defining search patterns??)

[...]

PR #185, opened 2015-08-04
Fix define link

ngeiswei:
## Description
Attempt to generalize DefineLink so that it can be used for anything not just in the context of the pattern matcher. It is still under the directory `opencog/atoms/bind`, maybe it should be moved elsewhere (like in `opencog/atoms/core`)

* Modify DefineLink class so that the definition body can be anything (not only a ScopeLink)
* Add a static method DefinieLink::get_definition(const Handle& alias) that return the definition body given a definition alias as in (DefineLink alias body)
* Simplify BetaRedexLink. Warning it assumes the definition body is a ScopeLink, I didn't change that, ultimately I guess it should be able to handle non ScopeLink (constants).

I strongly suspect BetaRedexLink and ExecutionOutputLink (or ApplyLink/ApplicationLink as we might soon rename it) are the same thing and should be unified.

## UTest Results
```
```

### DIFF
--- a/opencog/atoms/bind/DefineLink.cc
+++ b/opencog/atoms/bind/DefineLink.cc
@@ -36,23 +36,17 @@ void DefineLink::init(const HandleSeq& oset)
 		throw InvalidParamException(TRACE_INFO,
 			"Expecting name and definition, got size %d", oset.size());
 
-	// The name must not have been previously defined before.
-	HandleSeq ename;
-	oset[0]->getIncomingSetByType(std::back_inserter(ename), DEFINE_LINK);
-	if (0 < ename.size())
-		throw InvalidParamException(TRACE_INFO,
-			"This is already defined; remove before redfining!");
+	_alias = oset[0];
 
-	_definition = ScopeLinkCast(oset[1]);
-	if (NULL == _definition)
-		_definition = createScopeLink(*LinkCast(oset[1]));
-	if (NULL == _definition)
-	{
-		Type t = oset[1]->getType();
-		const std::string& tname = classserver().getTypeName(t);
-		throw InvalidParamException(TRACE_INFO,
-			"Expecting a ScopeLink, got %s", tname.c_str());
-	}
+	// The name must not have been previously defined before.
+	IncomingSet defs = _alias->getIncomingSetByType(DEFINE_LINK);
+	for (LinkPtr def : defs)
+        if (def->isSource(_alias))
+	        throw InvalidParamException(TRACE_INFO,
+	                                    "This is already defined; "
+	                                    "remove before redfining!");
+
+	_definition = oset[1];
 }
 
 DefineLink::DefineLink(const HandleSeq& oset,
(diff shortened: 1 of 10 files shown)


ngeiswei:
@linas if you could have a look at it that would be cool. Otherwise I'll merge it myself in 48h. Thanks.

ngeiswei:
This pull request partially (or wholly) addresses issue #74.

Also, in the case of a function definition, we may want to add support for defining the variables directly in the DefineLike, that is instead of
```atomese
DefineLink
   SchemaNode "f"
   LambdaLink
      VariableNode "X"
      PlusLink
          VariableNode "X"
          NumberNode "1"
```
one would write
```atomese
DefineLink
   VariableNode "X"
   SchemaNode "f"
   PlusLink
      VariableNode "X"
      NumberNode "1"
```
Similarly to Haskell as one may define f using 2 forms
```haskell
f = \x -> x+1
f x = x+1
```
The practical benefit is to save a LambdaLink.

linas:
The BetaRedexLink and the ExecutionOutputLink are definitely NOT the same thing.  See comments on issue #74 to see why.

Note also: there are a bunch of other issues tangled up that need to be solved, again see comments on issue #74 to see what these are.   Until they are resolved, the DefineLink will be kind-of crippled.

opencog/atoms/bind/DefineLink.cc
```diff
@@ -84,4 +78,23 @@ DefineLink::DefineLink(Link &l)
 	init(l.getOutgoingSet());
 }
 
+Handle DefineLink::get_definition(const Handle& alias) {
+	// Get all DefineLinks associated with that alias, beware that it
+	// will also return DefineLink with that alias as definition body.
+    IncomingSet defs = alias->getIncomingSetByType(DEFINE_LINK);
+
+    // Return the first (supposedly unique) definition
```
linas: you can throw an error if defs.size() > 1

opencog/atoms/bind/DefineLink.cc
```diff
@@ -84,4 +78,23 @@ DefineLink::DefineLink(Link &l)
...
+    for (LinkPtr defl : defs) {
+	    DefineLinkPtr def(DefineLinkCast(defl->getHandle()));
+	    if (def->get_alias() == alias)
```
linas: it would be really bizarre if this equality check every failed ... it would be a weird internal bug.

linas:
Yes, OK about collapsing DefineLink and LambdaLink.  However: study Haskell less, and prolog and MUMPS more. the atomspace is very unlike haskell, and much more similar to prolog/mumps.

opencog/atoms/bind/DefineLink.cc
ngeiswei: No, consider the following case
```atomese
DefineLink
   ConceptNode "B"
   ConceptNode "A"
DefineLink
   ConceptNode "C"
   ConceptNode "B"
```
although B is used only once as alias for something else, defs.size() would be 2, that is why I need to go through each of them and check if the alias matches.

opencog/atoms/bind/DefineLink.cc
ngeiswei: I don't think so, it would merely raise the exception "Cannot find defined hypergraph for atom %s".

ngeiswei:
OK, I'll take the time to revive my understanding of Prolog (and look into mumps as well, sounds like shumps, maybe it's fun).

opencog/atoms/bind/DefineLink.cc
linas: ah, right!

opencog/atoms/bind/DefineLink.cc
linas: yeah, but how could that ever happen?    There's no way to construct the define without the definition, and there is no way to erase the definition without also erasing the define link.  Right?

linas: well, nevermind, having this check is harmless, because stranger things have happened before.

ngeiswei: It's actually necessary for cases I've given above when the alias is being used as definition body by another alias.

Issue #155, opened 2015-07-17
Rule names not populated with the new URE configuration format


eddiemonroe:
@ngeiswei 
Nil, with the new URE config scheme format, Rule objects are no longer populated with their names. I think we need the rule names for, e.g., inference trail history, debugging, etc.

eddiemonroe:
@ngeiswei btw, I am not seeing a way to populate the names through a C++ code addition with the current configuration scheme, because the rules get associated with the rbs in the atomspace as the rule Bindlinks, which do not include the rule name info.

ngeiswei:
@eddiemonroe I decided to do that (put the BindLink directly as member) cause DefineLink wasn't completely implemented and EquivalenceLink seemed inappropriate.

I'm thinking of actually completing DefineLink, then fixing that.

linas:
what more do you need for DefineLink? ??

ngeiswei:
@linas this then DefineLink has been fully implemented so it's alright.

PR #189, opened 2015-08-07
Re enable rule name

ngeiswei:
## Overview
1. Rule have now a name, DefineLink use is actually madatory (I'll soon update the wiki about that), so if fixes issue #155 
2. Replaces cost in rule implementation by weight, as that is what that is
3. other misc bugs.

## UTest results
```
```


ngeiswei:
Looks like we don't need to paste the utest results anymore :-)

ngeiswei:
I've updated the wiki http://wiki.opencog.org/w/URE_Configuration_Format#Rules.

Is it wise to force the user to give a rule a name? I don't know. If it's too constraining I guess we can relax that later on.

ngeiswei:
There's a large number of nameless R2L in the opencog repo. Either someone should give them names, or I should add support for nameless rules.

I can take care of naming the PLN rules however...

ngeiswei:
Tests are passing on my machine, maybe the ci test is buggy, I'm merging for this time.

PR #190, opened 2015-08-09
Remove "stack setup" from cmake.

MarcosPividori:
Hi, with this pull request, I am completing the work on ocpkg installation script. (https://github.com/opencog/ocpkg/pull/19)
Thanks,
Marcos


ngeiswei:
I'm merging. Although I guess we can improve that by having cmake output a message explaining to the user what to do the first time or something. We can deal with that later.

williampma:
This is the PR that for some reason started killing CircleCI I think.

ngeiswei:
Hmmm, isn't stack setup run by ocpkg?

ngeiswei:
I see the environment is set by running
```
docker run -v $PWD:/atomspace -it opencog/opencog-deps /bin/sh -c "cd /atomspace && /tmp/octool -cbet"
```
I have no idea what that is actually doing...

@MarcosPividori At least cmake shouldn't include Haskell binding if `stack setup` hasn't been properly run once. Could you look into that? Since I see the following being triggered
```
No GHC found, expected version 7.8.4 (x86_64) (based on resolver setting in /atomspace/opencog/haskell/stack.yaml). Try running stack setup
```
I suppose that should be doing, though I don't know the details. Thanks!

MarcosPividori:
Hi,
How does CircleCI work? That shouldn't happen if you run octool from [ocpkg](https://github.com/opencog/ocpkg) repository, with the flag "-d" (install dependencies).
Anyway, I have made a pull request to ignore haskell bindings if proper ghc version is not installed. (users can install it on their owns or running "stack setup" or using octool with the -d flag).
Thanks,
Marcos

AmeBel:
CircleCI uses opencog/opencog-deps docker image. As you can see [here](https://github.com/opencog/docker/blob/master/opencog/base/Dockerfile#L14) it `-d` is used already.

MarcosPividori:
Hi, AmeBel, thanks for your comment.
Does the line https://github.com/opencog/docker/blob/master/opencog/base/Dockerfile#L64 means octool is run as root?? If is run as root, "stack setup" won't be run.
Please, could you point me where I can see the output of octool by docker?
Thanks,
Marcos

AmeBel:
Got it, I will add `-h` flag for haskell bindings for better control. But the suggestion made by @ngeiswei is still pertinant.
> cmake shouldn't include Haskell binding if stack setup hasn't been properly run once.

MarcosPividori:
Hi!
 - Yes! @ngeiswei suggestion was fixed in this pull request: https://github.com/opencog/atomspace/pull/197
 - Anyway, I would like to know why octool failed, probably because it was run as root, and global ghc installation failed. It would be really useful to see the stdout of octool when run with the -d flag. Where can I see that output from docker?
Thanks,
Marcos

AmeBel:
`docker build` outputs https://hub.docker.com/r/opencog/opencog-deps/builds/bmtjvgwyugpatdjunjk5k34/

MarcosPividori:
Great! Thanks.
Got it. On line:

"[1;34m[octool] Global ghc installation failed. It will be locally installed by Stack.[0m
Please run without sudo. Stack need to be run in user space."

So ghc global installation fails and "stack setup" is avoided because of root user.
This should be fixed if you run ocpkg without root.
Thanks,
Marcos

MarcosPividori:
Hi,
I think after https://github.com/opencog/ocpkg/pull/20 it should work fine, without changes to Dockerfile.
Marcos.

MarcosPividori:
Ok. Yes I can see it: https://hub.docker.com/r/opencog/opencog-deps/builds/b49bqsuchxrkwuqzuhjdkwf
Thanks,
Marcos

Issue #132, opened 2015-07-09
Evaluatable in non-logical link inside BindLink clause


williampma:
Just thought I would throw this out there.  Doing `cog-bind` on the following craziness

```
(BindLink
	(VariableList (stv 1.000000 0.000000)
		(VariableNode "$A") ; [80]
		(VariableNode "$B") ; [92]
		(VariableNode "$C") ; [103]
	)
	(MemberLink (stv 1.000000 1.000000)
		(BindLink (stv 1.000000 0.000000)
			(VariableList (stv 1.000000 0.000000)
				(TypedVariableLink (stv 1.000000 0.000000)
					(VariableNode "$A") ; [80]
					(TypeNode "ConceptNode") ; [100]
				) ; [101]
				(TypedVariableLink (stv 1.000000 0.000000)
					(VariableNode "$B") ; [92]
					(TypeNode "ConceptNode") ; [100]
				) ; [102]
				(TypedVariableLink (stv 1.000000 0.000000)
					(VariableNode "$C") ; [103]
					(TypeNode "ConceptNode") ; [100]
				) ; [104]
			) ; [105]
			(AndLink (stv 1.000000 0.000000)
				(InheritanceLink (stv 1.000000 0.000000)
					(VariableNode "$A") ; [80]
					(VariableNode "$B") ; [92]
				) ; [106]
				(InheritanceLink (stv 1.000000 0.000000)
					(VariableNode "$B") ; [92]
					(VariableNode "$C") ; [103]
				) ; [107]
				(NotLink (stv 1.000000 0.000000)
					(EqualLink (stv 1.000000 0.000000)
						(VariableNode "$A") ; [80]
						(VariableNode "$C") ; [103]
					) ; [108]
				) ; [109]
			) ; [110]
			(ExecutionOutputLink (stv 1.000000 0.000000)
				(GroundedSchemaNode "scm: pln-formula-simple-deduction") ; [111]
				(ListLink (stv 1.000000 0.000000)
					(InheritanceLink (stv 1.000000 0.000000)
						(VariableNode "$A") ; [80]
						(VariableNode "$B") ; [92]
					) ; [106]
					(InheritanceLink (stv 1.000000 0.000000)
						(VariableNode "$B") ; [92]
						(VariableNode "$C") ; [103]
					) ; [107]
					(InheritanceLink (stv 1.000000 0.000000)
						(VariableNode "$A") ; [80]
						(VariableNode "$C") ; [103]
					) ; [112]
[...]

PR #191, opened 2015-08-10
Backward Chainer: rule's input as premise

williampma:
Added the missing step in the Backward Chainer design:

* when no premise matches the input of a rule, add the partially grounded rule's input as potential target

Contains some basic temporary work-around for some unsolved issue, like #46, #132


williampma:
CircleCI is dead.  Merging.

PR #199, opened 2015-08-13
atom.sql fixed

VictorPhilipp:
Running the queries of the atom.sql file failed because the table "Atoms" references to "Spaces" before "Spaces" is being created.


linas:
thanks!

PR #205, opened 2015-08-17
use include/opencog/* as install path

AmeBel:
The other option, that I know of, is to refactor all dependent project's preprocessor declarations. That maybe the 'right' thing to do, I don't know, or revert the project name change.


linas:
The need for this change is a side-effect of my merge of Hendy's "obvious" project-name fix. Sigh.

linas:
BTW, you almost surely need something for cog-utils, as well.  Moses, though, perhaps, we can leave as-is?

AmeBel:
already done for cogutils

PR #206, opened 2015-08-18
Fiddling with behavior tree sequences

linas:
A minor fix for misbehaving code.


linas:
I'm self-merging mostly cause I figure this affects no one at this time...

PR #207, opened 2015-08-18
FC enhancemnet 

misgeatgit:
-Separated the unification functions
-Added sub-atom unification method as an addition feature
-Added map for holding rule and and its derived sets after source is matched

**Todo**
 -Make use of the features inside the main forward chainer logic
 -Enhance the interface

### DIFF
--- a/opencog/rule-engine/forwardchainer/DefaultForwardChainerCB.cc
+++ b/opencog/rule-engine/forwardchainer/DefaultForwardChainerCB.cc
@@ -46,76 +46,140 @@ DefaultForwardChainerCB::DefaultForwardChainerCB(AtomSpace& as,
 }
 
 /**
- * Choose rules based on premises of rule matching the source.
- * Uses temporary atomspace to limit the search space. It adds
- * each members of the implicant(members being identified by
- * checking if they are wrapped in Logical links or not) and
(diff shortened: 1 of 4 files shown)


opencog/rule-engine/forwardchainer/DefaultForwardChainerCB.cc
```diff
@@ -46,76 +46,140 @@ DefaultForwardChainerCB::DefaultForwardChainerCB(AtomSpace& as,
...
+    HandleSeq derived_rules;
+    HandleSeq impl_members = rule->get_implicant_seq();
+
+    UnorderedHandleSet output_expanded;
+    for (Handle h : impl_members) {
+        UnorderedHandleSet hs = get_all_unique_atoms(h);
+        hs.erase(h); //Already tried to unify this.
+
+        HandleSeq result = unify(source, rule);
+        derived_rules.insert(derived_rules.begin(), result.begin(),
```
williampma: `output_expanded` is unused.  It looks like you will unify the same thing multiple times when the same atom appear inside multiple clause.

misgeatgit: Thanks, I will fix those things.

PR #209, opened 2015-08-18
PythonEval: allow multiple atomspaces

linas:
Some quick-n-dirty code to allow multiple atomspaces in python, while also introducing some minimal thread-safety, too.  I'm hoping this fixes/works-around Yi-Shan's MineCraft scripting issues.


linas:
Self-merged; this seemed like a long-standing emergency.

PR #214, opened 2015-08-20
FC enhancements

misgeatgit:
-Fixed issues in sub atom unification
-Made some changes to unify so it could be called from sub_atom_unify
-Started adding test case for subatom_unify

### DIFF
--- a/opencog/rule-engine/forwardchainer/DefaultForwardChainerCB.cc
+++ b/opencog/rule-engine/forwardchainer/DefaultForwardChainerCB.cc
@@ -66,17 +66,12 @@ vector<Rule*> DefaultForwardChainerCB::choose_rules(FCMemory& fcmem)
 
     vector<Rule*> chosen_rules;
     auto rules = fcmem.get_rules();
-    vector<Rule*> visited;
 
     for (Rule* rule : rules) {
-        HandleSeq derived_rules = { };
-
-        HandleSeq results = unify(source,rule);
-        derived_rules.insert(derived_rules.end(), results.begin(),
-                             results.end());
+        HandleSeq derived_rules = unify(source,rule);
 
         //Chosen rule.
-        if (not (derived_rules.empty()))
+        if (not derived_rules.empty())
         {
             chosen_rules.push_back(rule);
             rule_derivations[rule->get_handle()] = derived_rules;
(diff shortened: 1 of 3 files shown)


opencog/rule-engine/forwardchainer/DefaultForwardChainerCB.cc
```diff
@@ -165,21 +165,22 @@ HandleSeq DefaultForwardChainerCB::unify(Handle source,Rule* rule){
...
 
-    UnorderedHandleSet output_expanded;
     for (Handle h : impl_members) {
         UnorderedHandleSet hs = get_all_unique_atoms(h);
         hs.erase(h); //Already tried to unify this.
 
-        HandleSeq result = unify(source, rule);
-        derived_rules.insert(derived_rules.begin(), result.begin(),
-                             result.end());
+        for (Handle h : hs) {
```
williampma: This is still over-unifying the same atom that appear in multiple clause right?

If the rule's inputs are
```
(InheritanceLink $A $B)
(InheritanceLink $B $C)
```
$B will be unified twice.

misgeatgit: Shouldn't get_all_unique_atoms return unique list of variables? and the unit test also asserts this assumption of mine. Am I missing sth here?

misgeatgit: Now I got what you mean. I have added a fix for that.

ngeiswei:
@misgeatgit I don't understand why we need subatom_unify. Could you briefly explain to me or point me to the conversation that explains it? Thanks.

ngeiswei:
I've disabled it, no unit test seem to rely on it, neither on the atomspace of opencog repo. I really can't understand why we want that. Any hint is appreciated.

ngeiswei:
The commit introducing that is https://github.com/misgeatgit/atomspace/commit/c8d24ea2d88a6bb5bd9ebc42f164c76efa5bf24c

I found a conversation off-list that talks about it. William asks a question (which really looks like this implementation, i.e. unifying a source with subatoms of the **rule**), and Eddie and Ben misunderstand the question and answer about unifying a rule with subatoms of the **source**.

Anyway, by now I'm almost sure this code is wrong, but some inputs would still be appreciated.

ngeiswei:
I need to move forward. @misgeatgit if you have some useful information about the reason of this code feel free to post it here. Otherwise it's about to go away.

ngeiswei:
Hmm, I'm hesitant to remove it just yet, what if it used by R2L? Please let me know.

misgeatgit:
Hi Nil, I think you are right. You may remove it. R2L doesn't even do
matching just applies all the rules on a predefined focus set.

ngeiswei:
OK, I see, so as long as I get this piece of code

```c++
    // Relex2Logic uses this. TODO make a separate class to handle
    // this robustly.
    if(_potential_sources.empty())
    {
        apply_all_rules(_search_focus_Set);
        return;
    } 
```

behave qualitatively similar, it shouldn't break R2L.

Thanks for the very useful info!

PR #219, opened 2015-08-21
Fix Template Haskell details

MarcosPividori:


MarcosPividori:
Everything seems to be ok with circleci test on haskell bindings:
 [100%] Built target haskell-atomspace-lib

Issue #162, opened 2015-07-22
Bug or feature? deleting an atom using a variable in scm shell does remove the atom from the atomspace but not from the scheme enviroment.


AmeBel:
### 1. Inpute atoms
```
opencog> (clear)
Entering scheme shell; use ^D or a single . on a line by itself to exit.
guile> 
guile> (cog-prt-atomspace)

guile> (define x (cog-new-node 'ConceptNode "abc")) 

guile> (cog-prt-atomspace)
(ConceptNode "abc")

guile> (define y (cog-new-node 'ConceptNode "def")) 

guile> (cog-prt-atomspace)
(ConceptNode "def")
(ConceptNode "abc")

guile> (define l (cog-new-link 'ListLink x y)) 

guile> (cog-prt-atomspace)
(ListLink
   (ConceptNode "abc")
   (ConceptNode "def")
)
```
### 2. delete the link 
```
guile> (cog-delete l)
#t
guile> (cog-prt-atomspace)
(ConceptNode "def")
(ConceptNode "abc")

```
### 3. part that caused me to doubt what the scheme shell does, 
@williampma informed me that the following was always the case.
```
guile> l
(ListLink
   (ConceptNode "abc")
   (ConceptNode "def")
)

guile> (cog-prt-atomspace)
(ConceptNode "def")
(ConceptNode "abc")

guile> (cog-handle l)
59
guile> (cog-outgoing-set l)
((ConceptNode "abc")
 (ConceptNode "def")
)
guile> (cog-atom (cog-handle l))
#<Invalid handle>

guile> (SetLink l)
(SetLink (av 0 0 0)
   (ListLink
      (ConceptNode "abc")
      (ConceptNode "def")
   )
)

guile> (cog-prt-atomspace)
(SetLink (av 0 0 0)
   (ListLink
      (ConceptNode "abc")
      (ConceptNode "def")
   )
)

```
My assumption (used to think that I knew, thus this issue) was that on evaluating the varaible `l` , after deleting the atom, the print out on the shell was a confirmation that the atom has been reinserted into the atomspace, which is not the case. So, is this a bug or a feature? @linas

AmeBel:
woops I should have opened this issue in opencog.

williampma:
Yeah, first discovered https://github.com/opencog/opencog/pull/825#issuecomment-45301736

I guess it could be a problem, as you and Mandeep said, when there are multiple processes working on the AtomSpace.  Any scheme variables will become invalid at any moment.

[...]

PR #222, opened 2015-08-21
Add support for multiple AtomSpaces.

MarcosPividori:
Also, I enable partial evaluatiin of computations in the atomspace. This is useful when working in interactive mode, such as using GHCi.

### DIFF
--- a/opencog/haskell/OpenCog/AtomSpace.hs
+++ b/opencog/haskell/OpenCog/AtomSpace.hs
@@ -1,11 +1,15 @@
 -- GSoC 2015 - Haskell bindings for OpenCog.
-{-# LANGUAGE TypeOperators        #-}
+{-# LANGUAGE TypeOperators #-}
 
 -- | This library defines Haskell Bindings for the AtomSpace.
 module OpenCog.AtomSpace
     (
     -- * AtomSpace Environment
       AtomSpace
+    , AtomSpaceObj
+    , newAtomSpace
+    , onAtomSpace
+    , (<:)
     , runOnNewAtomSpace
     -- * AtomSpace Interaction
     , insert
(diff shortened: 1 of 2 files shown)


ngeiswei:
Excellent!

linas:
I was unable to tell from the commit how to tell which atomspace "expands" on which other atomspace. That is, an atomspace B expands on A if everything in A is also visible in B. This makes an atomspace look like an "environment", and typical usage is to create a temporary B, do some calculations in it, then discard B.  Thus, a useful feature is to have a function that, given B, reports what its base A is.

MarcosPividori:
Hi Linas! Thanks for your comments! Yes, you were right. In this pull request only implements independent multiple atomspaces. I have been working on this, and I finally implemented hierarchy of AtomSpaces! It is on the pull request: https://github.com/opencog/atomspace/pull/225

linas:
OK, thanks!   I wasn't sure if you forgot, hadn't noticed or were working on it.

linas:
BTW, totally unrelated, but I just fixed a bug in the scheme bindings #162, and I wonder if perhaps it hits the haskell bindings.  In scheme it  looks like this:
```
(define x (ConceptNode "foobar"))
```
creates a symbol x which has an associated value that is the atom (ConceptNode "foobar") (well, actually, the scheme binding store the handle pointing to this atom...)  The user then deletes it: 
```
(cog-delete! x)
```
which does remove the atom from the atomspace; but I had forgotten to zero out the handle itself, which continued to point at the atom (even though it is no longer in any atomspace).  This gets confusing when you got to reference x again ...  If you do something like this, you may want to check.

Note also that  delete and purge are two similar functions, one deletes from atomspace and from backing store, the other just removes from atomspace, leaving backing store alone...

MarcosPividori:
Hi Linas,
Thanks for that information. Because of the characteristics of Haskell language (such as immutable data), I am not storing the handler pointing to an atom. All the atoms you use in Haskell code are just haskell data structures without a reference to the atomspace. When you do some operation over the atomspace, such as insert/remove/get, etc, haskell bindings gets the proper handle based on the atom information, and then works using it. So I don't have the problems such that you mentioned.
I know this is not optimal. Maybe this can be improved in future, but right now, it is not clear how to properly improve that in Haskell. I have been considering different options, for example, Persistent, a popular haskell library for data bases, considers a specific Key associated with the data type. So, when you want to update some information in the DB, you use that key:  [persistent_update](http://www.yesodweb.com/book/persistent#persistent_update)
But I think using keys (like handles) will make the code confusing and complex. We want high level bindings in Haskell.
Thanks,
Marcos

PR #227, opened 2015-08-22
pattern matcher unique results

jswiergo:
This partly addresses problems  #148. The duplicated results of the query are made unique by default grounding callback. I have analysed how to prune search for duplicated results at earlier phase of pattern matching but found it complicated in general for example for hierarchical choice links and other cases, so I suggest this simple solution for now.

### DIFF
--- a/opencog/query/Implicator.h
+++ b/opencog/query/Implicator.h
@@ -47,20 +47,31 @@ namespace opencog {
  *
  * The 'var_soln' argument in the callback contains the map from variables
  * to ground terms. 'class Instantiator' is used to perform the actual
- * grounding.  A list of grounded expressions is created in 'result_list'.
+ * grounding.  A set of grounded expressions is created in 'result_set'.
+ * Note that the callback may be called many times reporting the same
+ * results. In that case the 'result_set' will contain unique solutions.
  */
 class Implicator :
 	public virtual PatternMatchCallback
 {
+	protected:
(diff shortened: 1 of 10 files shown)


jswiergo:
```
```

opencog/query/Implicator.h
```diff
@@ -47,20 +47,31 @@ namespace opencog {
...
+ * grounding.  A set of grounded expressions is created in 'result_set'.
+ * Note that the callback may be called many times reporting the same
+ * results. In that case the 'result_set' will contain unique solutions.
  */
 class Implicator :
 	public virtual PatternMatchCallback
 {
+	protected:
+		UnorderedHandleSet _result_set;
+		HandleSeq _result_list;
```
linas: So  get_result_list() returns a HandleSeq; that HandleSeq will be constructed on-stack, and returned to the caller. There is no point in caching _result_list: you are just forcing a copy from the HandleSeq instance in this class instance, to the callers stack frame. 

I guess you are thinking that, byt using the _result_changed flag, you are saving some cpu time. if the user asks for the result_list more than once ... but I don't see that being likely?

jswiergo: I wanted to avoid copying from unordered set each time, but did it overcomplicated here:
https://github.com/opencog/atomspace/blob/master/opencog/query/Implicator.cc#L77-L78

PR #228, opened 2015-08-22
ChoiceLink unneeded special case

jswiergo:
At the present implementation of PatternMatchEngine this piece of code is not needed any more. The case described in the comment is covered by ChoiceLinkUTest - top-disco test. After removing the code it works fine and it is easier to introduce next changes.

```
```


linas:
Huh. OK.

PR #229, opened 2015-08-22
Pattern matcher term tree traversing

jswiergo:
This changes are dependent and may be merged after:
#227 pattern matcher unique results
#228 ChoiceLink unneeded special case

This are next changes that partly address problem #148 (Pattern matcher branch exploration problem)

The idea of solution is to preprocess pattern query and build internal tree structure of unique terms. Each term corresponds to one atom handle. If the same atom appears many times in the query then it corresponds to many unique terms, one for each occurrence. Having that structures, pattern matcher traverses the pattern tree exactly as before but uses term pointers instead of atom handles to know its exact position. Then it keeps states of unordered links and choice links in the language of term pointers (e.g. permutation of unique pointers instead of permutation of not unique handles).

Test status after merging #227 and #228:
```
```

This changes are ready and stable. More work is needed to solve the problem entirely.

### DIFF
--- a/opencog/atoms/bind/PatternLink.cc
+++ b/opencog/atoms/bind/PatternLink.cc
@@ -71,6 +71,8 @@ void PatternLink::common_init(void)
 		// _pat.cnf_clauses = _components[0];
 	   make_connectivity_map(_pat.cnf_clauses);
 	}
+
+	make_term_trees();
 }
 
 
@@ -161,6 +163,8 @@ PatternLink::PatternLink(const std::set<Handle>& vars,
 
 	make_connectivity_map(_pat.cnf_clauses);
 	_pat.redex_name = "Unpacked component of a virtual link";
+
+	make_term_trees();
 }
 
 /* ================================================================= */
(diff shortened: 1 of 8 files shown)


opencog/atoms/bind/PatternLink.cc
```diff
@@ -639,6 +643,35 @@ void PatternLink::make_map_recursive(const Handle& root, const Handle& h)
 	}
 }
 
+// Hack alert: Definitely it should not be here. Though some refactoring
+// regarding shared libraries circular dependencies (liblambda and libquery)
```
linas: Ugh. the circular dependencies are killing me. If you have any good ideas, let me know.

opencog/query/PatternTerm.h
```diff
@@ -0,0 +1,137 @@
...
+class PatternTerm;
+typedef std::shared_ptr<PatternTerm> PatternTermPtr;
+typedef std::vector<PatternTermPtr> PatternTermSeq;
+
+class PatternTerm
+{
+	protected:
+		Handle _handle;
+		PatternTermPtr _parent;
+		PatternTermSeq _outgoing;
```
linas: Can you add a note explaining how these three members are related to one-another?

opencog/query/Pattern.h
```diff
@@ -79,6 +80,15 @@ struct Pattern
...
 
+	// Each atom of the pattern as well as their corresponding pattern terms
+	// may appear in many clauses. Moreover the same atom may be replicated
+	// under the same clause root in many instances. Each occurence has its
+	// own unique PatternTermPtr. We need to keep the mapping beetwen atoms
+	// and clause roots to the list of atoms occurences. Typically the list
+	// of PatternTermPtr has single element, but as said above when given atom
+	// is replicated under one root then the list has many elements.
+	typedef std::map<std::pair<Handle,Handle>, PatternTermSeq> ConnectTermMap;
+
```
linas: what are the two members of the pair?  how are they related to each-other?

linas:
thanks!

opencog/atoms/bind/PatternLink.cc
jswiergo: Seems #642 solves that issue without expensive refactoring.

PR #232, opened 2015-08-23
post review fixes

jswiergo:


opencog/query/PatternTerm.h
```diff
@@ -32,6 +32,26 @@
 
 namespace opencog {
 
+/*
```
linas: If you start a comment with /**  then doxygen will find it and extract it as part of the class documentation.
Likewise, any line starting with three slashes is published by doxygen.

I don't use doxygen at all, but for some people this seems important, so I figure its good practice.

Issue #221, opened 2015-08-21
Hierarchical AtomSpace same atom multiple copies problem


williampma:
Finally bump into this problem of multiple atomspace outlined on the wiki

>However, if an atom is added to B that does not exist in A, then it will be in atomspace B only. If later it is added to A, then two independent copies shall exist: one in B and one in A. Its not quite clear if this is the 'right' thing to do; but defacto, this is what happens.

This code
```
		AtomSpace as;
		AtomSpace ex(&as);
		AtomSpace ns;

		Handle haa = as.add_node(CONCEPT_NODE, "PLN");
		Handle hab = as.add_node(CONCEPT_NODE, "URE");

		Handle hea = ex.add_link(LIST_LINK, haa, hab);
		Handle heb = ex.add_node(CONCEPT_NODE, "VAR");
		Handle hec = ex.add_link(IMPLICATION_LINK, hea, heb);

		as.add_link(LIST_LINK, haa, hab);

		Handle hna = ns.add_atom(hec);
		Handle hec_copy = ex.get_atom(hna);

		std::cout << hec->toShortString() << std::endl;
		std::cout << hna->toShortString() << std::endl;
		std::cout << hec_copy->toShortString() << std::endl;
```
gives seg-fault on printing hec_copy, becuase hec_copy is actually returned UNDEFINED.

I ran into this type of code path when handling member-to-inheritance-rule and inheritance-to-member-rule in two different steps of BC.

Ideally, it would be great if when the atom is added to A, handles that refer to B's copy will automatically change to refer to A's copy.  Not sure how do-able is this.

linas:
Well, the seg-fault should be fixed, first. The C++ code above is perfectly valid, and its well-defined, and hec_copy should have returned exactly hec; the two should have been pointers to the same atom.

linas:
Regardng the idea: 
>Ideally, it would be great if when the atom is added to A, handles that refer to B's copy will automatically change to refer to A's copy. Not sure how do-able is this.

I agree, except for the "automatically" part.  There tends to be a high overhead for "automatic" things.   Thus, it would be better if a method was added to the AtomTable -- lets call it "promote" that, given atom Xb, looks to see if the equvalent Xa is in the parent(s) A, and, if so, removes Xb from B, and replaces all uses of Xb in B by A's version Xa.

[...]

PR #233, opened 2015-08-24
Switched getHandle internal order

williampma:
This fixes the crash at https://github.com/opencog/atomspace/issues/221

Doesn't do any complicated promote stuff yet.


linas:
good catch!

PR #237, opened 2015-08-25
Improve multithreading. Also, add an example of that.

MarcosPividori:


ngeiswei:
Wow, that wasn't in the job description. A last minute bonus then. Quite educational to me (well, except for the example I don't understand a thing). Thanks!!!

MarcosPividori:
Thanks!! :)

PR #238, opened 2015-08-25
Prepare inclusion of persist/zmq/atomspace structure

ceefour:
Prepare inclusion of persist/zmq/atomspace structure. While still work in progress, atomspace can still build with it. Hendy @ceefour will maintain.


ceefour:
P.S. Let me know if this pull has too many commits, I'll squash them into a new commit or even break the changes apart.

opencog/persist/zmq/atomspace/ZMQClient.cc
```diff
@@ -1,11 +1,11 @@
+
 /*
- * opencog/atomspace/ZMQClient.cc
- *
- * Copyright (C) 2008-2010 OpenCog Foundation
```
linas: why are you deleting copyright notices?

opencog/persist/zmq/atomspace/ZMQClient.h
```diff
@@ -1,11 +1,11 @@
...
- * opencog/atomspace/ZMQClient.h
+ * FUNCTION:
+ * Base class for ZeroMQ-backed persistent storage.
  *
- * Copyright (C) 2008-2010 OpenCog Foundation
- * All Rights Reserved
- *
- * Written by Erwin Joosten
+ * HISTORY:
+ * Copyright (c) 2015 Erwin Joosten, Hendy Irawan <ceefour666@gmail.com>
```
linas: I am pretty sure Erwin did not work on this in 2015!  He probbly worked on this in 2009 or 2010 and then assigned copyright to the OpenCog Foundation .. whose copyright you removed ...

opencog/persist/zmq/atomspace/ZMQClient.cc
ceefour: this was oversight on my part, the "more correct" copyright is in `ZMQClient.h`.
The reason why it was this way is because I originally copied structure from your SQL Persistence, then realized that parts of implementation is in ZMQClient, so I moved the contents from Erwin's ZMQClient to my ZMQClient, forgotting to update ZMQClient.cc in the process (but I did update ZMQClient.h).

opencog/persist/zmq/atomspace/ZMQClient.h
ceefour: It should read: (also for ZMQClient.cc)

```
Copyright (C) 2008-2015 OpenCog Foundation
All Rights Reserved

Written by Erwin Joosten, Hendy Irawan <ceefour666@gmail.com>
```

Would this be okay?

opencog/persist/zmq/atomspace/ZMQClient.cc
```diff
@@ -1,11 +1,14 @@
...
- * opencog/atomspace/ZMQClient.cc
+ * FUNCTION:
+ * Persistent Atom storage, ZeroMQ-backed e.g. Neo4j.
  *
- * Copyright (C) 2008-2010 OpenCog Foundation
+ * Copyright (C) 2008-2015 OpenCog Foundation
  * All Rights Reserved
  *
- * Written by Erwin Joosten
+ * Written by Erwin Joosten, Hendy Irawan <ceefour666@gmail.com>
```
ceefour: @linas I've fixed the copyright notices.

linas: OK, thanks!

AmeBel:
@ceefour push your remaining changes to this branch no need to create a separte pr. Holding on merging untill the remaining updates  are pushed.

ceefour:
@AmeBel changes pushed here.

AmeBel:
Thanks, will test.

opencog/persist/zmq/atomspace/ZMQPersistSCM.cc
```diff
@@ -0,0 +1,227 @@
...
+}
+
+ZMQPersistSCM::~ZMQPersistSCM()
+{
+	delete _backing;
+}
+
+void ZMQPersistSCM::do_open(const std::string& networkAddress)
+{
+	_store = new ZMQClient(networkAddress);
```
linas: Based on the string name, I guess this is just an IP address. But how do you provide the username, the password or other crypto login credentials? How do you specify which database is to be connected to?  I assume a single server can host multiple different databases, right?

opencog/atomspace/Atom.h
```diff
@@ -81,6 +81,7 @@ class Atom
     friend class TLB;             // Needs to view _uuid
     friend class CreateLink;      // Needs to call getAtomTable();
     friend class DeleteLink;      // Needs to call getAtomTable();
+    friend class ProtocolBufferSerializer; // Needs to de/ser-ialize an Atom
```
linas: why is this needed? what private/protected elements is the serializer trying to get access to?

ceefour: If I remember correctly ._uuid, .getAtomTable() , probably some others. Part of implementation is still the same as it was before I came. For those accessors which are public now, I replace them with public ones, but some still are protected hence the 'friend request'. :)

opencog/persist/zmq/atomspace/ZMQPersistSCM.cc
ceefour: IP address **and** port. But you're right, ZeroMQ supports a concept called "topic", to share same IP and port with different, err, topics. It's a good feature for the future, but currently, multiple DBs can (probably?) be supported by using different ports.

linas: So Neo4J has no concept of security or authentication?  Any user who can get to a given port number on a given machine has full read-write access to all data? That can't be right...

AmeBel: It only added aurhentication recently. There is an option to disable it .

ceefour: @linas in this case it's ZeroMQ not Neo4j. So Ilve read that ZMQ 4.0 supports CurveZMQ for username & password authentication. This can probably be used for further improvement in security, I'll make note of this.

linas: I'm not talking about zmq, I'm talking about Neo4J.  When you go to access the database itself, via zmq or via rest or via whatever, you have to be able to prove that you are you, as otherwise,  any russian botnet could log into your database and rape it.  There has to be some way of protecting your data from hostile actors....

linas: Not just botnets. So assume you are running neo4j behind some firewall, so its not open to the internet.  Say its a corporate firewalll, you have 100 employees, and you hire employee 101 and on the first day, he goes "gee I wonder what this is", and starts deleting/mangling data in the database. Its almost a guarantee.  (Back at IBM, when the token ring broke, the sysadmin would walk up and down the hallways, asking "did we hire someone today?")

Now, you can hide a server in a corporate environment, by disabling all tcpip connections to it, and then setting up an ssh port-forwarder. That way, only the people with the correct ssh credentials can get access to the server. But managing ssh port-forwarding is painful, tricky, error-prone, non-automated.

linas: (My first day at qualcomm, I ran bit-torrent to download something. Bad idea, it freaked out their firewall, and I almost got fired. How could I know?)

ceefour: Neo4j over REST mandates auth by default:
http://neo4j.com/docs/stable/rest-api-security.html

However we're not using REST, at all.
We expose Neo4j through OpenCog's own protobuf through ZeroMQ. The OpenCog
protobuf is specific to OpenCog, and is hopefully "agnostic" with regard to
database. In this case auth needs to be done via ZMQ, which I assume can be
done with CurveZMQ but aside from knowing its name, I haven't studied much
about it (yet).

PR #240, opened 2015-08-27
Crude, ugly stab at DefineLink support

linas:
More wrong than right, but one basic example works... Implements part of issue #74 considerably more work is needed to make this usable.

### DIFF
--- a/examples/pattern-matcher/define.scm
+++ b/examples/pattern-matcher/define.scm
@@ -3,6 +3,8 @@
 ;
 ; Demonstrate the use of DefineLink to give names to things.
 ;
+; XXX caution: under construction, unstable, known buggy.
+;
 (use-modules (opencog))
 (use-modules (opencog query))
 
@@ -39,7 +41,7 @@
 (define elect-parts
 	(DefineLink
 		(DefinedPredicateNode "Electrical Part Of")
-		(AndLink
+		(PresentLink
 			(InheritanceLink
 				(VariableNode "$x")
 				(ConceptNode "electrical device"))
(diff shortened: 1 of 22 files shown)


examples/pattern-matcher/define.scm
```diff
@@ -0,0 +1,63 @@
...
+					(VariableNode "$x")
+					(VariableNode "$y"))))))
+
+;; Define a pattern to find the electrical parts of things
+;; At this time, an explicit variable declaration has to be done;
+;; this needs to be fixed...
+(define get-elect
+	(GetLink 
+		(VariableList (VariableNode "$x") (VariableNode "$y"))
+		(DefinedPredicateNode "Electrical Part Of")))
```
ngeiswei: In your example "Electrical Part Of" is defined as a body with free
variables, then the variables are scoped by the GetLink. Wouldn't be
better if the variables where scoped in the DefineLink, such as

```
(define elect-parts
	(DefineLink
		(DefinedPredicateNode "Electrical Part Of")
        (LambdaLink
            (VariableList
                (VariableNode "$x")
                (VariableNode "$y"))
		    (PresentLink
                (InheritanceLink
				    (VariableNode "$x")
				    (ConceptNode "electrical device"))
			    (EvaluationLink
				    (PredicateNode "PartOf")
				    (ListLink
					    (VariableNode "$x")
					    (VariableNode "$y")))))))
```

then the GetLink can take the predicate alone

(GetLink (DefinedPredicateNode "Electrical Part Of"))

That way we're sure the variable names are consistent.

linas: Yes, that would be better. This, however, has been one of the issues confusing me. It requires a more sophisticated "pasting" rule, when assembling a pattern out of define links.  The pasting rule seems to become: "if a lambda link is found, then paste the body of the lambda, not the lambda itself".  It's ... seems kind of strange at first, but it seems to be the right thing to o, I guess.

linas: Fixed in pull req #242 -- you can use a lambda link to name variables, else they are automatically extracted.

PR #243, opened 2015-08-29
Quoted variables matching refactor

jswiergo:
This solves problems described in the comments:
https://github.com/opencog/atomspace/blob/master/opencog/query/PatternMatchEngine.cc#L812-L819
https://github.com/opencog/atomspace/blob/master/opencog/query/PatternMatchEngine.cc#L196-L198

The right answer for QuoteUTest : infloop is 4. Previously it grounded variable $x by atoms that contain this bound variable unquoted, so it might caused infinite loops, but after present changes this sort of cases are not possible.

```
100% tests passed, 0 tests failed out of 82
```


linas:
Thanks...

Issue #172, opened 2015-07-27
Cannot create BindLink in one go (unadded Handles comparison problem)


williampma:
```
    Handle v(createNode(VARIABLE_NODE, "$A"));
    Handle blv(createLink(VARIABLE_LIST, v));
    Handle blb(createLink(AND_LINK, v));
    Handle blt(createLink(LIST_LINK, v));
    HandleSeq bll = {blv, blb, blt};
    Handle bl_final(createLink(BIND_LINK, bll));

    as->add_atom(bl_final);
```

gives me exception

```
The variable (VariableNode "$A") ; [18446744073709551615]
 does not appear (unquoted) in any clause! (/home/william/atomspace/opencog/atoms/bind/PatternLink.cc:365)
```

This is causing some problem with Instantiator when a mapping mapped a variable to a BindLink on an evaluatable (so the thing has to be added to `_temp_aspace`).

I came across this before but worked around it at the time.  The problem is that Handles that haven't been added to the AtomSpace yet cannot be compared (no UUID).

ngeiswei:
For that reason I always use AtomSpace::add_{node,link} and never create{Node,Link}, but it seems it would be nice if one could create and manipulate hypergraphs without having to insert them in the AtomSpace.

linas:
For the most part, you can create atoms outside of the atomspace, and
things should mostly work.  There are a few "obvious" limitations:

-- There is no way of gauaranteeing the uniqueness of atoms unless you
place them in the atomspace. Viz, witout the atomspace, you can have
multiple copies of "atom foo", and this can potentially cause
problems, depending on your algo.

--- UUID's are only issued whten atoms are placed in the atomspace.
Since the python bindings use UUID's (they should not, but they do),
you cannot use python on atoms that are not in the atompsace.

-- The only way to share atoms on different machines (in an
unambiguous fashion) is via the atomspace.

-- In parinciple, the pattern matcher does not really depend on the
atomspace; however, the new design splits pattern matching into two
parts: one-time-only initialization, and the rest, and perhaps the
first-time initialization is tangled into the atomspace. This could
perhaps be fixed; its not a fundamental limitation.

Linas.

linas:
I just tried this, it seems to work at this time. Maybe one of the recent pull requests fixed it.

jswiergo:
Yes, it is fixed by #244 and can be closed. BindLink constructor used FindUtils which did not support atoms not in the atomspace yet (having UUID=MAXINT).

[...]

Issue #210, opened 2015-08-18
Pattern Matcher unexpected exception


linas:
The following throws an exception, which is not expected:
```
(EvaluationLink
   (ConceptNode "arkle")
   (ConceptNode "barkle")
   (ConceptNode "curry"))

(EvaluationLink
   (ConceptNode "glib")
   (ConceptNode "blab"))

(define bnd
(BindLink
   ;(VariableList (VariableNode "$a") (VariableNode "$b")
   ;   (TypedVariableLink (VariableNode "$lnk")
   ;       (TypeNode "EvaluationLink")))
   (AndLink
       (VariableNode "$lnk")
       (EvaluationLink
           (VariableNode "$a")
           (VariableNode "$b"))
       (EqualLink
           (VariableNode "$lnk")
           (EvaluationLink
               (VariableNode "$a")
               (VariableNode "$b"))))
   (VariableNode "$lnk")
))
(cog-bind bnd)
```

It does not throw when the VariableList is uncommented, so that the type restrictions are enforced.  This is unexpected, because both versions should return the same answer, although the version with the type declarations should run faster, as it prunes some un-needed searches.

jswiergo:
In this case the instantiator is evaluating following solution during pattern matching:
```
(EqualLink (stv 1,000000 0,000000)
  (BindLink (stv 1,000000 0,000000)
    (AndLink (stv 1,000000 0,000000)
      (VariableNode "$lnk") ; [9]
      (EvaluationLink (stv 1,000000 0,000000)
        (VariableNode "$a") ; [10]
        (VariableNode "$b") ; [11]
      ) ; [18446744073709551615]
      (EqualLink (stv 1,000000 0,000000)
        (VariableNode "$lnk") ; [9]
        (EvaluationLink (stv 1,000000 0,000000)
          (VariableNode "$a") ; [10]
          (VariableNode "$b") ; [11]
        ) ; [18446744073709551615]
      ) ; [18446744073709551615]
    ) ; [18446744073709551615]
    (VariableNode "$lnk") ; [9]
  ) ; [18446744073709551615]
  (EvaluationLink (stv 1,000000 0,000000)
    (ConceptNode "glib") ; [6]
    (ConceptNode "blab") ; [7]
  ) ; [18446744073709551615]
) ; [18446744073709551615]
```
[...]

PR #244, opened 2015-08-30
Fix for #210 instantiator problem

jswiergo:
```
100% tests passed, 0 tests failed out of 83
```

### DIFF
--- a/opencog/atomutils/FindUtils.h
+++ b/opencog/atomutils/FindUtils.h
@@ -185,7 +185,6 @@ class FindAtoms
  */
 static inline bool is_atom_in_tree(const Handle& tree, const Handle& atom)
 {
-	if (tree == Handle::UNDEFINED) return false;
 	if (tree == atom) return true;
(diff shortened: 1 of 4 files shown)


linas:
Thanks!

williampma:
Wait, I am having trouble understanding this fix.  How would removing the == Handle::UNDEFINED fix the handle comparison of atoms not inside the AtomSpace?  It will just compare UNDEFINED to UNDEFINED and always return true...

linas:
William, the issue is that most things should work correctly even if an atom has not yet been placed in the atomspace. Unfortunately, the current Handle design makes them invalid when the accompanying atom is not yet in some atomspace.  Removing the UNDEFINED check was correct; that way, teh utility will work for atoms in general.

Handle comparison is a giant mess.  This is one reason why I want to get rid of UUID's in the handles. 

p.s. I am working to fix bug #172, but this requires tweaking the handle class itself, and is turning into an unpleasant task.  Handles were a bad design idea. My fault.

PR #252, opened 2015-09-01
Fix error. On last commit they forgot to update some cases of fromRaw…

MarcosPividori:
According to Types.hs file


MarcosPividori:
@ngeiswei , here I check some details in Haskell Bindings.

ngeiswei:
Thanks Marcos!

PR #253, opened 2015-09-01
Fix PythonEvalUTest and remove call to RestApiUTest

pennachin:
The current PythonEvalUTest will run OK on people's dev machines but not on a fresh install, such as CircleCI. This is mostly due to cmake not linking the scheme-related libraries. I've hacked `CMakeLists.txt` by aping the scheme UTests, and it now runs on a fresh container. Hopefully it will run on CircleCI.


AmeBel:
:fireworks:

PR #256, opened 2015-09-02
FC

misgeatgit:
Mainly
  -  Attempted to reduce the code size
  -  Added additional focus set parameter so cog-fc needs a third addtion argument which could be empty   SetLink
  -  Changed the logic inside do_chain and do_step - I need to completely decouple these two method on next PR.

@eddiemonroe, @ngeiswei  and @williampma please make review of these changes.


linas:
Can you go back and do a pull request that does not merge master and/or rebase or whatever it is that is pulling in other people's commits?  E.g. Casio's recent changes, and my own.  We had a discussion on the mailing list about how rebasing is bad; this is a good example of how its bad.

PR #257, opened 2015-09-03
Rework handle comparison code

linas:
This is an extensive set of changes with the goal of allowing atoms to be compared directly, as atoms, reducing some of the excessive dependency on UUID's in the code.

This was prompted as a fix to issue #172.

This also breaks BackwardChainerUTest in the fashion described at the bottom of #172

### DIFF
--- a/opencog/atomspace/Atom.cc
+++ b/opencog/atomspace/Atom.cc
@@ -249,7 +249,7 @@ void Atom::setAtomTable(AtomTable *tb)
     if (NULL != _atomTable) {
         // Atom is being removed from the atom table.
         // UUID's belong to the atom table, not the atom. Reclaim it.
-        _uuid = Handle::UNDEFINED.value();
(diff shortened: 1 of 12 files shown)


linas:
Merged, even though this breaks backwardChainerUTest.  The VariableFulfillmentQuery seems clearly wrong ...

ngeiswei:
Has someone started looking into fixing BackwardChainerUTest?  If nobody has I might...

williampma:
Thanks Linas.

I already fixed that BC problem couple days ago on my local copy.  It's just the BC not yet adapted to the new exception being thrown by the PM.

I will do a PR maybe tomorrow.

ngeiswei:
Oh, that's cool @williampma!

Although, of course Linas should not have merged knowingly breaking the utest. Kinda defeat the point of having this nice automatic branch checking!

linas:
Its not the UTest that's broken, its the chainer code. --linas

PR #261, opened 2015-09-04
FC

misgeatgit:
Mainly

- Added additional focus set parameter so cog-fc needs a third addtion argument which could be empty SetLink
- Changed the logic inside do_chain and do_step - I need to completely decouple these two method on next PR.
- Removed DefaultForwardChainer and ForwardChainerCallBack classes and moved code to ForwardChainer class.

@eddiemonroe, @ngeiswei and @williampma please make review of these changes.


opencog/rule-engine/forwardchainer/ForwardChainer.h
```diff
@@ -24,53 +24,99 @@
...
 	Handle _rbs;                // rule-based system
-
 	UREConfigReader _configReader;
-
-
     FCMemory _fcmem;            // stores history
     Logger * _log;
     int _iteration = 0;
+    source_selection_mode _ts_mode;
+    std::map<Rule*, float> rule_weight;
```
williampma: Be consistent.  Why is it `rule_weight` and not `_rule_weight`?

opencog/rule-engine/forwardchainer/ForwardChainer.cc
```diff
@@ -144,19 +147,50 @@ void ForwardChainer::do_chain(ForwardChainerCallBack& fcb,
...
-        do_step(fcb);
+        if(rule_weight.empty())
+        {
+            //! Choose next source.
+            _log->info("[ForwardChainer] setting next source");
+
+            _fcmem.set_source(choose_next_source(_fcmem));
+
+            // Choose matching rules whose input matches with the source.
+            vector<Rule*> matched_rules = choose_rules(_fcmem);
```
williampma: Can't quite tell what happens to `matched_rules`.  It's not stored anywhere (I only see you storing weight in `rule_weight`).

It seems somewhere it is magically stored in URECommon `_rec`?

opencog/rule-engine/forwardchainer/ForwardChainer.cc
```diff
@@ -144,19 +147,50 @@ void ForwardChainer::do_chain(ForwardChainerCallBack& fcb,
...
+
+            //! If no rules matches the pattern of the source,
+            //! set all rules for candidacy to be selected by the proceeding step.
+            //! xxx this decision is made based on recent discussion.
+            //! it might still face some changes.
+            if (not matched_rules.empty()) {
+                _log->info("[ForwardChainer] Found matching rule");
+
+                for (Rule* r : matched_rules) {
+                    rule_weight[r] = r->get_weight();
```
williampma: Ah I see.  The rule itself is stored as part of the map's key...

opencog/rule-engine/forwardchainer/ForwardChainer.cc
```diff
@@ -62,81 +67,79 @@ Logger* ForwardChainer::getLogger()
...
+    UnorderedHandleSet products;
 
-    //!TODO Find/add premises?
+    HandleSeq derived_rhandles = derive_rules(_fcmem.get_cur_source(), r);
 
-    //! Apply rule.
-    _log->info("[ForwardChainer] Applying chosen rule %s",
-               (r->get_handle())->toShortString().c_str());
-    HandleSeq product = fcb.apply_rule(_fcmem);
+    //try sub-atom unification on failure to full unification
```
williampma: Why would full unification even fail here, when your chosen rule already passed full unification in the "chosen_rule" method?

williampma: humm... I guess you planned to include "sub-atom unification" result into `rule_weight`.

opencog/rule-engine/forwardchainer/ForwardChainer.cc
```diff
@@ -62,81 +67,79 @@ Logger* ForwardChainer::getLogger()
...
+            std::copy(temp_result.begin(), temp_result.end(),
+                      std::inserter(products, products.end()));
+        }
     }
 
-    _log->info("[ForwardChainer] updating premise list with the "
-               "inference made");
-    _fcmem.update_potential_sources(product);
+    //done with this rule.
+    rule_weight.erase(r);
```
williampma: And when will the other rules be needed?  Why not just clear the whole thing?

opencog/rule-engine/forwardchainer/ForwardChainer.h
misgeatgit: Done.

opencog/rule-engine/forwardchainer/ForwardChainer.cc
misgeatgit: because do_step is called from within do_chain which may call it again based on number of iterations. I will clear this confusion once I separate the two methods.

misgeatgit:
@williampma I want to merge this so that I could work on the comments you raised.

williampma:
Should the others working more on FC also review this first?  @eddiemonroe @ngeiswei

ngeiswei:
I'm too far from the FC code to offer critical feedback (I'm looking for getting back to URE development but there's always more tasks coming at me).

ngeiswei:
@eddiemonroe would be the right person for reviewing that PR.

eddiemonroe:
Misgana, I will review.

But regarding:

> @williampma <https://github.com/williampma> I want to merge this so that
> I could work on the comments you raised.


can't you work on the issues raised before merging?

opencog/rule-engine/forwardchainer/ForwardChainer.cc
williampma: Why would you want to do that?  That's different from what Eddie written on the FC document.

williampma: Is this some idea from some private discussion?  Otherwise I think there's a misunderstanding again.

eddiemonroe:
@misgeatgit What's the purpose of having an FCMemory class separate from the ForwardChainer class itself? 

It seems like the fc *functions* are put in class ForwardChainer and the fc *data* into a separate class FCMemeory. Unless there is some reason to do so, it seems it would make sense to have the fc data in the same class as the fc functions that are using the data. Otherwise it just adds unnecessary complexity. But perhaps there is something I'm missing in terms of having a separate data class.

opencog/rule-engine/forwardchainer/ForwardChainer.cc
eddiemonroe: William's idea and the adopted plan is to choose each rule (by tournament/weight first) and attempt to unify with the selected source one-by-one until a match is found rather than iterating through and attempting to unify all of the rules. Please see the spec doc. Does that make sense?

opencog/rule-engine/forwardchainer/ForwardChainer.cc
```diff
@@ -144,19 +149,51 @@ void ForwardChainer::do_chain(ForwardChainerCallBack& fcb,
...
+
+            _fcmem.set_source(choose_next_source(_fcmem));
+
+            // Choose matching rules whose input matches with the source.
+            vector<Rule*> matched_rules = choose_rules(_fcmem);
+
+            //! If no rules matches the pattern of the source,
+            //! set all rules for candidacy to be selected by the proceeding step.
+            //! xxx this decision is made based on recent discussion.
+            //! it might still face some changes.
```
eddiemonroe: What's the thinking here for why to do that (set all rules for candidacy if none match with the selected source)? Do we really want to do that?

opencog/rule-engine/forwardchainer/ForwardChainer.cc
eddiemonroe: @misgeatgit Does this mean try unification using sub-atoms of the source? Remember we agreed on the call last week that the full source atom should unify somewhere within the rule premises. Right?

opencog/rule-engine/forwardchainer/ForwardChainer.cc
```diff
@@ -62,81 +69,79 @@ Logger* ForwardChainer::getLogger()
...
+        derived_rhandles = derive_rules(_fcmem.get_cur_source(), r, true);
+
+    HandleSeq temp_result;
+    if (not _fcmem.get_focus_set().empty()) {
 
-    _log->info("PRODUCTS...");
-    for(const auto& p:product)
-    {
-        _log->info("%s ", p->toShortString().c_str() );
+        for (Handle rhandle : derived_rhandles) {
```
eddiemonroe: @misgeatgit Ben's original spec is to choose one of the rule positions to unify the source with rather than applying all possible unification possibilities. If we are going to apply all of them as is being done here, that should be discussed with the group. In general every change from the spec should be run by the group and changed in the spec doc.

opencog/rule-engine/forwardchainer/ForwardChainer.cc
```diff
@@ -62,81 +69,79 @@ Logger* ForwardChainer::getLogger()
...
+        }
+
+    } else {
+
+        for (Handle rhandle : derived_rhandles) {
+            temp_result = apply_rule(rhandle);
+            std::copy(temp_result.begin(), temp_result.end(),
+                      std::inserter(products, products.end()));
+        }
     }
```
eddiemonroe: @misgeatgit In place of:
```
if (not _fcmem.get_focus_set().empty()) {

   for (Handle rhandle : derived_rhandles) {
       temp_result = apply_rule(rhandle, true);
       std::copy(temp_result.begin(), temp_result.end(),
                 std::inserter(products, products.end()));
   }

} else {

   for (Handle rhandle : derived_rhandles) {
       temp_result = apply_rule(rhandle);
       std::copy(temp_result.begin(), temp_result.end(),
                 std::inserter(products, products.end()));
   }
}
```
How about:
```
bool search_in_focus_set = !_fcmem.get_focus_set().empty()

for (Handle rhandle : derived_rhandles) {
       temp_result = apply_rule(rhandle, search_in_focus_set);
       std::copy(temp_result.begin(), temp_result.end(),
                 std::inserter(products, products.end()));
```
search_in_focus set could be a class member variable (perhaps set to true when the focus_set is set), so it doesn't need to be re-evaluated at every inference step.

opencog/rule-engine/forwardchainer/ForwardChainer.cc
misgeatgit: Not sub-atom of the source but sub-atoms of each premises of the rule.

opencog/rule-engine/forwardchainer/ForwardChainer.cc
misgeatgit: Agreed. if you look at apply_rule ,it has already got that flag default being false.Just forgot to use that flag passed to do_step. I will add that.

opencog/rule-engine/forwardchainer/ForwardChainer.cc
misgeatgit: I agree, the question being how to determine which unification will result in a useful inference on rule application.

opencog/rule-engine/forwardchainer/ForwardChainer.cc
bgoertzel: @eddiemonroe I could be wrong, but I believe the point is that for the application of the forward chainer to RelEx2Logic we want a different behavior than for the application of the forward chainer to PLN.  For R2L, what we want is to iterate through all the R2L rules, one after the other, applying each one to the {Atoms that came out of NL comprehension and are being processed via the R2L rules}.   For this use case, choosing rules via tournament selection is not appropriate and slows things down.   Recall that the R2L rules are similar to (though simpler than, and more abstract than) the RelEx2Frame rules you built a long time ago... your RelEx2Frame rule engine didn't use tournament selection but simply iterated through all the rules, applying them to the RelEx output, right?

I don't this @misgeatgit aims to replace tournament selection with "iterating through all the rules" for the PLN use-case, only for the R2L use-case....  The forward chainer should provide both options...

We discussed this extensively in the office  in HK yesterday, though the conversation was not captured and recorded online...

opencog/rule-engine/forwardchainer/ForwardChainer.cc
eddiemonroe: Great, thanks for the clarification.

opencog/rule-engine/forwardchainer/ForwardChainer.cc
eddiemonroe: My take is that's something to be implemented in the future (weighting of the different rule position unifications). 

Ben's specs say: Select among the potential positions via stochastic selection (in general this would also be done based on the weights of the rule positions in the current context).

At present we don't have weights for the rule positions, so for now I think it's just a random selection with equal weighting. 

BTW, I'm not saying that applying all the possible unifications (rather than picking a single one) is necessarily a bad idea. Just that if we deviate from the specs (especially Ben's) that it should be discussed first with Ben and the group rather than just being done without discussion.

opencog/rule-engine/forwardchainer/ForwardChainer.cc
eddiemonroe: Got it, so we need to have 2 different rule selection/application methods (1) tournament selection (of a single rule) and (2) iterate through all rules.

Let's add this to the google spec doc so we're all on the same page.

@misgeatgit, @williampma , @ngeiswei should we discuss how to best implement for the different use cases? Perhaps a PLNForwardChainer class and a R2LForwardChainer class that inherit from ForwardChainer? I think the 2 rule selection/application methods (single selection vs all rules) should be implemented separately from the use cases (PLN vs R2L), because the methods should be available to other future use cases.

misgeatgit: I like the idea of having two separate classes and it used to exist in my initial design( separating ForwardChaining mechanism from Policies like DefaultChaining and R2Lchaining...whatever that comes in the future.) but should we not complete this revision first? the design could be considered on next iteration. Rather we should focus on making sure things work according to the specs.

eddiemonroe: @misgeatgit Sure, so focusing on making sure things work according to the current specs, are you understanding how the code here is not working according to the specs?

PR #262, opened 2015-09-05
Throw error if bad atom type.

linas:


linas:
This addressees the opaque error message seen in one variant of a fix for the AnaphoraUTest crash.

PR #260, opened 2015-09-04
Make toShortString print AtomTable uuid

williampma:
Kind of useful for debugging stuff working in multiple atomspace.

PR #263, opened 2015-09-06
Revert "Make toShortString print AtomTable uuid"

pennachin:
Reverts opencog/atomspace#260


linas:
Please don't do this.

ngeiswei:
Please don't do what?

PR #279, opened 2015-09-09
Update scheme examples

pennachin:


pennachin:
Fixing `cog-bind-example.scm`, thanks @williampma

Issue #281, opened 2015-09-09
RemoveLink appears not to work in assign.scm Scheme example


pennachin:
Running this example in guile:

```
scheme@(guile-user)> (use-modules (opencog exec))
scheme@(guile-user)> (define (show-eval-links)
...         (cog-map-type (lambda (h) (display h) #f) 'EvaluationLink))
scheme@(guile-user)> (define to-be-added
...         (InsertLink
...                 (TypeNode "EvaluationLink")
...                 (PredicateNode "some property")
...                 (ListLink
...                         (ConceptNode "thing A")
...                         (ConceptNode "B-dom-ness"))))
scheme@(guile-user)> (show-eval-links)
$1 = #f
scheme@(guile-user)> (cog-execute! to-be-added)
$2 = (EvaluationLink (av 0 0 0)
   (PredicateNode "some property")
   (ListLink
      (ConceptNode "thing A")
      (ConceptNode "B-dom-ness")
   )
)

scheme@(guile-user)> (show-eval-links)
(EvaluationLink (av 0 0 0)
   (PredicateNode "some property")
   (ListLink
      (ConceptNode "thing A")
      (ConceptNode "B-dom-ness")
   )
)
$3 = #f
scheme@(guile-user)> (define to-be-removed
...         (RemoveLink
...                 (TypeNode "EvaluationLink")
...                 (PredicateNode "some property")
...                 (ListLink
...                         (ConceptNode "thing A")
...                         (ConceptNode "B-dom-ness"))))
scheme@(guile-user)> (cog-execute! to-be-removed)
$4 = #<Undefined atom handle>

scheme@(guile-user)> (show-eval-links)
(EvaluationLink (av 0 0 0)
   (PredicateNode "some property")
   (ListLink
      (ConceptNode "thing A")
      (ConceptNode "B-dom-ness")
   )
)
$5 = #f
```

The comments in the source suggest that the final call to `(show-eval-links)` shouldn't include that `EvaluationLink`.

pennachin:
Note: I'm running as of 00300be, a one line change to include module `(opencog exec)`. See #297

linas:
It would seem that RemoveLink is broken, and that there is no unit test for it (there is one for InsertLink and AssignLink.)

Now that we have PutLink, I'm thinking that maybe all three of these are obsolete...

linas:
well, I have a fix for this, but github is broken again, I can't merge it.  

github was working earlier today.

[...]

PR #282, opened 2015-09-10
Fix for issue #281

linas:


pennachin:
CircleCI build finished, reopening so it can be merged.

PR #286, opened 2015-09-11
AtomTable: Correctly add custom atom type with its own internal storage

williampma:
Custom atom types with its own internal Handle storage (such as VariableList) will store incorrect Handles if everything is added in one go.  This fixes that (see the unit test).

### DIFF
--- a/opencog/atomspace/AtomTable.cc
+++ b/opencog/atomspace/AtomTable.cc
@@ -432,27 +432,25 @@ Handle AtomTable::add(AtomPtr atom, bool async)
     Handle hexist(getHandle(atom));
     if (hexist) return hexist;
 
-    // If this atom is in some other atomspace, then we need to clone
-    // it. We cannot insert it into this atomtable as-is.  (We already
-    // know that its not in this atomspace, or its environ.)
-    AtomTable* at = atom->getAtomTable();
-    if (at != NULL) {
-        LinkPtr lll(LinkCast(atom));
-        if (lll) {
-            // Well, if the link was in some other atomspace, then
-            // the outgoing set will probably be too. (It might not
-            // be if the other atomspace is a child of this one).
-            // So we recursively clone that too.
-            HandleSeq closet;
-            for (const Handle& h : lll->getOutgoingSet()) {
-                closet.push_back(add(h, async));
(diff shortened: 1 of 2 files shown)


opencog/atomspace/AtomTable.cc
```diff
@@ -284,6 +284,9 @@ AtomPtr AtomTable::factory(Type atom_type, AtomPtr atom)
     } else if (BIND_LINK == atom_type) {
         if (NULL == BindLinkCast(atom))
             return createBindLink(*LinkCast(atom));
+    } else if (PATTERN_LINK == atom_type) {
+        if (NULL == PatternLinkCast(atom))
+            return createPatternLink(*LinkCast(atom));
```
linas: This probably should not be done, although the rest of the patch seems OK. PatternLink is meant to be a base-class, and not actually instantiated.

linas:
Same github configuration bug as the other pull request: all the circleci tests pass, there are no file-change conflicts, this pull request is fully mergeable, but the merge button remains still greyed out. This used to work just fine a few weeks ago.    Something is broken in the new github configuration.

linas:
The point here is that, operationally, it would be wrong to merge from master into this branch, because that would mess up the pull request. However, the github advice is recommending exactly what would be the wrong thing to do.  Viz: it states:

```
All checks have passed
1 successful check
Details Required  ci/circleci — Your tests passed on CircleCI!
This branch is out-of-date with the base branch
Merge the latest changes from master into this branch.
```

Of course, the branch is not out-of-date at all.

williampma:
Yeah, it doesn't seem to conflict with any of the new changes, but I guess CircleCI is assuming integrating with the newer stuff might not work, and so requires a merge with master anyway.  Looks like in the future, any newer PR that is merged will cause all unmerged older PR to be unmergeable until they are updated..................

Anyway, I rebased.  Hopefully it will be merge-able after the check complete.

I am leaving the PatternLink stuff in for now, since removing it will break the unit test.  Maybe it can be removed in some later PR.

linas:
I dunno. We'll see how this goes. If circleci+git == bzr, then its another black eye for circle-ci.

PR #295, opened 2015-09-14
Fix inheritance of the FunctionLink 

linas:
Now that we have a usable LambdaLink, the FuncitionLink should inherit from that.


linas:
Do not merge yet; this is a temporary holder for upcoming work.

linas:
Work don on this part, ready to merge

PR #296, opened 2015-09-14
AtomSpace: Clear AS on destruction

williampma:
Maybe this is a bad idea (too slow maybe), not sure.  I don't fully understand why this is happening yet.

It seems it is possible sometimes that the AtomSpace is not destroyed immediately, so if we have an extension

```
{ AtomSpace ex(&_as);   .... }
```

that goes out of scope, _as atoms' incoming set can still find atoms in `ex`.  It is like there's still some dangling shared pointer pointing to the atoms that was in `ex` (but in this case doing a `clear()` should not have affected it.....)  

In any case, adding a `clear()` to the destructor fixed it, so that atoms in _as will not see the destroyed one in `ex` after `ex` went out of scope.

### DIFF
--- a/opencog/atomspace/AtomSpace.cc
+++ b/opencog/atomspace/AtomSpace.cc
@@ -62,6 +62,7 @@ AtomSpace::AtomSpace(AtomSpace* parent) :
 
 AtomSpace::~AtomSpace()
 {
+    clear();
     // Be sure to disconnect the attention bank signals before the
     // atom table destructor runs. XXX FIXME yes this is an ugly hack.
     bank.shutdown();


linas:
Sounds weird. The AtomSpace destructor should call the AtomTable destructor, which should call the destructor for _atom_set which is an std::unordered_map which holds the actual AtomPtr's. The std::unordered_map destructor should then run and release these.

How about this: in ~AtomTable(), add a call to _atom_set.clear()  -- does that solve the problem? If so, it would be more efficient than the AtomSpace::clear()

williampma:
Hello Linas,

I just tried it, adding `_atom_set.clear()` at the end of ~AtomTable().  That does not fix it.  I guess the dangling copy is not the one in the _atom_set.  Maybe I will look at it some more to see where the real problem is at.

williampma:
Discovery 1:

Whatever these lines are doing https://github.com/opencog/atomspace/blob/master/opencog/atomspace/AtomTable.cc#L817-L821 is the one that is able to remove the dangling copy.  `clear()` calls `purge_atom()` and those lines are needed to really clean it.

williampma:
Humm... then I guess the better solution is add to ~Link() to make sure to purge itself completely from other atom's incoming set?

linas:
Interesting.  Yes, the lines L817-L821 are surely wrong.   Yes, they should be placed into ~Link() instead!  However, I don't quite understand why that would solve the issue, but depending on the case, maybe it could ... Does that solve the issue?

linas:
The diff wasn't obvious:

```
diff --git a/opencog/atomspace/Atom.h b/opencog/atomspace/Atom.h
index 3356b24..1ee864b 100644
--- a/opencog/atomspace/Atom.h
+++ b/opencog/atomspace/Atom.h
@@ -79,6 +79,7 @@ class Atom
     friend class Handle;          // Needs to view _uuid
     friend class SavingLoading;   // Needs to set _uuid
     friend class TLB;             // Needs to view _uuid
+    friend class Link;            // Needs to call remove_atom();
     friend class CreateLink;      // Needs to call getAtomTable();
     friend class DeleteLink;      // Needs to call getAtomTable();
     friend class ProtocolBufferSerializer; // Needs to de/ser-ialize an Atom
diff --git a/opencog/atomspace/AtomTable.cc b/opencog/atomspace/AtomTable.cc
index 0d56736..fa7d55a 100644
--- a/opencog/atomspace/AtomTable.cc
+++ b/opencog/atomspace/AtomTable.cc
@@ -813,12 +813,6 @@ AtomPtrSet AtomTable::extract(Handle& handle, bool recursiv
     nodeIndex.removeAtom(pat);
     linkIndex.removeAtom(atom);
     typeIndex.removeAtom(pat);
-    LinkPtr lll(LinkCast(atom));
-    if (lll) {
-        for (AtomPtr a : lll->_outgoing) {
-            a->remove_atom(lll);
-        }
-    }
     importanceIndex.removeAtom(pat);
 
     // XXX Setting the atom table causes AVChanged signals to be emitted.
diff --git a/opencog/atomspace/Link.cc b/opencog/atomspace/Link.cc
index 4559fd0..dd6a055 100644
--- a/opencog/atomspace/Link.cc
+++ b/opencog/atomspace/Link.cc
@@ -67,6 +67,10 @@ void Link::init(const std::vector<Handle>& outgoingVector)
 
 Link::~Link()
 {
+    for (AtomPtr a : _outgoing) {
+        a->remove_atom(LinkCast(shared_from_this()));
+    }
+
     DPRINTF("Deleting link:\n%s\n", this->toString().c_str());
 }
 
```

linas:
except that my last post wildly breaks the unit tests. Hmm very strange.

linas:
Oh, right. That is because, historically, the AtomTable was the thing that maintained the incoming set (and not the atoms themselves).  Thus, what is also missing is the add, which only the atom table does.  I'm experimenting now...

linas:
Ugh. here's what I (re-)discovered:  it is not valid to call shared_from_this() in a constructor or destructor.  So the naive approach breaks. Also, the insert code in AtomTable is a horrid mess.  Still trying to see what I can fix ...

linas:
Next: I (re-)discovered that you can't make a weak pointer out of anything except a strong pointer, and since the strong pointer is not yet available in the ctor, the Link ctor cannot insert itself into the incoming set of the atoms it holds.   Which means that it can only be done *after* the ctor runs. 

I think I can work around this for the ctor, by wraping createLink. (and the 20 other createBlahLink functions) however, it may be too late to work around the dtor.  I can't create a weak_ptr to the atom that is being destructed...  so this approach seems blocked.

williampma:
I see.  I am still trying to understand all these weak pointer stuff.

It seems strange that the incoming set will return the dead link, since https://github.com/opencog/atomspace/blob/master/opencog/atomspace/Atom.h#L319-L320 should have already prevented it by checking the weak pointer...

linas:
What do you mean by "will return a dead link"? its supposed to work like this:

1) the Atom contains a use-count. here by inheritance: `class Atom : public std::enable_shared_from_this<Atom>`

2) every 'strong pointer' (i.e. a regular 'shared pointer') increments the use count by one.

3) when a strong pointer is destroyed, the use count is decremented.

4) when the count reaches zero, the Atom is destroyed.

5) A weak pointer might still be pointing at a memory location that used to be occupied by the deleted atom.  Somehow it is able to tell if the atom has been deleted, I don't remember how.  You can read the code.

6) You can try to convert a weak pointer to a strong one. If the atom exists, the conversion will succeed. If the atom has been deleted, the conversion will fail.

So, yes, Lines L319-320 should prevent you from ever getting a pointer to an atom that has been deleted.  Can you prove that the atom destructor ran,  but that later you are getting a pointer to it (e.g. by print statements?)  

Note that you can directly view the use-count; I forget the name of the method, its something like count(). Also, the get() method returns the address of the thing being pointed at.  If I remember correctly, it should be impossible to see a count of zero, except maybe while the Atom destructor is running.

linas:
Here's how it works.  The code, for me, is in `/usr/include/c++/4.8/tr1/shared_ptr.h`

There are two use-counts: a regular use-count, and a weak use count.  Both are stored as part of Atom.

When the use-count drops to zero, the dispose() method prevents new shared pointers from being created.  Ergo, from then till forever, the use-count will stay at zero, so `use_count()` returns 0 and  `expired()` returns true. However, the Atom is not yet deleted; this happens only when the weak use count also drops to zero. (The weak use count is incremented by one with each weak pointer that points at the object).   Thus, the atom is only deleted when the incoming-set removal happens.

However ... as you pointed out L319-L320 should prevent you from getting a valid pointer to that atom.  So the atom may live on as a zombie, but you should not be able to actually point at it... 

The most efficient thing to do when destroyingthe AtomTable is to call just the lines L817-L821 and nothing else.  I can try this myself...

williampma:
Probably some code somewhere is keeping the use count from reaching 0.  Not sure where yet; maybe it's in my code.

Let me close this until I found out where it is.

linas:
William, I think that maybe pull req #303 might fix this for you.

I see now that the way the incoming set is handled is a huge ugly mess, but I don't see any fixes for it, as long as I can't call the needed methods from the Atom ctor and dtor.

PR #299, opened 2015-09-14
Allow Atom-valued terms in a SequentialAndLink

linas:
This set of changes allow atom-valued terms to appear in a SequentialAndLink.

It also redefines PutLink as a type of FunctionLink.


linas:
Do not merge yet; code is not complete; more is yet to come.

linas:
This batch of changes now complete.

PR #303, opened 2015-09-15
Incoming set hack

linas:
Possible fix for issue #296


williampma:
Thanks Linas.  This fix works.

Though I am still wondering what was keeping the shared_ptr from dying, so I probably will keep looking.

PR #307, opened 2015-09-16
PatternLink: Fixes circular share_ptr in PatternTerm

williampma:
Found the real culprit of https://github.com/opencog/atomspace/pull/296

PatternTerm usage at https://github.com/opencog/atomspace/blob/7ebc0384925db1a6965df2435feca11926e138a6/opencog/atoms/bind/PatternLink.cc#L723-L724 
combined with https://github.com/opencog/atomspace/blob/7ebc0384925db1a6965df2435feca11926e138a6/opencog/atoms/bind/PatternLink.cc#L732-L733
is circular.

 `ptm` parent is `parent`, while `parent` outgoing contains `ptm`.  This is keeping the handle stored within the PatternTerm object in memory even when everything is out of scope.

Converted `_outgoing` to use weak pointer fixed it (like how it's done on Atom's incoming set).

The fix at https://github.com/opencog/atomspace/pull/303 should no longer be needed.

### DIFF
--- a/opencog/query/PatternTerm.h
+++ b/opencog/query/PatternTerm.h
@@ -54,14 +54,16 @@ namespace opencog {
 
 class PatternTerm;
 typedef std::shared_ptr<PatternTerm> PatternTermPtr;
+typedef std::weak_ptr<PatternTerm> PatternTermWPtr;
 typedef std::vector<PatternTermPtr> PatternTermSeq;
+typedef std::vector<PatternTermWPtr> PatternTermWSeq;
 
 class PatternTerm
 {
 	protected:
 		Handle _handle;
 		PatternTermPtr _parent;
-		PatternTermSeq _outgoing;
+		PatternTermWSeq _outgoing;
 
 		// Number of QuoteLinks on the path up to the root including this
 		// term. Zero means the term is unquoted. Quoted terms are matched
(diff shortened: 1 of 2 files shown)


linas:
Looks OK to me. I can't merge this directly, and I can't think of any git-fu to work around the merge.  The instructions that github suggests, under the "an automatic merge cannot be performed, you can perform a manual merge on the command line." button do not actually work, I still get the same old CircleCI error as before.

opencog/query/PatternTerm.h
```diff
@@ -54,14 +54,16 @@ namespace opencog {
...
 typedef std::vector<PatternTermPtr> PatternTermSeq;
+typedef std::vector<PatternTermWPtr> PatternTermWSeq;
 
 class PatternTerm
 {
 	protected:
 		Handle _handle;
 		PatternTermPtr _parent;
-		PatternTermSeq _outgoing;
+		PatternTermWSeq _outgoing;
```
linas: It probably would have been easier to make the parent be weak.  Or, at least, that would make it consistent with the atomspace, where down-pointers are strong, and the up-ones (incoming set) are weak.

linas:
BTW, the fix #303 was "the right thing to do", anyway, irrespective of this problem, so I want to keep that.

I am not completely certain that the way incoming sets are constructed is 100% correct; the code in there is unfortunately a bit too complicated.

opencog/query/PatternTerm.h
williampma: I thought about it, but realized if parent is weak, there's no strong shared_ptr keeping the `root_term` in memory.

Maybe there are other ways to do it.

PR #317, opened 2015-09-19
experimental sketch of a pattern recognizer.

linas:
Persuant to discussions about AIML.


examples/aiml/recog.scm
```diff
@@ -0,0 +1,63 @@
...
+      (ConceptNode "I")
+      (VariableNode "$star")
+      (ConceptNode "you")
+   )
+   (ListLink
+      (ConceptNode "I")
+      (ConceptNode "love")
+      (VariableNode "$star")
+   )
+)
```
ngeiswei: Wow, so given an instance, the pattern recognizer will return all patterns in the atomspace that may have matched that instance, right? That definitely useful for the URE. Can it handle partial matches? Like if the **instance** is
```scheme
(ListLink
   (ConceptNode "I")
   (ConceptNode "love")
   (VariableNode "$star")
)
```
would it return
```scheme
(SetLink
   (ListLink
      (ConceptNode "I")
      (VariableNode "$star-1")
      (VariableNode "$star-2")
   )
   ...
)
```
Assuming
```scheme
(ListLink
   (ConceptNode "I")
   (VariableNode "$star-1")
   (VariableNode "$star-2")
)
```
is a pattern in the atomspace (I suppose present in a BindLink or such).

linas: Yes, that is the general idea.  The actual code is more like a "proof of concept"; it works in the simplest case and is probably broken in the more complex cases. I have not yet explored enough to even guess what might or might not work, what gotchas there might be in the code.  Yes, it was a Wow! moment for me too, when I first understood the possibilities; so yes, I'm excited by this. But more work is needed.

PR #323, opened 2015-09-22
Support disconnected optionals in Pattern Match

leungmanhin:
This allows us to have disconnected optional clauses in the pattern, but I am not sure if this is enough to handle every cases

### DIFF
--- a/opencog/query/PatternMatch.cc
+++ b/opencog/query/PatternMatch.cc
@@ -32,6 +32,7 @@
 #include "PatternMatch.h"
 #include "PatternMatchEngine.h"
 #include "PatternMatchCallback.h"
+#include "DefaultPatternMatchCB.h"
 
 using namespace opencog;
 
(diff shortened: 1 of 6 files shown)


linas:
Looks good to me. 

Note, however, that when it comes to mutli-component virtual links I believe that there are probably more  bugs relating to optionals. Doing this right took more time and energy than I had, back then.  So for example, use GreaterThankLink,, since this is virtual, the pattern will split into components.  The try something variations on this:

if (x < y) and inheritanceLink (y,z) and AbsentLink (z)

make x,y NumberNodes so that you can use GreaterThanLink.  Or use an EvaluationLink+ GroundedPredicateNode insted of the GreaterThanLink.   I'm fairly sure that if you mess with this, you'll find more bugs.   I bet there are also bugs involving ChoiceLink and AbsentLink, and possibly any kind of unordered link, although I think unordered links work fairly well, these days.  These are all dark corners.   

Don't try ChoiceLink until I finish a fix for #326 though, else it will be confusing.

opencog/query/PatternMatch.cc
```diff
@@ -378,13 +379,29 @@ bool PatternLink::satisfy(PatternMatchCallback& pmcb) const
...
 		clp->satisfy(gcb);
 
-		comp_var_gnds.push_back(gcb._var_groundings);
-		comp_term_gnds.push_back(gcb._term_groundings);
+		// Special handling for disconnected pure optionals -- Returns false to
+		// end the search if this disconnected pure optional is found
+		if (is_pure_optional)
+		{
+			DefaultPatternMatchCB* dpmcb = dynamic_cast<DefaultPatternMatchCB*>(&pmcb);
+			if (dpmcb->optionals_present()) return false;
```
linas: Misegana is right, This is guaranteed to crash if the callback is not the default callback! (the case will result in dpmcb in being null) ... some of the custom callbacks might not know what an "optional" is (maybe e.g. the fuzzy matcher??)

The reason that this is safe:
https://github.com/leungmanhin/atomspace/blob/master/opencog/query/Implicator.cc#L117-L120
is because, over there, we know, with 100% certainty that the default matcher is being used, because its all internal to that file; its just a utility. Ther is no way a non-default db could ever end up there.

PR #324, opened 2015-09-22
Backward Chainer: Temporary even more steps for the unit test

williampma:
Seems like the probability of fail is not low enough.

I have double checked again what could cause the unit test to be undeterministic, and it is the same as my previous conclusion.  Basically the usage of AtomPtr address ordering is a problem, since the address change between runs.

The first divergence between a passing test and a failing test is
```
[2015-09-21 07:12:59:223] [DEBUG] [BackwardChainer] Reverse grounded as (AndLink (stv 1.000000 0.000000)
  (InheritanceLink (stv 1.000000 0.000000)
    (VariableNode "$A-pln-rule-deduction") ; [1465][95]
    (VariableNode "$B-pln-rule-deduction") ; [1467][95]
  ) ; [1472][95]
  (InheritanceLink (stv 1.000000 0.000000)
    (VariableNode "$B-pln-rule-deduction") ; [1467][95]
    (ConceptNode "criminal") ; [52][1]
  ) ; [1523][95]
  (NotLink (stv 1.000000 0.000000)
    (EqualLink (stv 1.000000 0.000000)
      (VariableNode "$A-pln-rule-deduction") ; [1465][95]
      (ConceptNode "criminal") ; [52][1]
    ) ; [1524][95]
  ) ; [1525][95]
) ; [1526][95]
```
vs
```
[2015-09-21 07:12:48:417] [DEBUG] [BackwardChainer] Reverse grounded as (AndLink (stv 1.000000 0.000000)
  (InheritanceLink (stv 1.000000 0.000000)
    (VariableNode "$A-pln-rule-deduction") ; [1465][95]
    (VariableNode "$B-pln-rule-deduction") ; [1467][95]
  ) ; [1472][95]
  (NotLink (stv 1.000000 0.000000)
    (EqualLink (stv 1.000000 0.000000)
      (VariableNode "$A-pln-rule-deduction") ; [1465][95]
      (ConceptNode "criminal") ; [52][1]
    ) ; [1523][95]
  ) ; [1524][95]
  (InheritanceLink (stv 1.000000 0.000000)
    (VariableNode "$B-pln-rule-deduction") ; [1467][95]
    (ConceptNode "criminal") ; [52][1]
  ) ; [1525][95]
) ; [1526][95]
```

Basically, the same atom get added and assign different UUID & ordering (since unordered link's is resorted, and in this case by AtomPtr address since the AndLink was not in the atomspace).  AtomPtr usage is also in the different index inside AtomTable, which could also affect traversal ordering of Pattern Matcher.

Perhaps there's some way to restrict the unit test to be ran within the same virtual address, or force the OS to always assign the same address...  not sure.

### DIFF
--- a/tests/rule-engine/BackwardChainerUTest.cxxtest
+++ b/tests/rule-engine/BackwardChainerUTest.cxxtest
@@ -78,7 +78,7 @@ void BackwardChainerUTest::test_1rule_bc()
 	Handle soln = eval_.eval_h("(ConceptNode \"Fritz\")");
 
 	bc.set_target(target);
-	bc.get_config().set_maximum_iterations(150);
+	bc.get_config().set_maximum_iterations(300);
 	bc.do_chain();
 
 	VarMultimap results = bc.get_chaining_result();
@@ -112,7 +112,7 @@ void BackwardChainerUTest::test_1rule_impossible_bc()
 
 	// should NOT be possible to find the solution without deduction rule
 	bc.set_target(target);
-	bc.get_config().set_maximum_iterations(700);
+	bc.get_config().set_maximum_iterations(1000);
 	bc.do_chain();
 
 	VarMultimap results = bc.get_chaining_result();
@@ -147,7 +147,7 @@ void BackwardChainerUTest::test_2rules_bc()
 	Handle soln = eval_.eval_h("(ConceptNode \"West\")");
 
 	bc.set_target(target);
-	bc.get_config().set_maximum_iterations(700);
+	bc.get_config().set_maximum_iterations(1000);
 	bc.do_chain();
 
 	VarMultimap results = bc.get_chaining_result();
@@ -180,7 +180,7 @@ void BackwardChainerUTest::test_tvq_bc()
 	                 "   (ConceptNode \"green\"))");
 
 	bc.set_target(target);
-	bc.get_config().set_maximum_iterations(150);
+	bc.get_config().set_maximum_iterations(300);
 	bc.do_chain();
 
 	TSM_ASSERT("Incorrect strength after BC!",
@@ -217,7 +217,7 @@ void BackwardChainerUTest::test_tvq_impossible_bc()
 	                             "   (ConceptNode \"Frog\"))");
 
 	bc.set_target(target);
-	bc.get_config().set_maximum_iterations(150);
+	bc.get_config().set_maximum_iterations(300);
 	bc.do_chain();
 
 	TSM_ASSERT("Incorrect strength of target after BC!", target->getTruthValue()->getMean() < 0.1f);
@@ -269,7 +269,7 @@ void BackwardChainerUTest::test_focus_set()
 	Handle soln2 = eval_.eval_h("(ConceptNode \"Fritz\")");
 
 	bc.set_target(target, focus_set);
-	bc.get_config().set_maximum_iterations(150);
+	bc.get_config().set_maximum_iterations(300);
 	bc.do_chain();
 
 	VarMultimap results = bc.get_chaining_result();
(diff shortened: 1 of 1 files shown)


linas:
The OS always loads at a different address each time, this is "address space randomization" and is done to foil viruses, by making it harder to a virus to guess what address is where.

I suppose one could order unordered links by UUID, but note that UUID assignment is not fixed: with multiple threads, it would be first-come, first-served.  With a distributed atomspace, or with the sQL backend, UUID assignment could be even more chaotic.   Neither unit tests, nor anything else, should make any assumptions about what order UUID's are issued in.

ngeiswei:
There are 2 separate issues here:

1. Regarding the AtomSpace: we want to have high semantics syntactic correlation (that is why we want a reduct engine for that atomspace). Having a changing order across atomspace instances goes against that.

2. Regarding testing stochastic behaviors. If you can't turn it into a deterministic one, then pick up a convergent property and put a shit load of computation to almost surely get it. Could be a mean of distribution up to some error, or like in the this case the presence of a solution.

linas:
Seems odd to me; the whole point of making AndLink unordered is that nothing should depend on that order.  The results reported by the pattern matcher will not be order-dependent. So what part of the code is assuming that And is ordered?

williampma:
No, it doesn't assume And is ordered, but it will break out the stuff inside the And and add each as new potential target into a target set.  Target from the target set are selected via some basic roulette selection (using fixed seed for the random engine).  If the ordering inside the set is different, the target will be visited in different order.

Different target visiting order could result in different inference path, and some bad luck path would mean it takes longer (more steps) to reach the solution.

I guess increasing the number of steps is option 2 suggested above?  I will increase the number of steps if I see the unit test fail again, until I see no more.

ngeiswei:
@williampma yes, I guess you'll have to go with option 2 indeed. I still think we should strive to have reproducible behaviors as much as possible.

linas:
Ahh. Ugh.

linas:
One solution would be to compute some hash for atoms, and use that for comparison.  This wuold slow down addition of unordered atoms into the atomspace.  Also, unless the hash is large, then hash collisions are likely.  even a 64-bit hash would be too small to avoid collisions, we'd need a 128 bit or 256 bit hash, e.g. md5. that really starts getting slow.

ngeiswei:
I guess we could think of a hash without collision, using the hypergraph structure to come up with a total order and turn that into a number (maybe an arbitrary big one), but that would be very slow.

ngeiswei:
Hmm, or maybe not that slow! Because it would be done incrementally. That to determine the hash a link, you would compose it from the hashes of its outgoings.

So basically we're back to the concept of an UUID, expect this one would be determined by the AtomSpace structure, and it would determined regardless of whether the atom is added in an atomspace or not. Which would migrate a bit of the cost of inserting an atom in an atomspace to its creation. But then creation would be more costly, which isn't too good. But can we do otherwise!

The other option would be to have the UUID calculation done lazily, only when/if required!

ngeiswei:
Oh, I thinking of something else!!!

The problem with the UUID proposal above is that it would be really gigantic. So there is another option, we keep the UUID the way they are, but instead of comparing atoms without UUID based on their pointer, we compare based on their content (so based on a perfect order determined by the structure and content of an atom).

This would make comparison of atoms not in the any atomspace slow, would that turn into a bottleneck? I don't know...

ngeiswei:
Although it would only make it slow for atoms that are similar, as soon as a difference would be spotted the order would be determined, without having to travel along the atom's outgoings, so in practice that wouldn't be that slow. I start to think that this is the way to go...

linas:
Worth a try, I guess.  The atomspace has several indexes that use a less-than ordering for atoms that are probably very performance-sensitive.  You can use the `/opencog/benchmark/atomspace_bm` tool to measure before & after performance.

While designing the 'protoatom' (unrelated conversation, I discovered that a minor change to the atom less-than compare cut link insertion performance in half! Now, Ben really really wants this, but that's a big hit.

PR #325, opened 2015-09-22
Check if the PatternTerm is undefined before doing getHandle()

leungmanhin:
To avoid getting segfault, which can happen when doing fuzzy pattern matching.


linas:
Looks good to me, can't merge because of circleci.

PR #328, opened 2015-09-23
added distributed processing with gearman

yantrabuddhi:


opencog/guile/SchemePrimitive.h
```diff
@@ -140,6 +141,7 @@ class SchemePrimitive : public PrimitiveEnviron
 			H_HS,  // return handle, take handle and string
 			H_HHH, // return handle, take handle, handle and Handle
 			H_HTQ, // return handle, take handle, type, and HandleSeq
+			U_SSB, //return UUID, take string,string,boolean
```
linas: Can these be placed in alphabetical order?  here and line 100 and line 270

opencog/guile/SchemePrimitive.h
```diff
@@ -624,6 +638,7 @@ class SchemePrimitive : public PrimitiveEnviron
 		DECLARE_CONSTR_3(H_SQQ, h_sqq, Handle, const std::string&, const HandleSeq&, const HandleSeq&)
 		DECLARE_CONSTR_1(Q_H, q_h, HandleSeq, Handle)
 		DECLARE_CONSTR_3(Q_HTI, q_hti, HandleSeq, Handle, Type, int)
+		DECLARE_CONSTR_3(U_SSB, u_ssb, UUID,const std::string&,const std::string&, bool)
```
linas: ... and here and line 723 too

opencog/query/CMakeLists.txt
```diff
@@ -1,3 +1,30 @@
+IF (HAVE_GEARMAN)
+IF (HAVE_GUILE)
```
linas: This is wrong.  The test for gearman should not disable the test for guile.

linas: Oops sorry, I misread that; ignore previous commment. Can you change to to HAVE_GUILE AND HAVE_GEARMAN ?

opencog/query/CMakeLists.txt
```diff
@@ -1,3 +1,30 @@
...
+		Distributed/DistSCM.h
+		DESTINATION "include/opencog/query"
+	)
+ELSE(HAVE_GUILE)
+	MESSAGE(FATAL_ERROR "Cannot find guile")
+ENDIF (HAVE_GUILE)
+
+ELSE(HAVE_GEARMAN)
+	MESSAGE(FATAL_ERROR "Cannot find libgearman library")
+ENDIF(HAVE_GEARMAN)
```
linas: Also, can you move this closer to the bottom of the file, so it gets built after everything else?

opencog/query/Distributed/DistSCM.cc
```diff
@@ -0,0 +1,230 @@
...
+    self->init();
+}
+
+void DistSCM::init(void)
+{
+  //blocking slave mode
+	define_scheme_primitive("set-slave-mode", &DistSCM::slave_mode,
+                            this, "dist-gearman");
+  //exits the slave mode after last request is served
+	define_scheme_primitive("set-master-mode", &DistSCM::set_master_mode,
```
linas: indentation is inconsistent

opencog/query/Distributed/DistSCM.cc
```diff
@@ -0,0 +1,230 @@
...
+
+gearman_return_t DistSCM::worker_function(gearman_job_st *job, void *context)
+{
+  const char *workload= (const char *)gearman_job_workload(job);
+  const size_t workload_size= gearman_job_workload_size(job);
+
+  //std::cout << "Recieved " << workload_size << " bytes" << std::endl;
+  if (workload_size>MAX_SCM_CODE_STRING+1) return GEARMAN_ERROR;
+  char scm_str[MAX_SCM_CODE_STRING+1];
+  strcpy(scm_str,workload);
```
linas: why is there a limit on the size?? why do we need to copy the string?

opencog/query/Distributed/DistSCM.cc
```diff
@@ -0,0 +1,230 @@
...
+
+  SchemeEval evl((AtomSpace*)context);
```
linas: Its much more efficient if you call the static function 
```
SchemeEval* SchemeEval::get_evaluator(AtomSpace* as)
```
to get a copy of the evaluator. This avoids expensive setup.

opencog/query/Distributed/DistSCM.cc
```diff
@@ -0,0 +1,230 @@
...
+  evl.init_scheme();
+  //std::cout<<scm_str<<"\n";
+  Handle result=evl.eval_h(scm_str);//can we check if handle is valid? how to get current atomspace
+  UUID vl=result.value();
+  if (Handle::UNDEFINED==result)std::cerr<<"invalid handle \n";
```
linas: standard practice is the use the logger for error messages

opencog/query/Distributed/DistSCM.cc
```diff
@@ -0,0 +1,230 @@
...
+  const char *workload= (const char *)gearman_job_workload(job);
+  const size_t workload_size= gearman_job_workload_size(job);
+
+  //std::cout << "Recieved " << workload_size << " bytes" << std::endl;
+  if (workload_size>MAX_SCM_CODE_STRING+1) return GEARMAN_ERROR;
+  char scm_str[MAX_SCM_CODE_STRING+1];
+  strcpy(scm_str,workload);
+
+  SchemeEval evl((AtomSpace*)context);
+  evl.init_scheme();
```
linas: Do not do this ever; it will  be initialized by the time that your code is called.

opencog/query/Distributed/DistSCM.cc
linas: Actually, you almost surely should not check the handle here. If there is any error during processing, an exception is thrown. If the exception is in guile, its automatically converted to a C++ exception.

Since you never use the handle except for this error-check, I'm thinking there is no need for this error check at all

opencog/query/Distributed/DistSCM.cc
```diff
@@ -0,0 +1,230 @@
...
+  //std::cout<<scm_str<<"\n";
+  Handle result=evl.eval_h(scm_str);//can we check if handle is valid? how to get current atomspace
+  UUID vl=result.value();
+  if (Handle::UNDEFINED==result)std::cerr<<"invalid handle \n";
+
+  evl.eval("(sql-store)");
```
linas: This is going to be extremely inefficient; it will dump the entire atomspace to the database, which can take seconds or longer.  What you almost surely need to do is to store single atoms, using the store-atom function. If the atom is a link, it will automatically store the outgoing set for you.

opencog/query/Distributed/DistSCM.cc
```diff
@@ -0,0 +1,230 @@
...
+
+const std::string& DistSCM::slave_mode(const std::string& ip_string,const std::string& workerID)
+{
+	AtomSpace* atoms_ptr=SchemeSmob::ss_get_env_as("set-slave-mode");
+	//std::cout<<"atomspace "<<atoms_ptr<<"\n";
+
+    if ((worker= gearman_worker_create(NULL)) == NULL)
+    {
+      std::cerr << "Memory allocation failure on worker creation." << std::endl;
+      return false_string;
```
linas: The opencog convention is to always throw exceptions and never to return error values.  This convention makes it much much much easier to deal with failures and to debug errors.

linas: I mean, good exception support was probably the most important advance in programming languages that came out of the 1990's so it is kind-of crazy to write programs as if we are still in the 1980's.  Error checking is nuts; just assume the code works!

opencog/query/Distributed/DistSCM.cc
```diff
@@ -0,0 +1,230 @@
+/*
+ * DistSCM.cc
```
linas: This framework has essentially nothing to do with distributed queries at all; its a general distribution framework. it would be much better if you moved this all to the directory /opencog/persist  Yes, "persist" is a funny name for this, but that is where zmq sits, and its where hypertable sits, so its where this should sit as well.

opencog/query/Distributed/DistSCM.cc
linas: actually, I'm thinking that you should not do sql-anything at all, in this code. Let the scheme code that is running in the workers decide if it needs to call sql.  If the worker found zero atoms,, its pointless to try to "store" them. The worker is going to have a much better idea of what the correct thing to do is, than the dist framework will.

opencog/query/Distributed/DistSCM.cc
linas: I mean, Kurt Akeley personally showed me how crazy error checking was, in 1990. We had to engineer a hackfest around this, because at the time, compilers did not yet support exceptions. We invented something very clever: it was almost like an exception, but it did not need compiler support.  Later on, Kurt got awards for this and many other inventions.

yantrabuddhi: Hi Linas,
Thanks for the review, I will make those changes.
Regards,
Mandeep

yantrabuddhi:
Hi, I am abandoning this branch and making pull request from another due to some untraceable problem causing strange build behavior in this branch

PR #332, opened 2015-09-24
Remove a function definition from cython binding

AmeBel:
* The `scheme_wrapper.__init__(AtomSpace a)` is supposed to initialize something, which is not clear.
* The `scheme_wrapper.pyx` module is wrapping functions not a class, so doesn't make sense, and only results in confusion for a user, provided it doesn't have some other purpose.
* It could be their, to speedup the `SchemeEval` initialization stage, but couldn't observe any significant difference when running the python benchmark test, before and after the change. That could be b/c my setup isn't correct, not sure.


linas:
thanks.

PR #334, opened 2015-09-25
Extension to AtomUtils

misgeatgit:
I wanted to compare similarity of two atoms without creating bindlink and the pattern matcher.


opencog/atomutils/AtomUtils.cc
```diff
@@ -304,4 +304,58 @@ HandleSeq get_predicates_for(const Handle& target,
...
 
+bool are_similar(const Handle& h1, const Handle& h2,bool strict_type_match)
+{
+    if (h1 == h2) return true;
+
+    if (NodeCast(h1) and NodeCast(h2)) {
+        if (strict_type_match and (h1->getType() != h2->getType()) )
+            return false;
+        else
+            return true;
```
ngeiswei: you may write
```c++
return !strict_type_match or h1->getType() == h2->getType();
```
instead of if (strict...).

opencog/atomutils/AtomUtils.cc
```diff
@@ -304,4 +304,58 @@ HandleSeq get_predicates_for(const Handle& target,
...
+    }
+
+    LinkPtr lh1(LinkCast(h1));
+    LinkPtr lh2(LinkCast(h2));
+
+    if (lh1 and lh2) {
+        if(strict_type_match and (lh1->getType() != lh2->getType()))
+        {
+         return false;
```
ngeiswei: Maybe it's time to invest in a real editor (emacs, vim, etc). :-)

opencog/atomutils/AtomUtils.cc
misgeatgit: :+1: will do that.

opencog/atomutils/AtomUtils.cc
misgeatgit: Why?

ngeiswei: Oh, I mean the indentation of 'return false;' isn't correct, and I can only assume that your editor didn't auto-correct it as you typed. What development environment do you use?

misgeatgit: I use eclipse but that particular line I manually edited it and didn't correct it. I've fixed it now.

opencog/atomutils/AtomUtils.cc
```diff
@@ -304,4 +304,53 @@ HandleSeq get_predicates_for(const Handle& target,
...
+
+    if (NodeCast(h1) and NodeCast(h2)) 
+        return !strict_type_match or h1->getType() == h2->getType();
+    
+    LinkPtr lh1(LinkCast(h1));
+    LinkPtr lh2(LinkCast(h2));
+
+    if (lh1 and lh2) {
+        if(strict_type_match and (lh1->getType() != lh2->getType())){
+         return false;
```
linas: bad indentation

linas:
There is a much simpler implementation for that this is doing: if strict_type_match is stet to true, then its enough to just compare handles: if the atoms are in the atomspace, then they will be strictly identical. You don't need to walk the thing.

So, in fact, the only interesting case is where strict_type_match is false; so you don't need that a a flag.  If its false, you do need a recursive visit, but its enough to just compare the arity of each link; you don't need to get types.

linas:
Ooops, remove my previous remakr, its ioncorrect; I misunderstood what you were trying to do.

williampma:
The indentations on the unit test need to be fixed too.

misgeatgit:
I have corrected the indentation.

PR #337, opened 2015-09-26
distributed processing with gearman

yantrabuddhi:


yantrabuddhi:
fixed problems in #328 as suggested by @linas , some untraceable problem crept in and made unrelated tests fail so shifted the code to a new clean branch
- Changed string error returns to exceptions
- Removed unnecessary string size limits in code
- sorted position of new u_ssb signature in scheme primitive header
- improved cmake HAVE_GEARMAN if block
- corrected indentation to be consistent
- used SchemeEval* SchemeEval::get_evaluator(AtomSpace* as) instead of doing by constructor
- removed call to "sql-store", it is upto scheme string code to store created atoms
- removed unneeded init_scheme call
- updated readme.md

opencog/query/Distributed/DistSCM.cc
```diff
@@ -0,0 +1,214 @@
...
+      return GEARMAN_ERROR;//on gearman error return, no result is sent to master and job request stays in que
+    }
+    //std::cout<<"gearman data sent..."<<"\n";
+    return GEARMAN_SUCCESS;
+}
+
+const std::string& DistSCM::slave_mode(const std::string& ip_string,const std::string& workerID)
+{
+	  AtomSpace* atoms_ptr=SchemeSmob::ss_get_env_as("set-slave-mode");
+	  //std::cout<<"atomspace "<<atoms_ptr<<"\n";
```
linas: You still seem to have multiple indeptation issues; I guess your editor makes indentation invisible?

opencog/query/Distributed/DistSCM.cc
```diff
@@ -0,0 +1,214 @@
+/*
+ * DistSCM.cc
+ *
```
linas: This appears to have nothing at all to do with query, I will merge this pull request, but then I will immediately move this code to a new directory.

linas:
Argh. I just merged this, but it BREAKS THE BUILD. -- PLEASE TEST YOUR CODE SO THAT YOU DO NOT BREAK THE BUILD -- TESTING WITH CIRCLE-CI IS INSUFFICIENT TO DISCOVER BUILD BREAKS!

linas:
I will push a fix for this shortly.

linas:
... and of course, circle-ci is preventing me from pushing a fix for the build break.  Sorry about that, you will have to wait a while to get the fix for this. Arghh.

linas:
So .. pull request #338 fixes the indentation issues, and moves the code to new directories: opencog/persist/gearman and likewise for the unit test.  it also automates the CMakefile to look for the gearmand   It also hacks the unit test; now it passes for me. However, it still does not pass in CircleCI, and I cannot tell why; CircleCI does not display the test output.

yantrabuddhi:
Hi Linas,
That is very strange. I tested in three separate places including my local
machine, the docker container, then circleci. I wonder how it breaks the
build now. I had some strange issues in previous repository after merge, so
I shifted to a fresh one, but I tested it was working.
Regards,
Mandeep

linas:
So, in my cleanup #338, I added a check to the CMakefile that detects whether `gearmand` is running. It then reports a status; look for the message near the end of the CMakefile output; its immediately after the SQL checks, so maybe 50 or 60 lines in. Look at any CircleCI output list, it currently prints this:
```
CMake Warning at tests/persist/gearman/CMakeLists.txt:25 (MESSAGE):
  gearman-job-server is not running; will not run the GearmanUTest
```
This is enough to explain why the test fails in circleci.

linas:
Disregard the last post. It seems that cmake is run not once, but three times!  The third time is when building the unit tests. That time, the gearmand is found, and so the unit test is built and run!

I'm guessing that the reason its failing is because either the scm file is not found, or the library is not found.  This is probably because `make install` is not called before running the tests. One can over-ride this step by hard-coding paths into the GearmnUTest;  I tried adding the build paths, by guessing what they might be, but this was not enough to solve the problem.

PR #339, opened 2015-09-26
Attempt to fix the gearman unit test in CricleCI.

linas:


linas:
Never mind -- the log messages clearly state:
```
CMake Warning at tests/persist/gearman/CMakeLists.txt:25 (MESSAGE):
  gearman-job-server is not running; will not run the GearmanUTest
```

PR #341, opened 2015-09-28
more circle-ci guess-work

linas:
blind guess as to what might be wrong

### DIFF
--- a/tests/persist/gearman/CMakeLists.txt
+++ b/tests/persist/gearman/CMakeLists.txt
@@ -22,6 +22,7 @@ IF (GEARMAN_JOB_SERVER)
 		RESULT_VARIABLE RC)
 
 	IF (RC)
+		SET(GEARMAN_RUNNING 0)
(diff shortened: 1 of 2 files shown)


AmeBel:
@linas FYI, all tests that are run by circle-ci are run inside a container. The command run is https://github.com/opencog/atomspace/blob/master/circle.yml#L13 

Thus any change to https://github.com/opencog/ocpkg/blob/master/ocpkg or https://github.com/opencog/docker/blob/master/opencog/base/Dockerfile will propagate to the circle-ci after a rebuild is triggered @ https://hub.docker.com/r/opencog/opencog-deps/. when there is a merge on the ocpkg and docker repos a rebuild is triggered.

linas:
thanks, but the problem is that I cannot see what is actually failing, because the log messages are discarded.  So I have to guess what's wrong, which makes deugging difficult when circle-ci tests fail

AmeBel:
FYI, https://github.com/opencog/atomspace/pull/374 resolves the log access issue for this repo.

Issue #147, opened 2015-07-15
Replace all the dbgprt in the pattern matcher code by fine logging


ngeiswei:
It's all in the title. It's not convenient to uncomment DEBUG macros. And it's clearly something that is being done often.

linas:
the problem with fine logging is that it hurts performance significantly.  We really should turn both fine logging and debug logging into compile-time functions, so that no code is generated if they are disabled.

The other problem with the design of the logger is that you cannot turn it on only for some modules, but not others.   Typically, you don't want to turn on debugging for everything, but only for the system that you are actually debugging at the moment.   

I conclude: to make the logger actually useful, it needs to be radically redesigned; right now, its just not really usable. I usally avoid using it for exactly this reason: the #ifdef DEBUG method is much easier to use, and much more powerful, that the Logger class.

 So: do NOT convert to fine logging until the Logger class is fixed so that its .. well, not totally broken.

linas:
If I remember correctly, the glib libraries had a fairly flexible logging system (glib from gnome/gtk).  I do not recall if it was a compile-time option, though.   I'm guessing KDE has one too. Perhaps there is some other industry-standard logger maybe in boost, that is more sophisticated than what we have.

linas:
Anyway, swithcng to a higher-quality logger would not avoid the need to #define DEBUG, since you would still need to leave it turned off for the normal, non-debug case.

ngeiswei:
Agreed, switching to more powerful logging system would be good.

Regarding performances, I usually use conditionals with logger().isFineEnabled() to only execute the logging code if it has any computational load.

Being able to unable or disable at run-time is a good thing as it prevents from wasting time recompiling, but sure only if that doesn't introduce an overhead.

All that said, I find myself enabling/disabling PM debugs **very often**, I would personally find it more convenient to use the logging system, however limited it currently is, assuming it doesn't introduce any performance overhead, of course.

Anyway, I don't have time to do that now, and yeah, maybe it's about time to move to a better logging system.

linas:
I'm surprised .. why would you ever need to enable debug on the pattern matcher?

linas:
(much less "very often")

ngeiswei:
I don't know if that was meant to be a joke, but the URE relies enormously on the PM, debugging the URE often entails knowing what the PM does.

linas:
[...]

PR #343, opened 2015-09-29
Replace stdout debug by log in pattern matcher

ngeiswei:
This is to address issue #147 . All stdout debug messages have been replaced by fine debugging. The reason I use fine is because most pattern matcher tests enable debug level and this would slow down considerably run-time pattern matching. Generally I think we can assume that log levels behave that way

1. INFO: contain the bare minimum
2. DEBUG: contain useful information for debugging but do not slow down run-time substantially
3. FINE: may slow down run-time substantially and may not be used in production

The difference in performance is at DEBUG level is not significant, see statistics obtained by running scripts/query/benchmark_query.sh (which runs 20 times all pattern matcher unit tests, and spits out stats).

```
Before replacing stdout by log
========================
N   min    max    sum     mean     stddev
20  13.57  17.99  309.61  15.4805  1.43348
```

```
After replacing stdout by log
=======================
N   min    max    sum    mean    stddev
20  13.77  17.86  313.1  15.655  1.52332
```

The new version is 1% slower, but I don't think it is statistically significant given the standard deviation.

I think it's ready to be merged, but @linas I'll probably let you merge just in case.

### DIFF
--- a/opencog/query/InitiateSearchCB.cc
+++ b/opencog/query/InitiateSearchCB.cc
@@ -35,14 +35,6 @@
 
 using namespace opencog;
 
-// Uncomment below to enable debug print
-// #define DEBUG
-#ifdef DEBUG
-#define dbgprt(f, varargs...) printf(f, ##varargs)
-#else
-#define dbgprt(f, varargs...)
-#endif
-
 /* ======================================================== */
 
 InitiateSearchCB::InitiateSearchCB(AtomSpace* as) :
@@ -324,10 +316,11 @@ bool InitiateSearchCB::neighbor_search(PatternMatchEngine *pme)
 		_starter_term = ch.start_term;
 
 		_root = clauses[bestclause];
-		dbgprt("Search start node: %s\n", best_start->toShortString().c_str());
-		dbgprt("Start term is: %s\n", _starter_term == Handle::UNDEFINED ?
-		       "UNDEFINED" : _starter_term->toShortString().c_str());
-		dbgprt("Root clause is: %s\n", _root->toShortString().c_str());
+		LAZY_LOG_FINE << "Search start node: " << best_start->toShortString();
+		LAZY_LOG_FINE << "Start term is: "
+		              << (_starter_term == Handle::UNDEFINED ?
+		                  "UNDEFINED" : _starter_term->toShortString());
+		LAZY_LOG_FINE << "Root clause is: " <<  _root->toShortString();
 
 		// This should be calling the over-loaded virtual method
 		// get_incoming_set(), so that, e.g. it gets sorted by attentional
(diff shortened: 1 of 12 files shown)


opencog/query/InitiateSearchCB.cc
```diff
@@ -565,14 +558,12 @@ bool InitiateSearchCB::link_type_search(PatternMatchEngine *pme)
...
-	size_t i = 0;
-#endif
+	size_t i = 0, hsz = handle_set.size();
 	for (const Handle& h : handle_set)
 	{
-		dbgprt("yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy\n");
-		dbgprt("Loop candidate (%lu/%lu):\n%s\n", ++i, handle_set.size(),
-		       h->toShortString().c_str());
+		LAZY_LOG_FINE << "yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy\n"
+		              << "Loop candidate (" << i+1 << "/" << hsz << "):\n"
```
linas: You've introduced a bug here; i no longer increments.  I will fix this shortly n a different pull request.

ngeiswei: Oh sorry about that!

opencog/query/InitiateSearchCB.cc
```diff
@@ -667,16 +661,14 @@ bool InitiateSearchCB::variable_search(PatternMatchEngine *pme)
...
-	size_t i = 0;
-#endif
+	size_t i = 0, hsz = handle_set.size();
 	for (const Handle& h : handle_set)
 	{
-		dbgprt("zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz\n");
-		dbgprt("Loop candidate (%lu/%lu):\n%s\n", ++i, handle_set.size(),
-		       h->toShortString().c_str());
+		LAZY_LOG_FINE << "zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz\n"
+		              << "Loop candidate (" << i+1 << "/" << hsz << "):\n"
```
ngeiswei: Same bug here.

linas:
I dunno how much you use this, but I am getting to really hate this change. It makes debugging extra hard: there is much more work to do, much more typing, the logs are much hard to read.   This simply ws not a good idea.

This needs to be redesigned so that (a) it is easy to use,  (b) the log messages are readable.   The old way of doing things satisfied (a) and (b), the new way does not.

This is now the 5th or 6th time I have cursed this change, and I have Io curse it a few more time, I plan to revert it.

ngeiswei:
Much more work to do!!! It is so much easier! No need to recompile, just add a line of scheme code where you want to use it. I don't see how it's harder to use.

linas:
Its like 5 lines of scheme code, impossible to memorize, so you have to
either copy it somewhere and try to remember where you copied it, or you
need to reconstruct it from sratch which means using ,apropos, figuring out
which calls you need to make to set it up, and how to use them, and tehn
make several mistakes before you can actually use it.

Oh, and to add insult to injury, if you recompile the code, you have to do
this again .. and again and again, dozens of times a day.  cut-n-paste the
same boilerplate over and over. Its a major hassle.

With a compile time directive, its easy to find, you enable it once in the
morning you never have to touch it again, until you are done.   It would be
great if it was that easy, but its just not.

Oh, and then also everything else in the cogserver that is set to log fine
spoils the output, so you have to go in by hand, and change all those OTHER
users of fine(), and disable them, and recompile.   The logger is
absolutely horrific in terms of ease-of-use.

Issue #342, opened 2015-09-29
Can't load atoms from postgres


leungmanhin:
Got this error after loading all atoms at height 0:

```
	...
	Loaded 640000 atoms.
	Loaded 650000 atoms.
	Loaded 660000 atoms.
	Loaded 670000 atoms.
Loaded 670304 atoms at height 0
[2015-09-29 07:26:23:514] [ERROR] Caught signal 11 (Segmentation fault) on thread 140001601058880
	Stack Trace:
	2: basic_string.h:539	~basic_string()
	3: CogServerMain.cc:79	_Z7sighandi()
	4: ??:0	killpg()
	5: AtomTable.cc:273	  opencog::AtomTable::inEnviron(std::shared_ptr<opencog::Atom>)
	6: shared_ptr_base.h:545	~__shared_count()
	7: stl_vector.h:920	  std::vector<opencog::Handle, std::allocator<opencog::Handle> >::push_back(opencog::Handle&&)
	8: shared_ptr_base.h:545	~__shared_count()
	9: PersistModule.cc:90	  opencog::PersistModule::do_load(opencog::Request*, std::list<std::string, std::allocator<std::string> >)
	10: stl_list.h:378	~_List_base()
	11: CogServer.cc:297	  opencog::CogServer::processRequests()
	12: CogServer.cc:237	  opencog::CogServer::runLoopStep()
	13: CogServer.cc:200	  opencog::CogServer::serverLoop()
	14: CogServerMain.cc:137	main()
	15: libc-start.c:321	__libc_start_main()
	16: ??:0	_start()
```

linas:
I'm seeing a similar crash in the unit tests.  Looks like the unit tests have not been running.

linas:
Pull request #344 should fix this.  CircleCI is down right now, so I can't merge.  

Root problem is that the SQL tests haven't been run in a long time, and various brand new bugs have shown up, as a result.

leungmanhin:
Thanks Linas.

Just tested it again, many more atoms can be loaded successfully, but unfortunately it failed again while loading atoms at height 1... with the same exception:

```
	...
	Loaded 2280000 atoms.
	Loaded 2290000 atoms.
	Loaded 2300000 atoms.
	Loaded 2310000 atoms.
[2015-09-30 07:34:38:063] [ERROR] Caught signal 11 (Segmentation fault) on thread 139863691237440
	Stack Trace:
	2: basic_string.h:539	~basic_string()
[...]

PR #344, opened 2015-09-30
Really make sure that the unit tests get run.

linas:
Fixes for bug #342 -- SQL unit tests haven't been run in a long time, and a bunch of brand new bugs have accumulated, as a result.  The unit tests now pass.


linas:
CicrcleCI is failing with the following error message:
```
Looks like we had a bug in our infrastructure, or that of our providers (generally GitHub or AWS) We should have automatically retried this build. We've been alerted of the issue and are almost certainly looking into it, please contact us if you're interested in the cause of the problem.
```

PR #346, opened 2015-09-30
FC enhancemnet 

misgeatgit:
- Enabled single step controlled  forward chaining calling
- Added a better inference stat holder class for memoization
- And assorted cleanups


opencog/rule-engine/forwardchainer/ForwardChainer.h
```diff
@@ -48,39 +48,46 @@ class ForwardChainer {
...
-    void add_to_source_list(Handle h);
+    FCStat _fcstat;
 
-    void apply_all_rules(bool search_focus_set = false);
+    void init(Handle hsource, HandleSeq focus_set);
+    void setLogger(Logger* log);
+    Logger* getLogger(void);
 
+    void apply_all_rules(bool search_focus_set = false);
     void do_pm(const Handle& hsource, const UnorderedHandleSet& var_nodes);
```
williampma: Is `do_pm` an obsolete method now?  I can't find any usage.

misgeatgit: Yes, I removed it on my local branch since varialbe fulfillment queries should be handled by the backward chainer. Right?

PR #355, opened 2015-10-01
Remove the UUID from Handles.

linas:
This batch of changes removes the UUID from handles.  As a result, Handles will be smaller by maybe 30% and Links will be smaller by about the same amount. 

This is a major functional change for the atomspace, thus I will hold off on merging this for a few days.  If you think that this might affect you (it should not!), please take a look and review this code.


ngeiswei:
I tried to review it but I couldn't (I'm not deep enough in the AtomSpace code, and I'm to impatient to get back to my tasks). But if I understand well, you've removed UUID from Handle, but not from Atom, right?

Also, I really don't understand why Links will be smaller as well...

linas:
Yes, the Atom maintains a "master" copy of it's UUID.  There are now no other copies, other than an atomtable index. 

Links contain vectors of Handles. Shrink the Handle, the vector will shrink too. 

I'm not really asking for a review; it works, its the right thing to do. There are some messy areas that remain.  It significantly simplifies the code for adding atoms to the atomspace, which has always been painfully complex.  Its a good idea, I just don't want it to come as a surprise to anyone, as its a fairly big change in the overall design.

linas:
Merged. This was burning a hole in my pocket

tests/atomspace/AtomTableUTest.cxxtest
```diff
@@ -276,27 +275,4 @@ public:
...
-
-        VariableListPtr v_added_ptr(VariableListCast(hlist_added));
-        TS_ASSERT_EQUALS(v_added_ptr->get_variables().varset.count(hv1), 0);
-        TS_ASSERT_EQUALS(v_added_ptr->get_variables().varset.count(hv2), 0);
-
-        Handle hv1_added = table->add(hv1, false);
-        Handle hv2_added = table->add(hv2, false);
-        TS_ASSERT_EQUALS(v_added_ptr->get_variables().varset.count(hv1_added), 1);
-        TS_ASSERT_EQUALS(v_added_ptr->get_variables().varset.count(hv2_added), 1);
-    }
```
williampma: Humm... ok, if you think this unit test is bad...

But shouldn't there be a test somewhere that make sure custom Atom with custom C++ class (like LambdaLink, VariableList) stores the correct information inside the internal class storage when the atom are copied to another atomspace, when the whole atom was not inside the atomspace, etc?

Or does this UUID changes make this not a problem anymore?

linas: Perhaps I was hasty; you can test something like this, but the test above seemed to be enshrining weird bugs and undefined behavior.  You really should discard the pointers hv1, hv2 as soon as possible; createVariableList is not exactly a "public" api anyway.  

Anyway, atoms-specific tests should go in the /tests/atoms directory.

PR #357, opened 2015-10-02
Remove FCMemory.h from CMakeLists

leungmanhin:


AmeBel:
similar to https://github.com/opencog/atomspace/pull/359

PR #364, opened 2015-10-03
Handle execution of fully grounded outputs

misgeatgit:


opencog/rule-engine/forwardchainer/ForwardChainer.cc
```diff
@@ -124,10 +125,19 @@ void ForwardChainer::do_step(void)
...
 
     HandleSeq products;
-    //Applying all partial groundings found.
+    //Applying all partial/full groundings.
     for (Handle rhandle : derived_rhandles) {
-        HandleSeq hs = apply_rule(rhandle,_search_focus_Set);
-        products.insert(products.end(),hs.begin(),hs.end());
+        HandleSeq hs;
+        //Check for fully grounded outputs returned by derive_rules.
+        if (not contains_atomtype(rhandle, VARIABLE_NODE)) {
```
williampma: Is this check reliable?  What if the rule is creating SatisfyingSetLink with new variables (like evaluation-to-member-rule)?

misgeatgit: Its the implicand part which is checked https://github.com/opencog/atomspace/blob/master/opencog/rule-engine/forwardchainer/ForwardChainer.cc#L594 . So if implicand doesn't have variables after partial grounding, just implicand part is pushed to the derived rules vector and returned since I wasn't able to create able with no variables inside it.

williampma: Ah I see, so `derived_rhandles` is really `derived_implicands`.

williampma: Oh wait... `derived_rhandles` is sometimes of rules handle, and sometimes of implicand handles.... =_=|||

misgeatgit: Not all of them. if they have variables it will be BindLink else the Implicand. Do I have a better solution in mind?

williampma: I was going to suggest postponing the creation of the BindLink until apply_rule (since it doesn't look to be needed anywhere else anyway), but it seems it will take quite a bit of changes...

misgeatgit: I will put it on my todo list. However, since this is not a high priority issue, I prefer to focus on ways of speeding FC.

PR #370, opened 2015-10-05
Improve log in pattern matcher

ngeiswei:
Mostly log atoms on new lines so that the indentation isn't messed up.

### DIFF
--- a/opencog/query/InitiateSearchCB.cc
+++ b/opencog/query/InitiateSearchCB.cc
@@ -549,8 +549,10 @@ bool InitiateSearchCB::link_type_search(PatternMatchEngine *pme)
 		return false;
 	}
 
-	LAZY_LOG_FINE << "Start clause is: " << _root->toShortString();
-	LAZY_LOG_FINE << "Start term is: " << _starter_term->toShortString();
+	LAZY_LOG_FINE << "Start clause is: " << std::endl
+	              << _root->toShortString();
+	LAZY_LOG_FINE << "Start term is: " << std::endl
+	              << _starter_term->toShortString();
 
 	// Get type of the rarest link
 	Type ptype = _starter_term->getType();
--- a/opencog/query/PatternMatchEngine.cc
+++ b/opencog/query/PatternMatchEngine.cc
@@ -57,9 +57,10 @@ static inline bool log(const Handle& h)
 
 static inline void logmsg(const char * msg, const Handle& h)
 {
-	LAZY_LOG_FINE << msg << (h == nullptr ?
-	                         std::string("(invalid handle)") :
-	                         h->toShortString());
+	LAZY_LOG_FINE << msg << std::endl
+	              << (h == nullptr ?
+	                  std::string("(invalid handle)") :
+	                  h->toShortString());
 }
 
 /* ======================================================== */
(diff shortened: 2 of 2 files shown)


opencog/query/InitiateSearchCB.cc
```diff
@@ -549,8 +549,10 @@ bool InitiateSearchCB::link_type_search(PatternMatchEngine *pme)
 		return false;
 	}
 
-	LAZY_LOG_FINE << "Start clause is: " << _root->toShortString();
-	LAZY_LOG_FINE << "Start term is: " << _starter_term->toShortString();
+	LAZY_LOG_FINE << "Start clause is: " << std::endl
+	              << _root->toShortString();
```
linas: You can just use `\n` instead of a whole `endl`.

opencog/query/PatternMatchEngine.cc
```diff
@@ -164,8 +165,8 @@ bool PatternMatchEngine::variable_compare(const Handle& hp,
 
 	// Make a record of it.
 	LAZY_LOG_FINE << "Found grounding of variable:";
-	logmsg("$$ variable:    ", hp);
-	logmsg("$$ ground term: ", hg);
+	logmsg("$$ variable:", hp);
+	logmsg("$$ ground term:", hg);
```
linas: argh... the whitspace was there to explicitly align the two, so make it easier to read!  Can you put the whitespace back?

opencog/query/PatternMatchEngine.cc
```diff
@@ -57,9 +57,10 @@ static inline bool log(const Handle& h)
 
 static inline void logmsg(const char * msg, const Handle& h)
 {
-	LAZY_LOG_FINE << msg << (h == nullptr ?
-	                         std::string("(invalid handle)") :
-	                         h->toShortString());
+	LAZY_LOG_FINE << msg << std::endl
```
linas: you should not put the linefeed here, it wrecks some of the message formats .. e.g. when printing two aligned sequences of atoms on the same line... for example, the ones immediately below.   It is more important to have log messages that are easy to read, than to make them pretty.

opencog/query/PatternMatchEngine.cc
```diff
@@ -780,8 +781,8 @@ bool PatternMatchEngine::tree_compare(const PatternTermPtr& ptm,
 	if (not match) return false;
 
 	LAZY_LOG_FINE << "depth=" << depth;
-	logmsg("> tree_compare", hp);
-	logmsg(">           to", hg);
```
linas: again -- removing important formating!   These align the atoms so that they can be directly compared!

linas:
I'd merge this since I mostly don't care about such things, but in this case, you are removing formatting that aligns atoms side-by-side, to make them easier to read and compare.

opencog/query/PatternMatchEngine.cc
ngeiswei: It's exactly for readability reason I've put this newline, as otherwise it wrecks the indentation of h (when h is a link, which happens often). But after a second thought I agree its a bad idea, it's better to let the user decide by adding \n in the msg.

opencog/query/PatternMatchEngine.cc
ngeiswei: This wouldn't matter because there would be on the next line. Trust me, it doesn't make it more difficult to directly compare as the 2 atoms are still aligned. And at least there are correctly indented.

linas:
Soory, didn't mean to stop you from merging. I think I was i a bad mood when I wrote above.

ngeiswei:
No problem, I wasn't sure about it anyway as it forces the user to log the newline, not necessarily a good idea.

ngeiswei:
Actually I'm re-opening, there are nearly all mis-indented, which I find distracting (I've compared with and without so I can really tell).

PR #371, opened 2015-10-05
Another attempt to fix gearman

linas:


linas:
At last!!!  yayyy!  Thanks Nil!

PR #372, opened 2015-10-06
Changes To Haskell Bindings:

rTreutlein:
- Updated to the newest GHC version 7.10.2
- Added new Links (NotLink,ContextLink,LambdaLink,SubsetLink)
- Changed TypeConstraints on Links
    - ImplicationLink to accept any type of Atom:
        So it can be used in logical inplication between for example
        Evaluation links
    - EquivalenceLink to accept any type of Atom:
        So it will allow LambdaLinks which should be equivalent to
        predicates
    - InheritanceLink to accept any type of Atom
        So it also accepts Sets, as Concepts are just fuzzy Sets.
        It should be possible to use them interchangeably
    - SimilarityLink
        see InheritanceLink


lib/FindStack.cmake
```diff
@@ -11,7 +11,7 @@ IF (STACK_EXECUTABLE)
                     ERROR_QUIET)
     STRING(REGEX MATCH "[0-9]+\\.[0-9]+\\.[0-9]+" GHC_VERSION "${_GHCVERNO}")
 
-    IF ("${GHC_VERSION}" VERSION_EQUAL "7.8.4")
+    IF ("${GHC_VERSION}" VERSION_EQUAL "7.10.2")
```
ngeiswei: @LordMord Why do you need that? There wouldn't be anything wrong with this, except for the fact that ocpkg installs GHC 7.8.4 https://github.com/opencog/ocpkg/blob/master/ocpkg#L166

So, if you really want to go with the upgrade (cause say you're using some new features of GHC 7.10.2) you would have to have take care of ocpkg as well.

opencog/haskell/OpenCog/AtomSpace/Internal.hs
```diff
@@ -64,6 +64,10 @@ toRaw at = let atype = getType at
     QuoteLink a1             -> Link atype [toRaw a1] Nothing
     VariableList list        -> Link atype (map (appGen toRaw) list) Nothing
     BindLink a1 a2 a3        -> Link atype [toRaw a1,toRaw a2,toRaw a3] Nothing
+    ContextLink tv c e       -> Link atype [toRaw c,toRaw e] $ toTVRaw <$> tv
+    LambdaLink v a	     -> Link atype [toRaw v,toRaw a] Nothing
```
ngeiswei: Code formating matters ;-)

opencog/haskell/OpenCog/AtomSpace/Internal.hs
```diff
@@ -172,6 +176,25 @@ fromRawGen (Link araw out tvraw) = let tv = fromTVRaw <$> tvraw in do
...
+	case (v,e) of
+	  (Gen v,Gen e) -> Just $ Gen $ LambdaLink v e
+      (NotT ,[ar]) -> do
+	a <- filt ar :: Maybe (Gen AtomT)
+	case a of
+	  (Gen a) -> Just $ Gen $ NotLink tv a
+      (SubsetT , [sr,ar]) -> do
+	s <- filt sr :: Maybe (Gen AtomT)
+	a <- filt sr :: Maybe (Gen AtomT)
+	case (s,a) of
```
ngeiswei: The indentation from l.185 to l.196 is broken. Which editor do you use?

opencog/haskell/OpenCog/AtomSpace/Types.hs
```diff
@@ -156,6 +159,12 @@ data Atom (a :: AtomType) where
 
     BindLink            :: (v <~ VariableT,p <~ AtomT,q <~ AtomT) =>
                            Atom v -> Atom p -> Atom q -> Atom BindT
+    ContextLink         :: (c <~ AtomT,e <~ AtomT) =>
+                           TVal -> Atom c -> Atom e -> Atom ContextT
+    LambdaLink		:: (v <~ VariableT,a <~ AtomT) =>
+			   Atom v -> Atom a -> Atom LambdaT
+    SubsetLink          :: (a1 <~ AtomT, a2 <~ AtomT) =>
+			   TVal -> Atom a1 -> Atom a2 -> Atom SubsetT
```
ngeiswei: Same remark here, please try to keep the same formatting, otherwise it's gonna be a source of confusion and distraction.

opencog/haskell/stack.yaml
```diff
@@ -2,4 +2,4 @@ flags: {}
 packages:
 - '.'
 extra-deps: []
-resolver: lts-2.18
+resolver: lts-3.5
```
ngeiswei: Same remark regarding GHC version and ocpkg.

opencog/haskell/OpenCog/AtomSpace/Internal.hs
LordMord: I did those changes in gedit. Guess i should use vim for everything.

ngeiswei:
That's better, thanks. Yeah gedit just isn't worth investing your time in. I highly recommend emacs or vim (emacs + viper-mode actually encompasses vim goodies + some much more see https://www.youtube.com/watch?v=JWD1Fpdd4Pc).

Please update (push the Update button besides "This branch is out-of-date with the base branch"), so I can merge.

Also note that for the moment there is no way to use Scheme or Python if you use Haskell, that's because there is no way yet to call Haskell from Scheme or Python, or the other ways around, and Haskell isn't integrated in the cogserver. As a result you won't be able to use for instance PLN. We definitely should remedy that (I'm kinda busy with other things ATM).

MarcosPividori:
Hi, nice to see other people working on Haskell bindings!
Two remarks:
  - You made a mistake in your code. When you add/modify the atoms definition, you have to be sure that you respect the type declaration both in the Atom data type definition and in the toRaw/fromRaw serializing!
   You added a new atom type: "ContextLink". In the data type definition (https://github.com/rTreutlein/atomspace/commit/fdfb963a5a67c1bc231a70aa89902cbff54933ec#diff-f9f295eb8c464cbdbe5e57014de0405fR162) it accepts 2 any type of atoms, while in fromRaw, you constrain them to be: Concept and Evaluation (https://github.com/rTreutlein/atomspace/commit/fdfb963a5a67c1bc231a70aa89902cbff54933ec#diff-a64d1d4f8e3ca80a3b531ece35457c37R179). You should unify the code.
 - Take into account that one of the main purposes of Haskell Bindings, is to take advantage of the type system to impose type constraints in how atoms relate between them. If you generalize every atom declaration to accept any kind of atom, it will be useless.
Thanks!
Marcos

ngeiswei:
Thank you so much for your feedback, Marcos!

Yeah, it's definitely nice to see your code attracts some contributors :-)

rTreutlein:
Oh, yea i only generalized it in the data type definitions and not in the things in Internal.hs as well. Will fix that and submit another pull request.

Also i would like to take advantage of the type system. But in some place i have to use ContextNodes and Sets interchangeably. Which semantically is fine since a ContextNode is just a fuzzy Set. But the only parent Type they share is the general Atom (Sets are Links and ContextNode is a Node). So if i would like to accept both i have to accept an general Atom.

The other option of letting Sets inherit from ConceptNode as well as from Something-Link cause the c++ Unit-Test to fail. Also i couldn't exactly figure out why.

ngeiswei:
I'm not surprised that making SetLink inherit from ConceptNode breaks unit test. The problem is that those shallow types are mostly about syntax, while SetLink and ConceptNode relate to each other at a semantic level. The other option is to build the union type of SetLink and ConceptNode, Marcos tried hard to hack his way around but the result slows down Haskell type checking too dramatically.

For those reasons I'm thinking we probably want to write our own "deep type" checker. It would be better if type checking is part of the atomspace core so that any language can benefit from it. That doesn't necessarily mean it has to be coded in C++, it could actually be coded in Haskell. But anyway the first step is to design it, I have intended to do that a while ago, but came to the conclusion that in order to really know what we want is to complete our proto-URE/PLN first (but I might be just procrastinating).

linas:
Some short but maybe deep comments:

I'd strongly prefer typechecking to be deeply in the core atomspace,
because so many things depend on it. VarialbeList and TypeNode already
perform some minimal type-checking; SignatureLink is in unimplemented
proposal to add more Likewise FilterLink.  Note that PutLink already does
filtering...

I am also very nervous about getting too strict about type-checking too
early.  Some basic ideas about how things should work is still in flux:
e.g. AssignLink, RemoveLink were invented just 6 months ago, now they're
obsolete. ComposeLink was already created and bsoleted. BetaRedexLink was
renamed to PutLink.   Earlier today, I realized PutLink was mis-designed.
Also LambdaLink was mis-designed, which I just now fixed. ...

If all of these were type-checked too soon, to strictly, development could
grind to a halt.   Just be careful not to invent quick-setting cement ...

--linas

ngeiswei:
Agreed, well said.

PR #375, opened 2015-10-06
Log pm extra

ngeiswei:


ngeiswei:
@AmeBel It's annoying, I can't merge that because Cython unit test is failing. I don't think my changes have anything to do with Cython, it's probably a random failing (and it doesn't fail on my local repo). How can I re-trigger the check?

AmeBel:
login to circle-ci using github and click rebuild. You can do that to all the repos you have write access to

ngeiswei:
Got it, thanks!

PR #379, opened 2015-10-07
Fixed type in fromRaw to respect the generalizations

rTreutlein:


ngeiswei:
@rTreutlein you need to update your branch so I can merge it. Also, the fact that you missed that show you need to add some test or at least examples using those.

PR #380, opened 2015-10-07
Support scm logger format

ngeiswei:
The logger scheme binding now supports scheme string formatting which allows to log things like:
```scheme
(cog-logger-info "My concept ~a" (ConceptNode "A"))
```
See http://wiki.opencog.org/wikihome/index.php/Scheme#Logger

### DIFF
--- a/opencog/scm/logger-utils.scm
+++ b/opencog/scm/logger-utils.scm
@@ -1,0 +1,35 @@
+;
+; logger-utils.scm
+;
+;;; Commentary:
+;
+; Helper to use ice-9 format in logger functions
+;
+;;; Code:
+; Copyright (c) 2015, OpenCog Foundation
+;
+
+(use-modules (opencog))
+(use-modules (ice-9 format))
+
+(define (cog-logger-error msg . args)
+  (cog-logger-error-str (apply format #f msg args)))
+
+(define (cog-logger-warn msg . args)
+  (cog-logger-warn-str (apply format #f msg args)))
+
+(define (cog-logger-info msg . args)
+  (cog-logger-info-str (apply format #f msg args)))
+
+(define (cog-logger-debug msg . args)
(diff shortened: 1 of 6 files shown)


opencog/scm/logger-utils.scm
```diff
@@ -0,0 +1,35 @@
...
+
+(define (cog-logger-fine msg . args)
+  (cog-logger-fine-str (apply format #f msg args)))
+
+(define (export-logger-utils)
+  (export cog-logger-error
+          cog-logger-warn
+          cog-logger-info
+          cog-logger-debug
+          cog-logger-fine))
```
linas: The preferred method would be to move the entire contents of this file to `opencog/scm/opencog/logger.scm`

opencog/scm/rule-engine-utils.scm
```diff
@@ -86,4 +86,8 @@
      rbs)
 )
 
-(define (export-rule-engine-utils) (export ure-add-rules ure-set-num-parameter ure-set-fuzzy-bool-parameter export-rule-engine-utils))
```
linas: Likewise, all of this should go into `/opencog/scm/opencog/rule-engine.scm`

linas: I'll make this change in a few mintes...

linas: Pull request #383

opencog/scm/logger-utils.scm
ngeiswei: I actually thought of that, but I just wasn't sure, can you express reasons why it should be the "preferred method"?

linas: No particularly strong reasons, other than its pointless to have too many itty bitty files.  If its a module, put it in the module. there is no plausible scenario where a user would try to load that file stand-alone without having previously loaded the module.

PR #382, opened 2015-10-08
Distinguish between ScopeLink and LambdaLink

linas:
After discussions with Nil, its now clear (to me) these are two different but related concepts.  Fix PutLink so it uses the appropriate one.

### DIFF
--- a/examples/guile/filter.scm
+++ b/examples/guile/filter.scm
@@ -14,13 +14,14 @@
 ; a set of atoms.
 (define filter-it
 	(PutLink
-		; Declare a single variable: The variable must be a ConceptNode.
-		(TypedVariableLink
-			(VariableNode "%x")
-			(TypeNode "ConceptNode"))
+		(LambdaLink
+			; Declare a single variable: The variable must be a ConceptNode.
+			(TypedVariableLink
+				(VariableNode "%x")
+				(TypeNode "ConceptNode"))
 
-		; The body of the put is just the variable.
-		(VariableNode "%x")
+			; The body of the put is just the variable.
+			(VariableNode "%x"))
 
 		; This is the set of things that will be filtered.
 		(SetLink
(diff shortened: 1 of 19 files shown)


ngeiswei:
@linas I used to think that, but I'm now seeing the light. There are no useful distinctions between a ScopeLink and LambdaLink. A LambdaLink is merely an operator to abstract away the body and expose the variables of interest. Since what is typically done with a lambda is beta-reduction, then a lambda becomes a function constructor, but it doesn't have to be seen that way.

linas:
There is a good C++ analogy to explain the difference.

An opencog free variable is like a C++ global variable.
An opencog scoped variable is like a C++ block-scope variable.
An opencog bound variable is like a C++ function argument.

They are all variables, but you can only do beta-reduction on "function arguments".  You can do substitution on free variables, and you can do neither on block-scope variables, at least, not without fundamentally breaking your code.

AmeBel:
@linas with opencog free-variables,  do you mean a VraialbeNode like
```scheme
(EvaluationLink
     (PredicateNode "jumped")
     (ListLink
         (ConceptNode "Jerry")
         (VariableNode "x")))
```
If so, how will you use them without scoping? Let me rephrase, why have a non-scoped VariableNode, it seems you can't do anything within the atomspace(more correctly within the pattern matcher) without scoping?

linas:
Sure, by definition, variables in EvaluationLinks are always free. Not sure what you mean by "use a variable."; you don't have to use them; they just "are". So for example:

```
scheme@(guile-user)> (cog-satisfying-set (GetLink (VariableNode "$yyyy")
... (EvaluationLink
...      (PredicateNode "jumped")
...      (ListLink
...          (ConceptNode "Jerry")
...          (VariableNode "$yyyy")))))
$21 = (SetLink
   (VariableNode "x")
)
```

That searches for everything that Jerry might jump over, and the answer is "x".  In this example, "x" is free, and $yyyy is scoped.  

I want to distinguish between three different cases, now: free, scoped, and bound.  See the wiki page  http://wiki.opencog.org/w/ScopeLink  for a detailed explanation.  I hope that page will eliminate the lingering confusion.

williampma:
>An opencog free variable is like a C++ global variable.
An opencog scoped variable is like a C++ block-scope variable.
An opencog bound variable is like a C++ function argument.

Except for that to be really true in OpenCog, VariableNode need to be changed to have either no-name, or changed so each usage has unique name.  Otherwise you will have the exact same variable (as unique atom) that are both free and scoped and bound.

An example (yes I know the following won't happen since PM stops self-grounding)
```
scheme@(guile-user)>(ListLink
... (EvaluationLink
...      (PredicateNode "jumped")
...      (ListLink
...          (ConceptNode "Jerry")
...          (VariableNode "$x"))))
scheme@(guile-user)> (cog-satisfying-set (GetLink (VariableNode "$x")
... (EvaluationLink
...      (PredicateNode "jumped")
...      (ListLink
...          (ConceptNode "Jerry")
...          (VariableNode "$x")))))
$21 = (SetLink
   (VariableNode "x")
)
```

linas:
Yeah, I understand that. I haven't forgotten, I'm pondering ....

linas:
I forget what I might have said earlier in emails, but it would not be outrageous to generate unique variable names at the time that scoped links and lamda links  are added to the  atomspace.  This would not be technically hard.  I'm not sure what reprecussions this could have on other things.

PR #383, opened 2015-10-08
Reduce proliferation of module files...

linas:


linas:
closing -- nasty surprise -- the rule engine unit tests depend on these files, and I am too tired to try to figure this out right now.

PR #386, opened 2015-10-08
Make forall a lambda

ngeiswei:
Somewhat experimental, make ForAllLink inherit LambdaLink.

The practical side effect is that you can put a ForAllLink inside a PutLink.

### DIFF
--- a/opencog/atomspace/atom_types.script
+++ b/opencog/atomspace/atom_types.script
@@ -211,8 +211,8 @@ UNQUOTE_LINK <- ORDERED_LINK
 // Assorted PLN link types.
 
 // Logical quantifiers
-FORALL_LINK <- SCOPE_LINK "ForAllLink"
-EXISTS_LINK <- SCOPE_LINK
+FORALL_LINK <- LAMBDA_LINK "ForAllLink"
+EXISTS_LINK <- LAMBDA_LINK
 AVERAGE_LINK <- FORALL_LINK
 
 // Convert predicate into concept
--- a/opencog/atomspace/atom_types.script
+++ b/opencog/atomspace/atom_types.script
@@ -178,6 +178,12 @@ PUT_LINK <- ORDERED_LINK
 // beta-reducible.
 SCOPE_LINK <- ORDERED_LINK
 
+// A LambdaLink is a link whose bound variables mey be substituted,
+// that is, may be beta-reduced. It can be thought of as not only a
+// variable declaration (which is what ScopeLink does) but also has
+// the ability to substitute values for those variables.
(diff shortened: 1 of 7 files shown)


opencog/atomspace/atom_types.script
```diff
@@ -205,8 +211,8 @@ UNQUOTE_LINK <- ORDERED_LINK
 // Assorted PLN link types.
 
 // Logical quantifiers
-FORALL_LINK <- SCOPE_LINK "ForAllLink"
-EXISTS_LINK <- SCOPE_LINK
+FORALL_LINK <- LAMBDA_LINK "ForAllLink"
```
linas: well, we had this conversation, and this is clearly wrong, so I am confused about why you are changing this,,,

tests/atoms/PutLinkUTest.cxxtest
```diff
@@ -123,3 +127,74 @@ void PutLinkUTest::test_filtering()
...
+		  L(INHERITANCE_LINK,
+		    N(VARIABLE_NODE, "$y"),
+		    N(VARIABLE_NODE, "$x")));
+	Handle put =
+		L(PUT_LINK,
+		  lam,
+		  L(LIST_LINK,
+		    N(CONCEPT_NODE, "animal"),
+		    N(CONCEPT_NODE, "coyote")));
+
```
linas: !! But this is fundamentally wrong, you cannot do lambda substitution on a for-all link, it doesn't make sense.  That's the whole point of variable scoping -- it is to PREVENT this from happening!  $x and $y are bound in the for-all link!!

ngeiswei: It really make sense in my world. But anyway I'm reverting that change see PR https://github.com/opencog/atomspace/pull/387

AmeBel: @ngeiswei read Linas's explanation why it doesn't make sense @ https://groups.google.com/d/msg/opencog/P42OShf0stA/siAZDsjcAQAJ . Would you explain how it make senses to you? Trying to get a sense on the scoping vs bounding and effect of beta-redex on scoped vs bound variables.

ngeiswei: @AmeBel I probably shouldn't insist on that cause it might bring more confusion than anything. On the practical level I agree with @linas. On a more philosophical level, I think one can indeed define a formal system where forall inherits the properties of a lambda while not falling into inconsistencies.

AmeBel: @ngeiswei thanks

PR #387, opened 2015-10-08
Unmake forall a lambda

ngeiswei:
Turns out it doesn't make the universal instantiation code simpler, and,
although I still think a forall can easily be seen a lambda, I fear it's
gonna confuse others.


linas:
Also, please understand: the whole point of having C++ classes corresponding to atom types is to perform either some kind of pre-processing or caching or performance enhancement.   So unless that is what you need, then having the C++ class is pointless (and the atom factory in the atomtable slows things down).

So: for example: DefineLink, which throws an error if a second inconsitent Definelink is created, is a really great example of how a C++ class helps. What started this was the old BindLink, which has a two-phase design: (1) analyze the patttern (2) perform the search.  Part (1) needs to only be done once, ever, so its done in the BindLink constructor.  Then the c++ instance can be re-used to do 2 over and over.   Anything that has a natural (1)-(2) split can be/should be done in a C++ class, otherwise, its  .. pointless.

ngeiswei:
Yeah, I thought about that but I wasn't quite sure how to do otherwise. I suppose I could simply have AtomTable.cc instantiate a LambdaLink when it sees a ForAllLink, but then that might have been very confusing to the C++ user. Anyway, this is no longer needed so...

Issue #393, opened 2015-10-09
Unable to install atomspace


misgeatgit:
After pulling the latest atomspace and building, am getting the following error during 'sudo make install'

``
CMake Error at opencog/persist/zmq/atomspace/cmake_install.cmake:56 (FILE):
  file INSTALL cannot find
  "$HOME/OPENCOG/atomspace/opencog/persist/zmq/atomspace/ZMQMessages.pb.h".
``

AmeBel:
try now

PR #394, opened 2015-10-09
cmake install ZMQMessages.pb.h from build directory

AmeBel:
Fixes https://github.com/opencog/atomspace/issues/393

### DIFF
--- a/opencog/persist/zmq/atomspace/CMakeLists.txt
+++ b/opencog/persist/zmq/atomspace/CMakeLists.txt
@@ -27,7 +27,7 @@ INSTALL (TARGETS zmqatoms
 )
 
 INSTALL (FILES
-	ZMQMessages.pb.h
+	${CMAKE_CURRENT_BINARY_DIR}/ZMQMessages.pb.h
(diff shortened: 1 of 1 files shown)


opencog/persist/zmq/atomspace/CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ INSTALL (TARGETS zmqatoms
 )
 
 INSTALL (FILES
-	ZMQMessages.pb.h
+	${CMAKE_CURRENT_BINARY_DIR}/ZMQMessages.pb.h
```
linas: thanks.

I sometuimes think we are installing too many header files.  Many headers are really "internal use only"...

AmeBel:
The installed header files are the "official" c++ api for atomspace, right? If so, I think that is a great idea. Plus it will make life easier for who ever want to consume the api for their projects, or contribute a new binding/feature(some what dealing with the [paradox of choice](https://en.wikipedia.org/wiki/The_Paradox_of_Choice)).

PR #396, opened 2015-10-09
Add TrueLink, FalseLink

linas:
I resisted adding these for a long time, but need them to be able to handle DefinedSchemaNodes as if they were DefinedPredicateNodes.

Also: don't print default TV.

Also: try to keep python from spewing garbage into working directories.


AmeBel:
What are these for? And, is it a step/experimentation in the direction of implementing ProtoAtoms discussed @ https://groups.google.com/forum/#!topic/opencog/_F6z53Mr77U, since True/FalseLink are the atom representation of the stvs?

linas:
No, not at all, I need it for terminatting behavior trees.  

for example
https://github.com/opencog/ros-behavior-scripting/blob/master/src/btree.scm#L144

sequential and link is trying to step through a sequence of steps until one step is true or false,   unfortunaly, any schemanode yeilds atoms; I need to convert those into true/false.

re protoatoms: the code "works" but the speed of adding atoms to the atomspace is cut by half, so I don't want to merge that, not until it becomes more useful or more developed.

AmeBel:
If I got it, a TrueLink guarantees a continuation of execution, independent of what is returned by the definedschemanode and the truth-value of the returned atoms,  then what is the use of faleslink? I know it still in development, just curious.

linas:
SequentialOr will  sequentially execute terms until one finally returns true, and then it stops.  So FalseLink, if used before anything that might be true, guarantees execution.  In behavior trees, SequntialOr is called "fallback" or "selector" -- it selects the first true outcome, or alternately "falls back" from plan A plan B plan C until a plan finally works

linas:
Note also, as currently implemented, if a SequentialAnd has a schema in it, it will throw an error.  A reasonable alternative would be to look at the TV of the returned atom, and decide based on that; however, that seemed ... premature.   We don't yet really have good policy on how to handle truth values in the atomspace, and so, for example, its not clear that hard-coding a if (TV_mean > 0.5) into the pattern matcher.  i'd rather have people thin about what they are doping, and then use TrueLink/FalseLink to get the desired behavior.

AmeBel:
thanks

ngeiswei:
@linas Why are TrueLink and FalseLink links? Why aren't they nodes?

ngeiswei:
Hmm, maybe that's cause you may want then to inherits FunctionLink?

linas:
The are links because I need to convert a function returning a atom (e.g. a GroundedSchemaNode) into a crisp truth value that the SequentialAnd/OrLink  can operate on.  For this reason, these cannot be function links, since FunctionLinks return atoms not TV's. 

The core issue is this:  I want to execute a SequentialAndLink, until completion, or until one of the steps returns false.   Many of the things inside the SequentialAndLink are predicate-like things, e.g. EqualLink, GreaterThaLink, PresentLink, etc. and so these are TV-valued in a "natural" way.   However, I also wat to execute some things unconditionally, e.g. a PutLink, to place some data in the atomspace, or a DeleteLink, to remove it.  Using the TV o the PutLink makes no sense;  ad the DeleteLink doesn't even "return" anything. 

So -- I needed a way to unconditionally execute something, ad have it be TV-valued.

If you ca think of a better way, let me know.   A lot of  the stuff I've been doing lately is "experimental"; its the best I could think of, but sometimes it feels awkward.

ngeiswei:
OK, I got it. I can't think of a better way off the top of my head. But still why are those links and not nodes?

linas:
Because you cannot use a node to wrap an atom?  For example
```
(cog-evaluate! (FalseLink (PutLink ...))
```

causes the PutLink to be unconditionally executed, returning a TV of false.

ngeiswei:
Oh, got it!!!

ngeiswei:
I guess someone should add a wiki page for it.

PR #405, opened 2015-10-12
Allow constant terms in the default pattern matcher

linas:
Turns out this can be a useful feature... e.g. for the AIML work.


williampma:
Cool.  One less checking before using PM.

PR #412, opened 2015-10-14
Get a rule by its handle

misgeatgit:
Useful for referring all attributes related to a rule given a valid BindLink handle.


misgeatgit:
I would like to merge this PR.

PR #417, opened 2015-10-15
Fixed an infinite loop in HaskellBindings / Re-enable Haskell

rTreutlein:
Variables nodes are specially handled in the HaskellBindings to allow them to be used in place of any other atom.
The creation of a child Glob_Node to the VariableNode created an infinite Loop. (GlobNode was both parent and child of VariableNode.)

This has now been fixed.


linas:
Ah ha!  Thanks!  

FYI, there is some tiny chance I might need to add other types of glob nodes...

Note that variable nodes are typed, so they cannot be used "in place of any other atom"; only in place of those that have a matching type. (by default, variables have the loosest possibly typing, unless the user specifies something stricter)

Also, GlobNodes can be typed as well; the code is there; its probably buggy, and certainly untested.

linas:
Variable typing is described here:
http://wiki.opencog.org/w/TypeNode
http://wiki.opencog.org/wikihome/index.php/TypedVariableLink

the above are implemented and are in heavy use; a more general, more sophisticated kind of checking is described below, but not yet implemented:
http://wiki.opencog.org/wikihome/index.php/Type_signature
http://wiki.opencog.org/wikihome/index.php/Type_checking

rTreutlein:
Adding other types of glob nodes should not be a problem.

As for Typed Variable/Glob Nodes. Using those should work fine but they will not be type checked by ghc. Once Nil gets around to implementing type checking that's what will be used in the HaskellBindings.

MarcosPividori:
Great! Sorry for my disappearance these days. I had a very complicated week!

PR #424, opened 2015-10-16
Fcfix

misgeatgit:
- Bug fixes

### DIFF
--- a/opencog/rule-engine/forwardchainer/ForwardChainer.cc
+++ b/opencog/rule-engine/forwardchainer/ForwardChainer.cc
@@ -339,6 +339,20 @@ HandleSeq ForwardChainer::apply_rule(Handle rhandle,bool search_in_focus_set /*=
 
     //Check for fully grounded outputs returned by derive_rules.
     if (not contains_atomtype(rhandle, VARIABLE_NODE)) {
+        //Subatomic matching may have created a non existing implicant
+        //atom and if the implicant doesn't exist, nor should the implicand.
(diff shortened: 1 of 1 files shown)


opencog/rule-engine/forwardchainer/ForwardChainer.cc
```diff
@@ -345,58 +337,84 @@ HandleSeq ForwardChainer::apply_rule(Handle rhandle,bool search_in_focus_set /*=
...
+        //atom and if the implicant doesn't exist, nor should the implicand.
+        Handle implicant = BindLinkCast(rhandle)->get_body();
+        HandleSeq hs;
+        if (implicant->getType() == AND_LINK or implicant->getType() == OR_LINK)
+            hs = LinkCast(implicant)->getOutgoingSet();
+        else
+            hs.push_back(implicant);
+        //Actual checking here.
+        for (Handle& h : hs) {
+            if (_as.get_atom(h) == Handle::UNDEFINED)
```
williampma: What about checking the focus set atomspace?  You don't want the rule to be applied if the atom is not in the focus set.

misgeatgit: Yes, even if the grounded implicant is in the atomspace, we should check that it is a member of the focus set for the case where focus set is supplied. I will add that.Thanks.

misgeatgit: https://github.com/opencog/atomspace/pull/425 handles the case you mentioned.

PR #433, opened 2015-10-21
Haskelltest

rTreutlein:
Created a Test System for Haskell with 1 simple Test
and fixed various bugs in the Haskell Bindings


ngeiswei:
Excellent that you've added a test! Now you just need to make it pass.

linas:
Nil, can you review and merge when ready? @ngeiswei

ngeiswei:
@rTreutlein that looks good to me. You need to update you're branch. I've actually invited you to the organization so you don't have to suffer from the constant need to have someone ask you to update in order to merge.

ngeiswei:
Thanks a lot for setting that unit test, BTW. As narrow as it is for now, that was definitely a missing element in order to get the Haskell binding properly going.

PR #435, opened 2015-10-21
An ungrounded predicate evaluates to itself

ngeiswei:
Otherwise it breaks the PLNRulesUTest on the OpenCog repository.


ngeiswei:
@linas you may want to check that little commit just in case.

linas:
yes, that's reasonable/fine. I always wondered what to do about this, anyway, because right niow, little or nothing is probabilistic in that whole mass of code.

Anyway, can you merge latest branch, and also, move that check to the top of that code block? There is no reason to do the instation step, if you just return the TV...

PR #441, opened 2015-10-23
Bug fix

misgeatgit:
Add potential sources to focus set for all cases.

### DIFF
--- a/opencog/rule-engine/forwardchainer/ForwardChainer.cc
+++ b/opencog/rule-engine/forwardchainer/ForwardChainer.cc
@@ -336,8 +336,15 @@ HandleSeq ForwardChainer::apply_rule(Handle rhandle,bool search_in_focus_set /*=
 {
     HandleSeq result;
 
+    //Add source atoms to focus_set atomspace
+    if (search_in_focus_set) {
+        for (Handle h : _potential_sources)
+            _focus_set_as.add_atom(h);
+    }
+
     //Check for fully grounded outputs returned by derive_rules.
     if (not contains_atomtype(rhandle, VARIABLE_NODE)) {
+
         //Subatomic matching may have created a non existing implicant
         //atom and if the implicant doesn't exist, nor should the implicand.
         Handle implicant = BindLinkCast(rhandle)->get_body();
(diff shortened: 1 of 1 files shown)


eddiemonroe:
@misgeatgit Seems to me there is a lot of unnecessary redundancy looping through and re-adding the entire potential source set every time a rule is applied. How about (1) adding the initial sources at the very beginning and then (2) when a conclusion (or anything) is added to potential sources also add it to the focus set.

misgeatgit:
@eddiemonroe  If https://github.com/opencog/atomspace/blob/master/opencog/atomspace/AtomSpace.cc#L92 is a costly operation, you are right and remind you each iteration doesn't guarantee a unique inference as same rule might be chosen and same and the result set may hold previous inferences in which case what you suggested also ends up doing same redundant things. That case still the above check will be done. My design intention is to have a single responsible method for the particular operation. If you still see a serious performance penalty on it, lets change it.

eddiemonroe:
@misgeatgit I don't know how serious the performance penalty would be under different circumstances, but it certainly could be significant, and why do something that's so inefficient, completely unneeded, and easy to change? I guess you could do a benchmark test to see the actual impact, but what's the point when there's no benefit to doing the unneeded redundancy anyway. Why not just fix it.

You wrote: "My design intention is to have a single responsible method for the particular operation."
I think the change involves, instead of adding all the potential sources to the focus set at the beginning of apply_rule(), add just the conclusion results to the focus set at the end of apply_rule().

Maybe you're not thinking in terms of large atomspaces? Say the FC has been running and has generated 100K conclusions so far. Then, at each inference step, there would be 100K unnecessary calls to the function you link to, rather than one call for each conclusion generated at that step. And then that number increases as more conclusions are generated. Or substitute 1K or 1M conclusions generated... The cost of doing the operation a single time might be tiny, but it adds up when you're doing so many repetitions.

eddiemonroe:
Some quick numbers I'm pulling out of the air that might be off. Say we have the situation above of having generated 100K conclusions, and the function/operation takes 5ms (.005s) to run. 100K * .005 / 60 = 8.3 minutes of useless processing time taken for each of the following inference steps. If you want to do 1000 more inference steps, that's 8.3 * 1000 / 60 = 138 hrs of additional processing time with no benefit. My numbers may be off, but my point is that at these scales it adds up.

eddiemonroe:
Also, I understand your point about redundant conclusions at times being generated, and perhaps that's something that should be addressed also, but I think that's a separate issue with a performance impact on the process we're discussing that's orders of magnitude smaller .

bgoertzel:
If I understand correctly then I think I agree with Eddie's comment that 
 "@misgeatgit Seems to me there is a lot of unnecessary redundancy looping through and re-adding the entire potential source set every time a rule is applied. How about (1) adding the initial sources at the very beginning and then (2) when a conclusion (or anything) is added to potential sources also add it to the focus set."

misgeatgit:
@eddiemonroe I have made the changes you asked https://github.com/opencog/atomspace/pull/448.

PR #443, opened 2015-10-24
Fix cmake warnings

hedayat:
These changes doesn't seem to change cmake behavior. They fix two CMake warnings.


linas:
this looks reasonable; can you refresh the branch? (click on "refresh branch"  button).  This will need to be done after every merge, because CircleCI forces serialization :-(

hedayat:
Done. But it discourages multiple granular PRs. :(

linas:
yes it does. I don't particularly like circle ci; it seems to be incompletely thought out and poorly designed. but alas :-/

linas:
and now, the next one..

PR #444, opened 2015-10-24
Add check for existance of ZMQ C++ header file: zmq.hpp

hedayat:
ZeroMQ might be installed while its C++ header is not. This makes sure that zmq.hpp exists to enable building with ZeroMQ support.


linas:
this looks reasonable too; will need to refresh also, after the next pull req is merged.

hedayat:
merged.

PR #445, opened 2015-10-24
Fix compilation when cogutils is installed in a non-standard location

hedayat:


linas:
looks good to me ...

hedayat:
Updated.

PR #446, opened 2015-10-24
Fix compile flag for ZMQClient compilation

hedayat:
This will conflict with the other zmq related change. I can rebase it when the other PR is merged. I wonder what is the best approach in such situations. :P


opencog/persist/zmq/atomspace/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ ADD_CUSTOM_COMMAND(
 )
 
 ADD_LIBRARY (zmqatoms SHARED
-	ZMQMessages.pb
+	ZMQMessages.pb.cc ZMQMessages.pb.h
```
linas: This should not be required; cmake automatically (should be automatically) attaching the needed .cc to the end of the name string ... are you getting an error due to this?

hedayat: Yes, without .cc extension, CMake silently ignores ZMQMessages.pb.cc (verified by looking at 'make VERBOSE=1' output) and I got many 'undefined reference' errors for functions/variables defined in this file when trying to link opencog to this library. I don't know why (at least, it should print error message about not finding the source file). But, adding .cc extension explicitly fixes the problem I have.

(more commits will be added to this branch soon, since there are other problems too).

linas: strange, it builds for me ... i get the feeling you are using a different version of cmake..

hedayat: Notice that atomspace builds fine. I get undefined reference errors when trying to link opencog cogserver to libzmqatoms.so. Also, you can check if ZMQMessages.pb.cc is actually built and linked to libzmqatoms.so. 

I use CMake 3.3.2

hedayat:
Ready for merge.

pennachin:
Hmm, I can't run the cogserver from opencog HEAD, and I believe this is the cause? 
```
/build/opencog/server/cogserver: symbol lookup error: /usr/local/lib/opencog/modules/libPersistZmqModule.so: undefined symbol: _ZN7opencog13ZMQPersistSCMC1EPNS_9AtomSpaceE
```
@linas does the cogserver run for you from opencog master?

linas:
Hi Cassio,

it works for me, but I am not building with either ZMQ or persistence.  The link failure message you are seeing suggests that you are not building persistence, probably because you do't have UnixODBC installed -- and that should be OK, persistence is an optional component.

I would guess that something in ZMQ is trying to link to persistence, without using the `if HAVE_SQL` check in the right places.

pennachin:
Thanks, but I don't think that's the issue, I have UnixODBC installed and can access the DB from scheme, so I thought the issue might be related to the errors @hedayat was having.

linas:
Oh,

I dd not read the error message carefully. his should fix it:
```
--- a/opencog/modules/CMakeLists.txt
+++ b/opencog/modules/CMakeLists.txt
@@ -44,6 +44,7 @@ ADD_LIBRARY (PersistModule SHARED
 
 TARGET_LINK_LIBRARIES(PersistModule
        # server  # needed for definition of _ZTIN7opencog7RequestE
+       zmqatoms
        ${ATOMSPACE_LIBRARIES}
 )
 
```

AmeBel:
shouldn't https://github.com/opencog/opencog/blob/master/lib/FindAtomSpace.cmake#L72 have covered the `zmqatoms` library though?

pennachin:
I guess, Linas' suggestion gives me a linking error, though. Anyway, commenting out line 72 as @AmeBel suggested lets me get back to work...

hedayat:
The undefined reference error I get is when cogserver tries to link against libzmqatoms, so it differs from what @pennachin receives. Also notice that loadType() function in ZMQClient is declared but not defined, so if you don't receive undefined reference errors for it, you probably don't hit the other problem with functions defined in ZMQMessages.pb.cc  too. :P

pennachin:
Thanks, Hedayat. They may be different issues, but they seem related?

hedayat:
Yes, I'll check if I can reproduce your problem too.
Yes, I think this PR is ready for merge. Thanks!

PR #451, opened 2015-10-26
Handle TimeNode too

linas:


AmeBel:
FYI, if you login to circle-ci using your github account you will get a rebuild and cancel-build buttons

PR #453, opened 2015-10-26
Add MinusLink support

linas:
Glurg.  Basic arithmetic :-(

There must be a easier way ...


opencog/atoms/core/PutLink.cc
```diff
@@ -6,7 +6,7 @@
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU Affero General Public License v3 as
- * published by the Put Software Foundation and including the exceptions
+ * published by the Free Software Foundation and including the exceptions
```
ngeiswei: @linas That's why I always do interactive Replace. Doesn't Vim provide that? BTW, if you which to switch to Emacs you can use viper-mode, apparently some vim people have switched to Emacs because of it and are very excited about it https://www.youtube.com/watch?v=JWD1Fpdd4Pc

Issue #461, opened 2015-10-28
CircleCI + haskell is crazy/broken again...


linas:
CircleCI + haskell is crazy/broken again: chaging a README file causes CircleCI to break down in the Haskell tests

This test fails:
https://circleci.com/gh/opencog/atomspace/728

which is broken by a change to a readme file in pull req #460

linas:
well, whatever. Its not reproducible. Closing.

PR #463, opened 2015-10-29
Re-enable Haskell Test

rTreutlein:
It worked fine in my local docker container.


linas:
Can you re-merge? Sorry, I would have committed, but was't  paying attention ... 

Any clue why CircleCI failed earlier, but passes now?  I notice that, when it failed, the Haskell test ran 276 seconds, and then failed ...  which seems like much too long a time.

linas:
That is, as described in #461

ngeiswei:
@rTreutlein feel free to update and merge.

rTreutlein:
I don't know why it failed earlier. But the long runtime is because the Test is built at that point. And I am using TemplateHaskell to generate a bunch of Interfaces which unfortunately takes quite a bit of time.
@ngeiswei 
Should i maybe integrate my Test Library with the Haskell Bindings? So that during the test phase only the actual test have to be built.

ngeiswei:
@rTreutlein you may certainly feel free to move the more generic part of your test in the Haskell binding, yeah.

PR #466, opened 2015-10-30
Split Building and Running of the Haskell Test

rTreutlein:
Updated the cabal config (removed unnecesary dependency)


linas:
remerge ... I just merged ....

PR #476, opened 2015-11-05
Fix two build issues

hedayat:
1. zmqatoms library should be linked to protobuf library
2. ${CMAKE_BINARY_DIR} should be early in the include list, so that include files generated during build take precedence to those that might be installed previously in the system (this happens if atomspace is already installed, and you are going to build a newer version of it). 

Since CMAKE_BINARY_DIR is globally added to the include list, lots of "include_directory(${CMAKE_BINARY_DIR})" directives in different CMakeLists.txt files can be omitted, but it is not included in this PR.


pennachin:
Thanks!

Issue #472, opened 2015-11-03
Instantiator::walk_tree doesn't call the correct atom type constructor after subtitution


ngeiswei:
Towards the end of

```c++
Handle Instantiator::walk_tree(const Handle& expr)
```

a new link is constructed with its outgoings substituted by the arguments passed to the ExecutionOutputLink. See

```c++
	if (changed)
		return Handle(createLink(t, oset_results, expr->getTruthValue()));
```

However, createLink merely call the Link constructor, by-passing the possibly useful C++ inheritance hierarchy of atoms (ScopeLink, FunctionLink, etc).

The consequences is that when walk_tree recreates the ExecutionOutputLink right before execution, see

```c++
		ExecutionOutputLinkPtr eolp(createExecutionOutputLink(sn, args));
		return eolp->execute(_as);
```

the call

```c++
void FreeLink::init(void)
{
	_vars.find_variables(_outgoing);
}
```

is unable to properly ignore the bound variables after substitution. For instance in the following conditional, the conditional is always true

```c++
			ScopeLinkPtr sco(ScopeLinkCast(lll));
			if (NULL == sco)
				sco = createScopeLink(lll->getOutgoingSet());
```

I'm not exactly sure what is the right fix. My guess is that `createLink` in `Instantiator::walk_tree` should be replaced by a `AtomTable::factory`, `AtomTable::clone_factory` or something like that. But I'm not enough confident to be sure.

ngeiswei:
Hmmm, it seems I could just wrap AtomTable::factory around createLink but I can't access the atomtable (get_atomtable is private). There is a static AtomTable::do_clone_factory but it's not declared in the header... Puzzled...

I think meanwhile I'll just add the atom in the atomspace via add_link, but I won't commit anything, just do that so I can move forward...

ngeiswei:
Wrapping add_atom around the createLink works, but I won't commit anything...

ngeiswei:
It has a side effect though, the following 2 unit tests are failing
```
The following tests FAILED:
	 52 - ArcanaUTest (Failed)
	 58 - EvaluationUTest (Failed)
```
Which isn't too surprising, since it adds something new in the AtomSpace it creates rooms for altering the behavior.

I suppose we want to have something like add some method in the atomspace (maybe a static one), like create_atom...

linas:
[...]

PR #477, opened 2015-11-05
Implication scope

ngeiswei:
Warning, this PR is important.

1. Now instantiated arguments are added to the AtomSpace right before being passed to the ExecutionOutputLink. I tried to add them in a scratch atomspace but it apparently spawn the execution in an infinite loop. I do not know how bad it is w.r.t. to memory overhead to have those instantiation now in the AtomSpace.

2. I've removed ill-formed ImplicationLink from the Arcana example. Please make sure I didn't remove important test cases in the process.


ngeiswei:
There's been no feedback on that one, so I'm gonna merge it.

tests/query/arcana-repeat.scm
```diff
@@ -27,64 +27,6 @@
 	)
 )
 
-; Three parts, all three identical
-(ImplicationLink
-	(ListLink
```
linas: Nil, can you restore these tests, and just change ImplicationLink to some other ordered link type? The  link type here mostly does not matter,  what did matter was the structure of the patterns being explored. These tests are important; they're testing for some complex traversal scenarios.

opencog/atoms/execution/Instantiator.cc
```diff
@@ -275,7 +275,10 @@ Handle Instantiator::walk_tree(const Handle& expr)
 	HandleSeq oset_results;
 	bool changed = walk_tree(oset_results, lexpr->getOutgoingSet());
 	if (changed)
-		return Handle(createLink(t, oset_results, expr->getTruthValue()));
+	{
+		LinkPtr subl = createLink(t, oset_results, expr->getTruthValue());
+		return Handle(_as->add_atom(subl));
```
linas: This wasn't done before, and should not be  needed now, because the top-most result is added to the atomspace; the lower-level results do not need to be.

Kind-of off-topic, but I should mention this: the instantiator, as currently written, is a bit mis-designed; it does eager evaluation, and should really do lazy evaluation instead.

linas:
No feedback because I was travelling. I only got back online just now :-)

opencog/atoms/execution/Instantiator.cc
ngeiswei: This is to address issue #472

linas: Ah OK, yes, that makes sense.  So, several remarks:
-- You didn't really need an ImplicationLink C++ class; instead, the factory needed to be altered to put a `if classserver()isType(t, SCOPE_LINK) {createScopeLink(t, args); }` to find the variables.  Or if you want to keep it, then `classsever().isType(IMPLICATION_LINK...)` to pick up the extensional etc. links
-- The factory needs to be called ... 
-- The code needs to be changed to do lazy evaluation (and do a lazy factory as well)
-- A unit test case is needed for the new ImplicationLink so that when the lazy-eval stuff is added, the implicaitonlink is not broken.

opencog/atomspace/AtomTable.cc
```diff
@@ -285,6 +286,9 @@ AtomPtr AtomTable::do_factory(Type atom_type, AtomPtr atom)
     } else if (LAMBDA_LINK == atom_type) {
         if (NULL == LambdaLinkCast(atom))
             return createLambdaLink(*LinkCast(atom));
+    } else if (IMPLICATION_LINK == atom_type) {
+        if (NULL == ImplicationLinkCast(atom))
+            return createImplicationLink(*LinkCast(atom));
```
linas: per above, this probably should be `classserver.isA`, just like further below.

opencog/atoms/execution/Instantiator.cc
ngeiswei: Oh, indeed, thanks for the tip regarding using classserver().isType(IMPLICATION_LINK...) to pick up the extensional, etc.

PR #480, opened 2015-11-06
Boost signal for importance value change event 

misgeatgit:
We need this for analyzing ECAN.


misgeatgit:
This is wrong; I'm closing it.

PR #481, opened 2015-11-06
Added boost signal for Importance changes

misgeatgit:


opencog/atomspace/AtomSpace.h
```diff
@@ -673,22 +703,50 @@ class AtomSpace
     /** DEPRECATED! Do NOT USE IN NEW CODE!
```
AmeBel: Why extende a deprecated code?

@linas why are these deprecated?

misgeatgit:
I am closing this PR as well since https://github.com/opencog/atomspace/blob/master/opencog/atomspace/AtomSpace.h#L582 Does same thing.

opencog/atomspace/AtomSpace.h
linas: because code will get simpler and more readable if you just "do it yourself" instead of calling these wrapper ffunctions.

PR #482, opened 2015-11-09
Add rbs-rules scm function

AmeBel:
rbs-rules returns the alias for rules in a rule-base.


AmeBel:
@misgeatgit @williampma pls review, thanks

opencog/rule-engine/InferenceSCM.cc
```diff
@@ -131,6 +134,32 @@ Handle do_backward_chaining(AtomSpace* as,
 #endif
 }
 
+HandleSeq get_rulebase_rules(AtomSpace* as,
```
misgeatgit: May be you should move this to UREConfigReader class.

opencog/rule-engine/InferenceSCM.cc
```diff
@@ -131,6 +134,32 @@ Handle do_backward_chaining(AtomSpace* as,
...
+                          const Handle& rbs)
+{
+#ifdef HAVE_GUILE
+    if (Handle::UNDEFINED == rbs)
+        throw RuntimeException(TRACE_INFO,
+            "InferenceSCM::get_rulebase_rules - invalid rulebase!");
+
+	UREConfigReader ure_config(*as, rbs);
```
misgeatgit: Indentation.

opencog/rule-engine/InferenceSCM.cc
AmeBel: there is a get_rules already, you mean creating one that returns a `HandleSeq`?

misgeatgit:
@AmeBel  Whats the purpose these commits BTW? What is it solving?

AmeBel:
In OpenPsi an active-scheme-pool is a rulebase. Unlike the r2l or pln the rulebase isn't static it is dynamic,i.e, the set of rules change when goals changes. 

Yes, this searching can be done using scheme, but then should the definition of a rule-pattern change then I have to update scheme-function.

opencog/rule-engine/InferenceSCM.cc
misgeatgit: OK, I am wrong. Looks it is in the right place.

misgeatgit:
Thanks, now I understood.

Issue #485, opened 2015-11-11
InitiateSearchCB::variable_search only search over the rarest type


ngeiswei:
There is code in InitiateSearchCB::variable_search to only search grounding with the rarest variable type, see

```c++
	// Find the rarest variable type;
	...
```

and the code following.

I understand the drive to save computational resources behind this decision but that isn't what the user expects when using a TypeChoice. Clearly the pattern matcher should offer the possibility to perform a full search of all types.

I guess the right fix is to code another callback that overwrite variable_search to offer that.

Anyway comment or idea is welcome though.

ngeiswei:
Actually reviewing the code further I think the right fix should be done here, not in another callback. The right fix is to choose the variable with the rarest type set, not type singleton.

linas:
Yes, its a bug.  Each possible type choice should have been explored.  The patch looks good to me.

PR #486, opened 2015-11-11
Fix misc pm

ngeiswei:
Take a look at this PR, it is important and might change the URE behavior (I think for the better but perhaps at some computational cost).

* ~1. It allows to ground terms that have in them scoped variables that have the same name as the pattern. So for instance the following pattern
```atomese
GetLink
   VariableList
      TypedVariableLink
         VariableNode "$X"
         TypeChoice
            TypeNode "PredicateNode"
            TypeNode "LambdaLink"
      TypedVariableLink
         VariableNode "$Y"
         TypeChoice
            TypeNode "PredicateNode"
            TypeNode "LambdaLink"
   ImplicationLink
      VariableNode "$X"
      VariableNode "$Y"
```

will be able to deal with grounded terms containing `$X` or `$Y` like

```atomese
  ImplicationLink
     PredicateNode "P"
     LambdaLink
        VariableNode "$X"
        EvaluationLink
           PredicateNode "P"
           VariableNode "$X"
```

as there are scoped there is no ambiguity.

I know that we said that we want to have unique variable names, but frankly having the system break apart because variable names are reused sounds very brittle to me. At a formal level, it certainly shouldn't be required. At a practical level we may still decide to enforce it, but at least the code should give us the choice.

* ~2. When the pattern body is only composed of variables the search is operated over all the types, not just the rarest one. For instance

```atomese
GetLink
   VariableList
      TypedVariableLink
         VariableNode "$X"
         TypeChoice
            TypeNode "PredicateNode"
            TypeNode "LambdaLink"
      TypedVariableLink
         VariableNode "$Y"
         TypeChoice
            TypeNode "PredicateNode"
            TypeNode "LambdaLink"
   AndLink
      VariableNode "$X"
      VariableNode "$Y"
```

`$X` and `$Y`will be searched over PredicateNodes and LambdaLinks, not just the rarest type. It's more costly but is what the user expect.

This addresses the issue #485.

### DIFF
--- a/opencog/query/InitiateSearchCB.cc
+++ b/opencog/query/InitiateSearchCB.cc
@@ -607,9 +607,9 @@ bool InitiateSearchCB::variable_search(PatternMatchEngine *pme)
 
 	// Find the rarest variable type;
 	size_t count = SIZE_MAX;
-	Type ptype = ATOM;
+	std::set<Type> ptypes;
 
-	LAZY_LOG_FINE << "varset size = " <<  _variables->varset.size();
+	LAZY_LOG_FINE << "_variables = " <<  _variables->to_string();
 	_root = Handle::UNDEFINED;
 	_starter_term = Handle::UNDEFINED;
 	for (const Handle& var: _variables->varset)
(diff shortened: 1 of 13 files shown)


opencog/query/InitiateSearchCB.cc
```diff
@@ -620,40 +620,42 @@ bool InitiateSearchCB::variable_search(PatternMatchEngine *pme)
...
-			LAZY_LOG_FINE << "Type = "
-			              << _classserver.getTypeName(t) << " has "
-			              << num << " atoms in the atomspace";
-			if (0 < num and num < count)
+			for (const Handle& cl : clauses)
 			{
-				for (const Handle& cl : clauses)
+				// Evaluatables dont' exist in the atomspace, in general.
+				// Cannot start a search with them.
+				if (0 < _pattern->evaluatable_holders.count(cl)) continue;
```
linas: is there a 'contains' method that would be faster? or maybe `try {evaluatable_holders.at(cl); } catch(...)`

linas:
Sounds good to me. CircleCI says QuoteUTest failed, though.

opencog/query/InitiateSearchCB.cc
ngeiswei: There would be find, but since it's a set (as opposed to a multiset), count should be just as fast.

PR #497, opened 2015-11-22
Add unquote support

ngeiswei:
Add/complete unquote support according to http://wiki.opencog.org/wikihome/index.php/UnquoteLink

The code is spread across multiple places. I think (but I'm not 100% sure) one can simplify all that by setting clauses as PatternTerms. But I'm letting that for another PR, if any.


opencog/atoms/pattern/PatternLink.cc
```diff
@@ -589,7 +589,7 @@ void PatternLink::unbundle_virtual(const std::set<Handle>& vars,
 		bool is_virtual = false;
 		bool is_black = false;
 
-#ifdef BORKEN_DOESNT_WORK
+#ifdef BROKEN_DOESNT_WORK
```
linas: well, if it merely broken, then its not borken.  (c.f. Robert Bork, who was borked.)

linas:
wow that was a lot of changes.

linas:
There is one important spot that I think you missed: `atoms/execution/Instantiator.cc`

ngeiswei:
Fixed, thanks.

PR #499, opened 2015-11-23
Support Instantiator quotation level + utest

ngeiswei:


ngeiswei:
The result for the QuoteLink infloop utest has changed from

```
infloop is (SetLink
  (ListLink
    (ConceptNode "And the answer is ...") ; [179][1]
    (QuoteLink
      (VariableNode "$x") ; [178][1]
    ) ; [180][1]
  ) ; [181][1]
  (ListLink
    (ConceptNode "And the answer is ...") ; [179][1]
    (ConceptNode "And the answer is ...") ; [179][1]
  ) ; [188][1]
  (ListLink
    (ConceptNode "And the answer is ...") ; [179][1]
    (SetLink
      (ListLink
        (ConceptNode "And the answer is ...") ; [179][1]
        (QuoteLink
          (VariableNode "$x") ; [178][1]
        ) ; [180][1]
      ) ; [181][1]
    ) ; [186][1]
  ) ; [189][1]
  (ListLink
    (ConceptNode "And the answer is ...") ; [179][1]
    (ListLink
      (ConceptNode "And the answer is ...") ; [179][1]
      (QuoteLink
        (VariableNode "$x") ; [178][1]
      ) ; [180][1]
    ) ; [181][1]
  ) ; [190][1]
  (ListLink
    (ConceptNode "And the answer is ...") ; [179][1]
    (BindLink
      (VariableNode "$x") ; [178][1]
      (ListLink
        (ConceptNode "And the answer is ...") ; [179][1]
        (VariableNode "$x") ; [178][1]
      ) ; [183][1]
    ) ; [184][1]
  ) ; [191][1]
  (ListLink
    (ConceptNode "And the answer is ...") ; [179][1]
    (BindLink
      (VariableNode "$x") ; [178][1]
      (ListLink
        (ConceptNode "And the answer is ...") ; [179][1]
        (QuoteLink
          (VariableNode "$x") ; [178][1]
        ) ; [180][1]
      ) ; [181][1]
    ) ; [182][1]
  ) ; [192][1]
) ; [193][1]
```

to

```
infloop is (SetLink
  (ListLink
    (ConceptNode "And the answer is ...") ; [3][1]
    (VariableNode "$x") ; [2][1]
  ) ; [7][1]
  (ListLink
    (ConceptNode "And the answer is ...") ; [3][1]
    (ConceptNode "And the answer is ...") ; [3][1]
  ) ; [12][1]
  (ListLink
    (ConceptNode "And the answer is ...") ; [3][1]
    (ListLink
      (ConceptNode "And the answer is ...") ; [3][1]
      (VariableNode "$x") ; [2][1]
    ) ; [7][1]
  ) ; [13][1]
  (ListLink
    (ConceptNode "And the answer is ...") ; [3][1]
    (BindLink
      (VariableNode "$x") ; [2][1]
      (ListLink
        (ConceptNode "And the answer is ...") ; [3][1]
        (VariableNode "$x") ; [2][1]
      ) ; [7][1]
    ) ; [8][1]
  ) ; [14][1]
) ; [15][1]
```

Note how the QuoteLinks have disappeared in the new results. That is because the QuoteLinks going to level 1 are executed, and so disappear, just like it would in scheme.

ngeiswei:
The infinit loop detector also catches the infinite loop more earlier, but I'm not completely sure why, but I guess it's because the results look more like the infloop query, due to the QuoteLinks disappearing.

linas:
I'm slightly confused about the disappearing QuoteLinks, but given that I have no immediate plans for using this, I'll let it be.

ngeiswei:
It comes from l.122 of Instantiator::walk_tree. I.e. the variable substitution is itself walked, which makes the QuoteLink disappear. I think it is the correct behavior (although it's ultimately a matter of choice).

linas:
I saw the code, did not think hard about what it did or what it should do.

The result seemed odd, though, let me see if I can explain why: So, the bind link (in the QuoteUTest) is effectively saying "get me everything in the atomspace", since the body is just a single variable, it has no type constraints, and thus, it can match any/everything in the atomspace. 

One of the things it can match are bindlinks.  One of them ("crasher") just happens to have a QuoteLink in it. Why is it being re-written, so that the QuoteLink is removed? Why isn't it being returned verbatim?

ngeiswei:
In order to return verbatim one would just need to wrap the rewrite term of the BindLink in a QuoteLink. Although this would work unless the matches have unquote links in them, but that's just following the quasiquote semantics.

I don't really know whether it is ultimately what we want, but it looks like a coherent behavior, so I'm fine to keep it that way, till we realize it's bad, if ever.

linas:
Well, there's one potential trouble spot that I see: rule selection. So, for example @kinoc is writing rules to find rules: A given input text string is mashed into a bindlink, which is used to search for all possible bindlinks that might be triggered by that input sentence.  Then one or more of these is selected and run.  He is not currently using quotes for anything, but if he was, this could lead to trouble.

We may want to consider doing something similar for the rule engine, in general. One reason that relex is slow is it tries to apply *every* rule for each sentence.  This is rational, since we don't really know which rules might apply or not apply, but it would improve the situation if we could have "flows": attempt rule B only if rule A triggered.  Again, this might need bindlinks returning bind-links, and thus quote nesting could become an issue.

PR #504, opened 2015-11-25
Avoid the use of the export fuinction.

linas:
Somehow, its buggy. Perhaps it is being done in the wrong environment.
I can't tell. It results in weird undefined-function errors in sureal
and microplanning, for completely obscure reasons that I just cannot
isolate.

### DIFF
--- a/opencog/scm/opencog.scm
+++ b/opencog/scm/opencog.scm
@@ -52,7 +52,6 @@
 (load-from-path "core-docs.scm")
 
 (load-from-path "utilities.scm")
-(export-utilities)
 
 (load-from-path "apply.scm")
 (load-from-path "av-tv.scm")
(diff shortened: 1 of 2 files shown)


ngeiswei:
Very strange... @linas what error message do you get (when using export instead of define-public)?

ngeiswei:
Oh, I get it "weird undefined-function errors". Strange...

linas:
yeah, it was strange; I was trying to convert the chatbot into a guile module, and kept getting stuff like this:
```
;;; compiling opencog-git/build/..//opencog/nlp/sureal/sureal.scm
;;; WARNING: compilation of opencog-git/build/..//opencog/nlp/sureal/sureal.scm failed:
;;; ERROR: In procedure module-lookup: Unbound variable: stv
```
even though utilities.scm was obviously loaded...  it was mind-boggling.

PR #510, opened 2015-12-01
Extended ExecutionOutputLink to call functions from a shared, c-style Library

rTreutlein:
This is to be used mainly for Libraries created with Haskell.

GSN "lib: libname.so\\functionname"

This also includes the Changes to get the Haskell Bindings onto Hackage. Currently i am the only maintainer on Hackage if i should add anybody i will just need their Hackage Username.


ngeiswei:
That looks really cool, @rTreutlein ! :-) Do you think you can add a unit test for the lib schema execution?

opencog/atoms/execution/ExecutionOutputLink.cc
```diff
@@ -195,6 +196,49 @@ Handle ExecutionOutputLink::do_execute(AtomSpace* as,
...
+        std::string funcName = schema.substr(dotpos+1);
+
+        void *libHandle;
+        void * tmp;
+        UUID (*func)(AtomSpace* ,UUID);
+
+        //Try and load the Library and the Function
+        if ( (libHandle = dlopen(libName.c_str(), RTLD_LAZY)) == NULL ||
+             (tmp       = dlsym(libHandle, funcName.c_str())) == NULL )
+        {
```
linas: Interesting. What is this for?  What's the intended use?

opencog/atoms/execution/ExecutionOutputLink.cc
```diff
@@ -195,6 +196,49 @@ Handle ExecutionOutputLink::do_execute(AtomSpace* as,
...
+            throw RuntimeException(TRACE_INFO,
+                "LibName and FunctionName not seperated by a '\\'");
+        }
+        std::string libName  = schema.substr(pos,dotpos-pos);
+        std::string funcName = schema.substr(dotpos+1);
+
+        void *libHandle;
+        void * tmp;
+        UUID (*func)(AtomSpace* ,UUID);
```
linas: I've been discouraging the use of UUID's -- they really suck, for a large variety of reasons.  Use either Handles of AtomPtr for the API.

opencog/atoms/execution/ExecutionOutputLink.cc
rTreutlein: Well this will load a given library at run time to call 1 function from it. I wanted to be able to call Haskell Code from the AtomSpace and since Haskell is usually compiled we will have to load a library at runtime and call function from that instead of parsing the code. Of course this code will work with any kind of library that is in a C-style.

opencog/atoms/execution/ExecutionOutputLink.cc
rTreutlein: Well that wasn't my decision, it's how the Haskell Bindings are written but i could rewrite them to use handles at some point.

rTreutlein:
@linas @ngeiswei Unless you have any other comments i will merge this later today.

linas:
Fine with me I guess. It would probably make more sense to change the dlopen interface to something like :haskell: instead of the name of the so.  You'd also want to cache the dlopen results: each invocation of dlopen/load causes the glibc ld.so to run its course, which is an eyeblink but still a very non-trivial number of cycles.  But you can save this for a later pull.

Lest you think think that the dlopen interface is usable by "anybody" not just haskel .. banish that thought.  The UUID interface is a design error,  and, well, even if it changed to use Handle or AtomPtr, I don't want to encourage that style. It doesn't seem wholesome; it sort of screams "I'm too efing lazy to do it right so I'm gonna write spaghetti code which someone else can fix after I've quit this project".  not the dlopen code -- that's nice in itself; I mean if anyone builds some big scheme program that is then using this dlopen to call into random ELF binaries.   We've got the right way to add scheme wrappers: the scheme primitive thing. Use that.

rTreutlein:
@linas I don't see another way to write and then call haskell at runtime. Theoretically it could be interpreted but that is very slow and should only be used during development.

Also why souldn't the Atomspace be able to call any library? Maybe we will get some bindings to another language in the future than this would be an easy way to call that code from the AtomSpace.

linas:
Eh? I just told you why, read what I wrote above!  Also, I did not say ""run haskell interpreted" I said "change the format of the URL".

rTreutlein:
So it should only check in a specific directory for the Haskell libraries? I find it hard to understand what exactly you mean.

linas:
I mean make it work with a URL like `haskell:name_of_function_to_be_called`   or maybe something like `haskell://module_name:name_of_function_to_be_called` or maybe `hkl:name_of_function_to_be_called@module_name`   where module_name can be found in `libmodule_name.so` or something like that.   The actual file-system paths to be searched needs to be outside of all this, maybe in some config file, or pre-loaded into some haskell search-and-load mechanism. 

Also -- the actual library name also needs to be avoided: on Apple Mac OSX, libraries are called libblah.dylib and not libblah.so ; also OSX does some other  they weird ELF format conventions. 

I also have vague hopes of porting opencog to android, to run it on cell phones ... 

The point is: if you need to load haskel code, load haskell code; don't pretend you're doing something generic, cause your not.

linas:
When you say "I don't see another way to write and then call haskell at runtime." are you telling me that the haskell people never ever invented a way for C++ code to call haskell?  That their attitude is just "use dlopen, sayonara sucker"? I find that hard to believe.  Surely they have developed something more elegant than that.

rTreutlein:
Other than linking it during compile time there isn't any other way to call Haskell Code from C++ that i am aware of.

Issue #508, opened 2015-11-30
 Unbound variable: ConceptNode Afer running (use-modules (opencog)) in SchemeEval


AmeBel:
What am I missing?

#### Test class
```c++
class ActionUTest : public CxxTest::TestSuite
{
private:
    AtomSpace _as;
    SchemeEval _scm;
public:
    ActionUTest() : _scm(&_as)
    {
        // Print output to standard
        logger().setLevel(Logger::DEBUG);
        logger().setPrintToStdoutFlag(true);
        logger().setPrintLevelFlag(true);
    }

    ~ActionUTest()
    {
        // erase the log file if no assertions failed
        //if (!CxxTest::TestTracker::tracker().suiteFailed())
        //    std::remove(logger().getFilename().c_str());
    }

    void setUp()
    {
        cout <<_scm.eval("(add-to-load-path \"/usr/local/share/opencog/scm\")");
        cout << _scm.eval("(use-modules (opencog) (opencog rule-engine))");
        cout << _scm.eval("(load \"test-rbs.scm\")");
    }


    // Why this check? B/c PatternLink api's behavior might change.
    // What is being checked? That the derived rule doesn't have any form of
    // virtual-terms.
    void test_constuctor() {
   ...
   ...
    }
};
```

#### Content of `test-rbs.scm`
```
(ConceptNode "abc")
```

#### Output of running the test
```bash
opencog@4fd61e755acb:/opencog/build/tests/planning$ ctest -VV ActionUTest
UpdateCTestConfiguration  from :/opencog/build/tests/planning/DartConfiguration.tcl
UpdateCTestConfiguration  from :/opencog/build/tests/planning/DartConfiguration.tcl
Test project /opencog/build/tests/planning
Constructing a list of tests
Done constructing a list of tests
Checking test dependency graph...
Checking test dependency graph end
test 1
    Start 1: ActionUTest

1: Test command: /opencog/build/tests/planning/ActionUTest
1: Test timeout computed to be: 9.99988e+06
1: 
1: 
1: Backtrace:
1: In ice-9/boot-9.scm:
1:  157: 13 [catch #t #<catch-closure 24f5c60> ...]
1: In unknown file:
1:    ?: 12 [apply-smob/1 #<catch-closure 24f5c60>]
1: In ice-9/boot-9.scm:
1:  157: 11 [catch #t #<catch-closure 24f5b40> ...]
1: In unknown file:
[...]

PR #511, opened 2015-12-01
Proposed fix for issue #508

AmeBel:
### WIP Don't Merge
I am not sure whether this is a good approach for solving https://github.com/opencog/atomspace/issues/508 thus a proposal. 

**Intention:** be able to run `(use-modules (opencog))` without having to deal with hardcoding path in tests . Detail of usability and technical issue see https://github.com/opencog/atomspace/issues/508#issuecomment-160695583 and https://github.com/opencog/atomspace/issues/508#issuecomment-160706429 

**Proposed Solution:** use `GUILE_LOAD_PATH` to set the path to be searched for modules. That way path variables need not be hard-coded in source-files, and cmake could be used to set the value of `GUILE_LOAD_PATH`.

As it is right now ths fixes the https://github.com/opencog/atomspace/issues/508, but ParallelUTest fails, and using cmake's `SET_TESTS_PROPERTIES` as used for setting "PYTHONPATH" in other tests seems to have no effect. Don't know why for both.

### DIFF
--- a/opencog/guile/SchemeSmob.cc
+++ b/opencog/guile/SchemeSmob.cc
@@ -9,9 +9,11 @@
 #ifdef HAVE_GUILE
 
 #include <cstddef>
+#include <cstdlib>
 #include <libguile.h>
 
 #include <opencog/atomspace/AtomSpace.h>
+#include <opencog/util/StringManipulator.h>
 #include "SchemePrimitive.h"
 #include "SchemeSmob.h"
 
@@ -45,6 +47,25 @@ void SchemeSmob::init()
 	if (is_inited.test_and_set()) return;
 
 	init_smob_type();
+
+	// If one is importing this class then the functions defined in
+	// .scm files associated with '(opencog)' module should be available.
+	// Assuming installation in a standard path.
+	scm_c_eval_string("(add-to-load-path \"/usr/local/share/opencog/scm\")\n");
+
+	// Add paths set through GUILE_LOAD_PATH
+	std::string env_str;
+
+	if (const char* _env = std::getenv("GUILE_LOAD_PATH")) {
+		env_str = _env;
+		auto output = StringManipulator::split(env_str, ":");
+
+		for (const auto& i : output) {
+			std::string scm_command = "(add-to-load-path \"" + i + "\")";
+			scm_c_eval_string(scm_command.c_str());
+		}
+	}
+
 	scm_c_define_module("opencog", register_procs, NULL);
 	scm_c_use_module("opencog");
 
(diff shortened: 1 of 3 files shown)


opencog/guile/SchemeSmob.cc
```diff
@@ -45,6 +47,25 @@ void SchemeSmob::init()
 	if (is_inited.test_and_set()) return;
 
 	init_smob_type();
+
+	// If one is importing this class then the functions defined in
+	// .scm files associated with '(opencog)' module should be available.
+	// Assuming installation in a standard path.
+	scm_c_eval_string("(add-to-load-path \"/usr/local/share/opencog/scm\")\n");
```
linas: This should be CMAKE_BINARY_INSTALL_DIR or something like that -- wherever cmake installs, which defaults to `/usr/local`

opencog/guile/SchemeSmob.cc
```diff
@@ -262,6 +283,26 @@ void SchemeSmob::register_procs(void*)
...
+	register_proc_from_scm("config.scm");
+
+	register_proc_from_scm("core-docs.scm");
+
+	register_proc_from_scm("utilities.scm");
+
+	register_proc_from_scm("apply.scm");
+	register_proc_from_scm("av-tv.scm");
+	register_proc_from_scm("file-utils.scm");
+	register_proc_from_scm("debug-trace.scm");
```
linas: Ugh. I can see that this works. Its vaguely annoying.  Its a shell-game.  Today, these loads are done in the SCM_PRELOAD section of the opencog.conf file.  All you've done is to move them from the conf file, to here.  However ugly the conf file is, its still easier to use than altering C++ code, and recompiling...

opencog/guile/SchemeSmob.cc
```diff
@@ -45,6 +47,25 @@ void SchemeSmob::init()
...
+
+	if (const char* _env = std::getenv("GUILE_LOAD_PATH")) {
+		env_str = _env;
+		auto output = StringManipulator::split(env_str, ":");
+
+		for (const auto& i : output) {
+			std::string scm_command = "(add-to-load-path \"" + i + "\")";
+			scm_c_eval_string(scm_command.c_str());
+		}
+	}
```
linas: I like this, but it would be better to put it here: https://github.com/opencog/atomspace/blob/master/opencog/guile/load-file.cc#L82  which is the set of paths currently being searched.

Or better yet, maybe `load_scm_file_relative()` should be completely removed, and replaced by `scm_c_primitive_load_path()` and by this add-to-load-path mechanism.

PR #512, opened 2015-12-02
Misc improve tv

ngeiswei:
* Move TV code to its own folder /opencog/truthvalue
* Set the lookahead globally for all TV types
* Fix Indefinite TV scheme binding + add mean/count/confidence get


opencog/guile/SchemeSmob.cc
```diff
@@ -232,6 +232,9 @@ void SchemeSmob::register_procs(void*)
 	register_proc("cog-ptv?",              1, 0, 0, C(ss_ptv_p));
 	register_proc("cog-ftv?",              1, 0, 0, C(ss_ftv_p));
 	register_proc("cog-tv->alist",         1, 0, 0, C(ss_tv_get_value));
+	register_proc("cog-tv-mean",           1, 0, 0, C(ss_tv_get_mean));
+	register_proc("cog-tv-confidence",     1, 0, 0, C(ss_tv_get_confidence));
+	register_proc("cog-tv-count",          1, 0, 0, C(ss_tv_get_count));
```
linas: I like all of the rest of this patch, except for this. It would be much better to implement this in scheme code, 

The utilities.scm file already implements tv-mean and tv-conf and tv-count. You can rename these to cog-tv-mean etc if you wish.  They take a TV by default, but maybe you can implemnt a version that takes atoms directly.

opencog/guile/SchemeSmobTV.cc
```diff
@@ -460,5 +460,32 @@ SCM SchemeSmob::ss_tv_get_value (SCM s)
...
 }
 
+/**
+ * Return the truth value mean
+ */
+SCM SchemeSmob::ss_tv_get_mean(SCM s)
+{
+	TruthValue *tv = verify_tv(s, "cog-tv-mean");
+	return scm_from_double(tv->getMean());
+}
```
linas: OK .. so you work with TV ... just edit utilities.scm and rename tv-mean to cog-tv-mean and you'll get exactly the same thing, with no new C++ code.

linas:
I'm all for this, I've thought about doing this too. Just don't glom up the over-sized c++ api for guile, and its good.

linas:
I've also thought about moving the base Atom Node and Link defintions to their own directory.   As a result, the atomspace dir would contain ONLY the actual database code, and nothing else.

linas:
Oh, and a third remark: per talks with Ben for creating a generic truth value type thing for storing generic float pt values, I started work on a "ProtoAtom" which is just like an Atom, but is missing a TV and an AV and the UUID.  There is code in the protoatom branch. It works. you can try it. Its incomplete, because I can't decide the best way to place protoatoms into links. Should they have handles or not?  etc. See issue .. Hmm There is no protoatom issue!

Anyway, a long-term goal might be to use protoatoms instead of hard-coded TV, although this could be slower as a result.  Hard-coded TV has real advanteages for performance.

linas:
I opened #513 to track protoatom status

opencog/guile/SchemeSmob.cc
ngeiswei: I wasn't aware of these helpers, but they do not work the same. The ones I've coded calls the overloaded TruthValue::getMean(), TruthValue()::getConfidence() and TruthValue()::getCount(), while your helpers call the TV alist from the scheme representation. Those calls are different! In my case I can apply cog-tv-mean, etc. to any TV type inheriting TruthValue, even if it's missing from its alist. I need this to easily experiment with TV type conversion.

ngeiswei: I suggest to remove the scheme helpers and use my direct bindings to the C++ calls instead. Do you agree?

opencog/guile/SchemeSmobTV.cc
ngeiswei: As I said, I don't think I would get the same (but wait, I'm gonna try).

ngeiswei: I confirm, the scheme helpers tv-mean, tv-conf and tv-count return #t when given an IndefiniteTV.

These helpers are only used in the OpenPsi, and in all instances it would shorten the code a bit as cog-tv-mean, etc, are called directly on the atom.

So I suggest to remove the helpers. I assume you agree and I'll proceed.

ngeiswei: It is used in microplanning as well, but my remark still holds.

linas: Well, the scheme code should be fixed up to do "the right thing" even for ITV. No one has used the ITV code before, so it wasn't supported.

The scheme wrappers are used by the language-learning code to store mutual information and counts in TV values, and I fear that TV merge strategies will screw these up.  You can't just merge MI unless you know what you're doing and why you're doing it. This is why I take a very dim view of doing the TV merge in C++ code -- it is almost guaranteed to do the wrong thing at the wrong time for the wrong reason.  This needs to be fixed.

linas: I mean, if you want to merge this branch, do it, this is not a show-stopper.  Just in general, I would prefer not to glom up the C++ API with useless code that would be better done in scheme, directly, where you'll get better performance (since the scheme code is compiled! (into bytecode) whereas the C++ code requires three slow and clunky calls across the guile-C++ interface. The bytecode will run much faster than two or three API calls could -- approximately 10x or 20x faster, based on benchmark measurements in the /opencog/benchmark directory)

ngeiswei: Actually cog-tv-mean works on the TV not the atom (so my remark about having it shorten some scheme code isn't valid). But it is sill more generic than the helpers.

ngeiswei: Oh, I didn't know about the clunky guile-C++ interface!

I wouldn't mind to have them coded in scheme, expect that given the current bindings I cannot. However, maybe I can keep the existing tv-mean, etc. I removed them to avoid duplicating code, but if that slowing things down that would be a good reason not to remove them.

linas: I think you're missing the point: the scheme code will be FASTER than the C++ code. Here's why.  One call across the guile-to-scheme interface takes about 60 microseonds (on my machine, yours will be different) Lets assume that call returns all 3 TV values (mean confidence count).  Unpacking those three values from the a-list takes about a microscond each, so grand total 63 microseconds.   By contrast, if you do three calls across the C++-guile interface to get three single floats, it will take you 60+60+60 == 180 microseconds. Its a bad deal.  Do as much as possible in scheme; minimize boundary crossings.

ngeiswei: OK, I understand. So what I should do is have one C++ binding for {mean, confidence, count}, and then use scheme helpers to extract them.

Alternatively, I can add these mean, confidence, count fields to the indefinite TV scheme alist representation, which would make the tv-mean, tv-conf and tv-count work. There's a catch though, requesting the count of an itv might be computationally expensive (it isn't today, but the algorithm is wrong), and so is better done only when necessary.

So I'm a bit puzzled about what to do.

But wait!!! cog-tv->alist isn't memoized, is it? If not then calling each once tv-mean, tv-conf and tv-count should be roughly equivalent to calling cog-tv-mean, cog-tv-conf and cog-tv-count (even slightly faster), since tv-mean, etc rely on cog-tv->alist.

linas: well, I don't think its limited to guile-c++, similar numbers apply to cython/python. 

Hmm, as I write this, I realize some of the numbers I quoted are wrong.

I'm thinking that I might have mis-quoted the numbers, -- there's a big difference between entering gule for the first time, and entering it again a second time -- there's a fast track for the second time, since everything is already set up. Likewise for the reverse direction.   You'd have to measure carefully and ponder.

linas: Don't confuse memoizng and compiling.  Memoizing simply means that all lambdas have been expanded, or whatever. Don't know the details.  Compiling is much much more: the expressions are converted into byte-code, which is then cached in the file-system.  The byte-code runs at typical VM speeds -- comparable to Java I suppose, which is typically 1/3rd the speed of native assembly code.  I have not measured.  By default, all scheme code is compiled, so tv->alist would be compiled, and presumably the optimizer would unwind and unroll references & etc.  whatever it does ... there is a way to turn off compilation and memoization and always run  interpreted, and I did measure that once for opencog, it wasn't horrible but it wan't great.  (It wasn't horrible mostly because the guile snippets in opencog are always rather small and short.  If we were writing giant programs in guile, it would be a huge difference. But most code in e.g. utilities.scm is small.)

ngeiswei:
Perhaps, instead of having cog-tv-mean, etc, we could have scheme bindings for TV type converters, like
```scheme
tv->stv
tv->itv
...
```
so if one wants the mean, confidence and count of a TV it can convert it to an stv and look at its alist. Such converters will likely be useful elsewhere too.

ngeiswei:
I'm gonna revert the removal of tv-mean, tv-conf and tv-count, but keep the cog-tv-mean, cog-tv-confidence and cog-tv-count for now. Besides, I'm not sure what is gonna be the right way to use the various tv types so I prefer to decide later.

linas:
alists are just key-value thingies.  You can jam additional itv keys into the alist.  You could do something like if alist count is not found, then check to see if its itv, and if so then compute the count.

You can even avoid the cost of the if-statements for the default path, by throwing exceptions for the non-default case, and catching them, and doing any extra heavy computation there. 

Whatever, merge this now, the details can be sorted out later.

linas:
yeah, please don't remove tv-conf, tv-count.

PR #514, opened 2015-12-03
Bump version number, for the truth-value library.

linas:


linas:
whatever. pointless and unused.

PR #516, opened 2015-12-04
Fuzzy matcher updates

leungmanhin:
First batch of the updates, this should fix the problem of not finding the best match. I'll update the unit test to be more meaningful soon.


opencog/query/PatternMatchCallback.h
```diff
@@ -238,6 +238,16 @@ class PatternMatchCallback
 		                       const std::map<Handle, Handle> &term_soln) = 0;
 
 		/**
+		 * Called when we are comparing unordered links. It would be useful
+		 * if we want to skip this grounding and move on to the next right away
+		 * without having to compare all of its possible permutations.
+		 */
```
linas: There is an easier way to do this, without having to alter the pattern matcher.

What you do in the code below is to always declare a mis-compare, whenever you encounter an unordered link (you return false, when skipping permutation compares)  

You can accomplish the same thing by changing  FuzzyMatcher::link_match() to return false whenever the link type is unordered.

leungmanhin: I don't know if I understood this correctly. For example if there are two groundings G1 and G2, both are unordered links, and G2 is the one we want but not G1. Currently if G1 reaches the grounding first, it seems that we have to go through all permutations of G1 before G2 reaching grounding?

linas: I was only looking at what your code actually did, not what you want it to do.  By looking at what it actually did, I suggested a different way of writing it.   What your code did was to completely skip and always fail the grounding of unordered links.

leungmanhin:
ok I've updated the code and removed that callback from the pattern matcher.

linas:
thanks!

PR #519, opened 2015-12-04
Fix build break.

linas:
Fix a terribly OBVIOUS build break which circle-CI didn't catch. Once again, what good is circle-CI, if it can't even perform basic checks?


linas:
I don't know how you guys are building; this still breaks the build for me.

PR #520, opened 2015-12-04
Fix build break.

linas:
Another build break which circle-CI failed to detect. WTF?


AmeBel:
closing b/c changes are included in https://github.com/opencog/atomspace/issues/525

PR #515, opened 2015-12-03
Enabled Haskell Tests

rTreutlein:
Not sure why they where disabled.
Improved 1 Haskell Test

PR #522, opened 2015-12-06
Misc build fixes

pennachin:


pennachin:
So I've run CircleCI on this pull request twice, after the initial build, for a total of three builds. The first build (the automatic one triggered by creating the PR) passes. Then the second build failed. This is exactly what happened with #515, it passed for the first build, and then failed afterwards, but it had been merged by then.

The third build, without caching, also fails, so this doesn't seem to be a cache issue. I see a similar pattern in my local environment: the first build passes, subsequent builds fail. I have no idea what is causing that, so I'm reverting #515 so we get a usable master again.

PR #524, opened 2015-12-07
Revert 523 revert 515 gsn call lib with Fix

rTreutlein:
Apparently you can't run 2 stack builds at the same time which was causing the error. As a workaround one of the tests now depends on the other so that they will be built after each other.

Also fixed a small issue with the cleanup in the build-script running to late.


rTreutlein:
I am letting circlci build this a 2. time just to be sure.

PR #525, opened 2015-12-07
circle-ci configuration for running persitance tests

AmeBel:
Includes changes in https://github.com/opencog/atomspace/pull/520 and caching of build directory

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -13,7 +13,7 @@ dependencies:
 
 test:
     override:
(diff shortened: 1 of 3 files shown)


pennachin:
Shouldn't the build dir be /atomspace/build instead?

AmeBel:
It is inside the container, but on circle-ci's host container it isn't

pennachin:
Hmm, but CircleCI is complaining:
```
Warning: circle.yml specified cache directory: /home/ubuntu/atomspace/build but it does not exist
```

AmeBel:
Just found out tha caching is only for dependencies

PR #541, opened 2015-12-21
More updates for the fuzzy matcher

leungmanhin:
Try to make it a bit more simpler and flexible


linas:
Please keep the FuzzyMatch code in the FuzzyMatch directory.  It does not belong in the main directory, it is a very distinct set of code that does a distinct set of things, and is not part of the core pattern matching system.

leungmanhin:
ok I've just put them back to FuzzyMatch

linas:
Thanks, but I think you want to restore the README as well? It seems to have gotten lost.

leungmanhin:
That README seems to be a little bit outdated after refactoring the code, I actually wrote an updated version and put it in the class header file, I am thinking to also put it on a wiki page soon...

linas:
Oh. OK.

PR #542, opened 2015-12-21
Enable relative path resolution

misgeatgit:
This will enable PythonEval to properly resolve relative paths provided to PYTHON_PRELOAD_FUNCTIONS . Still fails if cogserver is not invoked directly from inside the build directory. May be we should define opencog specific environment variables to solve all issues related with path for once and for all.


linas:
CircleCI is preventing this from being merged; otherwise it is OK to merge.  Can someone PLEASE figure out what it takes to get CircleCI to work correctly for situations like this?   If we are going to use it, it should at least work right.

linas:
So: here's all that needs to be done to fix CircleCI for this situation: just get it to merge master before launching tests. We alreay know that there are no conflicts with master, so the merge will succeed without any warning or errors. Do the merge, run the tests, and report.  Don't block with the bogus/incorrect "branch out of date" error message.  The branch is not out of date, its just fine; CircleCI is confused.

PR #546, opened 2015-12-23
Update scheme primitive for fuzzy-matching

leungmanhin:


linas:
OK to merge this ... but I do say: it seems to me that you re making an API that is too complicated, if it needs so many different argument types.   The best API's have only one, maybe two arguments, usually all of the same type.  This seems like an awful random mixture ... can you think about your API some mre, and make sure there isn't some simpler way of doing things?

leungmanhin:
OK closing it, as we will not need it

PR #547, opened 2015-12-23
Robustify loading mechanism

misgeatgit:
- Added the capability for multiple directories to be specified
  for PYTHON_PRELOAD_FUNCTIONS separated by comma.
- Enforced loading order such that directories are loaded first
  so that no module import error occurs while trying to load file
  which imports from a directory that is not added to PYTHON_PATH.

Also avoids the need of appending to sys.path as in https://github.com/opencog/opencog/pull/1958


linas:
thank you!

PR #555, opened 2015-12-28
AtomUtils cleanup

linas:
This started as work to add more typing support, but ended up as a cleanup of the contents of AtomUtils.

Non-generic utilities were moved to the directories that contain the code that actually needs them.

Two poorly designed utilities were re-written for speed.


linas:
argh. This breaks main opencog. Working to fix this now.

PR #561, opened 2015-12-31
scheme bug fix and move of utility functions

AmeBel:
Move of functions, as per https://github.com/opencog/opencog/blob/master/opencog/nlp/relex2logic/rule-helpers.scm#L1

### DIFF
--- a/opencog/scm/utilities.scm
+++ b/opencog/scm/utilities.scm
@@ -46,6 +46,10 @@
 ; -- min-element-by-key -- Get maximum element in a list
 ; -- cog-push-atomspace -- Create a temporary atomspace.
 ; -- cog-pop-atomspace -- Delete a temporary atomspace.
+; -- check-name? -- Check if a there is a node with the given name.
+; -- random-string -- Generate a random string of given length.
+; -- random-node-name  -- Generate a random name for a node of given type.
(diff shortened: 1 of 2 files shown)


opencog/scm/utilities.scm
```diff
@@ -1019,8 +1023,63 @@
...
+
+; -----------------------------------------------------------------------
+
+(define-public (choose-var-name)
+"
+ Creates name for VariableNodes after checking whether the name is being
+ not used by other VariableNode.
+"
+    (random-node-name 'VariableNode 36 "$")
+)
```
linas: I'd like to revisit the variable-name thing, again.  I still have this gut sensation that variables are somehow being mis-used if there is a variable-naming collision.  Can you again, for the umpteenth time, send an email that describes the problem that unique variable names are supposed to solve?  I know we talked about it before, but maybe I have new insights.  I'll dregde through old emails, for now. Maybe we need a wiki on this topic.

AmeBel: https://github.com/opencog/atomspace/tree/master/opencog/rule-engine#things-need-to-be-implemented and https://groups.google.com/forum/#!topic/opencog/dKCYL47fpCQ.

PR #569, opened 2016-01-04
Make it explicit that AttentionValue objects are read-only

hedayat:
It seems to be valid now. i wonder if ImportanceIndex is expected to call decaySTI in future?!

### DIFF
--- a/opencog/truthvalue/AttentionValue.h
+++ b/opencog/truthvalue/AttentionValue.h
@@ -35,8 +35,10 @@ namespace opencog
  */
 
 //! stores attention in three components: short-term, long-term and very long-term
+//! This class is currently read-only, and assumed to be thread-safe. Beware of
+//! possible race conditions if it is changed
 class AttentionValue;
-typedef std::shared_ptr<AttentionValue> AttentionValuePtr;
+typedef std::shared_ptr<const AttentionValue> AttentionValuePtr;
 #define createAV std::make_shared<AttentionValue>
 
 class AttentionValue
@@ -78,9 +80,9 @@ private:
     //! Private, because changing the AV without emitting a signal
     //! will confuse everyone, so we sharply limit who is allowed
     //! to do this.
+    //! Note: this function is currently not used, not breaking read-only
+    //! assumption of this class
     void  decaySTI();
-
-    friend class ImportanceIndex; // The index can change the STI.
 
 public:
    /**
@@ -132,6 +134,6 @@ public:
 
 
 /** @}*/
-} // namespace opencog 
+} // namespace opencog
 
 #endif // _OPENCOG_ATTENTION_VALUE_H


pennachin:
@misgeatgit may know about decaySTI usage.

opencog/truthvalue/AttentionValue.h
```diff
@@ -35,8 +35,10 @@ namespace opencog
  */
 
 //! stores attention in three components: short-term, long-term and very long-term
+//! This class is currently read-only, and assumed to be thread-safe. Beware of
+//! possible race conditions if it is changed
 class AttentionValue;
-typedef std::shared_ptr<AttentionValue> AttentionValuePtr;
+typedef std::shared_ptr<const AttentionValue> AttentionValuePtr;
```
linas: well, but that's wrong, AV's are not meant to be read-only, they are supposed to be mutable, and regularly changing, as the ECAN "agent" runs.

opencog/truthvalue/AttentionValue.h
```diff
@@ -78,10 +80,10 @@ class AttentionValue
     //! Private, because changing the AV without emitting a signal
     //! will confuse everyone, so we sharply limit who is allowed
     //! to do this.
+    //! Note: this function is currently not used, not breaking read-only
+    //! assumption of this class
     void  decaySTI();
```
linas: yes, this function appears to be dead code.  It can probably be removed.

opencog/truthvalue/AttentionValue.h
hedayat: But currently, they are not changed. They are replaced instead. As you can see, AttentionValue's public interface is totally const. It has no set* methods. Notice that Agent has setAV/getAV methods to set/get AV values, but no way to change them. In the current state, I think it is nice to mark this class as read-only.

linas: No. Its old, crappy, code that is badly written.  We are not modifying the interface to be compatible and consistent with old crappy code.

pennachin: Linas, what Hedayat is working on is parallel execution of Agents and other background processes. The goal is to run ECAN and background PLN in parallel, as well as processing incoming requests such as ROS messages, chat, etc. Right now Misgana and Matt Iklé are tuning ECAN at the Atom level, and AFAIK they're not changing the immutable nature of the AV object. I don't know of any plans to change that in the near future, but perhaps you or @bgoertzel do?

If we're not changing that object, why not mark it as read-only?

linas: Ah, well, if the goal is to have atomic thread-safety, then this should be achieved in the atoms, and we should do the same for truth values, as well. I am pretty use that just marking the pointer itself as const probably breaks the scheme, python and presumably the haskell bindings too.  Let me take a look.

pennachin: Ah, understood, I didn't know this would break bindings, in which case this can surely wait and should be part of a larger effort...

linas: Curiously, it doesn't break bindings; apparently this was dealt with long ago.  I have created the pull req #570 which essentially does the same thing, and also updates assorted comments.

PR #575, opened 2016-01-06
Implement MapLink

linas:
As described in request #568 although the syntax is a bit different.

The syntax changed, so that the class hierarchy works consistently with what's in the C++ code base.

Issue #517, opened 2015-12-04
ParallelUTest fails most of the time.


mjsduncan:
ParallelUTest fails after `make test` on my current ubuntu 14.04 system:
````
The following tests FAILED:
	 19 - ParallelUTest (Failed)
Errors while running CTest
make[3]: *** [CMakeFiles/test] Error 8
make[2]: *** [CMakeFiles/test.dir/all] Error 2
make[1]: *** [CMakeFiles/test.dir/rule] Error 2
make: *** [test] Error 2
````
the funny thing is when i run it by itself, it passes the first time, and then fails when i try it again:
````
biocog@topdog:~/atomspace/build/tests/atoms$ ./ParallelUTest 
Running cxxtest tests (2 tests)[2015-12-04 17:17:59:442] [INFO] Loaded ../../opencog/atomspace/core_types.scm
[2015-12-04 17:17:59:498] [INFO] Loaded ../../../opencog/scm/utilities.scm
[2015-12-04 17:17:59:506] [INFO] Loaded ../../../opencog/scm/opencog/exec.scm
[2015-12-04 17:17:59:506] [DEBUG] BEGIN TEST: test_parallel
[2015-12-04 17:17:59:514] [INFO] Loaded ../../../tests/atoms/parallel.scm
0-count = 0

1-count = 1

2-count = 2

3-count = 3

[2015-12-04 17:18:05:516] [DEBUG] END TEST: test_parallel
.[2015-12-04 17:18:05:685] [INFO] Loaded ../../opencog/atomspace/core_types.scm
[2015-12-04 17:18:05:735] [INFO] Loaded ../../../opencog/scm/utilities.scm
[2015-12-04 17:18:05:743] [INFO] Loaded ../../../opencog/scm/opencog/exec.scm
[2015-12-04 17:18:05:743] [DEBUG] BEGIN TEST: test_join
[2015-12-04 17:18:05:749] [INFO] Loaded ../../../tests/atoms/parallel.scm
elapsed time = 5.001783 seconds
2-count = 2

[2015-12-04 17:18:10:752] [DEBUG] END TEST: test_join
.OK!
biocog@topdog:~/atomspace/build/tests/atoms$ ./ParallelUTest 
Running cxxtest tests (2 tests)[2015-12-04 17:18:26:021] [INFO] Loaded ../../opencog/atomspace/core_types.scm
[2015-12-04 17:18:26:070] [INFO] Loaded ../../../opencog/scm/utilities.scm
[...]

PR #576, opened 2016-01-06
Use robust sleep to address ParallelUTest flakiness

pennachin:
As noted on #517 and #575, the calls to `sleep()` weren't reliable, and the cause seems to be interrupts causing it to return sooner than expected. This implements a utility function that uses `clock_nanosleep()` to retry sleeping when interrupted. It also relaxes the range for `test_join()`, as I once saw it fail. `ParallelUTest` now reliably passes on my machine.

I didn't change `SleepLink` to use the same workaround, as I'm not sure what it should do when interrupted. It should be easy enough to move the utility sleep function elsewhere and use it in `SleepLink` if desired.

### DIFF
--- a/tests/atoms/ParallelUTest.cxxtest
+++ b/tests/atoms/ParallelUTest.cxxtest
@@ -39,6 +39,8 @@ private:
         AtomSpace *as;
         SchemeEval* eval;
 
+    int robustSleep(int secs);
+  
 public:
     ParallelUTest(void)
     {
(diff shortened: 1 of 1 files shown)


pennachin:
@ngeiswei, @AmeBel, @leungmanhin and any others who have seen flaky ParallelUTest runs, can you test this PR locally? I ran it a dozen times on my machine without any failures.

pennachin:
This has passed three consecutive CircleCI builds, so I think we're good, but will wait for another person to run it locally.

PR #578, opened 2016-01-07
Ure config fixes

ngeiswei:
Use DefinedSchemaNode instead of Node.

WARNING: I'm not sure, this might break the opencog repo, I don't understand why.

@linas If you want to restrict DefineLink to use only DefinedSchemaNode, DefinedPredicateNode or DefinedTypeNode I think you now can.

### DIFF
--- a/examples/rule-engine/chaining/frog.scm
+++ b/examples/rule-engine/chaining/frog.scm
@@ -23,7 +23,7 @@
 	)
 )
 
-(define rule1-name (Node "rule1"))
+(define rule1-name (DefinedSchemaNode "rule1"))
 (DefineLink rule1-name rule1)
 
 (define rule2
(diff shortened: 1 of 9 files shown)


linas:
Thanks!

ngeiswei:
Wait, I've got a twin PR on the opencog branch on the fire, things might break there till then

ngeiswei:
Yeah, it breaks ActionUTest on the opencog repo. I've got to understand why...

linas:
I already fixes ActionUTest. It was a bad use of DefineLink i.e. using Node instead of DefinedSchemaNode

linas:
p.s. I will open pull reqs to master, but am no longer merging them. So you and anyone else using the master branch are on your own.

ngeiswei:
Fair enough.

PR #579, opened 2016-01-07
Add type checking for DefineLink

linas:


linas:
This mergeds the wrong branch, use pull req #582 instead

PR #580, opened 2016-01-07
Add examples for the new MapLink

linas:


linas:
this should be merged only after #582 is merged

PR #582, opened 2016-01-07
Add type checking for DefineLink

linas:
This should be pull req #579, except that pull req used the wrong branch

### DIFF
--- a/opencog/atoms/core/DefineLink.cc
+++ b/opencog/atoms/core/DefineLink.cc
@@ -37,18 +37,16 @@ void DefineLink::init()
 	// Perform some additional checks in the UniqueLink init method
 	UniqueLink::init(false);
 
-#if ALREADY_BROKEN
-	// Type-check. Probably not needed, probably too strict, but what
-	// the heck, since we are here... Heh. There's already code that
-	// defines shit more loosely than this. Bummer.
+	// Type-check. The execution and FunctionLink's only expand
+	// definitions anchored with these type; other definitions won't
+	// work during execution.
 	Type dtype = _outgoing[0]->getType();
 	if (DEFINED_SCHEMA_NODE != dtype and
 	    DEFINED_PREDICATE_NODE != dtype and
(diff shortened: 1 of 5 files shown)


ngeiswei:
This PR introduces a bug in the Haskell binding or utest. I'm on it (although since I've never touched the Haskell utest code that might take a while).

rTreutlein:
@ngeiswei I know what causes the issue. But i think the best way to actually fix this would be to just throw out the Haskell Atom Types, if we want to do that anyway. Because then this test becomes obsolte.

ngeiswei:
@rTreutlein thank you! Although I don't know what that test is doing, I think throwing it away seems a bit extreme, a utest even testing something trivial is still valuable. I'm looking into it...

rTreutlein:
@ngeiswei The test makes sure that the conversion from the Raw AtomType to the Haskell Atom Type is correctly impleneted since that has to be done manually. Of course if there is Haskell Atom Type anymore then the test does nothing.

ngeiswei:
@rTreutlein alright, I trust you on that. So I guess we just disable the test for now (do you want to do it?)

Of course I strongly recommend that once you've done your changes on the Haskell binding, you write a new test, even a trivial one for it. Thanks.

ngeiswei:
@rTreutlein I'm pushing a PR disabling that test.

ngeiswei:
It's strength, github doesn't let me update that branch, so I can't merge it either.

AmeBel:
@ngeiswei here are the details on protected branches https://help.github.com/articles/defining-the-mergeability-of-pull-requests/

ngeiswei:
@AmeBel I don't think the fact that I cannot update @linas ' branch has to do with protected branches. But it obviously makes sense that github wouldn't allow me to update **his** branch (I don't know what I was thinking).

So we're blocked, either @linas update his branches, or I clone them all and creates new PRs.

AmeBel:
oh I misunderstood

Issue #587, opened 2016-01-09
I'm getting "double free or corruption" errors on multiple UTests after make install


pennachin:
Running the UTests from latest master I get the following failures (snipped `make test` output):

```
      Start 59: ArcanaUTest
*** Error in `/atomspace/build/tests/query/ArcanaUTest': double free or corruption (!prev): 0x00000000028b74a0 ***
59/94 Test #59: ArcanaUTest ......................***Exception: Other  4.22 sec
      Start 61: GetLinkUTest
*** Error in `/atomspace/build/tests/query/GetLinkUTest': double free or corruption (fasttop): 0x0000000002259840 ***
61/94 Test #61: GetLinkUTest .....................***Exception: Other  0.16 sec
      Start 66: EvaluationUTest
*** Error in `/atomspace/build/tests/query/EvaluationUTest': double free or corruption (!prev): 0x0000000000db63d0 ***
66/94 Test #66: EvaluationUTest ..................***Exception: Other  0.22 sec
      Start 79: SudokuUTest
*** Error in `/atomspace/build/tests/query/SudokuUTest': double free or corruption (!prev): 0x0000000001fdc8d0 ***
79/94 Test #79: SudokuUTest ......................***Exception: Other  0.17 sec
      Start 83: ForwardChainerUTest
*** Error in `/atomspace/build/tests/rule-engine/ForwardChainerUTest': double free or corruption (fasttop): 0x000000000269c720 ***
83/94 Test #83: ForwardChainerUTest ..............***Exception: Other  0.52 sec
      Start 86: UREConfigReaderUTest
*** Error in `/atomspace/build/tests/rule-engine/UREConfigReaderUTest': double free or corruption (fasttop): 0x0000000002772170 ***
```

I did a debug build and ran valgrind on ArcanaUTest:

```
valgrind --tool=memcheck --track-origins=yes --keep-stacktraces=alloc-and-free ./tests/query/ArcanaUTest
```

Here's the relevant part of the output:

```
==16314== Invalid free() / delete / delete[] / realloc()
==16314==    at 0x4C2C2BC: operator delete(void*) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==16314==    by 0x52D26B3: __gnu_cxx::new_allocator<opencog::FunctionWrap*>::deallocate(opencog::FunctionWrap**, unsigned long) (new_allocator.h:110)
[...]

PR #589, opened 2016-01-10
Another fix for bug #587

linas:
Another fix for bug #587


pennachin:
Thanks, this one did it.

PR #596, opened 2016-01-21
Fixed duplicate Python Atom identifier

inflector:
Special-cased the ATOM type so that it doesn’t create an Atom function
in core_types.pyx that conflicts with the Atom Python object.


linas:
thanks inflector!

PR #597, opened 2016-01-22
PM readme misc

ngeiswei:
1. Update query/README and query/README-Algorithm to reflect the current state of the code.
2. Minor formatting changes to be more markdown compatible.
3. Minor typos fixes.


ngeiswei:
@linas I don't think I've screwed anything but you may want to have a look at it. If you don't have time, no problem, just merge.

ngeiswei:
Never mind, I'm merging.

opencog/query/README.md
```diff
@@ -267,21 +268,17 @@ ignored.  This is because constant clauses are "trivial": the
...
 
-The set of clauses defining a pattern must be connected by means of
-common, shared variables.  A pair of clauses are connected if they
-share a common variable. A set of clauses is connected if there is a
-transitive path through all of them.  The need for this requirement
-was mentioned before: disconnected graphs are always just a trivial
-product of their connected components; thus, it is pointless to
-explicitly support disconnected graphs.
+The clauses are partitioned in to connected components, clauses that
+share directly or indirectly the same variables.
```
linas: The old text had a definition of connectivity and transitivity. I think that it is useful for the readers to develop a clear idea that connectivity is transitive (i.e. its the transitive closure of the graph)  Mostly, I just want to use these mathematical terms, since school textbooks already define what the transitive closure of a term algebra is, and I want the reader to just go, "oh yeah, I remember that".  (or look up "transitive closure" on wikipedia, or whatever.)

opencog/query/README.md
```diff
@@ -356,23 +353,23 @@ Currently Implemented Algorithm
    Thus, for example: the clause _subj(row, Steve) has _subj at the
    root of the tree, with "row" and "Steve" as leaves.
 
-   Distinct trees may have common vertices.  The common vertices
-   act to join together the trees into a single, connected graph.
-   Disconnected graphs are not allowed; the algorithm will fail to
-   find and ground disconnected components.  The graph as a whole may
-   contain loops (the decomposition into trees keeps the algorithm
-   from having to explicitly accommodate loops).
```
linas: Well, you removed the bit about loops. Its important, since an naive implementation results in an infinite loop, if the graph has loops.  The pattern matcher detects and breaks such infinite loops.  There's some unit test for this.

PR #598, opened 2016-01-25
Apply Linas remarks in PR #597

ngeiswei:


ngeiswei:
It looks like Haskell binding is randomly (though rarely) crashing. I recall someone has already observed that and hypothesized that it was due to too much memory consumption. I guess circle-CI doesn't a swap and the memory limit is a hard one.

linas:
thanks!

PR #599, opened 2016-01-25
Initial commits for atomspace C++ API tutorial.

misgeatgit:


misgeatgit:
Seems my code compiled with an old installation of the atomsapce. AttentiontionValue.h has been deleted from the codebase and in effect build is failing here. I will fix this.

wenwei-dev:
The commit from HR Bot is a bot managed by Jenkins running Hanson
Robotics's office server. I will replace it with a OpenCog bot later when
the CI integration is done.

On Monday, 25 January 2016, Misgana Bayetta <notifications@github.com>
wrote:

> Seems my code compiled with an old installation of the atomsapce.
> AttentiontionValue.h has been deleted from the codebase and in effect build
> is failing here. I will fix this.
>
> —
> <https://github.com/opencog/atomspace/pull/599#issuecomment-174473978>.
>


-- 
Sent from Gmail Mobile

misgeatgit:
@wenwei-dev could you check the build failure error message? seems to be related with docker image naming.

wenwei-dev:
retest this please

misgeatgit:
@wenwei-dev all tests are passing on circle-ci.

wenwei-dev:
I see. My last commit is a phrase that triggers jenkins to rerun the test. So please ignore it. Sorry for that.

examples/c++-api/AtomSpaceEventSubscribeExample.cc
```diff
@@ -0,0 +1,125 @@
...
+    AVChangedSignalConn = as.AVChangedSignal(
+            boost::bind(&AVChangedCBHandler, _1, _2, _3));
+    TVChangedSignalConn = as.TVChangedSignal(
+            boost::bind(&TVChangedCBHandler, _1, _2, _3));
+    AtomsRemovedSignalConn = as.removeAtomSignal(
+            boost::bind(&AtomRemovedCBHandler, _1));
+    AtomAddedToAttentionalFocusSignalConn = as.AddAFSignal(
+            boost::bind(&AtomAddedToAFCBHandler, _1, _2, _3));
+    AtomRemovedFromAttentionalFocusSignalConn = as.RemoveAFSignal(
+            boost::bind(&AtomRemovedFromAFCBHandler, _1, _2, _3));
```
linas: I am very very concerned about the addition of this example, for many reasons.  It encourages code debt!  Code debt is bad!

https://en.wikipedia.org/wiki/Technical_debt

1)  Most programmers should not be using the C++ at all.
2) The performance implications of the atomspace event signals are disasterous!  These signals should not be used, if at all possible.
3) We should not be writing examples for an API whose use we want to discourage!  
4) It is very easy to write code for an API, it is much much harder to remove code for an API. If someone starts to use this API, it will be quite hard to remove: someone ill have to do the work to redesign the code!!  Who is going to do that work?   It is much better to simpl not create the code to begin with, than too start by incurring a huge liability.

Basically, dont go into code debt on day one.  Try to avoid code debt.  These examples will encourage debt! 

 https://en.wikipedia.org/wiki/Technical_debt

misgeatgit: I agree on most of the issues you have raised. However, these signals seem to be useful in some cases; For instance, the Hebbian creation part of ECAN mainly relies on atom added signal. I wonder if there is some other better way to implement signals in the atomspace rather than discouraging their use. Probably by generalizing the approach take in the [HebbianCreation code](https://github.com/opencog/opencog/blob/master/opencog/attention/HebbianCreationModule.cc#L77) which uses separate thread for processing the signal.

bgoertzel:
Hmm ... it seems there are a couple issues combined here..

A) should there be a tutorial example for interacting with the Atomspace using C++ .   I think there should be, which is why I asked Misgana to write one....   I agree that most OpenCog development should not use C++ directly.  But some of it will need to, and for that reason, it is useful to have a basic tutorial on how it's done, in my opinion...

B) is the Atomspace signals API bad and worthy of deprecation.    That I don't have an opinion about at the moment...

Linas, can you clarify why you think it's so slow?   Is there some test code somewhere measuring its speed?  Is it always objectionably slow or does that depend on how it's used (i.e. is it actually too slow in the ECAN context where Misgana is using it?) ...

Also, I don't know if using a separate thread would help whatever slowness is there; maybe the slowness would be in the main Atomspace thread ... I don't remember how the signaling works in detail, but I could take a look on Friday when I'm at POly U...

linas:
The performance profile is forced by the design. To understand it, consider this scenario: some subsystem is adding and removing atoms at a fairly rapid rate: let say, some visual subsystem dealing with movement.  The atoms have to be added to the atomspace for at least a little while, because some decision-making has to be done (for example, creepers are attacking the minecraft bot; a fight-or-flight decision has to be made)  For ever atom added, the signal gets invoked, and the HebbianCreation agent runs. 

Does ECAN really need to care about *every* atom?  Or just some of them? How many of them? 1 in 2? 1 in 50? 1 in 1000?   If it only cares about 1 in 1000, why are we slowing down 999 atom adds just to find that one?

Is it actually important to be notified absolutely immediately, synchronously (i.e. within 5-10 machine cycles) of the atom addition, or is it OK to batch up the notifications into sets of a few thousand atoms, and report these only every 10 or 20 milliseconds?  The signals provide a very low-latency notification mechanism, literally, just the overhead of  subroutine call, which is ideal if you need sub-microsecond reaction times. I don't see that ECAN needs sub-microsecond accuracy.

For now, ECAN can continue to use the signal API well into the foreseeable future; just be aware that it is designed to provide an extremely low latency synchronous API to every atom addition.  If you don't need these characteristics, then there  are other ways to find out about atom additions.

I view atom addition and atom-removal as a primary performance bottleneck; the signals API has a very immediate and direct negative influence on it.  The API is needed, but we should be careful not to use it for frivolous reasons.

misgeatgit:
@linas  ECAN only cares about atoms added to the attenational focus and the signal it listens to is added_to_af; not atom_added. Sorry for the confusion I've created in my previous comment.

Issue #201, opened 2015-08-13
Python Atom <--> Handle confusion


linas:
Some python functions take Handles, some take Atoms, its hard to tell which.   Its is easy to get a Handle from an atom, but hard/confusing to get an Atom from a Handle:
```
h = Handle
a = Atom( h, atomspace)
```

There is no reason to have both an Atom class, and a handle class in the python bindings. The two concepts are identical; one should be removed.

AmeBel:
So which should be the standard?

linas:
Atom, I think.   I see no purpose in having Handle.  It doesn't do anything useful.

inflector:
Using only Atoms will make it easier to understand so I concur with the simplification goal. Two caveats:

1) This is a fair bit of work if I recall, and may not be possible with Cython due to restrictions on the interface between Python and Cython types during Python object creation. There is a reason it was done this way that I understood at one time but which now eludes me. I'm not sure the current implementation was the best way, or the only way, just that it was more rational than it would appear. I believe it had to do with the way that Atoms are created from UUID lists internally.

Best way to find out is to do it.

2) This proposed change will likely break a fair bit of existing Python code, so it's the kind of change that ought to be done with a bit of warning so someone doesn't just pull the latest and have their world break without notice. Perhaps we ought to put a notice on the Google group for Python users and poll existing projects to see how much this will affect their code to give a heads up.

AndrewBuck:
Why isn't there a get_atom_by_handle() function in the atomspace class?  Honestly, any of the functions which take an atom as an argument (or a handle) should just have an overloaded method which takes the alternate version and does the conversion for you, but having a get_atom_by_handle() method makes it possible to do the conversion yourself manually if you want to.

This is something I have had trouble with in working on the minecraft bot where sometimes you do a function call and get the atoms themselves back and sometimes you get handles.  But in theory one should be as good as the other to any of the atomspace routines.

AndrewBuck:
Oh, I should also add that there is (at least in theory) a reason for allowing both atoms and atom handles.  If you are storing or processing large amounts of atom data outside of the atomspace (for example if the bot reads out a bunch of block atoms to look at them all or do some operation to all of them, etc) then you may end up making copies or references to a large number of atoms.  Being able to store or work with atom handles and then only access the atoms themselves for certain atoms would allow memory and possibly computation savings, depending on what you are doing with all those atoms.

linas:
[...]

PR #604, opened 2016-01-27
Remove cython handle

inflector:
Updated the Cython bindings and tests to no longer use the Python Handle object. Now there is only the Atom object which consolidates the former Handle and Atom object into one simpler interface.


inflector:
This Pull Request addresses Issue https://github.com/opencog/atomspace/issues/201.

linas:
This looks very reasonable; I think it can be merged now. The remaining work items are then:
-- a pull request for https://github.com/opencog/opencog so that its in sync with these changes
-- updated to http://wiki.opencog.org/w/Python to reflect these changes (e.g. the suff you wrote in the accompanying email)

linas:
oh, I see you updated the wiki page already.

linas:
need to merge master again...

inflector:
I didn't have to change anything in opencog/opencog as all the changes were confined to atomspace and related tests.

I did, however, build and run the tests in opencog/opencog to make certain I didn't break anything.

inflector:
The wiki changes are in process but mostly not related to the above as the tutorial examples didn't use handles.

inflector:
The last commit fixed a problem I found while running through the wiki tutorials. It also prepares the way for removing many of the deprecated atomspace methods that are only used by Cython. In particular, there are a lot of Cython Atom calls which deferred to AtomSpace calls instead of calling into the C++ Atom object. The example in the above checkin was a removal of AtomSpace::atom_as_string which was replaced by calls to Atom::toString and Atom::toLongString.

I will make further changes on another branch so I don't intrude on this Pull Request.

inflector:
Turns out I was wrong about opencog/opencog. I know I built it and ran all the tests, but after the last pull I am getting issues. A dependency problem in the make files?

I'll post here when I've resolved the issues with spacetime.pyx

inflector:
That last merge supports some changes required to support the new Pull Request https://github.com/opencog/opencog/pull/2007

There is still a failure with the anaphora test and there will probably be some changes required in anaphora or the anaphora test. However, it was already failing and I don't understand it so I don't want to try to update it, if needed, until it passes for whatever reason it was failing before my changes.

AmeBel:
the anaphora test depends on scheme files autogenerated for atom additions. The symlinking of those files doesn't always work( see [here](https://github.com/opencog/opencog/blob/master/lib/OpenCogFunctions.cmake#L108)). So easy fix is just remove all content from your build directory and rebuild.  I guess the long term fix is to get api parity with scheme, so as to not have to use scheme from python.

inflector:
@AmeBel Your comment implies that the anaphora test is passing on your system with the latest from master. Is that a correct assumption?

AmeBel:
yes, if you look at the test log at `build/tests/Testing/Temporary/LastTest.log` you will likely see `n procedure primitive-load-path: Unable to find file "opencog/*/*.scm"`

inflector:
@AmeBel I figured out a couple of problems:

1. The call to `SET_TESTS_PROPERTIES` was setting `ENVIRONMENT` twice which meant that the test would fail if you didn't have `PYTHONPATH` set because while it appeared to set `PYTHONPATH` the second call would overwrite the first, so it was only setting GUILE_LOAD_PATH. The fix was to use a ';' delimited string and set both `GUILE_LOAD_PATH` and `PYTHONPATH` in one string.

2. The setting for the test for `GUILE_LOAD_PATH` only worked if you ran `make install` before running `make test`. This is the only test that relied on that ordering. I generally run tests before installing since I don't want to bother installing until I know it works. I updated `GUILE_LOAD_PATH` so that it would find the files in their pre-install path by adding `${PROJECT_BINARY_DIR}` and now it will find `nlp_types.scm` even if you don't run `make install` before running the tests.

These changes are in the pull request https://github.com/opencog/opencog/pull/2007

inflector:
To confirm, the anaphora tests pass on my setup after these changes.

linas:
branch is out of date again...

inflector:
Updated again. Hopefully it will stay stable long enough to get merged :-)

PR #608, opened 2016-01-29
Remove deprecated atomspace functions

inflector:
Removed quite a few deprecated AtomSpace functions and moved them to Atom.
Updated the tests to match.

Will be removing atomspace.get_atoms_by_target_atom after this pull request is merged and I can update opencog/opencog to use the new atom.incoming_by_type in the few places where it is currently using.

### DIFF
--- a/opencog/atoms/base/Atom.h
+++ b/opencog/atoms/base/Atom.h
@@ -237,7 +237,7 @@ public:
         setAttentionValue(new_av);
     }
 
-    /** Change the Long-term Importance */
+    /** Change the Long-Term Importance */
     void setLTI(AttentionValue::lti_t ltiValue)
     {
         AttentionValuePtr old_av = getAttentionValue();
@@ -245,6 +245,17 @@ public:
             old_av->getSTI(),
             ltiValue,
             old_av->getVLTI());
+        setAttentionValue(new_av);
+    }
+
+    /** Change the Very-Long-Term Importance */
+    void setVLTI(AttentionValue::vlti_t vltiValue)
+    {
+        AttentionValuePtr old_av = getAttentionValue();
+        AttentionValuePtr new_av = createAV(
+            old_av->getSTI(),
+            old_av->getLTI(),
+            vltiValue);
         setAttentionValue(new_av);
     }
 
--- a/tests/cython/atomspace/test_atomspace.py
+++ b/tests/cython/atomspace/test_atomspace.py
@@ -430,25 +430,25 @@ class AtomTest(TestCase):
 
         l = Link(a1, a2)
 
-        last_uuid = 95
+        space_uuid = self.space.uuid
 
         # test string representation
-        a1_expected = "(Node \"test1\") ; [{0}][{1}]\n".format(str(a1.value()), last_uuid)
+        a1_expected = "(Node \"test1\") ; [{0}][{1}]\n".format(str(a1.value()), space_uuid)
         a1_expected_long = \
             "(Node \"test1\" (stv 0.500000 0.800000)) ; [{0}][{1}]\n"\
-            .format(str(a1.value()), last_uuid)
+            .format(str(a1.value()), space_uuid)
 
-        a2_expected = "(Node \"test2\") ; [{0}][{1}]\n".format(str(a2.value()), last_uuid)
+        a2_expected = "(Node \"test2\") ; [{0}][{1}]\n".format(str(a2.value()), space_uuid)
         a2_expected_long = \
             "(Node \"test2\" (av 10 1 1) (stv 0.100000 0.300000)) ; [{0}][{1}]\n"\
-            .format(str(a2.value()), last_uuid)
+            .format(str(a2.value()), space_uuid)
 
         l_expected = \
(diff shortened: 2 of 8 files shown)


opencog/atoms/base/Atom.h
```diff
@@ -248,6 +248,17 @@ class Atom
         setAttentionValue(new_av);
     }
 
+    /** Change the Very-Long-Term Importance */
+    void setVLTI(AttentionValue::vlti_t vltiValue)
+    {
```
linas: VLTI was designed to not be settable, only incrementable/decrenmentable.  I don't recall why.

tests/cython/atomspace/test_atomspace.py
```diff
@@ -358,31 +422,33 @@ def test_type(self):
...
 
-        a2 = self.space.add_node(types.Node, "test2")
+        a2 = Node("test2")
         a2.av = {"sti": 10, "lti": 1, "vlti": True}
         a2.tv = TruthValue(0.1, 0.3)
 
-        l = self.space.add_link(types.Link, [a1, a2])
+        l = Link(a1, a2)
+
+        last_uuid = 95
```
linas: this is the uuid of the atomspace, so I believe that you want `last_uuid = atomspace.get_uuid()` here.

linas:
looks good to me.  A long time ago, there was some convoluted discussion about why VLTI must not be settable, but I no longer recall the reason. It was possibly a silly reason.

opencog/atoms/base/Atom.h
inflector: My read of the documentation said that `vlti` is a flag that means don't delete for any positive values. So it acts like a 'permanent atom' flag. However, I didn't check the EVAN code to verify that is what it does.

There is an increment and decrement that I implemented as well, and I'd expect that to be the normal interface to `vlti` for setting the flag.

The actual behavior of the existing Python interfaces was inconsistent and hard to describe. So in the spirit of simplifying and making it easier for new people to understand I changed some inconsistencies.

A particularly confusing example I encountered when testing is setting with a dictionary like:
```
>>>atom.av = {'st': 10, 'lti' 5, 'vlti': 5}
>>>atom.av
{'sti': 10, 'lti': 5, 'vlti': True}
```
Internally, the code for setting from a dictionary called increment if the value for `vlti` was `> 0` it would do an increment and decrement for `< 0`. Then the code for retrieval was treating the value as if it was a boolean and ignoring all other values. So there was no way to find out what the actual value was. And you'd get wonderful behavior like:
```
>>>atom.av = {'st': 10, 'lti' 5, 'vlti': True}
>>>atom.av = {'st': 10, 'lti' 5, 'vlti': True}
>>>atom.av = {'st': 10, 'lti' 5, 'vlti': False}
>>>atom.av
{'sti': 10, 'lti': 5, 'vlti': True}
```
which is really confusing and looks like a bug.

It seems like you'd want to be able to restore prior states which is why I changed the new implementation to make it easier to understand what is going on and make it possible to restore states.  So now you can do:
```
>>> atom.av
{'sti': 10, 'lti': 5, 'vlti': 2}
>>>original_av = atom.av
>>>
>>>... do some work ...
>>>
>>>atom.av = original_av
>>>atom.av
{'sti': 10, 'lti': 5, 'vlti': 2}
```
and
```
>>>atom.av = {'st': 10, 'lti' 5, 'vlti': 1}
>>>atom.av = {'st': 10, 'lti' 5, 'vlti': 1}
>>>atom.av = {'st': 10, 'lti' 5, 'vlti': 0}
>>>atom.av
{'sti': 10, 'lti': 5, 'vlti': 0}
```
Also, when debugging the code if you want to know why an atom is not being deleted, then it would help to know that the `vlti` value is actually `3` rather than just `True`.

So if we want to only support increment and decrement then we ought to disallow the setting of `av` with dictionaries, I think, as this usage implies a set of the value not an increment or decrement, and the previous behavior is non-intuitive.

tests/cython/atomspace/test_atomspace.py
inflector: Thanks, I just checked in some code that changes that so now it won't fail every time you add another atom or two to the test case. Yay!

linas:
Attention values are currently almost completely unused. The upcoming work on ECAN is supposed to change that.  Certainly, the SQL backend ignores VLTI at this time, and its not clear that this isn't a bad choice, at this time.  

Anyway, there is considerable confusion about things: one the one hand, high AV values are supposed to guide PLN reasoning.  On the other hand, AV is supposed to be used for memory management, so that low AV gets deleted.  Since these two are inconsistent with each other, VLTI steps in to resolve the difficulty.  Almost all code ignores AV. Its quite the mess.

inflector:
As mentioned above, once this is merged, I'll fix opencog/opencog to remove some deprecated use of atomspace.get_atoms_by_target_atom and then come back here to opencog/atomspace and remove it so it isn't used again.

inflector:
@linas Ready for merge? Or do you want me to back out the changes permitting set of VLTI?

linas:
thanks

PR #610, opened 2016-01-30
Removed deprecated get_type, get_name, get_incoming, get_outgoing from AtomSpace

inflector:


inflector:
After this is merged, PR https://github.com/opencog/opencog/pull/2011 needs to be merged and the CircleCI and Jenkins builds restarted. I don't seem to have access to the button to restart CircleCI.

opencog/atoms/base/Atom.h
```diff
@@ -173,6 +173,11 @@ class Atom
 
     virtual ~Atom();
 
+    virtual const std::string& getName() const {
+        static std::string empty_string("");
+        return empty_string;
```
linas: It would be better to throw here; that way, usage errors are immediately detected.

opencog/atoms/base/Atom.h
```diff
@@ -190,6 +195,16 @@ class Atom
...
 
+    /** Returns the outgoing set.
+     *
+     * @return The outgoing set.
+     */
+    virtual inline const HandleSeq& getOutgoingSet() const
+    {
+        static HandleSeq no_atoms;
+        return no_atoms;
+    }
```
linas: Same as above: should be a throw.

linas: Oh, I see .. the old code that this replaced did not throw.  However .. we probably should. Do the unit tests pass if we throw?

opencog/benchmark/AtomSpaceBenchmark.cc
```diff
@@ -1178,7 +1178,7 @@ timepair_t AtomSpaceBenchmark::bm_getOutgoingSet()
 #if HAVE_CYTHONX
```
linas: should probably change this back to HAVE_CYTHON from HAVE_CYTHONX .. the X was added because the code was wonky and had to be disabled.

opencog/atoms/base/Atom.h
inflector: I can add that, but the C++ code was previously doing instance checking. i.e. NodeCast and then checking for null returns.

opencog/benchmark/AtomSpaceBenchmark.cc
inflector: The reason this is HAVE_CYTHONX is because it doesn't support testing of loops for Python. The same comment is around all the python testing code for the loop test, for some reason.

linas: OK

opencog/atoms/base/Atom.h
linas: yes, I noticed that after adding the comment.  However, silently failing when there is a coding error seems like poor design, which is why I suggest having a throw, instead.   There is also a very minor performance boost, because `try..catch` is faster than `if (empty==...)` for those cases where the programmer wants to explicitly check for the error condition.

inflector: We could always just return a default name, i.e. "unnamed Atom" or something like that.

I generally try to avoid having to do casts and instance checks as they're slower than the virtual function overhead and they clutter the code with dependencies which make it harder to make architectural changes to the class hierarchy, though in this case, that's not likely given the difference between Nodes and Links.

The issue here is that the Python Atom doesn't know about the class hierarchy of Nodes and Links. So I could special-case Python to do an instance check or catch the exception, however, that will slow down the normal case quite a bit and we'd still have to either: force the Python caller to deal with exceptions or instance check.

For something like getting the name, that seems like overkill. Names are generally used for either printing or lookup, so an empty name won't go unnoticed in new code. Existing code already does an instance check, I didn't take any out. And since the call was not virtual before, no one could have been using it on non-Node Atoms.

inflector: My comment crossed yours...

In C++ `try..catch` is very fast. Python has much higher overhead.

opencog/atoms/base/Atom.h
inflector: The unit tests would pass since the code could never have been executed before because the call wasn't available to Atom only through NodeCast of handles which were checked for null.

If we do a throw here, I'll have to catch them in Python and then do something else anyway since name is a **property** of Atom and needs to be accessible for all atoms. Forcing people to do is instance checks before using a property of an object would be very odd in Python.

inflector: So the way I looked at it was that calling getName() on a non-Node Atom isn't a programming error, it is new well-defined behavior. Atoms without names will return the empty string for getName(). Those that have names will return the name. 

If you need to know if something has a name or not and don't, perhaps because you are iterating over a bunch of handles in C++ or Atoms in Python, you can use an instance check on the class.

linas: All I am saying is that it *is* a programming error to ask for the name of an atom that is not actually a node.  By throwing an exception, the programmer who makes that mistake is alerted of their mistake immediately. It's got nothing to do with python -- its an error in any language binding ...C++ or guile or haskell or python.

As to performance, I was under the impression that the C++ cast to a NodePtr was "free" and was done at compile time. By contrast, a virtual method call costs run-time CPU cycles.  Are you saying I mis-understood the NodeCast?  I'm not sure if I ever looked at the assembly that gets generated, so maybe I'm wrong on this.  Any, that would be a good reason to NOT use virtual calls.

linas: Wow. I just did a disassembly of the code for this:
```
char * foo(const Handle& h) {
   NodePtr p(NodeCast(h));
   if (p) return "asdf";
   return nullptr;
}
```
and it is insanely, jaw-droppingly horrific! I am just flabbergasted!  I could have sworn that last time I did this, it was nice and clean and elegant, so I'm a bit flummoxed at the moment.

inflector: The C++ cast is free, it's the checking of the result for NULL before calling a function on that result that is not free. At any rate, a branch on zero is not expensive. It's more ugly and verbose than slow.

As to whether or not it's an error, that seems like something that can be decided rather than a given. I can certainly see myself writing code that will print out the name of all the atoms during debugging etc. And I'd not find it confusing that names didn't exist for non-nodes.

Nevertheless, I'm in no position to argue with you about this as you've got much more experience with the domain. Throwing an exception in C++ is fine if that's what you'd prefer. I'll do it.

However, this may break Python code. The code for the name property wasn't throwing exceptions before since atomspace.get_name() didn't throw, so there was no need to do a type check on the handle before calling atomspace.get_name(). If there are any problems, I'll fix the tests.

linas: All this is making my brain hurt. Once upon a time, there were these violent email discussions about how links could never have a name. (Personally, I thought links should have a name, but was kicked and poked and stabbed until it hurt) So that is what got implemented, and that is a deep core assumption that is very deeply embedded in the design of absolutely everything.  If we suddenly allowed links to have names, we'd have to rethink the design of EvaluationLinks, which are central and used everywhere. Go back and redesign the forward/backward chainers, the pattern matcher.  Redisgn the way we do the linguistics processing. Redesign the SQL database internals.  Vast tracts of code would have to change.  So that change is imponderable at this time.

The disasterous disassembly for NodeCast suggests a major design flaw, and that, indeed, we should keep the virtual methods on Atom and try to avoid casts everywhere. Its shocking and I don't understand it just right now.

PR #615, opened 2016-02-03
Fc minor improvements

ngeiswei:
1. Simplify the code
2. Disable the derived rule cache which is buggy

@misgeatgit my next PR will probably consist of bigger refactoring, so if you have some changes on the fire you probably want to commit them very soon.

### DIFF
--- a/opencog/query/PatternMatch.cc
+++ b/opencog/query/PatternMatch.cc
@@ -300,7 +300,7 @@ bool PatternMatch::recursive_virtual(PatternMatchCallback& cb,
  * predicate can be satisfied. Thus, for example, given the structure
  *
  *    BindLink
- *       ListLink
+ *       VariableList
  *          VariableNode "$var0"
  *          VariableNode "$var1"
  *       AndList
--- a/opencog/query/Recognizer.cc
+++ b/opencog/query/Recognizer.cc
@@ -37,7 +37,7 @@ namespace opencog {
  * Very rough, crude, experimental prototype for the idea that
  * pattern recognition is the dual of pattern matching.
  * That is, rather than using one query pattern to search a large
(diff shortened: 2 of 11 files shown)


ngeiswei:
Jenkins, could you retest that?

ngeiswei:
retest please

ngeiswei:
retest this please

ngeiswei:
@wenwei-dev could you add this trigger message as well "Jenkins, could you retest that?". Also, where is that documented?

misgeatgit:
@ngeiswei  I don't have any patches to send. What was the bug you found in d1cefbad2a0d51cb5?

ngeiswei:
@misgeatgit the bug isn't in here actually it is in its use

ForwardChainer. l.130

```c++
        if (_fcstat.has_partial_grounding(_cur_source))
```

should probably replaced by something else (maybe by is_matched as defined in l.217). Cause the source may have been used with deriving other rule but not this one.

But I don't want to fix it because I'm not sure if it's right. First it could be replaced by lru_cache or such, or maybe it shouldn't be here in the first place, I need to study more to see the big picture.

ngeiswei:
test this please

ngeiswei:
retest this please

AmeBel:
@pennachin`PersistUTest` test failure, after parallel test run is enabled.

pennachin:
@AmeBel I guess we should disable parallel unit tests until we debug this... The tests all run fine on my local (Docker-based) setup...

ngeiswei:
retest this please

AmeBel:
@ngeiswei give it some time for the docker images to build

ngeiswei:
retest this please

AmeBel:
retest this please

AmeBel:
retest this please

Issue #513, opened 2015-12-02
Finish implementing ProtoAtom


linas:
The current ProtoAtom code is experimental and can be found in the ProtoAtom branch. It "works".  However its unfinshed, some serious design issues remain and some important decisions with large reprecussions have to be made.  These are difficult decisions which is why they have not been made.

A ProtoAtom is like an ordinary Atom, but without a TV, AV or UUID. It is meant to store mutable floating-point data (and associate it with an atom). ProtoAtoms cannot be searched for in the AtomSpace indexes, although the atoms and keys that refer to them can be.

Difficult decisions include: 
* Should Handles point at protoatoms?
* How should casting between AtomPtr and ProtoAtomPtr be handled?
* How should ProtoAtoms work with Links? I.e. they belong in the outgoing set, right?  If so, then various parts of the code have to check for protoatoms, and cast or not cast them to atoms.  Should the outgoing set be a vector of protoatom?  How would casting work?

*Idea! maybe we should also have a special ProtoLink that stores a vector of protoatoms!!! That way, the usual Links are undisturbed, and the protolink handles the requires atom/key linkages to the value. The protolink would need to be stored in an atomtable index for fast lookup.

I like the last idea. However, based on experience, liking an idea, and liking its actual implementation are two very different things.

linas:
FWIW the current code is here:

https://github.com/linas/atomspace/tree/protoatom

linas:
And I pushed a copy to here https://github.com/opencog/atomspace/tree/protoatom for safe-keeping -- since deleting branches in github is real easy, and would be a major loss if it got deleted.

linas:
FWIW, the protoatom code *could* be merged into master today, since all unit tests pass.  I have refrained from merging because there is a (minor) performance hit in various places: there is extra casting and checking of handles and atom pointers and what-not.  Flip side, atomspace insertion code got simpler.

linas:
Notes:
* Casting AtomPtr to NodePtr and similar casts are very expensive (based on reading the disassembly only; not based on measurements. This should also be measured.).
* The above suggests that smart-pointer casts should be avoided at all costs. 
* The above implies that the ProtoAtom should have dummy virtual methods for everything that a regular Atom or Node or Link might need.  The dummies should throw exceptions.
* NodePtr, LinkPtr, AtomPtr should all be exactly same as ProtoAtomPtr.
* There should be isNode() and isLink() methods to ease decisions, e.g. in the pattern matcher.

[...]

PR #618, opened 2016-02-05
Avoid smart-pointer casts, where possible

linas:
Per discussion in issue #513.

Measured the performance of the cast; its large and dominates the cost of getOutgoingSet().

### DIFF
--- a/opencog/benchmark/diary.txt
+++ b/opencog/benchmark/diary.txt
@@ -1056,6 +1056,7 @@ gets removed now. The change in addNode and addLink is unclear.
 
 How fast is std::vector<Handle>, really?
 
+atomspace_bm -
 
    count             1            2           3            4
    rate           M-ops/sec    M-ops/sec   M-ops/sec    M-ops/sec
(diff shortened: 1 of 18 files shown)


opencog/benchmark/diary.txt
```diff
@@ -1063,6 +1065,18 @@ How fast is std::vector<Handle>, really?
...
+
+push_back: call the push-back method
+     example: HandleSeq oset; oset.push_back(ha); oset.push_back(hb);
+
+emplace-back: call the emplace_back method
+
+reserve: create vector of fixed size, then assign directly.
+     example: HandleSeq oset(2); oset[0] = ha; oset[1] = hb;
+
+Looks like direct assignment is a winner!
```
inflector: Did you try reserve of fixed size then emplace_back? I think that might save a copy.

opencog/benchmark/diary.txt
```diff
@@ -1087,7 +1101,39 @@ Again, ProtoAtom.
...
+LinkCast: perform LinkCast only.
+
+Comments: we expect the pointerCast and the getOutgoingSet values to be
+nearly identical, since getOutgoingSet is just a cast, followed by a
+very simple method call.  The NodeCast and Link Cast numbers might
+differ, but not much, due to the different ratios of links and nodes in
+the sample set.
+
+Ergo, the cost of getOutgoingSet is dominated by the LinkCast!  If we
+get rid of it, then what happens?  It goes up to 44.1M-ops/sec! Woww!
```
inflector: Goodness, that will really speed up the pattern matcher. Won't it?

opencog/benchmark/diary.txt
linas: Yes, it made no difference.  Neither emplace_back or push_back seem to make use of the reserved space, from what I could tell.

opencog/benchmark/diary.txt
linas: probably not.  Most of the heavy lifting done by the pattern matcher has already been moved out, to a distinct initialization phase. What's left is a traverse of the graphs.

linas: But benchmarks are needed.  Pattern matcher performance is not really understood.

opencog/benchmark/diary.txt
inflector: I did some more work on the benchmark using vector.reserve() instead of the constructor since that is the only method available when you don't know the size of the outgoing set ahead of time. Using vector.reserve() makes the three approaches the same speed.

See the changes to diary.txt here: https://github.com/opencog/atomspace/pull/691

PR #620, opened 2016-02-05
Try to fix python benchmark.

linas:
It still crashes.   Something is wrong with the way it's designed.

### DIFF
--- a/opencog/benchmark/benchmark.py
+++ b/opencog/benchmark/benchmark.py
@@ -34,7 +34,7 @@ Get and Traverse 100K - by type                 1.473µs         679,073
 Yield Get and Traverse 100K - by type           1.035µs         966,191
 Get Outgoing                                    3.490µs         286,569
 Get Outgoing - no temporary list                3.469µs         288,235
-
+Test scheme_eval_h(+ 2 2)                      49.944µs          20,022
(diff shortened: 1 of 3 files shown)


inflector:
I'm about to work on that as part of speeding up the scheme eval call. I want to measure progress.

opencog/benchmark/benchmark.py
```diff
@@ -4,10 +4,15 @@
...
-take much longer than their comparable Cython bindings.
+Displays the number of operations per second achieved under various
+scenarios utilizing the Python bindings that wrap the AtomSpace using
+Cython. Some of the tests compare the evaluation of nodes using the
+Python Scheme bindings which take much longer than their comparable
+Cython bindings. XXX Which is utterly false and insane -- the
+measurements are being done incorrectly, since the Scheme bindings
+normally run about 3x to 10x faster than the Python bindings do.
+Python is generally quite slow, and is very unlikely to be faster than
+scheme; something is borked with this test case.
```
inflector: I think this was poorly worded and was referring to the scheme eval overhead not the scheme bindings.

linas: fix the wording when you're ready.

PR #621, opened 2016-02-05
Add deterministic atoms less

ngeiswei:
This is not to merge immediately. I'm still working on it (mostly benchmarking). But I'm putting this already to get some feedback.

# Overview
This implements a deterministic less than operator for Handle. Deterministic in the sense that if the Handles are undefined it computes the less_than relationship based on contents rather than pointers.

It has the main advantage that it makes unit tests reproducible, assuming of course that the atoms are added in a deterministic manner. For instance map<Handle, float> will behave the same way every time.

Obviously the main drawback is that it has more overhead than comparing pointers, however, according to my understanding this should not compromise the performances of the AtomSpace  because the indexes Handles always have UUIDs, so never resorts to content comparison. So the overhead should only occurs when manipulating atoms that not in any atomspaces.

I tried to estimate that but couldn't precisely, ultimately it depends on the size of the atoms, what proportion isn't in any atomspaces, etc. I find it very hard to get a realistic figure of that.

So what I've merely done is to compare the performances of the atomspace repository unit tests, 20 times and average, I found no significant performance drop. Then, to be sure that there was at least some non negligible usages of comparison by content I intentionally introduced inefficiencies (time waster) in the comparison. Here are the results.

# Performances

## Pointer based atom less than

```
N   min   max    sum      mean     stddev
20  69.3  71.43  1411.99  70.5995  0.591648
```

## Content based atom less than

```
N   min    max    sum      mean     stddev
20  69.56  72.06  1416.01  70.8005  0.677266
```

## Content based atom less than + loop of 50K printing some short message (it's rederected to /dev/null I think, but I just wanted to be sure gcc doesn't optimize it away)

```
N   min     max     sum      mean     stddev
20  108.87  113.03  2214.56  110.728  1.21169
```

## Content based atom less than + loop of 100K printing some short message

```
N   min     max     sum      mean     stddev
20  155.64  171.82  3234.69  161.734  3.94332
```

I tried to work out some math to extract more information about the overhead but I ended with more variables than equations and I couldn't solve it. In other words the only way to probe deeper is gonna be run a real benchmarking tool. But again I don't know what is supposed to be realistically meaningful. I guess running the FC on some real world inference could give a meaningful answer.

### DIFF
--- a/opencog/atoms/base/Atom.h
+++ b/opencog/atoms/base/Atom.h
@@ -417,6 +417,13 @@ public:
      * @return true if the atoms are different, false otherwise.
      */
     virtual bool operator!=(const Atom&) const = 0;
+
+    /** Returns whether this atom is less than the given atom.
+     *
+     * @return true if this atom is less than the given one, false otherwise.
+     */
+    virtual bool operator<(const Atom&) const = 0;
+
 };
 
 /** @}*/
--- a/opencog/atoms/base/Handle.cc
+++ b/opencog/atoms/base/Handle.cc
@@ -76,6 +76,21 @@ bool Handle::atoms_less(const Atom* a, const Atom* b)
     return a < b;
 }
 
+bool Handle::deterministic_atoms_less(const Atom* a, const Atom* b)
+{
+    if (a == b) return false;
+    if (nullptr == a) return true;
+    if (nullptr == b) return false;
+    UUID ua = a->getUUID();
+    UUID ub = b->getUUID();
+    if (INVALID_UUID != ua or INVALID_UUID != ub) return ua < ub;
+
+    // If both UUID's are invalid, we compare them by content. This is
+    // expensive but useful when you really want a deterministic
+    // comparison.
+    return a->operator<(*b);
+}
+
 // ===================================================
 // Handle resolution stuff.
 
--- a/opencog/atoms/base/Handle.h
+++ b/opencog/atoms/base/Handle.h
@@ -131,16 +131,16 @@ public:
         return _ptr.get() != h._ptr.get();
     }
     inline bool operator< (const Handle& h) const noexcept {
-       return atoms_less(_ptr.get(), h._ptr.get());
+       return deterministic_atoms_less(_ptr.get(), h._ptr.get());
     }
     inline bool operator> (const Handle& h) const noexcept {
-       return atoms_less(h._ptr.get(), _ptr.get());
+       return deterministic_atoms_less(h._ptr.get(), _ptr.get());
     }
     inline bool operator<=(const Handle& h) const noexcept {
-       return not atoms_less(h._ptr.get(), _ptr.get());
+       return not deterministic_atoms_less(h._ptr.get(), _ptr.get());
     }
     inline bool operator>=(const Handle& h) const noexcept {
-       return not atoms_less(_ptr.get(), h._ptr.get());
+       return not deterministic_atoms_less(_ptr.get(), h._ptr.get());
     }
 
     /**
--- a/opencog/atoms/base/Handle.h
+++ b/opencog/atoms/base/Handle.h
@@ -130,18 +130,20 @@ public:
     inline bool operator!=(const Handle& h) const noexcept {
         return _ptr.get() != h._ptr.get();
     }
+#define DEFAULT_ATOMS_LESS deterministic_atoms_less
     inline bool operator< (const Handle& h) const noexcept {
-       return deterministic_atoms_less(_ptr.get(), h._ptr.get());
+       return DEFAULT_ATOMS_LESS(_ptr.get(), h._ptr.get());
     }
     inline bool operator> (const Handle& h) const noexcept {
-       return deterministic_atoms_less(h._ptr.get(), _ptr.get());
+       return DEFAULT_ATOMS_LESS(h._ptr.get(), _ptr.get());
     }
     inline bool operator<=(const Handle& h) const noexcept {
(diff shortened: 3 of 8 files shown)


linas:
Benchmarking the unit tests is a terrible way to measure performance.  The unit tests are dominated by a few long-running tests that don't do anything interesting.  The tests that actually stress the system are extremely short, and are dominated by the set-up and tear-down costs.  The insides of most of the unit tests take microseconds to run; the test itself might take a second only because of setup/teardown.

It would be better to run `./benchmark/atomspace_bm -A`  before and after. its also a terrible test, but its better than the unit tests.

I can no longer remember if less-than is used in any code-critical path.  Where is it used?  Unordered links, I guess.. anywhere else?

Atoms that are not in the atomspace are not really comparable.  I mean, you can compare them, but there is no way to guarantee that they are unique.

What problem are you trying to solve?  You hint that there is some problem about "reproducibility", but what is that problem?  Perhaps there are better ways of solving it?

ngeiswei:
I'm using benchmark_bm, but I'm observing huge variances in performances (like over 20%) even when I provide a constant random seed (and of course using the deterministic atoms_less). Any idea?

ngeiswei:
I should emphasize that even with a very large -n of 1M, I do get such large variances in performances.

ngeiswei:
I could verify that the AtomSpaces created are the same (for a given random seed), yet large jitters persist (over 20% of run-time), even when using 10M atoms (I can't try much more than that cause it reaches my machine RAM limit). This could come from the OS, but I suspect it comes from the AtomSpace indexes that rely on Atom pointers (to get their hash values).

All this makes benchmarking performances very noisy. I guess we should add some option in the atomspace_bm that repeats N times the benchmark to get a mean that actually converges.

I've got to say I'm not sure I like having indexes rely on the atom memory addresses. It doesn't feel right to me. Why don't we use their UUIDs? Would that be slower? I'll probably try tomorrow.

linas:
Multiple remarks.

1) The benchmark is very badly designed.  It was written by someone who had just no clue how to benchmark. I've been trying to repair it, but its a nasty tangle of code, so maybe a clean rewrite is in order. Assuming the person doing the re-write has a clue about benchmarking... else we'll just get a different mess again.

2)  Yes, it jitters like crazy for me, too. This is not at all unusual, in my experience (not just for this benchmark, but in general): you've got different patterns of cache and TLB misses each run, due to other processes and threads running; you have kernel address-space randomization effects (an anti-virus strategy); ethernet interrupts, etc.  Once upon a time, I noticed that the performance reported by this tool went up, when I also ran something else (e.g. a compile).  I believe that this had something to do with the `clock()` system call being called too frequently -- having a compile going in the background allowed the kernel scheduler to reschedule the benchmark more quickly, after each system call.  Last time I checked, this effect went away, though.   Not sure if its due to benchmark changes, or kernel changes or both.   This is part of what I mean by "understanding how to benchmark".

3) The indexes use pointers, because the whole point is to find the actual atom.  If the indexes used UUID's, then you would have to do two lookups, not one: first, to find the UUID, and second to find the actual atom that the UUID refers to.

4) I am planning on removing the UUID's from the atoms, anyway. The UUID's are meant  to synchronize atoms across multiple networked machines, and for the SQL persistent storage backend.  Otherwise, UUID's are not needed.   The protoatoms already don't have UUID's in them, because it makes no sense (same reason that having UUID's in TV and AV does not make sense).

linas:
The point is: the address in RAM is a de-facto universally-unique identifier, for that particular session, on that particular machine. 

Note also: RAM addresses are completely deterministic, and do NOT depend on anything random, as long as you do these three things: 

1) disable address-space randomization in the kernel.

2) make sure you are running single-threaded, so that all mallocs are in deterministic order.

3) Avoid anything else that is a source of non-determinism: e.g. don't use green threads or any other timer technique that would cause mallocs to happen in different order in different runs.   Don't use a random number generator to generate different numbers of atoms.  Don't use mallocs in the benchmark that malloc different sizes, based on the data:  e.g. mallocing a string for "10 per second" mallocs 1 more byte than the string for "9 per second". This difference of 1 byte can lead to non-repeateability in everything that gets malloc'ed afterward.

4) Disable the opencog logger from using threads.  It mallocs and prints in different trheads, in non-deterministic ways.

In other words, RAM addresses are both perfectly unique, and perfectly deterministic,   *if* you remove all sources of non-determinism.

The current integer UUID's are a little more deterministic (e.g. the logger does not affect them); however multi-threaded atomspace access will cause the UUID's to be alloc'ed in a non-deterrministic fashion.

I think you are chasing a ghost with this set of commits.  Eliminating non-determinism without impacting performance is more-or-less impossible in modern computing systems; its been impossible for 40+ years, basically, since interrupts and multi-tasking became popular.  This is generally known & accepted by anyone who has ever benchmarked anything, ever.

You can get deterministic UUID's with hashing, but hashing is very expensive; prohibitively expensive in this particular case, I believe.

ngeiswei:
This is awesome beyond words! I had no idea one could disable address space layout randomization!!!

All I had to do is type in a terminal

```bash
sudo sysctl -w kernel.randomize_va_space=0
```

to get identical indexes across runs!

Also, after setting logger sync flag to true, my URE bug (which was primary reason for that PR) is reproducible as well!

So this PR has no purpose anymore.

I still observe 20% of jitters when running the benchmark. It seems to be all or nothing, 0%, or 20%. There must be a reason, but since I don't need to benchmark anymore I'm gonna let that issue aside.

Thanks for sharing this important piece of knowledge!

PR #627, opened 2016-02-06
Fix benchmark.py crash and scheme loading issue

inflector:


opencog/benchmark/benchmark.py
```diff
@@ -7,12 +7,9 @@
...
 Cython. Some of the tests compare the evaluation of nodes using the
-Python Scheme bindings which take much longer than their comparable
-Cython bindings. XXX Which is utterly false and insane -- the
-measurements are being done incorrectly, since the Scheme bindings
-normally run about 3x to 10x faster than the Python bindings do.
-Python is generally quite slow, and is very unlikely to be faster than
-scheme; something is borked with this test case.
+Python scheme_eval_h which takes much longer than the comparable
+Cython bindings because of the overhead of creating and initializing
+a SchemeEval object.
```
linas: ??? Why'd you remove the comment?  When I measure python and compare it to scheme, I find that scheme is faster. So this just seems like some random partisan comment that does not reflect reality, at least, not the reality of the python 2.7.6 that I have installed on my system.

inflector: Because this is a benchmark timing Python bindings and execution. The statement says that calling Python bindings compared to scheme_eval_h takes longer than the comparable Cython bindings. That is most definitely true.

There is a test which compare the Python bindlink against a call of cog-bind using scheme_eval_h.

This is not a statement about the relative speed of Python versus scheme at all. It is a comment about the speed of the scheme binding as currently implemented.

At any rate, I am working on fixing this so that the scheme binding doesn't take so much time using the suggestions you made in the request that Jim Rutt made on the email list. So this comment is short lived anyway. I'm going to take it away next week after the overhead is limited.

PR #630, opened 2016-02-08
Fix benchmark

ngeiswei:
Fix option -X + misc improvements


opencog/benchmark/README.md
```diff
@@ -9,19 +9,21 @@ the resident set size (memory) of the benchmark process.
...
-calculates some statistics on the times (min, max, mean, etc) and -f dumps
-the records to a file called addLink_benchmark.csv (the filename used is
-always the method name appended by "_benchmark.csv").
+This will run the addLink benchmark (-m "addLink"). It will also build
+a test AtomSpace before running the benchmark which consists of random
+graph whose size and link density can be changed on the command line.
+
+It will run and measure the speed of the addLink method 10,000 times
+(-n 10000), but between each it will add 100 more random atoms to the
+AtomSpace (-S 100) so that you can assess how performance scales with
```
linas: Interesting. Either -S doesn't work, or its not doing what is expected.

opencog/benchmark/README.md
```diff
@@ -9,19 +9,21 @@ the resident set size (memory) of the benchmark process.
...
+AtomSpace size. -k calculates some statistics on the run times (min,
+max, mean, etc) of the specified method, and -f dumps the records to a
+file called addLink_benchmark.csv (the filename used is always the
```
linas: -k seems broken too.   I just did got this:
```
Benchmarking AtomSpace's addLink method 400000 times ..........
47.040896 seconds elapsed (21258.10 per second)
Sum clock() time for all requests: 16100398 (16.1004 seconds, 62110.3 requests per second)
Per operation stats, in CPU clock ticks: 
  N: 200
  mean: 80501
  min: 59031
  max: 231048
  std: 14046
```

which seems to report 3 highly inconsistent rates: 21K per second, 62K per second, and 80K per second.  Which is it? What do these numbers mean?  I've never used -k before; this is deeply confusing.

ngeiswei: I don't think it's not broken, it's because -S adds atoms between the benchmarked method. But yeah, it's certainly very confusing.

linas: Well, with -S flag, I expect a sequence of measurements: viz "I started with N atoms and got blah per second  then added S atoms for a total of N+S and got blahblah atoms per second and then N+2S and got blahblahblah per second."   That is, there should be a report for N, for N+S, N+2S, N+3S, N+4S and N+5S -- without these separate, distinct reports, I'm not sure what do do with it.

It's not your job to fix this; I'm just sort of just expressing frustration with this benchmark. Its not trust-worthy.

PR #631, opened 2016-02-09
Add deterministic atoms less

ngeiswei:
logger().setSyncFlag(true) doesn't work as magically as I hoped (it seems to decrease the likelihood of indeterminism, but doesn't bring it to zero).

This PR is pretty much identical to #621, however deterministic_atoms_less has been renamed content_based_atoms_less and is disabled by default.


linas:
Well, OK, so I told you that UUID's are going to be removed from Atoms, and you merged anyway?

I am quite certain that this is a very wrong approach to whatever bug it is that you have.  This patch grossly abuses UUID's, doing bad things with them. This should not have been merged.  I really don't know how many times I can say the same thing over and over and over again.

ngeiswei:
@linas you misunderstood what this patch is doing, please read more carefully before starting to rant.

opencog/atoms/base/Handle.cc
```diff
@@ -80,6 +80,21 @@ bool Handle::atoms_less(const Atom* pa, const Atom* pb)
     return a < b;
 }
 
+bool Handle::content_based_atoms_less(const Atom* a, const Atom* b)
+{
+    if (a == b) return false;
+    if (nullptr == a) return true;
+    if (nullptr == b) return false;
+    UUID ua = a->getUUID();
+    UUID ub = b->getUUID();
```
ngeiswei: OK, I see that's the part that made you assume I have abused UUID, but this is a just a copy/paste of atoms_less. Sure I could have striven to not to copy/paste anything but I didn't want to refactor anything either so there it went.

opencog/atoms/base/Link.cc
```diff
@@ -149,3 +149,27 @@ bool Link::operator==(const Atom& other) const
...
+                    i++;
+                else
+                    return ll->operator<(*rl.atom_ptr());
+            }
+            return false;
+        } else
+            return arity < other_arity;
+    } else
+        return getType() < other.getType();
+}
```
ngeiswei: This function is the gut of the PR and it doesn't abuse UUID. It doesn't fix anything, it just helps me to reproduce a bug, give me another way and I'm happy to remove that code. Please keep in mind that this code is not used unless DEFAULT_ATOMS_LESS is define as content_based_atoms_less. In other words it doesn't bother anyone.

opencog/atoms/base/Handle.cc
linas: Ah, OK, sorry.  I probably wrote that code, and so shouldn't be so touchy. The meta-issue is that I seem to spend too much time struggling with pointless stupid bugs, and not enough time advancing the high theory and important functions.  Its like, the rocket can't launch because a bolt fell out.  So I get snappy when I see anything resembling a loose bolt.

PR #632, opened 2016-02-09
Remove UUID abuse

ngeiswei:
Not really an abuse, but if we're to compare by content and we're not
that concerned about performances, then it's even better to not use UUID
at all as the comparison will then even be independent of the order of
atom creation.


linas:
OK.   I can't recall any more if this occurs frequently in some performance-critical path.  That's the nly thing here that would concern me.

ngeiswei:
No, it should not concern you or anybody cause, as long as DEFAULT_ATOMS_LESS is not set to content_based_atoms_less, this code is not used at all.

PR #636, opened 2016-02-11
Add Jenkins job config file

wenwei-dev:
This is the job config file that is currently used in Jenkins CI


wenwei-dev:
The config files will go to github.com/opencog/docker/jenkins.

Issue #595, opened 2016-01-21
make test errors


beew:
Hi, just tried to compile atomspace, configure and make seemed fine but make test failed with errors

```
/atomspace/build$ cmake ..
-- The C compiler identification is GNU 5.2.1
-- The CXX compiler identification is GNU 5.2.1
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Build type: Release
-- Boost version: 1.58.0
-- Found the following Boost libraries:
--   date_time
--   filesystem
--   program_options
--   regex
--   serialization
--   system
--   thread
-- Boost version 105800 found.
-- Looking for include file opencog/util/cogutil.h
-- Looking for include file opencog/util/cogutil.h - found
-- Found CogUtil version 2.0.1 at /usr/local/lib/libcogutil.so
-- CogUtil found.
-- Found PythonInterp: /usr/bin/python (found version "2.7.10") 
-- CxxTest found.
-- Gearman found.
-- Guile (2.0.11 >= 2.0.9) was found.
-- Found these hypertable libs: /opt/hypertable/current/lib/libHyperCommon.a;HYPER_DFS_BROKER-NOTFOUND
-- Found hypertable install path: /opt/hypertable/current
-- Found hypertable config file: /opt/hypertable/current/conf/hypertable.cfg
-- Hypertable was found.
-- Found SIGAR: /usr/local/lib/libsigar-x86-linux.so;dl
-- Found these protobuf libs: /usr/lib/x86_64-linux-gnu/libprotobuf.so
-- Found PROTOBUF: /usr/lib/x86_64-linux-gnu/libprotobuf.so  
-- Protobuf was found.
-- Found PythonLibs: /usr/lib/x86_64-linux-gnu/libpython2.7.so (found suitable version "2.7.10", minimum required is "2.7") 
-- Python libs found.
-- Found Cython: /usr/bin/cython (Required is at least version "0.19.0") 
-- Cython ( >= 0.19.0) found.
[...]

PR #642, opened 2016-02-13
Add circular dependecies problem workaround (#595)

jswiergo:
Changes for Ubuntu 15.10 support (cmake version = 3.2.2 gcc version = 5.2.1)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,6 +76,10 @@ IF (CMAKE_COMPILER_IS_GNUCXX)
 		SET(CMAKE_C_FLAGS_DEBUG "-ggdb3 -fstack-protector")
 		SET(CMAKE_C_FLAGS_PROFILE "-O2 -g3 -fstack-protector -pg")
 		SET(CMAKE_C_FLAGS_RELEASE "-O3 -g -fstack-protector -flto")
+		LINK_LIBRARIES(pthread)
+
+		# Workaround for circular dependencies problem
+		SET(CMAKE_CXX_LINK_EXECUTABLE "${CMAKE_CXX_LINK_EXECUTABLE} <LINK_LIBRARIES>")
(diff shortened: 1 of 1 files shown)


linas:
Thanks. I merged ... but I don't understand how this works. Can you explain?

jswiergo:
CMAKE_CXX_LINK_EXECUTABLE is a variable available in CMake >=3.0. It tells the rule how to generate linker command line (link.txt files in build subdirs). The default rule is:
```
<CMAKE_CXX_COMPILER>  <FLAGS> <CMAKE_CXX_LINK_FLAGS> <LINK_FLAGS> <OBJECTS>  -o <TARGET> <LINK_LIBRARIES>
```
After the change the link libraries list is repeated twice:
```
<CMAKE_CXX_COMPILER>  <FLAGS> <CMAKE_CXX_LINK_FLAGS> <LINK_FLAGS> <OBJECTS>  -o <TARGET> <LINK_LIBRARIES> <LINK_LIBRARIES>
```
According to ld documentation (https://sourceware.org/binutils/docs/ld/Options.html#Options):
>The linker will search an archive only once, at the location where it is specified on the command line. If the archive defines a symbol which was undefined in some object which appeared before the archive on the command line, the linker will include the appropriate file(s) from the archive. However, an undefined symbol in an object appearing later on the command line will not cause the linker to search the archive again.
>
>You may list the same archive multiple times on the command line. 

Suppose we have circular dependency, for example libA -> libB -> libC -> libD -> libA. We generate repeated libraries list: libA libB libC libD libA libB libC libD. This ensures that the linker will find objects for each symbol, because each library appears on the list after each other.

linas:
Yes, excellent, that is exactly the right fix!  Wow. I think this makes me feel much better, because this has been a miserable, thorny issue that I could not quite see past, before.

Issue #644, opened 2016-02-15
How to make a choice while mapping in atomese?


AmeBel:
When using a GetLink the returned result pattern could be variable. What I wanted to do is deal with each result depending on what the result of the get could be. In other words, how to do an `if-elif-else` pattern?
## Scheme functions

``` scheme
(define (rand-set)
    ; This is for simulating a return from a GetLink

    (define pseudo-get
        (RandomChoice
            (SetLink
                (SetLink (ListLink (ConceptNode "foo")))
                (SetLink
                    (ListLink
                        (ConceptNode "bar")
                        (NumberNode 1)))
            ))
    )

    (cog-execute! pseudo-get)
)

(define (map-test a-set)
    (define map-link
        (MapLink
            (ImplicationLink
                (VariableList
                    (TypedVariable (Variable "$x") (Type "ConceptNode"))
                    (TypedVariable (Variable "$y") (Type "NumberNode")))
                (ChoiceLink
                    (ListLink
                        (Variable "$x")
                        (Variable "$y"))
                    (ListLink
                        (Variable "$x"))
                )
                (Variable "$x")
            )
            a-set
        ))
        (display "The set = \n")
        (display a-set)
        (display "The mapping = \n")
        (cog-execute! map-link)
)
```
## Present Output

``` scheme
guile> (map-test (rand-set))
The set =
(SetLink
   (ListLink
      (ConceptNode "foo")
   )
)
The mapping =
(SetLink
)

guile> (map-test (rand-set))
The set =
(SetLink
   (ListLink
      (ConceptNode "bar")
      (NumberNode "1.000000")
   )
)
The mapping =
(SetLink
)

```
## Expected Output

``` scheme
guile> (map-test (rand-set))
The set =
(SetLink
   (ListLink
      (ConceptNode "foo")
   )
)
The mapping =
(SetLink
      (ConceptNode "foo")
)

guile> (map-test (rand-set))
The set =
(SetLink
   (ListLink
      (ConceptNode "bar")
      (NumberNode "1.000000")
   )
)
The mapping =
(SetLink
      (ConceptNode "bar")
)

```

linas:
Sounds like a bug to me.  Or rather, a missing feature; the code doesn't know or care about ChoiceLink.

However, you might be able to achieve the same thing with TypeChoice.  Let me try ...

linas:
Pull request #645 fixes this.  merge whenever it finishes doing its thing

AmeBel:
thanks

[...]

PR #645, opened 2016-02-15
fix for bug #644

linas:
Also, some fair amount of unrelated cleanup, e.g. removal of un-needed casts.


AmeBel:
retest this please

PR #653, opened 2016-02-18
skip some initialization of transient atomspaces.

linas:
The temporary atomspaces used as scratch space during evaluation do not need to do a full initialization.  This should make things run a little bit faster.

### DIFF
--- a/opencog/atomspace/AtomSpace.cc
+++ b/opencog/atomspace/AtomSpace.cc
@@ -51,9 +51,17 @@ using namespace opencog;
 
 // ====================================================================
 
+/**
+ * Transient atomspaces skip some of the initialization steps,
+ * so that they can be constructed more quickly.  Transient atomspaces
+ * are typically used as scratch spaces, to hold temporary results
+ * during evaluation, pattern matching and inference. Such temporary
+ * spaces don't need some of the heavier-weight crud that atomspaces
+ * are festooned with.
+ */
 AtomSpace::AtomSpace(AtomSpace* parent, bool transient) :
     atomTable(parent? &parent->atomTable : NULL, this, transient),
-    bank(atomTable),
+    bank(atomTable, transient),
     backing_store(NULL)
 {
 }
@@ -67,7 +75,7 @@ AtomSpace::~AtomSpace()
 
 AtomSpace::AtomSpace(const AtomSpace&) :
     atomTable(NULL),
-    bank(atomTable),
+    bank(atomTable, true),
     backing_store(NULL)
 {
      throw opencog::RuntimeException(TRACE_INFO,
--- a/opencog/atomspace/AtomSpace.cc
+++ b/opencog/atomspace/AtomSpace.cc
@@ -51,8 +51,8 @@ using namespace opencog;
 
 // ====================================================================
 
-AtomSpace::AtomSpace(AtomSpace* parent) :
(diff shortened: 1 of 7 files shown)


linas:
retest this please

inflector:
I did some work to get gprof working. 

Here's a pastebin of 100,000 calls to bindlink:

http://pastebin.com/aKgnA2Mg

Is there any reason we can't cache and use the same DefaultPatternMatchCB? Most of the time is spent in constructors and destructors for that object.

linas:
Most of the cost of DefaultPatternMatchCB  appears to be in the constructor for the scratch atomspace, which this path should reduce somewhat.

The problem with caching the callback is that its a run-time thing: if you have two different threads using it, clearly they cannot share it. Also, you can't just cache it in the BindLink itself, since two different threads might be running the same bindlink at the same time.

You could try caching it per-thread, in thread-local sotarge, but its also a per-atomspace thing, so you need a thread-local storage for each different atomspace.  If an atomspace is destroyed, then you have to run the destructor for that  TLS cache as well, which means you have to  .. I dunno, register destructor callbacks somwhere. Which introduces yet more overhead, and which requires locking, and things get confusing: what if thread A is altering the callback sttructure that thread B is trying to remove an atomspace from?  It gets to be a hard-to-debug tangle, real quick.

linas:
93% of the cost of the DefaultPatternMatchCB ctor was the AtomSpace ctor.  Half the cost of the AtomSpace ctor was the AtteentionBank, that should now be zero after this patch.   The other 1/2 the cost is the AtomTable ctor, 

over 1/3 of the cost of the AtomTable ctor is the TypeIndex; another 1/3 is the ImportanceIndex. curiosly the ctor for NodeIndex is 10x faster (either that, or the statistics are flawed).

linas:
100% of the cost of TypeIndex and ImportanceIndex ctors are in std::vector<std::unordered_set<opencog::Atom*>>  -- Note that one std::unordered_set is created for *each* atom type, there are maybe 100 or 200 atom types.    Find some other way of building hese indexes, and the ctor cost for DefaultPatternMatchCB will shrink yet more.

linas:
anyway, perhaps you can re-measure. I'm guessing that this patch put a big dent in your original profile.

linas:
I can't quite figure out who is calling opencog::FixedIntegerIndex::size() -- this might be mis-reported.

inflector:
I'm checking out your changes.

I've got a pull request I just submitted, I want you (and perhaps Amen
since it is build related) to support a Profile build. I added notes for
how to do it to the PR.

linas:
Pull req #655 should also improve the gprof trace that you posted (that is, if I correctly guessed what the FixedIntegerIndex gunk was about.)

inflector:
I just submitted pull request #666 which implements a multi-threaded cache of transient atomspaces. The cache is currently set at 8 maximum atomspaces, but I can set that value to some other config variable or another value based on your suggestion.

Rather than deleting atomspaces it clears them (i.e. does most of the cleanup of ~AtomSpace) and places them into a stack of cached atomspaces. Access to the cache is protected during the process of grabbing an atomspace from the cache and placing one into the cache by a mutex.

Assuming you are okay with the design and implementation, I'll add a test to the multi-threaded unit test to perform multiple simultaneous bindlink searches, to test the mutex implementation here.

PR #654, opened 2016-02-19
Add support for Profile builds and profiling for bindlink.

inflector:
Makes the library type default SHARED in normal circumstances and STATIC under a Profile build. It now works in both profile and normal SHARED library which scheme and the cogserver need for dynamic loading sample code. It also changes the library and archive destinations to not specify the type so that it works in either case since the location doesn't need to change and just setting DESTINATION works fine.

```
mkdir profile
cd profile
cmake -D CMAKE_BUILD_TYPE=Profile ..
make
```
Now builds a profile build. There is a new file in benchmark called profile_bindlink.cc which contains a profile exe for bindlink profiling. 

Running:

`./profile_bindlink`

from the benchmark directory in profile runs for a few seconds and then creates a file:

`gmon.out`

when it quits. T     then running:

`gprof profile_bindlink gmon.out > analysis.txt`

outputs the results to `analysis.txt`.


opencog/benchmark/profile_bindlink.cc
```diff
@@ -0,0 +1,101 @@
...
+        "  (VariableNode \"$var\")\n"
+        "  ;; The pattern to be searched for\n"
+        "  (InheritanceLink \n"
+        "    (VariableNode \"$var\") \n"
+        "    (ConceptNode \"animal\")\n"
+        "  )\n"
+        "  ;; The value to be returned.\n"
+        "  (VariableNode \"$var\")"
+        ")";
+    return scheme->eval_h(animals_query.c_str());
```
linas: You realize, of course, that this is an absurdly simple search over a ridiculously small atomspace, and so is going to end up over-emphasizing the setup and tear-down costs of running the pattern matcher.  The "real life" queries, such as those in relex2logic, are far far more complex than this, and the atomspaces of the empty chatbot is 3K or 5K atoms; language atomspaces are in the million-atom range.

linas: A good query to measure might be for example SVIO1.scm  here: https://github.com/opencog/opencog/blob/master/opencog/nlp/relex2logic/rules/SVIO1.scm

here's one with AbsentLink
https://github.com/opencog/opencog/blob/master/opencog/nlp/relex2logic/rules/todo5.scm

inflector: We certainly ought to have a sweet of different queries that test more typical use cases. I will add these to the benchmark suites and do a profile there as well as it is much more important, obviously, to optimize the case we care about.

Yes, I do realize this query is overly simplistic. It also happens to be the exact query that slowed from 18µs to 25µs in one week and then to 350µs in one checkin. So it a useful test for that specific problem.

The Python benchmarks, from which this query was taken were also originally designed to profile the bindings more than the underlying functions. Though I've been adding more general useful benchmarks there as well.

PR #657, opened 2016-02-21
Added section on profiling and profile builds to benchmark README

inflector:


inflector:
Retest this please. This was just a change to a README.md file, shouldn't have failed the test.

wenwei-dev:
There is a test not stable. It fails now and then and I don't understand. 
I suggest we don't rebuild the docker image so frequently, and install a specific cogutils at the testing time. This helps reduce the chance of mysteries happening.

inflector:
@wenwei-dev Intermittent failures are hardest to find bugs, in general. 

This particular failure is likely due to elements of random source like the memory-partitioning which changes the way that the atoms sort, or random seeds being different from test to test. It might be helpful to nail down the specific source of randomness for this by printing out sufficient data for the test that is failing so we can reproduce that specific test failure reliably, i.e. for a start, let's print out the randseed values for random generation if there is any for the test. Then we can run the same test with that specific randseed and see if the failure is 100% reproducible with that value.

Issue #648, opened 2016-02-17
DefinedType not allowed in TypedVariableLink


AmeBel:
Is the following a missing feature or intentional? Just curious.

First modifying the type-signature example at [line 62](https://github.com/opencog/atomspace/blob/master/examples/pattern-matcher/type-signature.scm#L62) to be
```scheme
(DefineLink
    (DefinedType "predicate-type")
    (Signature
        (EvaluationLink
            (TypeChoice
                (TypeNode "PredicateNode")
                (TypeNode "AnchorNode"))
            (ListLink
                (Type "ConceptNode") (Type "ConceptNode")))))

; A search pattern that looks for predicates or grounded predicates.
(define predicate-search
	(GetLink
		(TypedVariable
			(Variable "$x")
            (DefinedType "predicate-type"))
		(AndLink (Variable "$x"))))
```
Then loading it gives, the following error
```scheme
scheme@(guile-user)> (load "type-signature.scm")
;;; note: source file /atomspace/examples/pattern-matcher/type-signature.scm
;;;       newer than compiled /home/opencog/.cache/guile/ccache/2.0-LE-8-2.0/atomspace/examples/pattern-matcher/type-signature.scm.go
;;; note: auto-compilation is enabled, set GUILE_AUTO_COMPILE=0
;;;       or pass the --no-auto-compile argument to disable.
;;; compiling /atomspace/examples/pattern-matcher/type-signature.scm
;;; compiled /home/opencog/.cache/guile/ccache/2.0-LE-8-2.0/atomspace/examples/pattern-matcher/type-signature.scm.go
ERROR: In procedure cog-new-link:
ERROR: Throw to key `C++-EXCEPTION' with args `("cog-new-link" "Unexpected contents in TypedVariableLink\nExpected type specifier (e.g. TypeNode, TypeChoice, etc.), got DefinedTypeNode (/tmp/atomspace-master/opencog/atoms/core/VariableList.cc:228)")'.

Entering a new prompt.  Type `,bt' for a backtrace or `,q' to continue.
```

linas:
Its a bug, an incomplete implementation.

linas:
pull request #649 should fix this.  I did not test it very carefully.  It could use a unit test, I suppose.

linas:
I guess that a unit test for this is still needed.

AmeBel:
Would #659 suffice.

[...]

PR #659, opened 2016-02-23
Unbunding DefinedType in TypedVariableLink test

AmeBel:
test solution for #648


tests/atoms/VariableListUTest.cxxtest
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    void tearDown(void)
+    {
+        as->clear();
+    }
+
+    // Test to make sure that the unbundling of DefinedType in
+    // TypedVariableLink is taking place.
+    void test_unbundle_defined_type_node()
```
linas: Rather than creating a new unit test, could you just copy this single subroutine into DeepTypesUTest -- that is where all the other similar tests are performed.

linas:
thanks!

PR #666, opened 2016-02-25
Speedup bind with a multi-threaded transient atom space cache.

inflector:
Linas's prior changes to bindlink made it 80% faster. It was originally:

```
Test                                        Time per op  Ops per second
----                                        -----------  --------------
Bind - bindlink - Cython                      360.240µs           2,775
```
and after Linas's changes:
```
Test                                        Time per op  Ops per second
----                                        -----------  --------------
Bind - bindlink - Cython                      202.252µs           4,944
```

Implementing a multi-threaded cache for the transient atomspaces made binklink 5X faster. It is now:

```
Test                                        Time per op  Ops per second
----                                        -----------  --------------
Bind - bindlink - Cython                       39.998µs          25,001
```

I originally tried caching the entire DefaultImplicator but couldn't get that working for some reason and decided to check in the atomspace cache as that was the biggest likely savings. 

I've left the mostly implemented implication cache code in #ifdef CACHED_IMPLICATOR blocks and will add it back in later. But I've got more important work, at the moment.

### DIFF
--- a/opencog/atoms/execution/Instantiator.h
+++ b/opencog/atoms/execution/Instantiator.h
@@ -76,6 +76,18 @@ private:
 public:
 	Instantiator(AtomSpace* as) : _as(as) {}
 
+	void ready(AtomSpace* as)
+	{
+		_as = as;
+		_halt = false;
+	}
+
+	void clear()
+	{
+		_as = NULL;
+		_vmap = NULL;
+	}
+
 	Handle instantiate(const Handle& expr, const std::map<Handle, Handle> &vars);
 	Handle execute(const Handle& expr)
 	{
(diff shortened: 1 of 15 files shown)


inflector:
Jenkins retest this please

linas:
Merged. Please do add a multi-threaded pattern-matcher test. This is something that is under-tested.

There are multi-threaded atomspace tests in `tests/atomspace`.   There are multi-threed scheme tests in `tests/scm/MultiThreadUTest`

There is a `tests/scm/SCMExecutionOutputUTest` which might hit your code, because it tests execution, which also uses temporary atomspaces.  Except I guess you did not cover that case with this patch ... 

Finally, it might be best to switch to `cogutils/opencog/util/concurrent_stack.h` because I'm pretty sure that code actually does work safely.  It is also close to the proposed future C++ standards that would add this to std c++.

linas:
Care to take a look at issue #676 ?  One of the pattern matcher unit tests, EinsetinUTest, went from 10 seconds to 95 seconds on my machine.   I am guessing wildly that its due to this change.   That unit test has been a canary for pattern-matcher changes in the past.

inflector:
Pull request #680 fixes this slowdown.

PR #673, opened 2016-02-26
Larger fc improvements

ngeiswei:
1. Make code more modular
2. Fix a few bugs
3. Optimize a few things


examples/rule-engine/crisp-deduction.scm
```diff
@@ -27,9 +27,9 @@
 (load-from-path "crisp-deduction-config.scm")
 
 ; Define knowledge base
-(define A (ConceptNode "A" (stv 1 1)))
-(define B (ConceptNode "B"))
-(define C (ConceptNode "C"))
+(define A (PredicateNode "A" (stv 1 1)))
+(define B (PredicateNode "B"))
+(define C (PredicateNode "C"))
 (define AB (ImplicationLink (stv 1 1) A B))
```
linas: p.s. these days, you can just say `Concept`  or `Predicate` and it just works.

opencog/query/InitiateSearchCB.cc
```diff
@@ -653,7 +653,7 @@ bool InitiateSearchCB::variable_search(PatternMatchEngine *pme)
 	bool all_clauses_are_evaluatable = true;
 	for (const Handle& cl : clauses)
 	{
-		if (0 < _pattern->evaluatable_terms.count(cl)) continue;
+		if (0 < _pattern->evaluatable_holders.count(cl)) continue;
 		all_clauses_are_evaluatable = false;
```
linas: EinsteinUTest recently slowed down from 10 seconds to 95 seconds, on my machine. It is either due to pull req #666 or possibly to the above change on other changes in this pull req.  See bug #676

PR #680, opened 2016-02-27
Fixed speed problem with EinsteinUTest

inflector:
Moved the `clear` code for atom space to the atom table since it was using only atom table methods anyway. 

Then special-cased `clear` to use the faster `clear` for transient atom spaces. Also moved the guts of `clear_transient` out into a separate function `clear_all_atoms` that is used for both `clear` and `clear_transient`.

### DIFF
--- a/opencog/atomspace/AtomTable.cc
+++ b/opencog/atomspace/AtomTable.cc
@@ -129,15 +129,32 @@ void AtomTable::ready_transient(AtomTable* parent, AtomSpace* holder)
     Handle::set_resolver(this);
 }
 
-
 void AtomTable::clear_transient()
 {
     if (not _transient)
         throw opencog::RuntimeException(TRACE_INFO,
-                "AtomTable - clear called on non-transient atom table.");
+                "AtomTable - clear_transient called on non-transient atom table.");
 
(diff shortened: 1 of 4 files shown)


opencog/atomspace/AtomTable.cc
```diff
@@ -168,10 +185,40 @@ void AtomTable::clear_transient()
     // the last shared_ptr referecence, and set the size of the set to 0.
     _atom_set.clear();
```
linas: Technically, this is not entirely right, because it will leave weak pointers in all sorts of incoming sets.  That doesn't affect the system at least, not at first, since the weak pointers can never resolve.  However, after using the system for a while, the weak pointers will accumulate on more and more atoms, siting there, eating up space, and pointing at non-existent atoms.  So there is some trouble that lies in this path.

The problem is that, when an atom dtor runs, its too late to clean up the incoming sets; the dtor has already destroyed too much stuff.  So the incoming-set cleanup has to happen earlier, which is why that stupid extract step is done.  Its kind of ugly and all.

It would be good it you added some sort of warning/todo about this in the above code, as otherwise it will be discovered years later, and will be a bit of a puzzle to understand.

Issue #658, opened 2016-02-22
cog-bind-single should return first N matches


linas:
The `cog-bind-single` call should be modified to compute and return the first N matches.  There also needs to be a first-n version of `cog-satisfying-set` as well.

This is a clone of issue opencog/opencog#1586

linas:
closing as fixed,  thanks @jswiergo

PR #681, opened 2016-02-28
#658 cog-bind-first-n, cog-satisfying-set-first-n

jswiergo:
I have implemented functions that return first N matches. Scheme bindings are ready. I plan to add python bindings later. Unit tests contain example calls.

If you have better ideas for names or signatures of this functions let me know, I will refactor.
Not sure if previous implementation of bind-single will be still useful or should be removed, because new functions are more general. If so, I can remove the older version.

### DIFF
--- a/opencog/query/PatternSCM.cc
+++ b/opencog/query/PatternSCM.cc
@@ -69,7 +69,7 @@ void PatternSCM::init(void)
 	// BindLink containing variables, a pattern and a rewrite rules.
 	_binders.push_back(new FunctionWrap(bindlink, "cog-bind", "query"));
 
-	// Identical to do_bindlink above, except that it only returns the
+	// Identical to bindlink above, except that it only returns the
 	// first match.
 	_binders.push_back(new FunctionWrap(single_bindlink,
 	                   "cog-bind-single", "query"));
--- a/opencog/query/PatternSCM.cc
+++ b/opencog/query/PatternSCM.cc
@@ -74,6 +74,12 @@ void PatternSCM::init(void)
 	_binders.push_back(new FunctionWrap(single_bindlink,
 	                   "cog-bind-single", "query"));
 
+	// Identical to bindlink above, except that it only returns the
+	// first N matches, assuming that N is the first argument and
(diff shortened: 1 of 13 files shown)


opencog/query/PatternSCM.cc
```diff
@@ -69,11 +69,17 @@ void PatternSCM::init(void)
 	// BindLink containing variables, a pattern and a rewrite rules.
 	_binders.push_back(new FunctionWrap(bindlink, "cog-bind", "query"));
 
-	// Identical to do_bindlink above, except that it only returns the
+	// Identical to bindlink above, except that it only returns the
 	// first match.
 	_binders.push_back(new FunctionWrap(single_bindlink,
 	                   "cog-bind-single", "query"));
```
linas: as you mention elsewhere, the code for this should be ripped out and replaced by the one-liner
```
(define-public (cog-bind-single atom) (cog-bind-first-n 1 atom))
```
in query.scm

opencog/query/PatternSCM.cc
```diff
@@ -69,11 +69,17 @@ void PatternSCM::init(void)
...
 
+	// Identical to bindlink above, except that it only returns the
+	// first N matches, assuming that N is the first argument and
+	// the second is a BindLink handle.
+	_binders.push_back(new FunctionWrap(first_n_bindlink,
+	                   "cog-bind-first-n", "query"));
+
```
linas: Well, and it probably makes sense to also remove the code for `cog-bind` as well, and replace it by
```
(define-public (cog-bind atom) (cog-bind-first-n -1 atom))
```

Notice that I used `-1` in the above instead of unsigned-max.  The reason for this is that I fear bugs due to some crazy mis-match between the C++ UNSIGNED_MAX, and the scheme version of the same. But maybe this is an old-fashioned way of thinking; I should just trust the modern-day systems to do the right thing...

Figuring out unsigned-max in guile might be hard -- its normally arbitrary precision; it has no max....

Your pick in how to handle this. My goal is to minimize the total amount of code; the fewer lines of code the better -- the less confusing things will be.

PR #683, opened 2016-03-02
Refactoring of cog-bind-* functions

jswiergo:
I have added changes suggested in review #681 and changed BindLink API to more compact version.


linas:
Looks excellent to me!  Many Thanks!

PR #685, opened 2016-03-03
Add cog-bind-first-n python bindings

jswiergo:
Python bindings for functions introduced recently. This the last thing solving the issue #658.


linas:
test this please

PR #690, opened 2016-03-05
Phase I - PostgreSQL link fix

inflector:
This phase includes refactoring necessary to support:

1) Separate storage objects for ODBC and PostgreSQL drivers in use at the same time.
2) Separate guile access to both storage systems.
3) Direct use of `libpq` PostgreSQL library (to bypass ODBC)

Significant changes were:

1) Created abstract superclass `AtomStorage`
2) Renamed `AtomStorage` to `ODBCAtomStorage` which now inherits from `AtomStorage`
3) Created new `PGAtomStorage` which also inherits from `AtomStorage`
4) Added helper routines to `get_all_atoms`, `get_all_nodes`, and `get_all_links` from an `AtomSpace`

And testing enhancements to support more comprehensive testing of the storage subsystems:

1) Pulled the random node and link generation out into a separate class, `RandomAtomGenerator`,  for reuse.
2) Implemented comparison operators and function for `AtomSpace` comparisons in testing.
3) New standalone executable `test_pg_storage` to test PostgreSQL storage system.

NOTE: This phase does not yet handle the bug itself. That's coming in phase II soon. `PGAtomStorage` also is largely a copy of `ODBCAtomStorage` for this pull request, i.e. it still uses the same ODBC library and calls to write out the atoms.

I wanted to submit a pull request for these changes separately since there are quite a few changes that shouldn't affect the semantics of AtomSpace at all in this set. That way the next phase will be mostly confined to changes to `PGAtomStorage` itself.

@ngeiswei - Note above the abstraction of random atom generation into it's own class, `RandomAtomGenerator`,  if you still have a need for this sort of thing for benchmarking in the future. I too needed it for benchmarking at reasonable size and testing that database backing stores were functioning properly.


inflector:
@AmeBel Jenkins fails this pull request because it can't find the PostgreSQL main header: `libpq-fe.h`

I'm not getting this error but I manually installed the PostgreSQL developer package, `libpq-dev`, at some point, so I suspect that the docker container image that Jenkins uses doesn't install `libpq-dev`. 

Does Jenkins use `octool` for package installs, so the packages here in PACKAGES_BUILD:

https://github.com/opencog/ocpkg/blob/master/octool#L26

define what gets installed?

If so, then `libpq-dev ` should be defined right before or after `unixodbc-dev`. We'll want to support both methods, i.e. using the current ODBC and the direct to PostgreSQL library approach for a while.

If not, where are the dependencies defined for the Jenkins docker containers for testing?

What's the current practice for getting packages added to this list? Does Jenkins pick up the latest version of `octool` off github using wget before installing dependencies? Do I submit a pull request for the ocpckg repository?

bgoertzel:
you may want to ask @wenwei-dev  about Jenkins stuff as well, he set up the Jenkins stuff for OpenCog ...

wenwei-dev:
@inflector Jenkins pulls the docker image that built by octool to run the tests. So if there is new dependency, you need to update the octool, and this I think will trigger the docker image building chain. So after several hours the docker images are updated, you can ask Jenkins to retest this PR.

AmeBel:
test this please

AmeBel:
https://github.com/opencog/ocpkg/pull/57 adds the dependency you requested. And the docker images are updated. But there is a cmake error

AmeBel:
test this please

inflector:
@AmeBel The gnu sasl library, `gsasl`, also needs to be added as a dependency.

@wenwei-dev If you can point me to the Jenkins script or the command for the exact docker image that Jenkins pulls, I can test this on my end.

inflector:
I've updated FindPGSQL.cmake to look for the specific library in libgsasl7, `libgsasl.so.7`, and have updated the error processing so that it will continue to build if that library is missing, it just won't build the PostgreSQL persistence system. So it won't be a fatal error if the library isn't installed.

Test this please.

inflector:
@AmeBel The tests pass because the CMake files now properly handle a missing Gnu SASL library.

However, we still need to add the dependency for `libgsasl7` or the code for the non-ODBC PostgreSQL persistence won't be built.

linas:
re: libpq-fe and SASL: there should be a file called `lib/FindPostgres.cmake` similar to the other Find*.cmake files in that directory. If postgres (and it's pre-req's e.g. SASL) is not found, then that subsystem is not built. 

Chances are very high that a `FindPostgres.cmake` already exists, and some clever googling should be able to find it/them. Then just copy that to the `lib` directory, and then just add `IF (HAVE_POSTGRES)` to the Cmakefiles.

inflector:
@linas - I had already made one. I called it FindPGSQL.cmake. There was a bug in it that I fixed when the library was missing.

After the last couple of checkins, the subsystem is now not built when the required libraries are missing.

linas:
To avoid confusion, it might be good to have two directories: `opencog/persist/odbc-sql` and  `opencog/persist/pgsql` 

Note also that the now-obsolete `hypertable` directory also contains SQL-like code; this is because the hypertable interfaces resemble SQL.  Thus, if we ever wanted to make this code work again, it could -should be rejiggered to inherit from this AtomStorage class.

Hmm. Looking at the existing hypertable code, its .. radically crappy. Oh well.

opencog/atoms/base/Atom.cc
```diff
@@ -227,6 +228,21 @@ void Atom::markForRemoval(void)
...
+void Atom::setChecked(void)
+{
+    _flags |= CHECKED;
+}
+
+void Atom::setUnchecked(void)
+{
+    _flags &= ~CHECKED;
+}
+
```
linas: Ewww

opencog/atomspace/AtomSpace.cc
```diff
@@ -99,6 +99,156 @@ AtomSpace& AtomSpace::operator=(const AtomSpace&)
          "AtomSpace - Cannot copy an object of this class");
 }
 
+bool AtomSpace::compare_atomspaces(const AtomSpace& space_first,
+                                   const AtomSpace& space_second,
+                                   bool check_truth_values,
+                                   bool emit_diagnostics)
+{
```
linas: Shouldn't this be a part of some unit test, instead of being here, in the main namespace?

opencog/atomspace/AtomSpace.h
```diff
@@ -358,6 +374,57 @@ class AtomSpace
...
+     * Example of call to this method:
+     * @code
+     *         std::list<Handle> atoms;
+     *         atomSpace.get_all_atoms(atoms);
+     * @endcode
+     */
+    void get_all_atoms(HandleSeq& appendToHandles) const
+    {
+        // Defer to handles by type for ATOMs and subclasses.
+        get_handles_by_type(appendToHandles, ATOM, true);
```
linas: I really really don't like one-line subroutines.   If something takes one line of code to implement, we should not be wrapping it in a wrapper.  It does not save any typing, and it does make it harder to understand the code.

opencog/atomspace/BackingStore.cc
```diff
@@ -46,3 +48,13 @@ bool BackingStore::ignoreAtom(Handle h) const
 	return false;
 }
 
+void BackingStore::registerWith(AtomSpace* atomspace)
+{
+	atomspace->registerBackingStore(this);
```
linas: again -- single-line implementations of methods.  Why? this only makes the code harder to understand, and doesn't save any typing... is this some weird work-around for public vs. private method calls??  Maybe the private methods should not be private, if this work-around is needed?

opencog/persist/sql/AtomStorage.h
```diff
@@ -49,154 +41,54 @@ namespace opencog
...
+		static void reserveMaxUUID(unsigned long maxUUID)
+			{ TLB::reserve_upto(maxUUID); }
+		static unsigned long getMaxUUID()
+			{ return TLB::getMaxUUID(); }
+		
+		// For accessing Atom through friend relationship in subclasses.
+		static void setAtomUUID(AtomPtr atom, UUID newUUID)
+			{ atom->setUUID(newUUID); }
+		static AtomTable* getAtomTable(AtomPtr atom)
+			{ return atom->getAtomTable(); }
```
linas: single-line implementations .. why?

opencog/persist/sql/test_pg_storage.cc
```diff
@@ -0,0 +1,492 @@
...
+		fprintf(stdout," NO     grouping\n");
+	fprintf(stdout, "Starting inserts ");
+	print_time();
+	
+	// Get the starting times.
+	std::chrono::steady_clock::time_point begin = std::chrono::steady_clock::now();
+	clock_t start_time = clock() ;
+
+	// Generate the insert part.
+	snprintf(insert_part, INSERT_BUFFER_SIZE, "INSERT INTO Atoms "
```
linas: The very newest postgres version 9.5 has a combined INSERT/UPDATE statement:  `UPSERT` -- actually `ON CONFLICT DO UPDATE`  http://www.postgresql.org/docs/9.5/static/sql-insert.html

http://www.craigkerstiens.com/2015/05/08/upsert-lands-in-postgres-9.5/

https://www.compose.io/articles/postgresqls-future-is-looking-upsert/

Using this would be nicer than the current select craziness, but postgres 9,.5 is not available in trusty 14.04 so that makes it kind-of unusable for the short term.

linas:
I'm gonna merge right now .. can I get you to reconsider the one-line-long methods for some future update?  And also moving the compare-atomspace code to some unit-test directory?

opencog/atomspace/AtomSpace.h
inflector: These were added because there are a lot of examples of inefficient usage of the other methods, for example, ones that pass in a back_inserter rather than using the more efficient pass of the list directly which can then use the code which allocates the storage all at once.

For example:
```
atomSpace->get_handles_by_type(back_inserter(links), LINK, true);
```
would be much faster as:
```
atomSpace->get_handles_by_type(links, LINK, true);
```
because the `AtomSpace::get_handles_by_type` does a reserve on the entire size all at once before creating it's own back_inserter and passing that into `AtomTable::get_handles_by_type`. This was several times faster in my testing last year, if I recall properly. I planned on going back through the code to remove the spurious back_iterators to replace these as a separate step.

The new routines I made don't take iterators, they only take the `HandleSeq&` which means that future users won't be tempted to use the slower way out of ignorance. 

Further, the vast majority of `get_handles_by_type` code looks only for `ATOM`s, `NODE`s or `LINK`s. So by separating the common ones from more specific usages, the methods can now be optimized separately. For example, `get_all_atoms` can iterate directly from AtomTable::_atom_set and bypass the type-specific iterator used by `AtomTable::get_handles_by_type`.

opencog/atomspace/BackingStore.cc
inflector: The alternative was to add the subclasses of AtomStorage as friends to all the places where AtomStorage was added, Atom and TLB etc.

Subclasses of friends can't see the same thing that the friend class can see, due to the way that C++ defines friends. So by adding protected accessors to AtomStorage the subclasses can now do what needs to be done without having to be explicitly added as friends.

This isn't so important with a couple of classes, but I suspect there will be future MySQL and SQLite classes as well.

opencog/persist/sql/AtomStorage.h
inflector: Same friend / subclass issue here.

inflector:
I can certainly move the comparisons to a test utilities and I'm happy to remove the one-liners if you don't think they are useful or would prefer to change the access semantics via adding explicit friends.

opencog/atomspace/AtomSpace.h
linas: OK, realize also that a lot of code has suffered through multiple backwards-compat re-writes, and so as the API changed, the code that uses it did not. As a result, it tends to look as if its badly written.  Some of it was, from the very begining; the worst offenders have now been removed; they were in the embodiment module.  So when you see poor code, there's often a historical reason for that.

opencog/atoms/base/Atom.cc
inflector: Not having a way to mark the atoms means you'd have to search for atoms in both directions for AtomSpace equality checking. For atomspaces A and B you'd have to make sure each atom in A was in B, and then check to see that each atom in B was also in A. By setting a flag, you only need to make sure that every atom in B has the flag set when you are done.

I thought a general-purpose marking method would be useful for other test generation cases. And we already have `_flags` being used for removal, and it looked like other temporary uses in the past. So it doesn't add storage space to Atoms.

I can certainly take this out if you really don't like it.

opencog/atomspace/BackingStore.cc
linas: Hmm OK

opencog/persist/sql/test_pg_storage.cc
AmeBel: A repo could be added, https://wiki.postgresql.org/wiki/Apt

inflector:
And I'll create separate directories for the different cases: ODBC and PostgreSQL.

opencog/persist/sql/test_pg_storage.cc
inflector: I've installed 9.5 on 14.04 using that method, it's not too bad, but I don't know how easy it is to addd repos to `octool`.

linas:
The atomspace comparison function kind-of gets on my nerves.  It seems both to not be generally useful, except for unit testing, and it also seems ill-defined: if two atomspaces are connected to the same backend, aren't they by definition equivalent? The fact that one has cached atoms that the other doesn't shouldn't matter.

what if two atomspaces have the same atoms, but have managed to issue different UUID's for them? this can play havoc for storage solutions. 

what if I create a child atomspace, but don't add any atoms to it. Should it be considered to be equivalent to its parent, or should it be considered distinct?

What if there are two children, both from the same parent, both had the same atom added to them, but due to mechanics, that same atom was issued distinct different UUID's. Is it really the "same" atom? Because changing the TV on one does not change the TV on the other, and they won't/can't be unified until they are both pushed into the same backend ... 

stuff like that.  There's crazy grey areas in there, both incomplete, undefined semantics, which we haven't figured out because no one is doing anything quite that crazy, yet.

opencog/atomspace/AtomSpace.h
inflector: Oh, I figured it was historical. 

I myself should have gone through the code myself last year and removed the no-longer-necessary `back_iterator` wrappers when I added the optimization to `reserve` only once and made the non-iterator method much faster for large return sizes.

inflector:
I don't see any uses outside of testing either. So moving it to another directory is a good call.

opencog/atomspace/AtomSpace.cc
inflector: Moving this somewhere else makes sense.

We don't really have a test-specfic utilities directory. Should I create called `testutils`? 

Or would you prefer that I just create a file TestUtilities.h and place that in `atomspaceutils` like I did with `RandomAtomGenerator`?

inflector:
Retest this please.

AmeBel:
@inflector I am getting
```
/usr/bin/ld: cannot find -lldap
/usr/bin/ld: cannot find -lkrb5
/usr/bin/ld: cannot find -lgssapi_krb5
collect2: error: ld returned 1 exit status
make[2]: *** [opencog/persist/sql/libpersist-pgsql.so] Error 1
make[1]: *** [opencog/persist/sql/CMakeFiles/persist-pgsql.dir/all] Error 2

```
what are the full list of required dependencies?

inflector:
@AmeBel I didn't have to install anything to get those so I assumed they were part of the normal linux libraries that came with the container. But it's very likely I built using a different dockerfile than you did. I built my images using the one that has since been removed from the main OpenCog repository. The only one I had to add explicitly was sasl.

Those libraries are part of LDAP, and Kerberos (with associated GSSAPI). They're pretty common in lots of networking applications.

They're found in these packages:

libldap2-dev - contains libldap.so
krb5-multidev - contains both libkrb5.so and libgssapi_krb5.so

I installed Postgres 9.5 a while back too, so it may be that they added those libraries as part of the dependencies for 9.5 that aren't part of the Postgres 9.3 that gets installed as part of OpenCog's normal dependencies.

That's why I've been asking about the docker images that Jenkins uses so I can see what comes with a clean container that Jenkins builds. Then I can run octool myself and make sure all the dependencies are accounted for.

AmeBel:
The Dockerfiles used for the images are found @ https://github.com/opencog/docker/tree/master/opencog, autobuilt and hosted @ https://hub.docker.com/u/opencog/ and, the command for running the test are https://github.com/opencog/atomspace/blob/master/circle.yml#L11 and https://github.com/opencog/atomspace/blob/master/circle.yml#L16

opencog/atomspace/AtomSpace.cc
linas: Well, it is not needed by multiple unit tests in multiple directories .. so, for now, put it in `/tests/persist/sql` and link it into those tests that need it.  If, for some reason, you have tests in `/tests/atomspace` tat need it, then maybe something fancier is needed. I dunno. Keep it simple.

inflector:
Thanks @AmeBel. I found two problems:

1) The package  `libldap2-dev` was not installed. It needs to be added to `octools`.

2) The package `krb5-multidev` was already installed but in a location that the linker didn't find. So I updated the FindPGSQL.cmake script to look in the location where the package installs the library. This update is reflected here: https://github.com/opencog/atomspace/pull/691

To get `test_pg_storage` to build, you need to add `libldap2-dev` to `octools` as a required dependency.

Once this is done, the pull request branch https://github.com/opencog/atomspace/pull/691 should find PGSQL and build `test_pg_storage` successfully on Jenkins.

PR #691, opened 2016-03-07
Updated FindPGSQL to look for Kerberos

inflector:
This is needed to fix the link problem outlined by @AmeBel in the comments for https://github.com/opencog/atomspace/pull/690

It appears that octools already installed `krb5-multidev` but the package puts it in a non-standard library location:
```
/usr/lib/x86_64-linux-gnu/mit-krb5/
```
FindPGSQL now looks in that location and also checks to see that those libraries are installed there.


linas:
"This branch is out-of-date with the base branch" -- you have to hit the "merge" button.

inflector:
(Copied from https://github.com/opencog/atomspace/pull/690)
I found two problems:

1) The package libldap2-dev was not installed. It needs to be added to octools.

2) The package krb5-multidev was already installed but in a location that the linker didn't find. So I updated the FindPGSQL.cmake script to look in the location where the package installs the library. This update is reflected here: #691

To get test_pg_storage to build, you need to add libldap2-dev to octools as a required dependency.

Once this is done, the pull request branch #691 should find PGSQL and build test_pg_storage successfully on Jenkins.

linas:
test this please

inflector:
The failure of this build will be fixed only when `libldap2-dev` is added to `octools` and the changed docker images have propagated through so Jenkins finds them.

inflector:
Test this please.

linas:
test this please

inflector:
Test this please.

inflector:
It sure does take a long time for the docker images to update. It's still using the old ones from yesterday.

AmeBel:
it will fail but got upated 10 minutes ago, missed another, missed `libkrb5-dev` for a dependancy.

inflector:
@AmeBel I didn't have to do an `apt-get install` on `libkrb5-dev` or even  `krb5-multidev` as the libraries I needed were already installed on the opencog/cogutils image used by Jenkins. The only package I needed to install was `libldap2-dev`.

What makes you say that we need `libkrb5-dev`?

AmeBel:
after rebulding opencog/opencog-deps, trying a clean build failed.

inflector:
The image opencog/cogutils already had  `krb5-multidev` installed. That's why I changed the FindPGSQL script so it can find it.

Did you build the branch from this pull request for your clean build? Otherwise it will fail, not because the library is missing, but because the linker can't find it.

I should also add the directory where `libkrb5-dev` places it as I just noticed they install the libraries in a different location for that package than they do for `krb5-multidev`.

AmeBel:
I was testing on master, retesting using this pr

inflector:
To be more explicit: both `libkrb5-dev` and `krb5-multidev` have the required link libraries: `libkrb5.so` and `libgssapi_krb5.so`. 

`krb5-multidev` was already installed. The problem was that my script didn't look in the right place. This pull request fixes that problem.

So even if you have the packages which have the libraries installed, you need the changes from this pull request or make won't know where to find the libraries.

AmeBel:
All is good on my end too. One thing though any reason `test_pg_storage` isn't part of the test-suite?

inflector:
Yay! Just verified that `PGSQL` was found and so the `test_pg_storage` that actually has the `libpq` code in it linked successfully.

The executable `test_pg_storage` is designed to do a lot more intensive benchmarking and such. More like `benchmark_bm` than a unit test. The fix / changes  are also only partly implemented.

I'll move some of the  guts into a new test suite when I've got things working.

AmeBel:
Got it, :smile:

PR #699, opened 2016-03-11
Postgres link fix - Phase II - Now handles links of arbitrary size

inflector:
**Phase II** notes:

- There is an open bug #699 for a degenerate fetch of links where the height is more than 15 or so.

- The changes to move some code around suggested in comments to Phase I have only been partially implemented to keep this pull request smaller. In particular, I haven't moved the code of our AtomSpace to the test folders. I wanted to check in the code that works to fix the open bug before doing those changes. The changes for this pull request are extensive enough already.

- I've not done any performance optimizations for loading the Atoms table. Only for loading edges. The single atom fetch case can be improved even for non-degenerate links.

- All the code still uses the ODBC except for some parts of the test harness. I wanted to fix the bug and do the reword before moving to the libpq library that way I'll be able to refactor and use this newer code for both ODBC and Postgres with only a few database specific items in each of the respective library-specific directories and classes.

**Phase III** - will be a brief cleanup

**Phase IV** - will finish up with libpq


inflector:
The tests are failing because there is a new table Edges and a couple of columns that need to be added to Atoms for the test PostgreSQL database `opencog_test`. The schema in `pg_atom.sql` is compatible with both the ODBC (aka current) way of handling edges and the new PostgreSQL handling in the edges table. If we use pg_atom.sql for the test database, it will work for both. That's how I have things setup here.

I've kept both ways working so I could test back and forth easily during the same tests, so even PGAtomStorage can use the old way if you set the flag `_save_edges` to `false` with `setStoreEdges()`.

@leungmanhin Are you in charge of the test database scheme for the Jenkins tests?

AmeBel:
retest this please

AmeBel:
https://github.com/opencog/docker/pull/88 updated to `pg_atom.sql`. Testing locally on master works

linas:
So .. is this ready to be merged, or not?

inflector:
retest this please

inflector:
@AmeBel @leungmanhin This test is still not putting the out_hash column into the testing database. The database is missing the `out_hash` column for the table `Atoms` which is causing two failures. 

@AmeBel - You updated the script to use pg_atoms.sql but that script doesn't appear to be being applied to the test database. Do something else need to be refreshed? Updated?

AmeBel:
you have made  changes to pg_atom.sql that isn't part of the `opencog/postgres` image. I guess a new image which have everything installed within a single image has to be created for testing purposes or the `opencog/postgres` image has to be modified so as to deal with such situations. I think the first option is easier. @wenwei-dev any ideas?

AmeBel:
retest this please

AmeBel:
ok I disabled the postgres test, b/c this change does not only change in `pg_atom.sql`  but also move it to a differnet directory. I will enable it once this is merged. If you want all  to be configured then you have to deal with it yourself or @wenwei-dev could deal with this. 

I have tested it locally with postgres enabled and it passes.

FYI, for fast paced development a dev branch has been created a while back.

PR #700, opened 2016-03-12
Fix rule-engine installation.

linas:


linas:
This undoes some craziness related to bug #508

PR #703, opened 2016-03-14
Update the guile module loading code.

linas:
This should remove any remaining confusions about having to load "utilities.scm" - it should no longer be required.


linas:
test this please

linas:
test this please

linas:
Can someone who knows jenkins please explain why this is failing?  Please merge this, if possible.

linas:
test this please

linas:
I beleive that this should resolve issue #508

wenwei-dev:
retest this please

AmeBel:
I think the failure is because atomspace isn't installed and doesn't find the scheme files. The recurring output is,
```
Backtrace:
In ice-9/boot-9.scm:
 157: 1 [catch #t #<catch-closure 13d73c0> ...]
In unknown file:
   ?: 0 [apply-smob/1 #<catch-closure 13d73c0>]

ERROR: In procedure apply-smob/1:
ERROR: In procedure primitive-load-path: Unable to find file "core_types.scm" in load path
```

linas:
why would the atomspace not be installed?

linas:
I looked and don't see any obvious errors.  Is there an error message that precedes this, and shows why the install failed?  I can't really guess the problem from that one error message; there must be some earlier error message, too!?

AmeBel:
You can click on 'show all checks -> details' next to 'some checks have failed' message , authenticate with github and see the console outputs and the test logs.

Here is the log http://61.92.69.43:8080/job/ci-atomspace/415/artifact/build/tests/Testing/Temporary/LastTest.log

AmeBel:
On why the atomspace isn't installed, that is because a typical workflow is to go through 'change -> build -> test ' cycle instead of 'change-> build -> install -> test' cycle.

linas:
My browser window doesn't have that button(s).

One must ALWAYS install before testing; the unit tests won't work unless they have an install step before them!  Its been like that since forever!?    I mean, its impossible for the unit tests to even come close to not crashing, if you skip the install step.

linas:
I don't have permissions to view http://61.92.69.43:8080/job/ci-atomspace/415/artifact/build/tests/Testing/Temporary/LastTest.log

AmeBel:
The log https://gist.github.com/AmeBel/e421bcf80618764fe28f

pennachin:
Wait, I don't think one shouldn't have to install a build in order to check if it passes the tests, that seem like the wrong order. One should be able to run `make test` before `make install`, and this is how the build has worked for many months...

linas:
well, except that, for the last 8 years, you weren't able to run the tests unless you installed first. Currently, its a mandatory pre-req.   You can change that, but there's a pile of unit tests that would need to be refactored.

wenwei-dev:
@linas Have you tried to login with Github account? That is required to use Jenkins web. If you still can't see the log, let me know.

pennachin:
@linas the master branch runs `make test` without running `make install` first, and has done so for a while. This is the sequence implied by the README. The unit tests run ok.

inflector:
>I don't think one shouldn't have to install a build in order to check if
it passes the tests.

Installing into the container for the test is part of the test process
itself. Defining the order to run the steps in order to simulate the
desired tested action. Installing is a prerequisite of running the test
itself.

pennachin:
> Installing into the container for the test is part of the test process
itself. Defining the order to run the steps in order to simulate the
desired tested action. Installing is a prerequisite of running the test
itself.

This shouldn't be a requisite. One should be able to run tests without installing a potentially broken build. Of course this doesn't matter on ephemeral containers, but it matters for people not using containers.

I also don't understand why you and Linas say it currently is a requisite. The test process used by Jenkins does not include installing before testing, just check `octool`. I have run the tests by hand many times without installing and I seriously doubt I'm the only one.

inflector:
There are two distinct "running tests".

1) Running of tests by the Jenkins machine to test the code that is
potentially broken, and thus something that should not be installed.

In this case, for example, if you have an updated cogutils and that the
AtomSpace code changes requires, and you don't first install the new
cogutils after building it, the test will ALWAYS fail. There are many cases
where there are changes we make to AtomSpace that have corresponding
changes in OpenCog, if you don't install the AtomSpace changes, then
OpenCog won't build.  When you are testing three components like we do,
then you need to build the two prerequisite components for OpenCog (namely
AtomSpace and CogUtils) first. An automated script needs to install the
dependent components first before testing OpenCog. This is a linear
pass-fail process.

2) Running of tests by users of OpenCog. As an individual, you build and
test and then install OpenCog itself, yes. The same with AtomSpace, etc.
This is an interactive and iterative process.

pennachin:
@inflector Of course you have to install dependencies (cogutils to test atomspace, both to test opencog, etc), the point is that you shouldn't have to install atomspace to test atomspace, no matter the scenario (Jenkins, developers running it by hand, etc)

linas:
Cassio, your creating a new requirement where one never existed before.  Install has been a requirement since about forever, now you want to run tests without installing. This requires extra work from someone -- someone has to crawl through all of the unit tests, and figure out how to fix each one that breaks.  Who is going to do that? Who has the spare time for this?  If you have a warm body to assign to this task, then assign it.   But otherwise, its just a bunch of make-work.

inflector:
@cassio Tests are users of the libraries being tested... So the AtomSpace
tests run tests as AtomSpace as users would test it. That's why there have
been two separate make commands. So you do a:

make && make install && make test

to test the atomspace, because you are testing that the install itself puts
the files in the right place so that the tests can find them acting as
users of AtomSpace would do. Tests are in user-of-library space because
that's the tests that matter to users.

Suppose the install parts of the makefiles didn't work? Perhaps some file
wasn't install properly, or not placed into the proper CMakefile.txt file,
you want the tests to pick that up. So install must be part of the tested
process.

If you add a file dependency that requires a new install line, how are we
to know that work without the current process?

pennachin:
@linas as I keep saying, the master branch has had the unit tests pass without installing for a long time. So if it wasn't explicitly discussed before that unit tests should be run without installing, it was assumed in practice by Amen and myself. Aside from this PR, there is no work that's required to get us to follow the common practice of testing before installing.

@inflector I see your point about testing the actual install, but unit tests aren't release tests, they're development tests. If we want to explicitly test the install process, we should have a (hopefully much smaller) set of release tests. Right now this is done indirectly (and inadequately, of course, I'm not proposing it as sufficient) by the `opencog` build and tests, which require `atomspace` to be installed.

pennachin:
@inflector On the topic of testing the release, this is what the automake docs suggest:

https://www.gnu.org/software/automake/manual/html_node/Checking-the-Distribution.html#Checking-the-Distribution

and the suggested target for post-install tests, which makes it clear one is testing the installation, not the code:

https://www.gnu.org/software/automake/manual/html_node/Install-Tests.html#Install-Tests

linas:
Cassio, what you are saying is simply not true. That's the problem. The `make install` step has been a pre-requisite for maybe five years now, and it is more or less impossible that the unit tests ever could have passed without the install step running first. There are multiple reasons for this. One is that the header file hierarchy is fairly garbled; the other is that, during the transition to the scheme modules, and away from just raw scheme files, the way that the files got loaded got very garbled, leading to all sorts of really crazy and wild and hard-to-debug problems.  Issue #508 is the tip of the iceberg on all of the things that were going wrong.  Ask Amen, ask William Pa, they suffered through all the file-loading and module craziness.  I'm trying to end this, and return to a semblance of normal operations, where the code is dependable and works well, and doesn't suffer from fragile breakdowns due to some muffed file load.

This pull request does some of the last remaining cleanup to get things fully modularized, and stop the file-loading insanity. To suddenly walk in and create a brand new requirement where none existed before is just plain obstructionism.    This is the umpteenth time where you are asserting your top-dog, alpha-male personality, and every time, it just leads to arguments, damages the project, and pisses everyone off.  You need to figure out why you have to dominate every discussion, and figure out some less expensive way to interact.

pennachin:
> Cassio, what you are saying is simply not true. That's the problem. The make install step has been a pre-requisite for maybe five years now, and it is more or less impossible that the unit tests ever could have passed without the install step running first. 

This is very puzzling. I have just spun up a cogutils container (which uses `ocpkg` to install all dependencies and `cogutils`, but NOT `atomspace`). Then, using the current master HEAD, I did:

```
cd /atomspace
mkdir build
cd build
cmake ..
make -j4 
make test
sudo make install
make test
```

and got identical test execution outputs from before and after installing. Test runtimes are also nearly identical, which suggest the same tests being run.

What am I doing wrong in the above? Did you try running these tests on a fresh VM or container recently? 

> This pull request does some of the last remaining cleanup to get things fully modularized, and stop the file-loading insanity. 

And I agree this is a good thing.

> To suddenly walk in and create a brand new requirement where none existed before is just plain obstructionism. 

I am not doing that. Nowhere in the build documents or READMEs it says that there is a requirement to install before running tests. This PR discussion is the first time I hear of the requirement that unit tests can only be run after installing the build. Since neither Amen nor I knew about this, every PR merged since September has run the unit tests without installing the build first.

AmeBel:
FYI, the configuration for running continuese integration tests was inhertied from the time of buidbot. I am new to this stuff so I don't know what the "right" thing is. So, during subsequent migration to circle-ci and now jenkins that configuration has remained. The configuration structure being "build -> test " instead of "build -> install -> test". The most recent buildbot conf file is [this](https://github.com/opencog/docker/blob/master/buildbot/master/master.cfg) (presently only used for building doxygen docs hosted at docs.opencog.org) and the oldest from 5 years ago is [this](https://github.com/opencog/atomspace/blob/master/lib/buildbot-master.cfg).

PR #710, opened 2016-03-18
Minor bc hack

ngeiswei:
Set of minor hacks and simplifications in the backward chainer.


ngeiswei:
I've also added some code to increase determinism (optionally of course). And better debugging in gdb. I'm updating the wiki and I'll send emails on the list about that.

PR #711, opened 2016-03-18
Remove assorted pointless LinkCasts

linas:
Turns out that LinkCast was a performance killer.  We don't need it, and using it makes the code more complicated. So, remove them.


linas:
The diffs for this pull request have been garbled beyond recognition due to the circle-ci merge craziness.  This used to be a very simple and small patch set when it started.  Oh well.

linas:
This pull request was originally supposed to consist of only 3 patches,  however, after jenkins/circle-ci got done with it, it turned into dozens.  Originally, the first patch was this:  9c8ddb9af9f8a36642de8f4ce6a444f62233a032 everything before this commit belongs to some other pull request.   The second patch was this: 218bc7646559d9fd8f64e96024c366e956694194  The third part was supposed to be this: 218bc7646559d9fd8f64e96024c366e956694194  no other patches are meant to be a part of this pull request.

Why do I need to waste my time to do this? This should be automated.  This is fucked up.

Issue #705, opened 2016-03-17
cog-execute! is mangling BindLinks


linas:
Spin-off from bug #704 -- cog-execute, when given a BindLink, manges the bindlink. Example:

```
(use-modules (opencog))
(use-modules (opencog exec))
(use-modules (opencog query))

(define z-dsn (DefinedSchemaNode "OpenPsi: Energy-action-rule-Default"))
(DefineLink
   z-dsn
   (BindLink
      (VariableList
         (TypedVariableLink
            (VariableNode "Demand" (stv 0.96491325 0))
            (TypeNode "ConceptNode")
         )
      )
      (AndLink
         (DefinedPredicateNode "OpenPsi: demand-pattern")
         (EqualLink
            (VariableNode "Demand" (stv 0.96491325 0))
            (ConceptNode "OpenPsi: Energy" (stv 0.66960984 0.99999982))
         )
      )
      (ExecutionOutputLink
         (GroundedSchemaNode "scm: foo")
         (ListLink
            (VariableNode "Demand" (stv 0.96491325 0))
         )
      )
   )
)

(define (foo node) node)
```

Then  `(cog-execute! z-dsn)`  returns
```
(BindLink
   (VariableList
      (TypedVariableLink
         (VariableNode "Demand")
         (TypeNode "ConceptNode")
      )
   )
   (AndLink
      (DefinedPredicateNode "OpenPsi: demand-pattern")
      (EqualLink
         (VariableNode "Demand" )
         (ConceptNode "OpenPsi: Energy")
      )
   )
   (VariableNode "Demand" )
)
```
which is more or less junk.   This is related to bug #686

linas:
easier said than done. Doing this promptly elicits infinite recursion in the two unit tests that check for it.  It also sends the forward chainer into infinite recursion, as well.

linas:
pull request #713 fixes this.  After loading the above, and the other code from #704, the following output is generated:
```
(cog-execute! z-dsn)
$11 = (SetLink
   (ConceptNode "OpenPsi: Energy" (stv 0.66960984 0.99999982))
)
```

which I beleive is correct, given the lack of a no-exec link type.

linas:
Fixed in pull req #713

PR #712, opened 2016-03-18
Scope Links scope the variables they declare.

linas:
Therefore, do not perform beta-reduction on the hidden (alpha-hidden, alpha-converted) variables.

PR #713, opened 2016-03-19
Bug-fix for issue #705

linas:
This wraps up changes needed to fix bug #705


linas:
Due to the requirements of circle-ci testing, this pull request, which was a very simple and straight-forward set of patches,  ended up mangling together four other pull requests into a tangled mess: #712 #711 #708 and #707

These other pull reqs have clean development lines, individually, but these cannot be preserved when there are more than one or two pending pull requests.  This is another example of everything wrong with circle-ci methodology: by design, it is unable to maintain a clean history of the work being done, because you have to keep clicking the "grey Merge" button over and over, so that the testing can advance to the next level.  Here, there were 5 pull requests, each neatly stacked, one after the other.  The result of the repeated button clicks is like a felt wool carpet, with the threads intertwined every-which way.

I feel like this is madness, and is the wrong way to develop code, and is clearly not scalable even when there is only just one developer. What happens if there are multiple developers committing code daily?  Its like you may as well not use source control at all.  Circle-ci negates everything that is good about source code control, and brings us back to the 1970's, when this stuff was basically broken.

I don't understand why we are using 1970's technology in 2015.  This circle-ci stuff is utter, complete, insanity.

linas:
The two commits that really matter for this are these:  

commit/5208b9d5c792cd6dcf7330b7055f3c8df5b0d6e2

commit/f71ff1fdaedbb38cde34da28103900f1f076e9d7

linas:
everything after f71ff1fdaedbb38cde34da28103900f1f076e9d7 is not a part of this pull request, but is part of some other pull request that got mangled in with this one.

Issue #704, opened 2016-03-16
Internal patterns passed on evalution of a DefinedSchemaNode by a GroundedPredicateNode


AmeBel:
### Setup scheme file content
```scheme
(clear)
(use-modules (opencog))
(use-modules (opencog exec))
(use-modules (opencog query))

(InheritanceLink
   (ConceptNode "OpenPsi: Energy" (stv 0.66960984 0.99999982))
   (ConceptNode "OpenPsi: Demand")
)

(InheritanceLink
   (DefinedSchemaNode "OpenPsi: Energy-action-rule-Default")
   (ConceptNode "opencog: action")
)

(MemberLink (stv 1 0.99999982)
   (DefinedSchemaNode "OpenPsi: Energy-action-rule-Default")
   (ConceptNode "OpenPsi: Energy" (stv 0.66960984 0.99999982))
)

(EvaluationLink
   (PredicateNode "OpenPsi: default_value")
   (ListLink
      (ConceptNode "OpenPsi: Energy" (stv 0.66960984 0.99999982))
      (NumberNode "0.710000")
   )
)

(DefineLink
   (DefinedPredicateNode "OpenPsi: demand-pattern")
   (PresentLink
      (InheritanceLink
         (VariableNode "Demand" (stv 0.96491325 0))
         (ConceptNode "OpenPsi: Demand")
      )
      (EvaluationLink
         (PredicateNode "OpenPsi: default_value")
         (ListLink
            (VariableNode "Demand" (stv 0.96491325 0))
            (VariableNode "default_value")
         )
      )
   )
)

(define z-dsn (DefinedSchemaNode "OpenPsi: Energy-action-rule-Default"))
(DefineLink
   z-dsn
   (BindLink
      (VariableList
         (TypedVariableLink
            (VariableNode "Demand" (stv 0.96491325 0))
            (TypeNode "ConceptNode")
         )
      )
      (AndLink
         (DefinedPredicateNode "OpenPsi: demand-pattern")
         (EqualLink
            (VariableNode "Demand" (stv 0.96491325 0))
            (ConceptNode "OpenPsi: Energy" (stv 0.66960984 0.99999982))
         )
      )
      (ExecutionOutputLink
         (GroundedSchemaNode "scm: foo")
         (ListLink
            (VariableNode "Demand" (stv 0.96491325 0))
         )
      )
   )
)

(define (foo node) node)

(define z-get
    (GetLink
       (TypedVariableLink
          (VariableNode "x")
[...]

PR #714, opened 2016-03-20
fix for bug #704

linas:
This patch-set adds a DontExecLink that can be used to solve the issue described in #704.


linas:
test this please

PR #716, opened 2016-03-21
More groping toward lazy execution.

linas:
Eager executation is still required for some things. But the situation seems to be getting better.

Also remove more un-needed casts.


linas:
test this please

PR #722, opened 2016-03-24
Fix for '$' suffix of prefix in node names.

inflector:
The parser for PostgreSQL does a linear search for $$ rather than an outward-in search so it fails when there are $$ next to another $ for the [PostgreSQL dollar escaping](http://www.postgresql.org/docs/current/interactive/sql-syntax-lexical.html#SQL-SYNTAX-DOLLAR-QUOTING). Reverting to the more verbose $ocp$ escaping mechanism as $ocp$ is not likely to be in a node name.


inflector:
Retest this please.

inflector:
Retest this please.

PR #733, opened 2016-03-26
Separate degenerate-case from deep link tests

inflector:
The degenerate link case created a depth^2 traversal time for recursive visiting of atoms in the recursive outgoing set. So `AtomSpace::add_link` for depths more than about 20 took long since `add_link` required a traversal checking for atoms to ignore which should not be added to the backing store.

So I created a separate test which tests deep links where each link has an outgoing set that contains two atoms: a single node shared by all links, and the prior link. Then I bumped the second test to depth 1000 to test high-depth links.


linas:
Looks OK to me. Merge this if you are ready.

PR #740, opened 2016-03-28
Remove many LinkCasts and NodeCasts

linas:
It turns out that LinkCast, NodeCast are slow.  With the newer virtual methods, they are also un-needed.  So remove them.

NOTE: This also fixes a bug in the BackwardChainer, which seemed to have reversed a boolean by accident.

### DIFF
--- a/opencog/atoms/base/atom_types.script
+++ b/opencog/atoms/base/atom_types.script
@@ -280,7 +280,7 @@ SIGNATURE_LINK <- ORDERED_LINK
 FUZZY_LINK <- SIGNATURE_LINK
 
 // A type binder: it binds a type to a VariableNode. Always arity-2,
-// the first position must be a Variablenode, the second must be a type.
+// the first position must be a VariableNode, the second must be a type.
 // Note that this is always scoped, i.e. used in lambdas and places
 // where variables are not free. Thus, unlike TypedAtomLink, this is
 // never globally unique.
--- a/opencog/atomspace/AtomSpace.cc
+++ b/opencog/atomspace/AtomSpace.cc
@@ -278,15 +278,11 @@ Handle AtomSpace::add_atom(AtomPtr atom, bool async)
     if (_backing_store and not _backing_store->ignoreType(t))
     {
         AtomPtr ba;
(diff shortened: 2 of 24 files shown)


ngeiswei:
@linas do you have an idea of how much slower are the cast + call versus directly calling the virtual methods?

linas:
@inflector found and measured this, about 1-2 months ago; its in the mailing list somewhere, or some pull req ...  actually, there had been some previous hints that it was bad: here

https://github.com/opencog/atomspace/blob/master/opencog/benchmark/diary.txt#L1111-L1139

although that was tangled with other work.  Anyway, if you read the assembly that is generated, it about 100LOC or something like that ... and the default path through it was not obvious, but was clearly more than just a few cycles.  By contrast, a virtual table lookup does require a function call indirection, and the resulting cache hit ... but it was lots faster.   Maybe @inflector can find the numbers.

linas:
See pull req #610

linas:
The results were from conversations in the days immediately preceding pull req #610

linas:
actually the performance is here:
https://github.com/opencog/atomspace/blob/master/opencog/benchmark/diary.txt#L-1138-L1150

To interpret this:  on my machine: `LinkCast(h)->getOutgoingSet()` is 7.5M per second  while `h->getOutgoingSet()` is 38M/second.  In other words, the LinkCast was (1/7.5-1/38) microsecs, or (1/7.5 - 1/38) / (1/7.5) = 1 - 7.5/38 = 80% of the total time was in LinkCast, and only 20% actually getting the outgoing set.    This is an extreme case, of doing this in a tight loop, but clearly, its a big win over-all.

ngeiswei:
Alright so the bottom line is that using `h->getOutgoingSet()` is 5 times faster than `LinkCast(h)->getOutgoingSet()`. That's very significant.

inflector:
Yes, Linas found the timings just as I was digging into this last night. The virtual functions are about five to six times faster than the casts. It's a sub-microsecond savings, but since the casts are used all over the place and especially with recursive traversals of links, for example, the total savings can be significant.

PR #744, opened 2016-03-29
Cleanup python path loading

linas:
Addresses the issues observed in /opencog/opencog#2093


linas:
This requires a correspondnig pull to opencog, in order to work.

PR #749, opened 2016-03-30
Add handlemap and so

ngeiswei:
Rename very commonly used types by the URE and PM like map<Handle, Handle> by HandleMap.


ngeiswei:
BTW the main reason I doing that, besides the fact that it simplify the code, it because I'm adding gdb print helps for them all, and it's easier to identify the type to print.

PR #761, opened 2016-04-01
Fix weird virtual confusion pm bug

ngeiswei:


ngeiswei:
@linas although I'm confident enough to merge you may still want to have a look at

https://github.com/opencog/atomspace/pull/761/commits/cc4261f34c1ed3f69872d9d28317978630e46968

opencog/query/PatternMatchEngine.cc
```diff
@@ -1307,9 +1307,12 @@ bool PatternMatchEngine::clause_accept(const Handle& clause_root,
...
 
-	clause_grounding[clause_root] = hg;
-	logmsg("---------------------\nclause:", clause_root);
-	logmsg("ground:", hg);
+	if (not is_evaluatable(clause_root))
+	{
+		clause_grounding[clause_root] = hg;
+		logmsg("---------------------\nclause:", clause_root);
+		logmsg("ground:", hg);
+	}
```
linas: Not a bad solution, but i'm wondering if this is hiding another bug.  Evaluatable clauses are not supposed to be grounded, they're supposed to be evaluated, and so I'm somewhat surprised that they are even going through this particular block of code. 

I don't have time to dig into this right now; I think I need to open to issues to track this so I don't forget.

linas: Never mind, I did look into this. The patch does seem reasonable.

tests/query/PatternUTest.cxxtest
```diff
@@ -192,52 +195,52 @@ void PatternUTest::tearDown(void)
...
 {
 	// create a gnarly graph
-	hconst_1 = as->add_node(CONCEPT_NODE, "some const concept node 1");
-	hconst_2 = as->add_node(CONCEPT_NODE, "some const concept node 2");
-	hconst_3 = as->add_node(CONCEPT_NODE, "some const concept node 3");
-	hsoln_a = as->add_node(CONCEPT_NODE, "desired solution concept node a");
-	hsoln_b = as->add_node(CONCEPT_NODE, "desired solution concept node b");
-	hsoln_c = as->add_node(CONCEPT_NODE, "desired solution concept node c");
+	hconst_1 = an(CONCEPT_NODE, "some const concept node 1");
+	hconst_2 = an(CONCEPT_NODE, "some const concept node 2");
```
linas: FYI, this is the first-ever, oldest unit test for the pattern matcher.

PR #766, opened 2016-04-04
Set number of counts to detect infinit loop to 100

ngeiswei:
Obviously looking at _pattern alone isn't enough. But since it is a
temporary workaround anyway it's OK. I set it to 100 so that the BC unit
test can pass.

### DIFF
--- a/opencog/query/InitiateSearchCB.cc
+++ b/opencog/query/InitiateSearchCB.cc
@@ -761,7 +761,7 @@ bool InitiateSearchCB::variable_search(PatternMatchEngine *pme)
 			if (prev != _pattern) { prev = _pattern; count = 0; }
 			else {
 				count++;
-				if (5 < count)
+				if (100 < count)
 					throw RuntimeException(TRACE_INFO,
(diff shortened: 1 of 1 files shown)


opencog/query/InitiateSearchCB.cc
```diff
@@ -761,7 +761,7 @@ bool InitiateSearchCB::variable_search(PatternMatchEngine *pme)
 			if (prev != _pattern) { prev = _pattern; count = 0; }
 			else {
 				count++;
-				if (5 < count)
+				if (100 < count)
 					throw RuntimeException(TRACE_INFO,
```
linas: OK, but if you are hitting this more than 1x, you've got a bug.  I thought 5x was overkill ... so 100x makes no sense at all, I don't think.  Or am I missing something?

ngeiswei: I don't think hitting more than 1x is a bug. We're talking about the invariance of _pattern only, I don't think that is the only thing to look at. If it was really looping ad infinitum, 2, 5 or 100 would amount to the same result.

PR #768, opened 2016-04-08
DualLink: a dynamic Rete algorithm tool.

linas:
The earlier implementation of the Rete algo for the AIML rules was kind-of-ish hacky. These cleans it up.  It also expands the AIML example, concluding with a mini-AIML system.  Effectively, this demonstrates Rete-selection of rules, cause that's more-or-less all that AIML is.

Anyway: I'm not quite sure how the current forward and backward chainers implement rule selection.  Maybe they are efficient, and maybe they are not.  The DualLink is the "opposite" of pattern matching: it reverses the direction of search, to find patterns, given input data (as opposed to finding data given a pattern).


ngeiswei:
@linas I think rule selection is implemented efficiently in the URE, but the code is convoluted. I'll carefully take a look at it, maybe it can simplify the URE a bit. My main concern is whether this code can deal with partial matching.

linas:
Yeah, don't know about partial matching, haven't tested yet, but it will be something I need (sooner or later). Something fuzzy too.   If you can define what sort of "partial matching" is needed, that would help; I've got some ideas for it but they are incomplete.

I can't merge the code yet, the PutLink unit test fails. I have to fix that.

ngeiswei:
FYI PutLinkUTest consistently fails on my system. BackwardChainerUTest also fails but it comes from the master (I apparently introduced a bug of something).

ngeiswei:
I'm having again this annoying problem with tab indentation when pasting in the guile interpreter...

ngeiswei:
Well, it actually does support "partial matching" as the URE would like it to be. For instance if I run the DualLink on `(ListLink (Concept "I") (Concept "love") (Concept "you"))` given the following rule in the atomspace
```
(BindLink
 (AndLink
  (ListLink
   (Concept "I") (Glob "$star") (Concept "you"))
  (ListLink
   (Concept "I") (Glob "$star") (Concept "her")))
 (ListLink
  (Concept "I") (Glob "$star") (Concept "you and her") (Concept "too")))
```

I do get the matching premise as a result

```
... scheme@(guile-user)> scheme@(guile-user)> (cog-execute! (DualLink sent))

$2 = (SetLink
   (ListLink
      (ConceptNode "I")
      (GlobNode "$star")
      (ConceptNode "you")
   )
)
```

So yeah, I think this functionality could potentially speed up and simplify the URE!

linas:
OK, cool. Yes, that's the idea, you get back the rule premise.  How to climb up from that, to the actual rule that the premise lives in ... for that, you are on your own (for now).

...  although  I may work on that at some point, because I need to be able to handle the situation where I have multiple premises, and I would like to find all of the rules they participate in, and rank them from best to worst.  This is kind-of what the fuzzy matcher does: finds the greatest overlap between one graph and another; except I want to find the greatest overlap between the premises that I have, and a rule that could handle them.

linas:
FYI, variables work fine, too. The examples only use globs because globs can match multiple items in a row in a ListLink. If you are matching just one item, just use ordinary variables like you ordinarily use.

I have never ever tried using a glob with an UnorderedSetLink!  Don;'t know if it would work. I wonder if that would allow matching subsets of SetLinks! Hmm...

linas:
Seems to work fine with unordered links, and does what is expected. For example:
```
(UnorderedLink
  (ConceptNode "this")
  (GlobNode "$star")
  (ConceptNode "is"))

(UnorderedLink
  (GlobNode "$star")
  (ConceptNode "this")
  (ConceptNode "test"))

(UnorderedLink
  (VariableNode "$star")
  (ConceptNode "this")
  (ConceptNode "test"))

(UnorderedLink
  (VariableNode "$star-b")
  (VariableNode "$star-a")
  (ConceptNode "this")
  (ConceptNode "test"))

(cog-execute!  (Dual
(UnorderedLink
  (ConceptNode "is")
  (ConceptNode "test")
  (ConceptNode "this")
  (ConceptNode "a"))))
```

PR #770, opened 2016-04-08
Clean the FC utest

ngeiswei:
Struggling with scheme load, it works then doesn't work, then works again. Very frustrating...


ngeiswei:
I think I know!!! When you delete/rename a scm file its compiled version stays in scheme's cache and so scheme doesn't complain for a while.

tests/rule-engine/ForwardChainerUTest.cxxtest
```diff
@@ -38,21 +38,20 @@ public:
 		// Disable the AF mechanism during testing!
 		_as.set_attentional_focus_boundary(AttentionValue::MINSTI);
 
-		eval.eval("(add-to-load-path \"..\")");
```
linas: Don't remove this: its needed to run the stand-alone unit test for the command line.  For example, if you `cd build` to your build directory, and then say `./tests/rule-engine/ForwardChainerUTest`  then the test will pass with the above line, and it will fail without the above line.

linas:
https://www.gnu.org/software/guile/manual/html_node/Compilation.html

Automatic compilation generally works transparently, without any need for user intervention. However Guile does not yet do proper dependency tracking, so that if file a.scm uses macros from b.scm, and b.scm changes, a.scm would not be automatically recompiled. To forcibly invalidate the auto-compilation cache, pass the --fresh-auto-compile option to Guile, or set the GUILE_AUTO_COMPILE environment variable to fresh (instead of to 0 or 1).


There is exactly one macro in opencog, I think, something that @williampma  set up somewhere. I tripped over it once.

ngeiswei:
I'm sure that was the problem, thanks!

You mention a macro [for fresh-auto-compile I presume]. I can't find any of that. But I guess all I need to know is the --fresh-auto-compile option...

linas:
I was just quoting from the web-page.

PR #777, opened 2016-04-22
Add Inheritance case in EvaluationLink::do_eval_scratch

ngeiswei:
That is because a PutLink also evaluates the resulting body, and if this
body had InheritanceLink in it it would crash. Honestly I don't
understand why it has to evaluate the resulting body but I guess there
is a good reason.

I've also added a unit test for it.


linas:
The PutLink probably should not evaluate, and that seems like a mis-design, mis-feature.  The code is even marked as such:  https://github.com/opencog/atomspace/blob/master/opencog/atoms/execution/Instantiator.cc#L213-L236  

According to those notes, it seems that the current behavior tree depends on that behavior; what it takes to fix or refactor is unclear.   Its possible/likely that switching over to a fully lazy-evaluation style will make this not needed, but the switch to lazy evaluation is only half-done.  I dunno, maybe there should be some other bug report to make note of this.

PR #778, opened 2016-04-26
More fc fixes

ngeiswei:


opencog/query/InitiateSearchCB.cc
```diff
@@ -761,7 +761,7 @@ bool InitiateSearchCB::variable_search(PatternMatchEngine *pme)
 			if (prev != _pattern) { prev = _pattern; count = 0; }
 			else {
 				count++;
-				if (100 < count)
+				if (1000 < count)
```
linas: this change is probably just temporarily hiding some design flaw.

PR #784, opened 2016-05-24
Make GlobNodes work with ImplicationLinks

linas:
GlobNodes are a special case of Variable Nodes. They need to work correctly with ImplicationLinks, too.

### DIFF
--- a/opencog/atoms/core/ImplicationLink.cc
+++ b/opencog/atoms/core/ImplicationLink.cc
@@ -61,26 +61,23 @@ ImplicationLink::ImplicationLink(Link &l)
 void ImplicationLink::extract_variables(const HandleSeq& oset)
 {
 	size_t sz = _outgoing.size();
-	// For now we assume no implicit scoping, so the number of
-	// arguments is 2 when sugar syntax isn't used (no scoping), or 3
-	// when the sugar syntax (scoping) is used.
+
 	if (sz < 2 or 3 < sz)
 		throw InvalidParamException(TRACE_INFO,
-			"Expecting an outgoing set size of at most two, got %d", sz);
+			"Expecting an outgoing set of size 2 or 3, got %d", sz);
 
-	// No sugar syntax
+	// If there is an explicit variable declaration, then use that.
+	// If there is no variable declaration, then assume that all of
+	// the free variables in the link are scoped.
+	ScopeLink::extract_variables(oset);
+
 	if (2 == sz)
 	{
-		_body = oset[0];
 		_implicand = oset[1];
 		return;
 	}
 
 	// If we are here, then the first outgoing set member should be
 	// a variable declaration.
-	_body = oset[1];
 	_implicand = oset[2];
-
-	// Initialize _varlist with the scoped variables
-	init_scoped_variables(oset[0]);
 }
--- a/opencog/atoms/core/MapLink.cc
+++ b/opencog/atoms/core/MapLink.cc
@@ -318,9 +318,10 @@ Handle MapLink::rewrite_one(const Handle& term, AtomSpace* scratch) const
 	if (not extract(_pattern->get_body(), term, valmap, scratch))
 		return Handle::UNDEFINED;
 
-	// Make sure each variable is grounded.
-	// Actually, not all variables need to be grounded ...
-	// re-writes might actually ignore ungrounded vars.
+	// Make sure each variable is grounded. Place the groundings
+	// into a sequence, for easy access. Not all variables need to
+	// be grounded, because the re-write might not use all variables.
+	// If it does use a variable, it must have a grounding.
 	bool partial = false;
 	HandleSeq valseq;
 	for (const Handle& var : _vars->varseq)
@@ -370,7 +371,7 @@ Handle MapLink::execute(AtomSpace* scratch) const
 	// Handle three different cases.
 	// If there is a single value, apply the map to the single value.
 	// If there is a set of values, apply the map to the set.
-	// If there is a link of values, apply the map to the link.
+	// If there is a list of values, apply the map to the list.
 	Type argtype = valh->getType();
 	if (SET_LINK == argtype or LIST_LINK == argtype)
 	{
--- a/opencog/atoms/core/MapLink.cc
+++ b/opencog/atoms/core/MapLink.cc
@@ -50,8 +50,8 @@ void MapLink::init(void)
 		Handle decl(createVariableList(fv.varseq));
 		_pattern = createScopeLink(decl, body);
 	}
-	_vars = &_pattern->get_variables();
-	_varset = &_vars->varset;
+	_mvars = &_pattern->get_variables();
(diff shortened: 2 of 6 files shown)


opencog/atoms/core/ImplicationLink.cc
```diff
@@ -61,26 +61,23 @@ ImplicationLink::ImplicationLink(Link &l)
...
+
 	if (sz < 2 or 3 < sz)
 		throw InvalidParamException(TRACE_INFO,
-			"Expecting an outgoing set size of at most two, got %d", sz);
+			"Expecting an outgoing set of size 2 or 3, got %d", sz);
+
+	// If there is an explicit variable declaration, then use that.
+	// If there is no variable declaration, then assume that all of
+	// the free variables in the link are scoped.
+	ScopeLink::extract_variables(oset);
```
ngeiswei: @linas this is wrong. In the following

```atomese
Implication
  Variable "$X"
  Predicate "P"
```

you don't want `$X` to become a scoped variable under the `Implication`.

Due to that the following PutLink fails

```scheme
(PutLink
   (LambdaLink
      (VariableList
         (TypedVariableLink
            (VariableNode "$X")
            (TypeNode "PredicateNode")
         )
         (TypedVariableLink
            (VariableNode "$Y")
            (TypeNode "PredicateNode")
         )
      )
      (AndLink
         (ImplicationLink
            (VariableNode "$X")
            (PredicateNode "is-human")
         )
         (ImplicationLink
            (VariableNode "$Y")
            (PredicateNode "is-human")
         )
         (EvaluationLink
            (PredicateNode "acquainted")
            (ListLink
               (VariableNode "$X")
               (VariableNode "$Y")
            )
         )
      )
   )
   (ListLink
      (PredicateNode "is-Self")
      (PredicateNode "is-Bob")
   )
)
```

This is a perfect example as to why implicit scoping is a bad idea (unless regulated by a set of rules allowing complete unambiguous expressiveness, which is not the case ATM).

ngeiswei: When I revert that changes I get numerous unit tests that fail, that scares me a bit

```
The following tests FAILED:
	 14 - MultiSpaceUTest (Failed)
	 18 - MapLinkUTest (Failed)
	 30 - FindUtilsUTest (Failed)
	 41 - PatternCrashUTest (Failed)
	 58 - RecognizerUTest (Failed)
	 59 - ArcanaUTest (Failed)
	 61 - GetLinkUTest (Failed)
	 66 - EvaluationUTest (Failed)
	 67 - QuoteUTest (Failed)
	 70 - BuggyQuoteUTest (Failed)
	 82 - TypeChoiceUTest (Failed)
	 85 - UREConfigReaderUTest (Failed)
	 86 - ForwardChainerUTest (Failed)
	 87 - BackwardChainerUTest (Failed)
```

Any idea?

ngeiswei: I sense that implicit scoping in Implications are probably all over the place. It might not be wise to change that [very wrong] assumption before the July demo. On the other hand I don't see any easy workaround that would allow me to keep progressing on reasoning over NLP if don't have that fixed.

Anyway, I deeply look into that, maybe I can get it fixed without screwing the rest...

ngeiswei: Yeah, like I'm thinking of hacking the extract variable code so that it ignores variables which are direct children of the Implication. It is an ugly technical dept, but it *might* not screw up the existing use of it for the time being.

linas: I'd like to remind you that you are the one who added ImplicationLink.cc in the first place ... I was sort of mystified as to why it was needed, or what the point of having it was.  Prior to that addition, there was no special treatment for the ImplicationLink.   Is this class really needed?  Does it really do something important that cannot be done otherwise?

The goal of having C++ classes was not to create or impose semantics, but simply to give a performance boost to something that was going to happen anyway.   Yet ImplicationLink seems to do nothing special for performance, but adds a lot of semantics -- so again, do we even need this C++ class at all? What problem was it meant to solve? Can we just get rid of it (and implement the semantics elsewhere?)

linas: If I remove ImplicationLink.cc and ImplicationLink.h entirely, then only two unit tests fail: 
```
	 18 - MapLinkUTest (SEGFAULT)
	 92 - BackwardChainerUTest (Failed)
```

The MapLinkUTest fails because .. well, let me explain below ... but the shorter question is: do we need ImplicationLink.cc at all? Can we just get rid of it entirely?  If we did, then I think it would solve the problem that you are having.

linas: OK -- now, to explain MapLink and how it currently interacts with ImplicationLink.  

MapLink can be thought of in three different ways:
* It is like "map" in scheme, haskell -- it applies a "function"  to a list or set.  (that's why its called MapLink)
* It is like UnPutLink, so that if "function f" is a pattern with variables in it, it can extract values for those variables.  Thus, its kind-of-like GetLink, except that GetLink searches the entire atomspace, but MapLink only searches the set/list that is it's argument.  You can see why that would be kind-of nice.
* MapLink combined with ImplicationLink is kind-of-like BindLink, but instead of working on the entire atomspace, it only works on the argument list/set. 

I hope the above makes it clear why MapLink is fun, useful, interesting. 

The MapLink unit test fails on this third, last thingy -- combining Map+Implication.   There is very little code that uses/needs this, but I believe that the AIML converter (or maybe the OpenPsi subset of it?) does depend on this working correctly.  So if we break that, the chatbot would break. 

However, there are alternatives. Note that BindLink is "exactly" equal to a GetLink followed by a PutLink.  We could "get rid of" BindLink, its kind of a historical accident that happens to be used a lot.  Recall that BindLink was originally created as a kind-of ImplicationLink that could be evaluated "on the spot".

Similarly, the combination of MapLink+ImplicationLink could probably be replaced by PutLink+MapLink.   That is, just like Put+Get is a kind-of-implication acting on all of atomspace, similarly, Put+Map is a kind-of-implication acting only on its input set/list.  Both of these   perform a re-write of P(x)->Q(x)  where BindLink searches the entire atomspace for x, while Put+Map searches only its input set/list for x.

This we could do this: nuke ImplicationLink.cc entirely, update the MapLinkUTet to use Put+Map instead, and update OpenPsi/AIML to use Put+Map instead of Map+Implication.

We could also create a new link type, e.g. ReWriteLink, that performs the P(x)->Q(x) operation.  We could also re-define MapLink so that if it has multiple args, the first arg would be P(x) and the second arg would be Q(x).   Any of these approaches seem plausible to me.

At any rate, I do not see why we need ImplicationLink.cc at all.  It seems to cause semantic problems.

linas: This discussion can/should continue in issue #822

linas: Seems to be related to #818 also.

PR #789, opened 2016-05-26
Schemepremitive support for signature of type void (Handle, short)

misgeatgit:
Needed for stimulating a handle with some stimulus value from a scheme code.

### DIFF
--- a/opencog/guile/SchemePrimitive.h
+++ b/opencog/guile/SchemePrimitive.h
@@ -123,6 +123,7 @@ class SchemePrimitive : public PrimitiveEnviron
 			UUID (T::*u_ssb)(const std::string&, const std::string&, bool);
 			void (T::*v_b)(bool);
 			void (T::*v_h)(Handle);
+			void (T::*v_hs)(Handle, short);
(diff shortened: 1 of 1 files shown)


opencog/guile/SchemePrimitive.h
```diff
@@ -132,6 +132,7 @@ class SchemePrimitive : public PrimitiveEnviron
 			                 const std::string&);
 			void (T::*v_t)(Type);
 			void (T::*v_ti)(Type, int);
+                	void (T::*v_hs)(Handle, short);
```
linas: could you place this in alphabetical order? h comes before t ....

opencog/guile/SchemePrimitive.h
```diff
@@ -566,6 +568,15 @@ class SchemePrimitive : public PrimitiveEnviron
 					(that->*method.v_ti)(t, i);
 					break;
 				}
+                                case V_HS:
```
linas: alphabetical order here too, and in 4 other places!

misgeatgit:
@linas I have corrected the alphabetic order.

PR #794, opened 2016-06-01
Assorted ProtoAtom fixes and general work.

linas:
Fix a few bugs, extend the scheme bindings, and, most importantly, explicitly place the values into the atom.


linas:
Pursuant to design #513

Issue #798, opened 2016-06-02
Add more convenient scheme bindings


ngeiswei:
I find myself often with the need to use the following functions coded in C++.

1. get_distant_neighbors in `<ATOMSPACE_ROOT>/opencog/atomutils/Neighbors.h`
2. has_free_vars in `<ATOMSPACE_ROOT>/opencog/atomutils/FindUtils.h`

These 2 functions are not that trivial to recode in scheme, of course they may, but then if say the definition of what makes a variable free changes then it requires to correct the scheme implementation as well, etc. For that reason I'd like to add them in the scheme bindings.

Let me know @linas if that's a bad or a good idea.

ngeiswei:
I've added binding for is_closed, see http://wiki.opencog.org/wikihome/index.php/Scheme#cog-closed.3F_atom

linas:
Sorry for late response .. sounds like a great idea to me.  Recoding these grungy algorithms in pure scheme would be a terrible idea.

The only plausible alternative I see might be to invent some sort of `IsClosedLink` or something like that, but its not clear how that could work.   The semantics of a DistantNeighborsLink would be even harder to imagine, but doing so could be enlightening.

ngeiswei:
Yeah, I'm letting aside distant neighbor for now as I can often replace it by a slightly more convoluted use of the pattern matcher.

linas:
Can we close this? I think it was solved in pull req #814

PR #801, opened 2016-06-07
Is closed

ngeiswei:
1. fix get_free_variables (to account properly for quotation)
2. add is_closed to check whether an atom has no free variable
3. add scheme binding cog-closed?

I think this is a sufficiently important function to have its binding.

This partially addresses #798.

### DIFF
--- a/opencog/atomutils/FindUtils.cc
+++ b/opencog/atomutils/FindUtils.cc
@@ -273,44 +273,39 @@ bool contains_atomtype(const HandleSeq& clauses, Type atom_type)
 	return false;
 }
 
-
-HandleSeq get_free_vars_in_tree(const Handle& tree)
+OrderedHandleSet get_free_variables(const Handle& h, int quotation_level)
 {
-	OrderedHandleSet varset;
+	Type t = h->getType();
 
-	std::function<void (const Handle&)> find_rec = [&](const Handle& h)
-	{
-		Type t = h->getType();
-		if (t == VARIABLE_NODE)
-		{
-			varset.insert(h);
-			return;
-		}
+	// Base cases
+	if (t == VARIABLE_NODE and quotation_level < 1)
+		return {h};
+	if (h->isNode())
+		return {};
 
-		// XXX FIXME Add support for UNQUOTE_LINK
-		if (t == QUOTE_LINK) return;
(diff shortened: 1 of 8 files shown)


opencog/atomutils/FindUtils.cc
```diff
@@ -273,44 +273,39 @@ bool contains_atomtype(const HandleSeq& clauses, Type atom_type)
...
+	OrderedHandleSet results;
+	OC_ASSERT(h->isLink());
+	if (t == QUOTE_LINK) quotation_level++;
+	else if (t == UNQUOTE_LINK) quotation_level--;
+	for (const Handle& ch : h->getOutgoingSet()) {
+		OrderedHandleSet ch_free_var = get_free_variables(ch, quotation_level);
+		results.insert(ch_free_var.begin(), ch_free_var.end());
+	}
+	// If the link was a scope link then remove the scoped
+	// variables from the free variables found.
```
linas: This is probably OK for now, but is kind-of dangerous . bound variables in a scope link behave as if they are quoted.  So, here is a nasty example that could occur in real-life:
```
(SomeLink
    (VariableNode  $x)   ;;; -- this is free
    (ScopeLink
           (Variable $x)  ;; --- this is bound, and is not really the "same" variable as above. 
           ...))
```

Note that the second $x can be safely renamed (alpha-converted) into some other variable, leaving the first $x free.  Thus, I suggest that the correct long-term behavior is to treat the ScopeLink as if it was a QuoteLink -- any bound variables inside the ScopeLink behave as if they were quoted, from the point of view of anyone outside of the scope.

I'm not sure, I thought that there already was code somewhere else that already did something like this or almost like this.

ngeiswei: The algorithm processes that correctly, in other words it will say that your example above is not closed. That is because the removal only occurs over the variables inside the children of the scope link.

There was almost code doing that but not quite. Ffirst the `FindAtoms` class does ignore quoted variables, but not scoped variables. Second, there was code that checks whether a given variable is scoped or quoted, but no code actually fetching unscoped, unquoted variables.

linas: ah, OK. I didn't look it through carefully.

PR #802, opened 2016-06-08
Fix CythonBindLink test

pennachin:
This unit test was broken by #744, which removed a `.conf` file, and cause the initialization code to unsuccessfully search for the default one. It doesn't seem to need a `.conf` file, but someone should double check that. If such file is needed, parts of #744 should be reverted.


linas:
thanks. The atomspace itself does not need/use a config file; only the cogserver does, so this change should be fine.  Strange that I missed this crash, before.

pennachin:
Cool, that's what I thought but wasn't 100% sure.

PR #804, opened 2016-06-09
Fix ure utest pass without install

ngeiswei:


tests/rule-engine/fc-config.scm
```diff
@@ -1,10 +1,21 @@
...
-; Load the rules
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; Load required modules and utils ;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(use-modules (opencog))
+(use-modules (opencog rule-engine))
+
+(load-from-path "utilities.scm")
+(load-from-path "av-tv.scm")
```
linas: After saying `(use-modules (opencog))`, one should not need to also load "utilities.scm" and "av-tv.scm" since everything in those files should already be publicly exported by the main opencog module.

ngeiswei: Yeah, I know, I guess it must be some cruft from a time that this assumption wasn't true and I didn't dare to challenge it. If you're confident it's gonna stay true in the future I'm happy to remove these lines.

linas: It will stay true -- the base opencog guile module provides the various common utilities, and I see no reason to change this, other than to periodically go in there and clean up whatever mess and cruft accumulates over time.

Anyway, this is very low priority, I just thought I'd mention it.

ngeiswei: Oh, I've just tried, the problem is that it's only true if the modules are installed on the system. So these are still useful if one wants to run these unit tests without installing opencog.

PR #807, opened 2016-06-13
Overload oc_to_string for LinkPtr

ngeiswei:
More GDB types support for poc

### DIFF
--- a/opencog/atoms/base/Handle.cc
+++ b/opencog/atoms/base/Handle.cc
@@ -183,6 +183,10 @@ std::string atomtype_to_string(Type type)
 	ss << classserver().getTypeName(type) << std::endl;
 	return ss.str();
 }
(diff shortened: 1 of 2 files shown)


opencog/atoms/base/Handle.cc
```diff
@@ -216,6 +220,10 @@ std::string oc_to_string(Type type)
 {
 	return atomtype_to_string(type);
 }
+std::string oc_to_string(const LinkPtr& gl)
+{
+	return lptr_to_string(gl);
```
linas: why are there two functions with different names, that do the same thing?

ngeiswei: Well, that's probably over cautious, some versions of GDB have problems with overloading so when that happens you can still call the non-overloaded version.

Issue #795, opened 2016-06-01
python should not use UUID's in it's internal representation


linas:
The use of UUID's in the python bindings is a design flaw, it results in various issues, e.g. #792

linas:
Closing, as this is fixed in pull req #922

PR #808, opened 2016-06-14
cython bug fix

AmeBel:
Just stumbled upon the bug while experimenting with `get_atom_with_uuid`. Maybe the fix is unnecessary as per https://github.com/opencog/atomspace/issues/795 .


AmeBel:
Ok `get_atom_with_uuid` is used by the restapi. So this fix is required.

linas:
Using UUID's is OK for inter-process communications (its currently unavoidable). Using UUID's internally is what #795 is complaining about.

AmeBel:
got it, thanks :smile:

PR #810, opened 2016-06-15
time-octomap signatures for scheme

yantrabuddhi:
updated schemeprimitive.h to hold 12 new signatures for timeoctomap(to be submitted) in opencog

### DIFF
--- a/opencog/guile/SchemePrimitive.h
+++ b/opencog/guile/SchemePrimitive.h
@@ -93,8 +93,13 @@ class SchemePrimitive : public PrimitiveEnviron
 			bool (T::*b_b)(bool);
 			bool (T::*b_hi)(Handle, int);
 			bool (T::*b_hh)(Handle, Handle);
+                        bool (T::*b_sddd)(const std::string&,double,double,double);
+                        bool (T::*b_sdii)(const std::string&,double,int,int);
+                        bool (T::*b_shddd)(const std::string&,Handle,double,double,double);
+                        bool (T::*b_siddd)(const std::string&,int,double,double,double);
 			double (T::*d_hht)(Handle, Handle, Type);
 			double (T::*d_hhtb)(Handle, Handle, Type, bool);
+                        double (T::*d_s)(const std::string&);
 			Handle (T::*h_h)(Handle);
 			Handle (T::*h_hi)(Handle, int);
 			Handle (T::*h_hh)(Handle, Handle);
@@ -102,7 +107,12 @@ class SchemePrimitive : public PrimitiveEnviron
 			Handle (T::*h_hs)(Handle, const std::string&);
 			Handle (T::*h_htq)(Handle, Type, const HandleSeq&);
 			Handle (T::*h_hz)(Handle, size_t);
-			Handle (T::*h_sq)(const std::string&, const HandleSeq&);
+                        Handle (T::*h_sddd)(const std::string&,double,double,double);
+                        Handle (T::*h_sh)(const std::string&,Handle);
+                        Handle (T::*h_shddd)(const std::string&,Handle,double,double,double);
+                        Handle (T::*h_shi)(const std::string&,Handle,int);
+                        Handle (T::*h_siddd)(const std::string&,int,double,double,double);
+               		Handle (T::*h_sq)(const std::string&, const HandleSeq&);
 			Handle (T::*h_sqq)(const std::string&,
 			                   const HandleSeq&, const HandleSeq&);
 			HandleSeq (T::*q_h)(Handle);
@@ -110,6 +120,7 @@ class SchemePrimitive : public PrimitiveEnviron
 			HandleSeq (T::*q_htib)(Handle, Type, int, bool);
 			HandleSeqSeq (T::*k_h)(Handle);
 			HandleSeqSeq (T::*k_hi)(Handle, int);
+                        int (T::*i_s)(const std::string&);
 			const std::string& (T::*s_as)(AtomSpace*,
 			                             const std::string&);
 			const std::string& (T::*s_s)(const std::string&);
@@ -126,6 +137,8 @@ class SchemePrimitive : public PrimitiveEnviron
 			void (T::*v_hs)(Handle, short);
 			void (T::*v_s)(const std::string&);
 			void (T::*v_sa)(const std::string&, AtomSpace*);
+                        void (T::*v_sh)(const std::string&,Handle);
+                        void (T::*v_shi)(const std::string&,Handle,int);
 			void (T::*v_ss)(const std::string&,
 			                const std::string&);
 			void (T::*v_sss)(const std::string&,
@@ -144,8 +157,13 @@ class SchemePrimitive : public PrimitiveEnviron
 			B_B,   // return boolean, take boolean
 			B_HI,  // return boolean, take handle and int
 			B_HH,  // return boolean, take handle and handle
+                        B_SDDD,
+                        B_SDII,
+                        B_SHDDD,
+                        B_SIDDD,
 			D_HHT, // return double, take handle, handle, and type
 			D_HHTB,// return double, take handle, handle, and type
+                        D_S,
 			H_H,   // return handle, take handle
 			H_HI,  // return handle, take handle and int
 			H_HH,  // return handle, take handle and handle
@@ -153,8 +171,14 @@ class SchemePrimitive : public PrimitiveEnviron
 			H_HHH, // return handle, take handle, handle and Handle
 			H_HTQ, // return handle, take handle, type, and HandleSeq
 			H_HZ,  // return handle, take handle and size_t
+                        H_SDDD,
+                        H_SH,
+                        H_SHDDD,
+                        H_SHI,
+                        H_SIDDD,
 			H_SQ,  // return handle, take string and HandleSeq
 			H_SQQ, // return handle, take string, HandleSeq and HandleSeq
+                        I_S,
 			Q_H,   // return HandleSeq, take handle
 			Q_HTI, // return HandleSeq, take handle, type, and int
 			Q_HTIB,// return HandleSeq, take handle, type, and bool
@@ -172,6 +196,8 @@ class SchemePrimitive : public PrimitiveEnviron
 			V_HS,  // return void, take Handle and short
 			V_S,   // return void, take string
 			V_SA,  // return void, take string, Atomspace
+                        V_SH,
+                        V_SHI,
 			V_SS,  // return void, take two strings
 			V_SSS, // return void, take three strings
 			V_T,   // return void, take Type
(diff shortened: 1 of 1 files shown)


opencog/guile/SchemePrimitive.h
```diff
@@ -93,23 +93,34 @@ class SchemePrimitive : public PrimitiveEnviron
 			bool (T::*b_b)(bool);
 			bool (T::*b_hi)(Handle, int);
 			bool (T::*b_hh)(Handle, Handle);
+                        bool (T::*b_sddd)(const std::string&,double,double,double);
+                        bool (T::*b_sdii)(const std::string&,double,int,int);
+                        bool (T::*b_shddd)(const std::string&,Handle,double,double,double);
+                        bool (T::*b_siddd)(const std::string&,int,double,double,double);
```
ngeiswei: @yantrabuddhi could fix the indentation, they should be tabs not whitespaces. (Curious, what text editor / IDE do you use?)

yantrabuddhi: Hi @ngeiswei ,
 Oh sorry about that , i will fix it, i used spaces by default, i was using vim

ngeiswei: See http://wiki.opencog.org/w/Development_standards#Coding_standards

opencog/guile/SchemePrimitive.h
```diff
@@ -144,17 +157,28 @@ class SchemePrimitive : public PrimitiveEnviron
...
 			D_HHTB,// return double, take handle, handle, and type
+                        D_S,
 			H_H,   // return handle, take handle
 			H_HI,  // return handle, take handle and int
 			H_HH,  // return handle, take handle and handle
 			H_HS,  // return handle, take handle and string
 			H_HHH, // return handle, take handle, handle and Handle
 			H_HTQ, // return handle, take handle, type, and HandleSeq
 			H_HZ,  // return handle, take handle and size_t
+                        H_SDDD,
```
ngeiswei: I think similar comments are worth adding, to counteract the inner crypticness of that code..

```c++
// return handle, take string and 3 double
```

linas:
Umm, I kind of object to all of this .. we had multiple discussions about this, before I left HK, and I thought that it was decided that Atoms would be used for the API.  I don't see any reason to change this decision.

Nil, the reason I wanted Atoms for the API, and not raw floats, etc. is so that both PLN, moses, attention-allocation, openpsi and etc. could have access to the data. Sticking this stuff into the scheme bindings makes it completely inaccessible to every subsystem that knows how to manipulate data.  There is no way to get access to the data, when its buried underneath the scheme bindings, like this.

So -- I really strongly suggest re-thinking the general idea of what the goal of all of this is. This smells like "junk code" to me, that will only have to be ripped out and replaced in the future. In terms of development time and resources, its a lot cheaper to do it right the first time, than to do it wrong, and redesign later.

bgoertzel:
Part of the reasoning, as I understand it, is as follows...

One requirement we have is: We want Atoms created by perceptual processes
(e.g. face tracking and audio localization) to be able to get created and
placed into the TimeSpaceMap, *without* necessarily being placed in the
Atomspace....   In this way Atoms can be created, rapidly put into the
TimeSpaceMap, and then rapidly removed -- without the overhead of Atomspace
insertion and removal...

So it seems workable to me, to have perceptual Atoms resultant from
perception processes put into the TimeSpaceMap directly via the Scheme API
--- and then these Atoms will be pushed into the Atomspace only when this
is required.

On the other hand, if the only way to get Atoms into the TimeSpaceMap is
via putting them into the Atomspace first, then we are (in my best guess)
going to have significant performance problems processing perceptual data
in real time.   Whereas the OctoMap inside the TimeSpaceMap is meant for
dealing with perceptual data in real-time and can handle it just fine; and
creating Atoms and deleting them, outside the context of the Atomspace, is
also fast...

I do think that we want an "Atomese" API for the TimeSpaceMap -- i.e. we
want to be able to do everything that is do-able with the TimeSpaceMap from
within DefinedSchemaNodes in the Atomspace, without having to wrap all the
TimeSpaceMap manipulations inside GroundedSchemaNodes ... i.e. we do want
the TimeSpaceMap API exposed in terms of Atoms....   However, I don't yet
see why it should be exposed *only* in terms of Atoms ... because it seems
to me that for some use cases (e.g. the putting of
potentially-very-temporary Atoms into the TimeSpaceMap via perception
processes) using an Atomese API will probably be insufficiently performant
and a Scheme API will be more satisfactory...

Now, if you see an easy, better way to do what Mandeep has done -- please
do it and check in the code, that will be great....   But bear in mind that
it shouldn't be significantly slower than the Scheme bindings approach, for
the perception use-case I've described...

Also, I'm sure you understand this is something we need working YESTERDAY
for Hanson Robotics application.   I really want to fuse audio localization
input and face tracking input in OpenCog, not in some external ROS node;
and to be able to do this with adequate performance, I need to have the
TimeSpaceMap involved....   Yes, we could do that just using the C++ API to
the TimeSpaceMap, but I really think it will be better to use Scheme than
C++ here...

Anyway we are open to there being a better solution, but Mandeep and I
could not quickly think of a better solution that would not also be a lot
slower .. and we do not have time to wait for a better solution to slowly
appear from somewhere...

You may be right that this code will be replaced in the future, though this
is not obvious to me at all.   But even if it IS replaced in the future, if
it enables us to get sensor fusion to work in OpenCog during the next few
weeks, it will still have served a very valuable function....   We need to
get this stuff working NOW ... I don't want to make horrible disgusting
hacks in order to make things work rapidly, but I am willing to make some
design choices that are not maximally elegant and beautiful if that's what
it takes...

-- Ben



On Thu, Jun 16, 2016 at 4:51 PM, Linas Vepštas <notifications@github.com>
wrote:

> Umm, I kind of object to all of this .. we had multiple discussions about
> this, before I left HK, and I thought that it was decided that Atoms would
> be used for the API. I don't see any reason to change this decision.
>
> Nil, the reason I wanted Atoms for the API, and not raw floats, etc. is so
> that both PLN, moses, attention-allocation, openpsi and etc. could have
> access to the data. Sticking this stuff into the scheme bindings makes it
> completely inaccessible to every subsystem that knows how to manipulate
> data. There is no way to get access to the data, when its buried underneath
> the scheme bindings, like this.
>
> So -- I really strongly suggest re-thinking the general idea of what the
> goal of all of this is. This smells like "junk code" to me, that will only
> have to be ripped out and replaced in the future. In terms of development
> time and resources, its a lot cheaper to do it right the first time, than
> to do it wrong, and redesign later.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/opencog/atomspace/pull/810#issuecomment-226426063>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AFolXNZ3AueOda_y0dujgZDrz91uAXC8ks5qMQ5-gaJpZM4I2AXP>
> .
>



-- 
Ben Goertzel, PhD
http://goertzel.org

Super-benevolent super-intelligence is the thought the Global Brain is
currently struggling to form...

bgoertzel:
Also, this comment

***
Nil, the reason I wanted Atoms for the API, and not raw floats, etc. is so
that both PLN, moses, attention-allocation, openpsi and etc. could have
access to the data. Sticking this stuff into the scheme bindings makes it
completely inaccessible to every subsystem that knows how to manipulate
data. There is no way to get access to the data, when its buried underneath
the scheme bindings, like this.
***

confuses me a bit...

The data is being stored in Atoms, which are being placed in the
TimeSpaceMap

The lodging of an Atom in the TimeSpaceMap can be recorded in the Atomspace
using AtLocationLinks and AtTimeLinks...

Queries into the TimeSpaceMap (and the Atoms contained therein) very
definitely need to be exposed within Atomese in a simple and elegant way

So, I don't think there is any data involved that is going to be "buried
underneath the scheme bindings"

Rather, the idea is to provide multiple ways to manipulate the data

-- using C++ (which already exists)
-- using Scheme (which Mandeep has just implemented)
-- using Atomese

The data will all be accessible to Atoms in the Atomspace .. .but sometimes
(as in real time perception processing) the most efficient and best course
is to manipulate this data via some other means than via Atoms in the
Atomspace, e.g. via Scheme or C++

Not having Scheme bindings just means that when one needs to do something
fast with the TimeSpaceMap one needs to use C++; and forcing C++ on the
developer doesn't seem the right design choice...

The above is my current understanding which may well be flawed.  However,
we're trying to get some complex stuff to work in a hurry under a lot of
pressure, and so I'm in the position of needing to make decisions based on
my current, possibly flawed understanding...

thx
ben


On Thu, Jun 16, 2016 at 5:18 PM, Ben Goertzel <ben@goertzel.org> wrote:

>
> Part of the reasoning, as I understand it, is as follows...
>
> One requirement we have is: We want Atoms created by perceptual processes
> (e.g. face tracking and audio localization) to be able to get created and
> placed into the TimeSpaceMap, *without* necessarily being placed in the
> Atomspace....   In this way Atoms can be created, rapidly put into the
> TimeSpaceMap, and then rapidly removed -- without the overhead of Atomspace
> insertion and removal...
>
> So it seems workable to me, to have perceptual Atoms resultant from
> perception processes put into the TimeSpaceMap directly via the Scheme API
> --- and then these Atoms will be pushed into the Atomspace only when this
> is required.
>
> On the other hand, if the only way to get Atoms into the TimeSpaceMap is
> via putting them into the Atomspace first, then we are (in my best guess)
> going to have significant performance problems processing perceptual data
> in real time.   Whereas the OctoMap inside the TimeSpaceMap is meant for
> dealing with perceptual data in real-time and can handle it just fine; and
> creating Atoms and deleting them, outside the context of the Atomspace, is
> also fast...
>
> I do think that we want an "Atomese" API for the TimeSpaceMap -- i.e. we
> want to be able to do everything that is do-able with the TimeSpaceMap from
> within DefinedSchemaNodes in the Atomspace, without having to wrap all the
> TimeSpaceMap manipulations inside GroundedSchemaNodes ... i.e. we do want
> the TimeSpaceMap API exposed in terms of Atoms....   However, I don't yet
> see why it should be exposed *only* in terms of Atoms ... because it seems
> to me that for some use cases (e.g. the putting of
> potentially-very-temporary Atoms into the TimeSpaceMap via perception
> processes) using an Atomese API will probably be insufficiently performant
> and a Scheme API will be more satisfactory...
>
> Now, if you see an easy, better way to do what Mandeep has done -- please
> do it and check in the code, that will be great....   But bear in mind that
> it shouldn't be significantly slower than the Scheme bindings approach, for
> the perception use-case I've described...
>
> Also, I'm sure you understand this is something we need working YESTERDAY
> for Hanson Robotics application.   I really want to fuse audio localization
> input and face tracking input in OpenCog, not in some external ROS node;
> and to be able to do this with adequate performance, I need to have the
> TimeSpaceMap involved....   Yes, we could do that just using the C++ API to
> the TimeSpaceMap, but I really think it will be better to use Scheme than
> C++ here...
>
> Anyway we are open to there being a better solution, but Mandeep and I
> could not quickly think of a better solution that would not also be a lot
> slower .. and we do not have time to wait for a better solution to slowly
> appear from somewhere...
>
> You may be right that this code will be replaced in the future, though
> this is not obvious to me at all.   But even if it IS replaced in the
> future, if it enables us to get sensor fusion to work in OpenCog during the
> next few weeks, it will still have served a very valuable function....   We
> need to get this stuff working NOW ... I don't want to make horrible
> disgusting hacks in order to make things work rapidly, but I am willing to
> make some design choices that are not maximally elegant and beautiful if
> that's what it takes...
>
> -- Ben
>
>
>
> On Thu, Jun 16, 2016 at 4:51 PM, Linas Vepštas <notifications@github.com>
> wrote:
>
>> Umm, I kind of object to all of this .. we had multiple discussions about
>> this, before I left HK, and I thought that it was decided that Atoms would
>> be used for the API. I don't see any reason to change this decision.
>>
>> Nil, the reason I wanted Atoms for the API, and not raw floats, etc. is
>> so that both PLN, moses, attention-allocation, openpsi and etc. could have
>> access to the data. Sticking this stuff into the scheme bindings makes it
>> completely inaccessible to every subsystem that knows how to manipulate
>> data. There is no way to get access to the data, when its buried underneath
>> the scheme bindings, like this.
>>
>> So -- I really strongly suggest re-thinking the general idea of what the
>> goal of all of this is. This smells like "junk code" to me, that will only
>> have to be ripped out and replaced in the future. In terms of development
>> time and resources, its a lot cheaper to do it right the first time, than
>> to do it wrong, and redesign later.
>>
>> —
>> You are receiving this because you are subscribed to this thread.
>> Reply to this email directly, view it on GitHub
>> <https://github.com/opencog/atomspace/pull/810#issuecomment-226426063>,
>> or mute the thread
>> <https://github.com/notifications/unsubscribe/AFolXNZ3AueOda_y0dujgZDrz91uAXC8ks5qMQ5-gaJpZM4I2AXP>
>> .
>>
>
>
>
> --
> Ben Goertzel, PhD
> http://goertzel.org
>
> Super-benevolent super-intelligence is the thought the Global Brain is
> currently struggling to form...
>



-- 
Ben Goertzel, PhD
http://goertzel.org

Super-benevolent super-intelligence is the thought the Global Brain is
currently struggling to form...

linas:
Ugh.  OK. Yes, I understand the idea that "it was needed yesterday" -- Just be aware that there is a concept of "accumulating technical debt", and like monetary debt, one pays interest on it, and the total cost rises over time. Sooner or later, you have to pay it back, or you go bankrupt.  Bankruptcy is easy to spot: its when you decide that its cheaper to throw away the code, and rewrite from scratch (or when developers stop working on a project, without throwing away the code).

yantrabuddhi:
Hi Linas,
 I have a question I am not clear about.
Saying the following in scheme api:
(AtTimeLink
    (TimeNode "time-string")
    (Node "some face")
)
will this produce atoms in atomspace or just the heap not attached to
atomspace?
If it is only in the heap I can attempt to pass just handles, however it
takes more time to make the code since there is no automatic type and
signature checking and anyone who codes this way for creating scheme
bindings needs to decode and verify the type and structure from the handle
passed requiring good amount of time to write the code, plus not being very
transparent and being unnecessarily verbose, with performance hit of run
time type/signature checking..
If i had a clear easy api for type/signature match[for atoms in/not in
atomspace] it would be heavily reused and transparent and would not require
us to invent lot of new signatures in scheme primitives. [However still we
will need to put signatures somewhere, only advantage of passing atoms is
it can be generic signatures and we don't need to put these in code files
but still need in data files]
In case the solution i made is sub-optimal, i am sorry , that is just what
came to my mind to get a working solution following patterns of what i
knew. I will improve my way as my understanding grows. There is technical
debt and will always be, we will not always be able to make best decisions
from the start, but yes if we already know and can do better we will take
that path.
Regards,
Mandeep

PR #815, opened 2016-06-24
bug fix in link.h and spatial query signature

yantrabuddhi:
We found a bug in link.h in and fixed it. Contains 2 added signatures for spacial query api.

### DIFF
--- a/opencog/atoms/base/Link.h
+++ b/opencog/atoms/base/Link.h
@@ -110,7 +110,7 @@ public:
         HandleSeq oset(3);
         oset[0] = ha;
         oset[1] = hb;
-        oset[3] = hc;
+        oset[2] = hc;
         init(oset);
     }
     Link(Type t, const Handle& ha, const Handle &hb,
--- a/opencog/guile/SchemePrimitive.h
+++ b/opencog/guile/SchemePrimitive.h
@@ -100,6 +100,7 @@ class SchemePrimitive : public PrimitiveEnviron
 			double (T::*d_hht)(Handle, Handle, Type);
 			double (T::*d_hhtb)(Handle, Handle, Type, bool);
 			double (T::*d_s)(const std::string&);
+                        double (T::*d_shhi)(const std::string&, Handle, Handle,int);
 			Handle (T::*h_h)(Handle);
 			Handle (T::*h_hi)(Handle, int);
 			Handle (T::*h_hh)(Handle, Handle);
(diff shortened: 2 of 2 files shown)


opencog/atoms/base/Link.h
```diff
@@ -110,7 +110,7 @@ class Link : public Atom
         HandleSeq oset(3);
         oset[0] = ha;
         oset[1] = hb;
-        oset[3] = hc;
+        oset[2] = hc;
```
ngeiswei: OMG!

ngeiswei:
Sounds good to me, I'm merging.

opencog/guile/SchemePrimitive.h
```diff
@@ -100,6 +100,7 @@ class SchemePrimitive : public PrimitiveEnviron
 			double (T::*d_hht)(Handle, Handle, Type);
 			double (T::*d_hhtb)(Handle, Handle, Type, bool);
 			double (T::*d_s)(const std::string&);
+			double (T::*d_shhi)(const std::string&, Handle, Handle,int);
```
linas: would it be possible to adopt a convention where the handles always come first?

yantrabuddhi: Hi Linas,
 In case you all think it is better to keep handles first convention then
let me know, I will change my code to match the convention.
Regards,
Mandeep

linas: Well, right now, from where I sit, it all looks very randomized, and programming API's tend to work better when they are orderly and organized.  So that's what I'm asking for. A string followed by a handle seems weird.

Orderly-and-organized is another reason I don't really like all of these additions. In guile,  its very hard to tell what type a given argument is, because guile does not have types. Thus, to create functions in guile that take four arguments, all with different types, makes them hard to use, hard to remember, hard to debug. 

I'm also concerned that, below, you are creating another function, very similar to this, but takes 5 arguments!  and it returns int instead of double! That will lead to even more confusion.   (Because, in scheme, you cannot tell what the return type of a function is, because, in scheme, there are no types...)

In general, in *any* programming language, it is very unusual to have more than two arguments to a function. Look around at almost any popular, famous, well-known, well-loved API and you will see that this is the case.   Thus, the fact that you are creating signatures with a large number of arguments worries me.  The fact that the signatures are inhomogenous (each argument is a different type) is even more worrisome. Normal, well-designed, easy-to-use API's just don't do things like that. 

Maybe you have a good reason for this, but please do ask yourself if there might be some easier, simpler way to accomplish things.

PR #820, opened 2016-06-28
Hack implication implicit scope

ngeiswei:
Make things like
```atomese
Implication
   $X
   Q
```
not consider $X as implicitly scoped by the implication. It really is a temporary hack, for instance
```atomese
Implication
   $X
   Evaluation
      Q
      $Y
```
will not consider $Y as being implicitly scoped either.

I think implicit scoping is dangerous so I have no interest in devising a rule set for it. But I prefer to introduce this hack rather than changing the assumptions and possibly breaking others code before the July demo.


ngeiswei:
BTW, OpenPsi utest breaks with or without this hack so I don't know if it actually breaks it...

PR #834, opened 2016-07-22
Added LibraryManager

rTreutlein:
To avoid having to constantly reaload librarys
Also allows haskell to stay in the same IO Context

### DIFF
--- a/opencog/atoms/execution/ExecutionOutputLink.h
+++ b/opencog/atoms/execution/ExecutionOutputLink.h
@@ -23,6 +23,7 @@
 #ifndef _OPENCOG_EXECUTION_OUTPUT_LINK_H
 #define _OPENCOG_EXECUTION_OUTPUT_LINK_H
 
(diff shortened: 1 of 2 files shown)


opencog/atoms/execution/ExecutionOutputLink.h
```diff
@@ -63,6 +64,15 @@ static inline ExecutionOutputLinkPtr ExecutionOutputLinkCast(AtomPtr a)
...
 
+class LibraryManager
+{
+private:
+    static std::unordered_map<std::string, void*> _librarys;
+    static std::unordered_map<std::string, void*> _functions;
+public:
+    static void* getFunc(std::string libName,std::string funcName);
+};
+
```
linas: could you move this class declaration out of the header file, and into the C file? I see no reason to make this public --  its not even a real class, as all of it's members are static. Its also specific to linux; a mac or windows port would work differently.

PR #838, opened 2016-07-26
Fix python block chunking

ngeiswei:
Partially revert commit https://github.com/opencog/atomspace/pull/837/commits/8dda9e4b3842b209af8355b41d49aab283de32cf so that code blocks with newlines in the middle don't get split.

### DIFF
--- a/tests/cython/PythonEvalUTest.cxxtest
+++ b/tests/cython/PythonEvalUTest.cxxtest
@@ -5,6 +5,8 @@
 #include <opencog/cython/PythonEval.h>
 #include <opencog/guile/SchemeEval.h>
 
+#include <cxxtest/TestSuite.h>
+
 using std::string;
 
 using namespace opencog;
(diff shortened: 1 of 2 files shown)


ngeiswei:
@linas I letl it to you to review and possibly merge.

ngeiswei:
I suggest to merge it. If something is wrong I should be able to fix today.

linas:
OK, well, additional fixes are needed: at this time,  starting the cogserver, telnetting to port 17001, and entering the python shell there  produces no output, either at the shell, or at the cogserver, for this:
```
for i in range(10) :
    print "foo", i
```

Before your change, the above worked.  I am not sure that anyone is acutally using the python shell for anything at this time, so maybe fixing this is low-priority.

tests/cython/PythonEvalUTest.cxxtest
```diff
@@ -297,4 +299,33 @@ public:
...
+        PythonEval::create_singleton_instance(as);
+        PythonEval* python = &PythonEval::instance();
+
+        // Define python functions with newline within the code block
+        string result = python->eval(
+            "def fun(x):\n"
+            "    y = x + 1\n"
+            "\n"
+            "    return y\n"
+            "\n"
```
linas: This is not valid python.  When I run this by hand, in python, I get an error:
```
>>> def fun(x):
...     y = x + 1
... 
>>>     return y
  File "<stdin>", line 1
    return y
    ^
IndentationError: unexpected indent
>>> 
```

linas: However, if I copy the above into a file, and run it from a file i.e. `python f.py`  then it runs without any errors. So it seems that python is inconsistent with itself, as to what it accepts as valid syntax.

Contents of f.py:
```
def fun(x):
    y = x + 1


    print "foo", y
    return y

fun(1)
print "bar"

```

ngeiswei: Yeah, since opencog python interpreter is supposed to work in non-interactive/script mode as well (as is the case now) we want to support newlines within blocks.

PR #839, opened 2016-07-28
Initial work of moving out attention code from Agent class

misgeatgit:

### DIFF
--- a/opencog/atoms/base/Atom.cc
+++ b/opencog/atoms/base/Atom.cc
@@ -201,6 +201,17 @@ void Atom::setAttentionValue(AttentionValuePtr av)
     avch(getHandle(), local, av);
 }
 
+void Atom::stimulate(AtomSpace* as, float stimulus)
+{
+    int sti = getAttentionValue()->getSTI();
+    int lti = getAttentionValue()->getLTI();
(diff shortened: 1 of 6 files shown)


opencog/atoms/base/Atom.cc
```diff
@@ -201,6 +201,17 @@ void Atom::setAttentionValue(AttentionValuePtr av)
     avch(getHandle(), local, av);
 }
 
+void Atom::stimulate(AtomSpace* as, float stimulus)
```
linas: this can't work, atoms must not depend on the atomspace ...

linas: The solution that makes sense is to add a method to AtomSpace something like 
```
AtomSpace::stimulate(const Handle& h, double stimulus)
```
that does what your code does, but flipping things around.

Also -- please don't use float -- use double.  floats are just 32 bit, with 23-bit mantissas and will plague you with rounding errors.

misgeatgit: I have made those changes.

linas:
thanks, looks great!

linas:
this pertains to bug opencog/opencog#2314

PR #840, opened 2016-07-28
Add lock_guard in PythonEval::eval

ngeiswei:


opencog/cython/PythonEval.h
```diff
@@ -177,7 +178,10 @@ class PythonEval : public GenericEval
 
         // The synchronous-output interface.
         std::string eval(const std::string& expr)
-            { begin_eval(); eval_expr(expr); return poll_result(); }
+        {
+            std::lock_guard<std::mutex> lock(_eval_mutex);
```
linas: this is fine, and semantically, it "does the right thing".  However,  the other mutexes should already have been having exactly the same effect as this, so there must be a bug with the placement of the other mutexes.

PR #842, opened 2016-07-29
temporary fix for race bug, root cause not known

yantrabuddhi:
This fixes crash, but might break adding of a node randomly since i do not know why its unable to find an atomspace. The reasons are related to netcat spawning threads and executing scheme eval to add a node


ngeiswei:
Ideally you would log some warning. But we're working under ideal conditions anyway. I'm merging it for what it is, a temporary fix.

linas:
Please stop merging code that is obviously incorrect.  This kind of behavior is making things worse, not better.

Lets start by opening a bug report which reports the acutal problem, and reports the stack traces and the log files associated with the crash.

PR #843, opened 2016-07-29
probable fixes

yantrabuddhi:
Hi Nil,
 It was earlier crashing less with temporary fix, but these new additions seem to solve the problem for me and it doesn't crash in a long while

### DIFF
--- a/opencog/guile/SchemeSmob.cc
+++ b/opencog/guile/SchemeSmob.cc
@@ -36,7 +36,7 @@ using namespace opencog;
  * simplify that code, and probably improve performance just a bit.
  */
 
-scm_t_bits SchemeSmob::cog_misc_tag;
+std::atomic<scm_t_bits> SchemeSmob::cog_misc_tag;
 std::atomic_flag SchemeSmob::is_inited = ATOMIC_FLAG_INIT;
 SCM SchemeSmob::_radix_ten;
 
--- a/opencog/guile/SchemeSmobAS.cc
+++ b/opencog/guile/SchemeSmobAS.cc
@@ -277,7 +277,7 @@ SCM SchemeSmob::ss_as(SCM satom)
 /**
  * Return current atomspace for this dynamic state.
  */
-SCM SchemeSmob::atomspace_fluid;
+std::atomic<SCM> SchemeSmob::atomspace_fluid;
 
 SCM SchemeSmob::ss_get_as (void)
 {
--- a/opencog/guile/SchemeSmob.h
+++ b/opencog/guile/SchemeSmob.h
@@ -28,6 +28,8 @@
 
 #include <opencog/atomspace/AtomSpace.h>
 
+#include <atomic>
+
 namespace opencog {
 /** \addtogroup grp_smob
  *  @{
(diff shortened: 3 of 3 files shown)


ngeiswei:
Still crashes for me but less frequently (from 1 minutes to 5 minutes).

ngeiswei:
BTW, the last log of my last crash was

```
[2016-07-29 11:48:15:389] [DEBUG] [GenericShell] dtor, wait for eval threads.
```

At that point I'm not trying to debug anything, I'm working on my demo instead.

ngeiswei:
Unfortunately this is not the right way to fix things. Although it crashes much less often, scheme code tends not to be executed till the end. Like in the middle of some psi-rule action, the interpretation is aborted and states (like ProcessFinish) cannot be properly set.

We need a proper fix, and I'm just not knowledgeable enough of the cogserver to to rapidly come up with one.

ngeiswei:
I'm thinking a big lock over the scheme eval interpreter SchemeEval::do_eval and the dtor SchemeEval::~SchemeEval, but I don't think it's work because multiple processes (open-psi, etc) run permanently do_eval in the background without returning the hand... I may try anyway, but I'm not holding much hope.

ngeiswei:
I tried and it doesn't work at all... I don't know what to do...

yantrabuddhi:
Hi Nil,
 I tried all day to fix the problems but it seems I am fixing in wrong
places, it worked for me some times but now, i kind of give up on it. I
hope you or linas would find a better fix. I made a last pull request still
under check from jenkins, you can try that out if it doesn't fix it for you
still then i guess I couldn't properly patch up properly at all

linas:
Please don't make random changes that don't make any sense, and then merge them into the main repo. Please debug and describe the real problem, and make changes that actually fix it.  Random changes only create even bigger, greater issues later on -- this is how we got into this mess to begin with.  Lets stop making things worse.  This is not the correct path.

opencog/guile/SchemeSmob.cc
```diff
@@ -36,7 +36,7 @@ using namespace opencog;
  * simplify that code, and probably improve performance just a bit.
  */
 
-scm_t_bits SchemeSmob::cog_misc_tag;
+std::atomic<scm_t_bits> SchemeSmob::cog_misc_tag;
```
linas: This is a nonsense fix.  `cog_misc_tag` is set once and only once ever, making it atomic cannot possibly solve any problem that you are having.

yantrabuddhi: Hi Linas,
 I am not going to make any further changes, I think I made certain changes
which might not be sensible, and I am not sure of the right way to fix the
problem. I hope you can get head installed soon and see the problem
yourself as I have been blocked by the issue to complete my tasks for
monday demo.

opencog/guile/SchemeSmobAS.cc
```diff
@@ -277,7 +277,7 @@ SCM SchemeSmob::ss_as(SCM satom)
 /**
  * Return current atomspace for this dynamic state.
  */
-SCM SchemeSmob::atomspace_fluid;
+std::atomic<SCM> SchemeSmob::atomspace_fluid;
```
linas: This is also a non-sense fix!  Its a fluid, which is another name for TLS or thread-local-storage. It does not make sense to make TLS storage atomic.

Again, the key point is that this storage location is initialized once and only once, and is never written to again.   Because it is a fluid, every thread gets it's own unique copy, making it effectively atomic in each thread.  

Anyway, as far as I am aware, there is only one atomspace in use in the entire system, so all of the fluids have exactly the same atomspace, which is the same atomspace as the cogserver, so even the TLS locations always contain exactly the same data.  So, again, this fix doesn't make any sense.

PR #844, opened 2016-07-29
possible fixes with btree-psi

yantrabuddhi:
these fixes make behavior tree psi run longer however please check if cmt and face tracker don't crash for you

### DIFF
--- a/opencog/guile/SchemeEval.cc
+++ b/opencog/guile/SchemeEval.cc
@@ -362,9 +362,11 @@ SchemeEval::SchemeEval(AtomSpace* as)
 
 }
 
+static std::mutex prot_mtx;
 /* This should be called once for every new thread. */
 void SchemeEval::per_thread_init(void)
 {
+  std::lock_guard<std::mutex> lck(prot_mtx);
 	/* Avoid more than one call per thread. */
 	if (thread_is_inited) return;
 	thread_is_inited = true;
@@ -374,7 +376,7 @@ void SchemeEval::per_thread_init(void)
 #endif /* WORK_AROUND_GUILE_185_BUG */
 
 	// Arghhh!  Avoid ongoing utf8 fruitcake nutiness in guile-2.0
-	scm_c_eval_string ("(setlocale LC_ALL "")\n");
+//	scm_c_eval_string ("(setlocale LC_ALL "")\n");
 }
 
 SchemeEval::~SchemeEval()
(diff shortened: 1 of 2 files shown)


linas:
None of these changes make sense for any problem that I know of.  Can you open a bug report, desrbiing the problem?

You are putting mutexes into code that is single-threaded -- its impossible for that code (or its supposed to be imposible) for that code to ever run in more than one thread.  It's explicitly designed to be per-thread.

The atoms-ace pointer should never ever be null, unless someone somewhere is explicitly setting the atomspace to null. However, in normal operation, ther is always an atomspace.  

Thus, neither batch of fixes make sense.

yantrabuddhi:
Hi Linas,
 I was trying to find reasons for crashes in several places, following gdb
and returned null pointer for atomspace crashes, i might have made some
less sensible code additions but I got it to stop crashing on my machine
with btree-psi. The point was to try and get it running though i did a
hacked up job since I do not understand the whole architecture. If you see
my changes maybe you could know what might be causing errors in those
places, but there were errors in those places due to race conditions or
null pointers being returned.

linas:
Can you create a bug report with a description of the problem, and the stack traces? 

My first guess for the null pointers is that you are referencing freed memory or corrupt memory -- this would occur if deleted objects are being referenced.

Its also a rule of thumb to always look for bugs in new and recently changed code, instead of looking at bugs in very old code.  The scheme smob code has half-a-dozen multi-threaded unit tests that pund on it, and those test have been passing for many years.  The code has not changed at all, in many years -- it seems very unlikely to me that the root cause of the bug is there.

By contrast, the generic shell and socket code is new -- I made major changes to in in the last few weeks. That is the most likely location of the bug -- that's the right place to bug-hunt.

The other place to bug-hunt is in the space-server code, which is also brand-new and untested.  One obvious question to ask and answer is: does it still crash, if you disable the space-server?

yantrabuddhi:
Hi Linas,
 We tried testing by removing space server code initially however the
crashes still occurred whenever netcat was happening. In case the
genericshell stuff fixed it, i will try it again. Also I have tested the
space server code directly in guile by loading the scheme module and it
never crashed there. unless the cog-delete is deleting stuff and octomap
checking for it is causing a problem, i will test it once i go to office
again

linas:
Well, I pushed a fix yesterdya .. are you saying you have not tried that ifix?

ngeiswei:
Linas we have tried your fix, I'm pretty sure it has improved the situation, but it hasn't entirely fixed it.

Yes, we should create an issue (although reproducing it requires HEAD).

PR #851, opened 2016-08-03
Remove obsolete, troublesome code.

linas:
Remove code that depended on the cogserver config files.  It was always a source of trouble.  Get rid of the core source of pain.


linas:
This is necessitated by opencog/cogutils#54

PR #856, opened 2016-08-11
ReAdded Loading from Config for AttentionBank with Defaults

rTreutlein:


opencog/atomspace/AttentionBank.cc
```diff
@@ -36,16 +38,15 @@ AttentionBank::AttentionBank(AtomTable& atab, bool transient)
...
-    LTIAtomWage = config().get_int("ECAN_STARTING_ATOM_LTI_WAGE");
-#endif
+    startingFundsSTI = fundsSTI = config().get_int("STARTING_STI_FUNDS",100000);
+    startingFundsLTI = fundsLTI = config().get_int("STARTING_LTI_FUNDS",100000);
+    stiFundsBuffer = config().get_int("STI_FUNDS_BUFFER",10000);
+    ltiFundsBuffer = config().get_int("LTI_FUNDS_BUFFER",10000);
+    targetLTI = config().get_int("TARGET_LTI_FUNDS",10000);
+    targetSTI = config().get_int("TARGET_STI_FUNDS",10000);
+    STIAtomWage = config().get_int("ECAN_STARTING_ATOM_STI_WAGE",10);
+    LTIAtomWage = config().get_int("ECAN_STARTING_ATOM_LTI_WAGE",10);
```
linas: This is OK, but you may want to ponder as to whether it might be better to store these config parameters in the atomspace itself. Not required, but it would provide a uniform API to such things.

PR #860, opened 2016-08-16
Changed Recursive to Normal Mutex in AttentionBank

rTreutlein:
Looking at the code I see no reason why this needs to be a recursive mutex. @linas please take a quick look and merge if it is okay.


linas:
the atomspace uses recursive mutexes because, while inserting a link, it might need to first insert the link's outgoing set, and it wants to continue to hold the lock while it does this.  I guess that does not affect things here.

PR #861, opened 2016-08-16
Async Atom addition/removal for ImportanceIndex

rTreutlein:
This requires a change in the cog utils namely an async_caller which accepts functions that take a "const" argument. Haven't found a nice way to do that without a lot of duplicate code.

@linas please take a look at this. If it's okay I can merge this once I have the changes to cog utils merged.
The code is basically a mirror of what already exists for the AtomTable.


rTreutlein:
I just checked and I could convert all uses of async_caller to async_caller_const. Atomspace still builds and passes tests. Could we need to call a non-const function from the async_caller?

linas:
OK to merge, I guess. So, does this solve opencog/opencog#2354 as discussed in opencog/opencog#2333 ? That is, if the `std::this_thread::sleep_for(std::chrono::nanoseconds(1));` is now removed, then things should still work OK, and the slow-chat-response should no longer happen.  That's the theory ... sometimes the theory can be wrong.

linas:
>  Could we need to call a non-const function from the async_caller?

That can be added easily enough, if it is ever needed.

PR #867, opened 2016-08-30
Extend  scm binding function signature list.

misgeatgit:
Needed for appending a boolean flag to nlp-fuzzy-match function argument list.

### DIFF
--- a/opencog/guile/SchemePrimitive.h
+++ b/opencog/guile/SchemePrimitive.h
@@ -107,6 +107,7 @@ class SchemePrimitive : public PrimitiveEnviron
 			Handle (T::*h_hhh)(Handle, Handle, Handle);
 			Handle (T::*h_hs)(Handle, const std::string&);
 			Handle (T::*h_htq)(Handle, Type, const HandleSeq&);
+			Handle (T::*h_htqb)(Handle, Type, const HandleSeq&, bool);
 			Handle (T::*h_hz)(Handle, size_t);
 			Handle (T::*h_sddd)(const std::string&,double,double,double);
 			Handle (T::*h_sh)(const std::string&,Handle);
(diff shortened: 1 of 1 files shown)


linas:
Please please please stop doing things like this!  This is a terrible abuse of the intent of having scheme bindings, and of opencog API design, in general.  Its going in the wrong direction, for opencog.

The overall philosophy should be that almost everything in opencog should be expressible in atoms and atomese. The reason for this is that we want various automated learning algorithms to be able to learn and manipulate atoms, and atomese is the *only* way in which this can be done!   

For a learning algorithm such as MOSES, or the pattern miner, these algos know how to crawl the atomspace, and change it in various ways.  Just like relex2logic can change the atomspace, or just like PLN can do forward-backward chaining in atomese.   Or even  open-psi.

Note that all of these are kind-of-like "programmming in atomese".  NONE of these subsystems know how to write code in scheme, or in C++ or in python. ALL of these systems know how to program in only one language: atomese. 

Whenever you put something in SchemePrimitve.h, you make it IMPOSSIBLE for moses, pattern-miner, relex2logic, pln or open-psi to edit, alter or change or even use the API you've just created.  The only things in the system that know how to use SchemePrimitive.h are humans -- human programmers.  Our goal, in opencog, is to build a system that does NOT require humans to write the code for it.  

Thus, any and every extension to SchemePrimitve.h is just one more breakage of the core philosophy of opencog.  Please try to avoid doing this -- it may seem easy now, but it will just be lots of pain later on.

This above note is for everybody -- including @yantrabuddhi who was the last person to make major extensions here, against my protests.  Mandeep -- don't forget -- you still have a todo item to figure out how to remove from SchemePrimitve.h all the stuff that you added.

misgeatgit:
@linas As you know most of the NLP work is done inside the scheme code and it utilizes some codes written in C++, like the fuzzy matcher, URE and Pattern matcher via the scheme bindings. Or are you saying in this particular PR the last argument should have been a boolean atom type instead of  a c++ boolean type?

linas:
I'm just saying that if it is a complicated API that needs an addition to SchemePrimitive.h then it is probably too complicated.  I have not looked carefully at this particular use case -- there are always exceptions --  but pretty much any API that takes anything more than a list of atoms is too complicated.

PR #869, opened 2016-09-07
Compute minMax STI updating inside the stimulate method.

misgeatgit:


linas:
Somebody who is working with the attention subsystem should review and comment as appropriate. Maybe @rTreutlein  ?

rTreutlein:
The Min/Max STI should be computed in the AVChanged Methode, not the stimulate Methode, since the stimulate Method is not the only way to change the STI.

misgeatgit:
@rTreutlein Where is the AVChanged method? and shouldn't agents be prohibited from directly changing AVs?

rTreutlein:
@misgeatgit: Well yes most agents are prohibited from changing the STI directly but not the Agents of the ECAN system. Rent and AttetionSpreading Agents for example modify the STI.
And the methode is here:
https://github.com/opencog/atomspace/blob/master/opencog/atomspace/AttentionBank.cc#L95

misgeatgit:
@rTreutlein I've moved the code to the AVChanged method now.

rTreutlein:
This looks fine now.

PR #875, opened 2016-09-09
Throw errors, instead of using NullTruthValue

linas:
This should help expose the core bug that is being hit in issue opencog/ros-behavior-scripting#102


rTreutlein:
@linas did you notice that the ci build is still running? It seems one of the tests produced an infinite loop. I have aborted it after it reached a runtime of 2 days and 7 hours.

PR #877, opened 2016-09-12
BC substance part I

ngeiswei:
This merge breaks the Backward Chainer (which was already broken, but it breaks it even more!).

I decided to merge now because it starts interacting to much with other parts, and I don't want to create too much mess by differing merging for too long.

I'm gonna wait for 12 hours before merging, in case someone relies on the old version (which I seriously doubt)..


ngeiswei:
Well, querying my memory a little bit more makes me more confidence that nobody currently relies on the old BC, so I'm merging now.

PR #878, opened 2016-09-15
Fix unintentionally small sized index.

misgeatgit:
A lot of atoms were falling in the same bin since the number of bins was only 32, which seems unintentional according to the comment left in the code. The actual intention was to have 2048 bins  holding atoms with STI range of 32 where 2048 bin size is the result of dividing 2^sizeof(short) by 32 assuming size of short is 2 bytes. So, the fix is based on my above understanding.

### DIFF
--- a/opencog/atomspace/ImportanceIndex.cc
+++ b/opencog/atomspace/ImportanceIndex.cc
@@ -31,7 +31,7 @@
 using namespace opencog;
 
 //! Each bin has STI range of 32 means 2048 importance bins.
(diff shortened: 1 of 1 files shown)


rTreutlein:
As far as I understand the code there were 2048 bins the problem was just that it only used the first 32.
Also, I really don't like setting a variable with value like this (1 << 5). Sure I know that's a shift operation but if you would just write 32 that's a lot easier to comprehend.

Anyway, that looks like what was originally intended.

Issue #874, opened 2016-09-08
Unification with variables on both sides


ngeiswei:
# Overview

I need a function that would perform unification between atoms that both contain variables.
# Context

The Backward Chainer requires that. The current BC has a unification function `BackwardChainer::unify` that is supposed to do that but does not.
# Specification

I'll describe the specification by examples. I'm giving a simplified type signature of the function, specifically I'm letting aside variable declarations, but they should of course be taken into account in the final implementation.
## Type signature

The type signature of the function is

``` c++
HandleMapSeq unify(Handle, Handle)
```
## Examples

1.

```
unify((Variable "$X"), (Concept "A"))
->
[{(Variable "$X"):(Concept "A")}]
```

2.

```
unify((Concept "A"), (Concept "$X"))
->
[{(Variable "$X"):(Concept "A")}]
```

3.

```
unify((Inheritance (Concept "A") (Concept "B")), (Variable "$X"))
->
[{(Variable "$X"):(Inheritance (Concept "A") (Concept "B"))}]
```

4.

```
unify((Inheritance (Concept "A") (Variable "$Y")),
      (Inheritance (Variable "$X") (Concept "B"))
->
[{(Variable "$X"):(Concept "A"),
  (Variable "$Y"):(Concept "B")}]
```

etc. I didn't give an example with multiple mappings but they would occur if for instance one uses an unordered link, thus containing multiple possible permutations. I think it's enough to give the idea.
# How to code it

So assuming such a function doesn't exist (I think not), the question is, how should it be implemented? My spontaneous answer is to run 2 instances of pattern matching, one where `lhs` is considered as pattern and `rhs` is put in a temporary atomspace as grounding, and vice versa (`rhs` is considered as pattern and `lhs` is put in a temporary atomspace as grounding), run these 2 instances of pattern matching and build the union of their mappings going both ways. The `node_match` and `variable_match` callbacks will have to be overloaded so that unification as described in example 4. can be possible (namely, a node needs to accept a variable of the same or inherited type as a match, etc).

I'm feel my spontaneous suggestion has shortcomings and will not encompass all cases, but might be good enough for a while. Any better suggestion is welcome.

linas:
I believe that there is a far more straight-forward implementation (based on your description): just compare the left and right sides, side-by-side, recursively. If one is a variable, and one is not, that's your unification.

The pattern matcher is kind-of overkill for this, because it tries to unify lists of clauses: i.e. 
```
[...]

PR #884, opened 2016-09-20
Implement both way unification

ngeiswei:
Supports all unit tests in issue #874 except this one https://github.com/opencog/atomspace/issues/874#issuecomment-247238510 which requires to support more kinda type deep.

The solution set is a set of typed partitions, where each typed partition is a map from *equality atom set* (atoms that must be equal in order to be unifiable) to type.

Adding substitutions to completely fulfill the unification should be trivial given the solution set, just a matter of selecting the substitutions out of the equality partitions according to some fitness, like minimizing the number of variables, etc.


ngeiswei:
@linas, there's a bunch of type checking related functions, these should ultimately be moved to the right place like `opencog/atomutils/TypeUtils.h` or such, not sure. Since I'm not sure I need to support full unification right now I didn't bothered studying your deep type code, it only relies on types defined by TypeNode and TypeChoice, nothing more.

ngeiswei:
Oh, and it doesn't support unordered links. This can't wait, thus my next task.

PR #887, opened 2016-09-21
Added way to get Atoms in the highest/lowest bin of the ImportanceIndex

rTreutlein:
This is needed for cheaply figuring out min and max STI values.

### DIFF
--- a/opencog/atomspace/ImportanceIndex.cc
+++ b/opencog/atomspace/ImportanceIndex.cc
@@ -28,6 +28,8 @@
 #include <opencog/atoms/base/Atom.h>
 #include <opencog/atomspace/AtomTable.h>
 
+#include <boost/range/adaptor/reversed.hpp>
+
 using namespace opencog;
 
 //! Each bin has STI range of 32 means 2048 importance bins.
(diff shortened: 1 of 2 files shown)


rTreutlein:
Didn't actually want both these commits in this pull request. But if there is no issue with either one we can just merge like this. If there is ill just have to rewrite history on my branch.

opencog/atomspace/ImportanceIndex.cc
```diff
@@ -123,3 +146,35 @@ UnorderedHandleSet ImportanceIndex::getHandleSet(
...
+        }
+    }
+
+    return ret;
+}
+
+UnorderedHandleSet ImportanceIndex::getMinBinContents()
+{
+    UnorderedHandleSet ret;
+    for (AtomSet s : idx)
```
linas: I am guessing that what you really want is to get rid of the for-loop, and just set
```
const AtomSet& s = idx[0];
```
and for the other one, 
```
const AtomSet& s = idx[IMPRTANCE_MAX_BINS-1];
```

rTreutlein: What I want to get is the smallest/biggest bin that is not empty. So iterate over all the bins starting from the bottom/top and if I find a bin that is not empty I immediately return that. Maybe you missed the return at line 159.

rTreutlein:
@linas: Yes this is an exponential bin size setup. But so that the size doesn't grow too quickly each binSize is part of a group that contains (in this case) 8 bins with this Size.

This bias the Index towards lower STI values and allows us to initialize the STI with a small random value so that not all new Atoms end up in the same bin.

Why exactly do you think this is not going to work? It seemed to do what I wanted it to when I was experimenting with it.

opencog/atomspace/ImportanceIndex.cc
linas: yes I did. OK

PR #892, opened 2016-09-24
Attempt to fix the python pipe bug

linas:
See opencog/ros-behavior-scripting/issues/108#issuecomment-249342902


linas:
never mind. Tis is wrong, I just reverted this.

PR #894, opened 2016-09-26
Optimize unify

ngeiswei:
Optimization for #874. Unfortunately `test_unify_unordered_8` (formerly `test_unify_unordered_7`) is only 2x or so faster, going from 1min to 20sec.

I don't think I can do better on the algorithmic side. I might be able to speed that up further by doing C++ optimizations but I'll only look into that if it is a problem in practice.

Something that I might do is add a `unifiable` function that only test satisfiability as opposed to finding all possible unifications. This should save a lot of time, but again better do that based on real use cases.

### DIFF
--- a/opencog/atomutils/Unify.cc
+++ b/opencog/atomutils/Unify.cc
@@ -43,6 +43,10 @@ UnificationSolutionSet unify(const Handle& lhs, const Handle& rhs,
                              const Handle& lhs_vardecl,
                              const Handle& rhs_vardecl)
 {
+	///////////////////
+	// Base cases    //
+	///////////////////
+
 	// Make sure both handles are defined
 	if (lhs == Handle::UNDEFINED or rhs == Handle::UNDEFINED)
 		return UnificationSolutionSet(false);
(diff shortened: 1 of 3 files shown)


linas:
unordered links are stunningly unpleasent to work with.

I cannot read through your code just right now, but keep this in mind: an unordered link with zero variables can always be given a canonical order, just by comparing pointer addresses in RAM.  Of course, there is nothing to unify in that case.  However, to unify a link with zero variables to one with one variable, place both in canonical order, and then that one variable can appear in just one slot at a time: comparing the two takes linear time, comparing all slots takes worst-case N^2 time, but probably NlogN average. This is a lot better than trying all N-factorial reorderings.

Perhaps you do this already, I don't know.

linas:
The NlogN would be for unifying just one variable. For two variables, you'd have to do this recursively, so it would be N choose k approx-eq N^k (times logish-powers of N) for k variables.

ngeiswei:
Thanks Linas. The algorithm does not do that at the moment. But I decided not to optimize further till I have a real case set, as I feel subsequent optimizations will trade speed for code elegance, which I'd rather avoid unless it is measurably worth it.

PR #900, opened 2016-09-29
Revert "Fix scopelink alpha equivalence"

### LINKED ISSUE #896: Bug in ScopeLink::is_equal
Due to `ScopeLink` assuming only a single body`ScopeLink::is_equal` fails to properly check the alpha-equivalence of scope links with multiple bodies, such as bind links. 

I suggest to replace
```c++
Handle ScopeLink::_body
```
by
```c++
HandleSeq ScopeLink::_bodies
```
and update `ScopeLink::is_equal` appropriately. This might also simplify the code of the classes inheriting from ScopeLink as we won't need special treatments for scope links with more than 1 body.

ngeiswei:
Reverts opencog/atomspace#899

I've merged that one too fast because I wanted to work on top of it. I've just realized that there is a much simpler way to fix #896 .

### DIFF
--- a/opencog/atoms/core/PutLink.cc
+++ b/opencog/atoms/core/PutLink.cc
@@ -104,14 +104,15 @@ void PutLink::init(void)
 		// If the body is just a single variable, and there are no
 		// type declarations for it, then ScopeLink gets confused.
 		_vardecl = Handle::UNDEFINED;
-		set_body(_outgoing[0]);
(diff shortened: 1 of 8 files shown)


ngeiswei:
BTW, the motive of my fast merge was really wrong, I didn't need to pollute the master just so I can work on top of it, I could have done that locally. Yet another case of misplaced laziness on my part...

PR #901, opened 2016-09-29
Fix for #896 + unit tests

ngeiswei:
@linas, this one is much simpler and I believe harmless but I let you review it and merge.


ngeiswei:
It's been 4 days, my next PR will depend on that code, it's really harmless so I think I can merge.

PR #902, opened 2016-09-29
Alpha conversion

ngeiswei:
@linas please have a look at this PR.

It implements `ScopeLink::alpha_conversion` that performs alpha-conversion with randomly generated new variables (maybe I should name it `rand_alpha_conversion`).

You may look in particular at `ScopeLink::factory`. I'm a bit unconfident about it as it's somewhat redundant with `AtomTable::factory` and I'm not sure how all this is supposed to work together. I would say perhaps `AtomTable::factory` could be decomposed into `FunctionLink::factory`, `ScopeLink::factory`, etc, but I'm really not sure, and it's another issue anyway, if it is at all.


ngeiswei:
retest this please

ngeiswei:
Closing, will make another one.

PR #904, opened 2016-10-03
Alpha conversion 2

ngeiswei:
Implements ScopeLink::alpha_conversion that performs alpha-conversion with randomly generated new variables (maybe I should name it rand_alpha_conversion).

@linas you may look at `ScopeLink::factory`. I'm a bit unconfident about it as it's somewhat redundant with `AtomTable::factory` and I'm not sure how all this is supposed to work together. I would say perhaps `AtomTable::factory` could be decomposed into `FunctionLink::factory`, `ScopeLink::factory`, etc, but I'm really not sure, and it's another issue anyway, if it is at all.

### DIFF
--- a/opencog/atoms/core/ScopeLink.cc
+++ b/opencog/atoms/core/ScopeLink.cc
@@ -178,7 +178,8 @@ bool ScopeLink::is_equal(const Handle& other) const
 	if (n_scoped_terms != other_n_scoped_terms) return false;
 
 	// Variable declarations must match.
-	if (not _varlist.is_equal(scother->_varlist)) return false;
+	if (scother == nullptr or not _varlist.is_equal(scother->_varlist))
+		return false;
 
 	// Other terms, with our variables in place of its variables,
 	// should be same as our terms.
--- a/opencog/atoms/core/ScopeLink.cc
+++ b/opencog/atoms/core/ScopeLink.cc
@@ -29,6 +29,11 @@
 #include <opencog/atoms/TypeNode.h>
 #include <opencog/atoms/core/FreeLink.h>
 #include <opencog/atoms/core/LambdaLink.h>
+#include <opencog/atoms/core/PutLink.h>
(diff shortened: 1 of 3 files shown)


linas:
well, I think I mentioned this before, but my general impression is that explict alpha conversion is a design mistake.

I'm behind on email, my graphics card is overheating/dying ... I'm hoping an earlier email provides some rational for why this is a good idea, and explains the problem being solved.  Once you go down this road, getting back out will be hard.

ngeiswei:
Imagine you have 2 programs and you merge them to create a third program, just for the sake of avoiding name collision you'd want to explicitly alpha convert them so they have different variable names. Otherwise, yes, I agree, implicit alpha conversion is better.

linas:
OK, I read the code now. I see. I have no objection to this.  Sorry.  Clearly, my failing graphics card is having a negative effect on my mental state.

One very minor remark, then: suggest that the alpha_convert function accept a suggested variable name, and generate a random one, only if one is not provided. Also suggest using the existing utility for creating randomly-named nodes:
```
Handle add_prefixed_node(AtomSpace&, Type, const std::string& prefix = "");
```

linas:
Another remark: you do NOT need an explicit factory: there already is a factory, its the constructor
```
ScopeLink::ScopeLink(Type t, const Handle& body,
                     TruthValuePtr tv, AttentionValuePtr av)
```

which, given the type t will do exactly the same thing as what the factory does .... and its future-proof.

linas:
Err, my last remark is not quite right. Hmmm.  I'm confused.

ngeiswei:
Thanks for the tip, I didn't know about `add_prefixed_node`. If I were to use it inside `ScopeLink::alpha_conversion` it would have to take an atomspace in argument and it would be guarantied to generate a unique name, which is very good, but it's also much slower and might loops forever if no name is available (which is extremely unlikely at it requires an atomspace at least 60^8 nodes).

PR #905, opened 2016-10-04
Deploy atom factories on a broader scale.

linas:
Now that everyone is using factories, we may as well use them correctly, everywhere.  Including the atomspace.


ngeiswei:
Great!

PR #918, opened 2016-10-10
Revert "Add utest failing ScopeLink::is_equal"

ngeiswei:
Reverts opencog/atomspace#917

Forgot to re-enable the other tests...

### DIFF
--- a/tests/atoms/AlphaConvertUTest.cxxtest
+++ b/tests/atoms/AlphaConvertUTest.cxxtest
@@ -44,17 +44,16 @@ public:
 
 	void tearDown() {}
 
-	// void test_untyped_variable();
-	// void test_variable();
-	// void test_body();
-	// void test_atomspace();
-	// void test_bindlink();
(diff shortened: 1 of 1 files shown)


tests/atoms/AlphaConvertUTest.cxxtest
```diff
@@ -44,17 +44,16 @@ public:
...
-	// void test_untyped_variable();
-	// void test_variable();
-	// void test_body();
-	// void test_atomspace();
-	// void test_bindlink();
-	// void test_rand_alpha_conversion();
-	// void test_names_alpha_conversion();
-	// void test_vardecl_alpha_conversion();
-	// void test_vardecl_bindlink_alpha_conversion();
-	// void test_values_alpha_conversion();
```
linas: if you just change the first letter to not be "t" then cxxtest doesn't think they are tests, and won't run them.  This might be easier than commenting out code.

Issue #920, opened 2016-10-10
Bug in Link::operator==


ngeiswei:
`ScopeLink::is_equal` does not robustly handle unordered links. To reproduce the bug run `AlphaConvertUTest::test_unordered` added in commit 4b37707.

EDIT: actually the bug is in `Link::operator==` see the comments below.

ngeiswei:
Replacing in Handle.cc l.155

```
#define DEFAULT_ATOMS_LESS atoms_less
```

by

```
#define DEFAULT_ATOMS_LESS content_based_atoms_less
```

fixes the problem. Although the problem is that it might slow down a bit the atomspace. I'm gonna try to measure that now. Another option would be to use the content_based_atom_less comparison only for scopes, I'll look into it...

ngeiswei:
I know using the unit tests for profiling is bad but replacing `atoms_less` by `content_based_atoms_less` goes from 52.471 sec to 50.819 sec in average (based on 7 trials each).

ngeiswei:
OK, my suggestion is to code in `Atom.h` and `Link.h` an alternate `operator==` that would compare the children of unordered links where their children are sorted according to a content based comparison. This should create absolutely no slow down for anything else than callers of `ScopeLink::is_equal`. Of course `ScopeLink::is_equal` will be slower, but I don't see another way unless we use `content_based_atoms_less` all over the place.

ngeiswei:
I know!!! In fact `Link::operator==` is buggy, links that have no UUID yet SHOULD have their children sorted based on content before being compared!

PR #921, opened 2016-10-10
Fix unordered is equal

ngeiswei:
Fix for issue #920 

I'm rather confident so I'll merge it tomorrow morning European time, but @linas you might want to have a look at it.

### DIFF
--- a/opencog/atoms/base/Handle.h
+++ b/opencog/atoms/base/Handle.h
@@ -277,8 +277,6 @@ struct content_based_handle_less
     }
 };
 
-typedef std::set<Handle, content_based_handle_less> ContentBasedOrderedHandleSet;
-
 struct handle_seq_less
 {
     bool operator()(const HandleSeq& hsl, const HandleSeq& hsr) const
--- a/opencog/atoms/base/Handle.h
+++ b/opencog/atoms/base/Handle.h
@@ -62,6 +62,7 @@ friend class Atom;
 friend class AtomTable;
 friend class AtomStorage;         // persistance
 friend class content_based_atom_ptr_less;
+friend class content_based_handle_less;
 
 private:
     AtomPtr _ptr;
@@ -122,6 +123,10 @@ public:
 
     // Cython can't access operator->() so we'll duplicate here.
     inline Atom* atom_ptr() {
+        return _ptr.get();
+    }
+
+    inline const Atom* const_atom_ptr() const {
         return _ptr.get();
     }
 
@@ -262,6 +267,17 @@ struct content_based_atom_ptr_less
         return Handle::content_based_atoms_less(al, ar);
     }
 };
+
+struct content_based_handle_less
+{
+    bool operator()(const Handle& hl, const Handle& hr) const
+    {
+        return Handle::content_based_atoms_less(hl.const_atom_ptr(),
+                                                hr.const_atom_ptr());
+    }
+};
+
+typedef std::set<Handle, content_based_handle_less> ContentBasedOrderedHandleSet;
 
 struct handle_seq_less
 {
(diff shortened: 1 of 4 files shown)


ngeiswei:
Don't merge that just yet there is a subtle bug that would occur if an unordered link contains duplicated atoms.

opencog/atoms/base/Handle.h
```diff
@@ -263,6 +268,17 @@ struct content_based_atom_ptr_less
     }
 };
 
+struct content_based_handle_less
+{
+    bool operator()(const Handle& hl, const Handle& hr) const
+    {
+        return Handle::content_based_atoms_less(hl.const_atom_ptr(),
+                                                hr.const_atom_ptr());
+    }
```
linas: Why are you passing Atom* here? why not just pass in the handles themselves?  Its kind-of dangerous to pass around raw, naked Atom* -- which is why the API mostly doesn't have Atom*'s in it anywhere -- except for this new usage.

opencog/atoms/base/Link.cc
```diff
@@ -137,21 +137,69 @@ bool Link::operator==(const Atom& other) const
...
+        it = lhs.begin(),
+        other_it = rhs.begin();
+
+    for (; it != lhs.end(); it++, other_it++)
+    {
+        if ((*it)->getType() != (*other_it)->getType())
+            return false;
+
+        if ((*it)->isNode())
+        {
```
linas: This is not required -- PrototAtom has a virtual `operator==()` in it already. So it should be enough to just say `*lhs == *rhs`  and be done with it. You can skip the type comparison, above, also.

The core issue turns out to be that NodeCast and LinkCast are insanely expensive.

opencog/atoms/base/Link.cc
```diff
@@ -137,21 +137,69 @@ bool Link::operator==(const Atom& other) const
...
+            LinkPtr ol(LinkCast(other_outgoing[i]));
+            if (*tl != *ol) return false;
+        }
+    }
+    return true;
+}
+
+bool Link::content_based_equal(const ContentBasedOrderedHandleSet& lhs,
+                               const ContentBasedOrderedHandleSet& rhs) const
+{
```
linas: I mean, maybe if I studied it hard, I could figure it out, but then I will forget, and the next person won't understand this, either...

opencog/atoms/base/Link.cc
ngeiswei: I've merely copied/pasted this code from the previous `Link::operator==`. But before I did that I tried to return `lhs[i] == rhs[i]` and for some reason it broke the unit tests, so I concluded that there's must be a reason for that and let it there. I'll investigate though.

ngeiswei: I found the problem, Handle::operator== directly compare the pointers, see https://github.com/opencog/atomspace/blob/master/opencog/atoms/base/Handle.h#L153 this is wrong too. Maybe I should fix it as well with a similar fix, that is if one of them have undefined UUID they might have different pointers but same content.

ngeiswei:
I'm merging. I have let `Handle::operator!=`, I fear it is buggy but I prefer to address when a problems will appear due to it.

opencog/atoms/base/Link.cc
linas: Ugh.  I made some progress yesterday in removing UUID's. (They will remain in SQL, and in anything else that needs them -- e.g. the rest server, the zmq server)

I retract the comment that started this conversation: it *should* be enough to just call the overloaded operator== but it turns out that this is difficult to implement correctly, for C++-ish reasons, so changing this is probably a low priority. The code works, for now.

PR #922, opened 2016-10-11
Remove UUID's from python bindings

linas:
While waiting for a compile to finish, I started work on this long-overdue task.  Requires a matching puull to opencog.


linas:
Requires a matching pull req opencog/opencog#2457

PR #924, opened 2016-10-11
Complete unify target

ngeiswei:
Code for unifying a rule conclusion with a target and return all its valid rule variations.


opencog/atomutils/TypeUtils.cc
```diff
@@ -258,32 +258,39 @@ Handle opencog::filter_vardecl(const Handle& vardecl, const Handle& body)
...
-	// Base case
+	// Base cases
+
+	if (vardecl.is_undefined())
+		return Handle::UNDEFINED;
+
 	Type t = vardecl->getType();
-	if (VARIABLE_NODE == t) {
-		if (is_unquoted_unscoped_in_any_tree(hs, vardecl))
+	if (vardecl->isNode()) {
```
linas: This adds extra CPU overhead, for no reason at all.

PR #926, opened 2016-10-11
Fixing opencog issue #2315 and #2322

Dagiopia:
fixing python shell handling and printing issue. 

The shell handling had problems with scopes. 
Printing wasn't working for obvious reasons.


linas:
Thanks for the work, but this is the wrong place to fix this issue!  The correct place to fix this issue is in the python shell, in the opencog/opencog repository.

The code here, the "python evaluator", is used to evaluate code in various embedded contexts, where there is no I/O at all, -- there is not stdout -- and thus print statements don't make sense, cause there is nowhere for them to go to.

The "python shell" is a subsystem of the cogserver: its a block of code that provides telnet (tcpip) access to a python command line. That shell uses this evaluator to perform the evaluation of python code ... and it is there, where the output of print statements need to be returned to the user.  ... and so it is there where the fix needs to be made.

Note, in particular, that the handling of blank lines is completely different in these two situations: when python is evaluated from a file, blank lines do NOT mean "end of function", and since this evaluator is used to load files, the changes you made would break existing code.  The only place where blank-lines mean "end of function" is at the shell prompt, so again -- that fix can only be made to the shell code, and not to this code.

Note that one of the difficulties is that some people want to write infinite loops, with print statements in them; for example the "agents" written in python are infinite loops, and are started in new python threads, and have prints in them.  Those print outputs need to be captured, and sent back on the shell socket.  

By contrast, there is no socket at all, in this code, so the dup2 does not make sense -- the only reason that dup2 is needed is to capture stdout, and send it across tcpip to the telnet connection.  If there is no socket, then there is no reason to use dup2!

Dagiopia:
yes the evaluator is the wrong place to work on outputs but the problem is that its the PyRun_StringFlags that is doing the outputing on stdout and it's this output that needs to be captured. any way I was wrong to just move the code in GenericShell without any modifications. 

about the shell handling issue though, I think this is the correct fix... U said ur self that modules should be loaded by python's own 'import xxx' when getting rid of loadpy. I also checked with scheme's (python-eval execfile) and it works fine because python script file that may have multiple empty lines within scopes goes to be processed directly by python's evaluator and the PythonEval in atomspace doesn't have to do the parsing. The pythonEval is only responsible for the python shell

Dagiopia:
I have been thinking of fixing the python output problem for more than a couple of months now and there doesn't seem to be a non-hackish way even that will not be enough... and since it's a singleton instance I don't know how one instance will be separate from others.

Issue #910, opened 2016-10-05
Feature or bug?


AmeBel:
### Content of `test.scm`

```scheme
(use-modules (opencog))
(use-modules (opencog query))

;; -----------------------------------------------------------------------------
;; Helper functions

(define (content-1)
    (MemberLink
      (ConceptNode "Socrates@81011e61-27a7-4001-a63b-3b569478bced")
      (SatisfyingSetLink
         (VariableNode "$X")
         (EvaluationLink
            (PredicateNode "breathe@ea723bda-70bb-47c0-8930-b344fb47a4d1")
            (ListLink
               (VariableNode "$X")
               (ConceptNode "air@76357ee3-d334-4b5f-b49e-323b294310b6")
            )
         )
      )
    )
)

(define (content-2)
; MemberLink having the same Variable name as rule
    (MemberLink
      (ConceptNode "Socrates@81011e61-27a7-4001-a63b-3b569478bced")
      (SatisfyingSetLink
         (VariableNode "$X-M2E")
         (EvaluationLink
            (PredicateNode "breathe@ea723bda-70bb-47c0-8930-b344fb47a4d1")
            (ListLink
                (VariableNode "$X-M2E")
               (ConceptNode "air@76357ee3-d334-4b5f-b49e-323b294310b6")
            )
         )
      )
    )
)

;; -----------------------------------------------------------------------------
; From "opencog/pln/rules/member-to-evaluation-rule.scm"
(define (member-to-evaluation-2-1-rule)
    (BindLink
        (VariableList
          (VariableNode "$B")
          (VariableNode "$C")
          (TypedVariableLink
             (VariableNode "$D")
             (TypeNode "PredicateNode")
          )
        )
        (MemberLink
          (VariableNode "$B")
          (SatisfyingSetLink
             (VariableNode "$X-M2E")
             (EvaluationLink
                (VariableNode "$D")
                (ListLink
                   (VariableNode "$X-M2E")
                   (VariableNode "$C")
                )
             )
          )
        )
        (ExecutionOutputLink
          (GroundedSchemaNode "scm: member-to-evaluation-formula")
          (ListLink
             (EvaluationLink
                (VariableNode "$D")
                (ListLink
                   (VariableNode "$B")
                   (VariableNode "$C")
                )
             )
             (MemberLink
                (VariableNode "$B")
                (SatisfyingSetLink
                   (VariableNode "$X-M2E")
                   (EvaluationLink
                      (VariableNode "$D")
[...]

PR #927, opened 2016-10-11
Prevent duplicate scoped links from being placed into the atomspace

linas:
This is a fix for bug #910.

This is a fairly serious, important change: amy time that you have a link type that is scoped (e.g. lambda-link, bind-link, etc.0 and you place it into the atomspace, then if there is another link already in there, that is exactly alike, except for different variable names, then you will get back the equivalent link (possibly with different, "alpha-renamed" variable names).


linas:
p.s. this pull require requires opencog/cogutils/pull/60 as a pre-requisite

williampma:
:+1:

PR #938, opened 2016-10-16
Removes UUIDs from the Haskell Interface

rTreutlein:
Also upgrades to GHC 8.0.1 and Stackage LTS-7.3

UUIDs have been replaced by pointers to Handle.


rTreutlein:
There seems to be a bug somewhere got to fix that first before this can be merged.

opencog/haskell/OpenCog/AtomSpace/Api.hs
```diff
@@ -57,51 +59,47 @@ foreign import ccall "AtomSpace_addNode"
   c_atomspace_addnode :: AtomSpaceRef
                       -> CString
                       -> CString
-                      -> Ptr UUID
+                      -> Handle
```
linas: was ptr UUID, why isn't it Ptr Handle? Maybe this is the bug?

linas: Unless Handle in Haskel is always the same thing as Handle* in C++ -- because I see this everywhere in the code.  If so, the naming convention is confusing.

opencog/haskell/OpenCog/AtomSpace/Internal.hs
```diff
@@ -7,23 +7,24 @@
...
 import Data.Functor                  ((<$>))
 import Data.Typeable                 (cast,Typeable)
 import OpenCog.AtomSpace.Sugar       (noTv)
 import OpenCog.AtomSpace.Types       (AtomName(..),AtomType(..)
                                      ,Atom(..),TruthVal(..))
 
--- Data type to hold atoms's UUID.
-type UUID = CULong
+type Handle = Ptr ()
+type HandleSeq = Ptr Handle
```
linas: Ah, OK< this answers my earlier question/comment

PR #941, opened 2016-10-17
Backward Chainer Subtance

ngeiswei:
That's the bulk of the Backward Chainer code. Still not functional due to a bug in the AtomSpace (in `Handle::operator==` when comparing atoms not in any atomspace), and probably other bugs.


ngeiswei:
@linas I'm merging but you may want to have a look at these 2 commits

https://github.com/opencog/atomspace/pull/941/commits/a582ca3f0d86819f21fba595c736f6f45cb95275
this one is rather inoffensive and pretty useful (I was surprised it didn't already exist).

and

https://github.com/opencog/atomspace/pull/941/commits/094fc569299de523fa69de14ebc615b922a9a61d
this one might be more harmful, it doesn't break any unit test but changes the semantics of Variables::extend, so that the new type is the intersection of the 2 types to merge. I think it makes a lot of sense, and that is what we want when build the back inference tree, that is we want to fulfill the intersection between the variable declaration of a variable in the target and what a rule conclusion can offer. But I don't know if that's the behavior we want in general.

linas:
Yeah, a582ca3f0d86819f21fba595c736f6f45cb95275 is pretty cool. I'm wondering why it hasn't been implemented before, and the only reason I can think of is that, in most cases, that info is already available elsewhere, e.g. in the ScopeLik from which the Variables instance was created.

linas:
The change 094fc569299de523fa69de14ebc615b922a9a61d make sense, although there are two types of extensions one could do: take a union of variables and union of types, or take a union of variables but intersection of types.  Both are valid, depending on what you want.

The only current user of extend() is `query/InitiateSearchCB.cc` and I'm reading the code now ... hard to tell what it expects, but it would seem that intersecting the types, as you did, is probably the right thing to do.

linas:
You might want to add a unit test to make sure someone does not later undo the type intersection.  This is the kind of thing that might cause someone to want to change it someday.

PR #942, opened 2016-10-17
Fix handle equal

ngeiswei:
@linas I believe this is the right thing to do but please have a look at it.

### DIFF
--- a/opencog/atoms/base/Handle.cc
+++ b/opencog/atoms/base/Handle.cc
@@ -124,7 +124,7 @@ inline Handle Handle::do_res(UUID uuid)
     return Handle();
 }
 
-inline bool Handle::operator==(const Atom* other) const noexcept {
+bool Handle::operator==(const Atom* other) const noexcept {
 	// If both are defined and are in no atomspace or different
 	// atomspaces, compare them by content
 	if (_ptr.get() and other) {
--- a/opencog/atoms/base/Handle.cc
+++ b/opencog/atoms/base/Handle.cc
@@ -124,6 +124,19 @@ inline Handle Handle::do_res(UUID uuid)
     return Handle();
 }
 
+inline bool Handle::operator==(const Atom* other) const noexcept {
+	// If both are defined and are in no atomspace or different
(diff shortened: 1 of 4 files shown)


linas:
What problem does this solve? Recall that there are two or three different types of equality: semantic and syntactic are two, I'll explain the third, shortly. In this case, I think the two are being needlessly confused (this is partly the fault of C/C++, which (unlike other languages) never bothered to define these two different equalities.)

So: in opencog, two atoms are (syntacticallly) equal if and only if they have the pointers. that's it. (same uuid) 

Two atoms are equivalent (semantically equal) if they have the same contents (but maybe different TV's, AV's, or belong to different atomspaces, or no atomspace at all).

Finally, they are "equ" if you can upcast one into the other, with the intent of putting them in the same atomspace.

Trying to overload both, or all three types onto operator==() is going to cause trouble -- e.g. because operator==() will return true even if the TV and AV are different.  So -- can we just add a method `equivalent()` for semantic (content) equality?

The main issue I am worried about is that operator == is used all over the place in the atomspace, to figure out if we've got an atom or not, and the add/remove it to the atomspace, to add/remove it from various lists, etc.  If we change operator== now, we risk inroducing all kinds of subtle, hard-to-debug bugs.

ngeiswei:
Actually some tests like MapLinkUTest fail (explode in RAM consumption). I don't understand why but it shows that it is indeed problematic.

linas:
I can guess why: the atomspace does a lot of handle compares to figure out if it already is holding an atom. If it is not holding an atom, it does a recursive content-compare, and adds the atoms that are missing. The original atom is then deleted, when the user lets go of it. But if you fool the atomspace, then you will get hunderds or millions of "identical" atoms, but which are not actually the same atom, and they cannot ever be freed, because something somewhere points to them.

Just rename operator==() to `equivlanet()` or `content_compare()` or `cont_equality() or something, then it is mergable.

ngeiswei:
I'm closing, I realize it's better to define a temporary atomspace to hold atoms rather than no atomspace.

PR #943, opened 2016-10-18
Disable ImplicationLink

linas:
It seems that ImplicationLink is not actually used anywhere -- at least, not in a way that is detectable by the unit tests.  So, stub it out, for now.


linas:
Nil I am not merging this, it is up to you to ponder this.  Its not critical for my work, but what it does do is remove code that I can't make sense out of.

linas:
never mind. I just read the other email. I get it now.

PR #944, opened 2016-10-18
Thread safe fixed integer index

rTreutlein:
@linas could you take a look at this. I was thinking about the Importance Index and underlying Fixed Integer Index when I realized that you don't actually need to lock the whole index if somebody wants to use it. It should be enough to lock the individual bins. So this change should reduce the lock contention for the Importance Index. But I am unsure if this is actually an improvement as i do not have any performance data.


rTreutlein:
The Type index does not use the threadsafe version because of it's iterator design. see #945

linas:
Looks good to me

PR #949, opened 2016-10-21
Add failing (but disabled) TypeChoice utest

ngeiswei:


tests/query/TypeChoiceUTest.cxxtest
```diff
@@ -146,5 +147,104 @@ void TypeChoiceUTest::test_disconnected()
...
+ *  (VariableList
+ *      (TypedVariableLink
+ *         (VariableNode "$TyVs")
+ *         (TypeChoice
+ *            (TypeNode "TypedVariableLink")
+ *            (TypeNode "VariableList")))
+ *      (VariableNode "$B")))
+ *
+ * (define forall-instantiation-body
+ *   (ForAllLink
```
ngeiswei: Obviously this is wrong, the Variable "$TyVs" bound to the ForAll now masks the "$TyVs" bound to the bind link. This might just work if the ForAll is quoted, and the variables in it unquoted.

Issue #940, opened 2016-10-17
AIML rules loading becomes very slow...


leungmanhin:
It used to take a few minutes to load all of the rules, but now it takes so much longer and it seems to get even slower over time and never finish...

To reproduce it:
- Download and extract the files: https://github.com/opencog/test-datasets/releases/download/current/oc_public_aiml.tar.gz
- Go to `Guile` and do:

```
(use-modules (opencog) (opencog nlp) (opencog openpsi))
(primitive-load "/home/leungmanhin/Downloads/oc_aiml/generic-Oct-20-2016.scm")
```

It's possibly related to the recent changes in ImplicationLink/ScopeLink, and it becomes slower after applying the changes made in https://github.com/opencog/atomspace/commit/2be97b30fd61676dbf26e2351021733c01ca52da

linas:
Ooof. That hurts. Yes, that is the most likely explanation for the slowdown.  Just when I thought I was done fixing emergencies...

linas:
BTW, I have the perfect solution for this, its half-implemented, but then my graphics card died, which lead to a major 4-day ordeal of an operating system backup and reinstall.  I've had a lot of major fails this last few weeks. (including the refridgrator...) will have a solution for this shortly.

linas:
I think/home that pull req #957 fuixes this, but have not yet tested.

leungmanhin:
thanks Linas, unfortunately it's still very slow to load the rules... (~40 mins for loading generic)

linas:
ok investigating this now

linas:
OK, I've got a partial fix, but can't check it in until tomorrow. When I converted to hashes, I forgot to remove a piece of expensive code (which was kind-of the whole point of having the hashes)

linas:
Pull req #970 fixes this.

AmeBel:
~34 sec now

leungmanhin:
yes fixed, thanks!

linas:
"~34 sec now" -- is this about the same as before? a little faster? A lot faster? The original report says "a few minutes" -- so is it actually 4x faster?

leungmanhin:
"~34 sec" is the time for loading only the generic rules on a desktop computer in the office, I tested it on the old Han demo laptop and it took roughly the same as before

Issue #950, opened 2016-10-21
TypeChoice fails over some pattern


ngeiswei:
See the unit added in #949 to reproduce it.

ngeiswei:
It is tied to this line

https://github.com/opencog/atomspace/blob/master/opencog/query/DefaultPatternMatchCB.cc#L295

I have to admit I don't understand it...

ngeiswei:
It's not related to `TypeChoice` it's related to `ScopeLink` again, I can reproduce it in a simpler way as follows:

```
scheme@(guile-user)> (ForAllLink (VariableNode "$X") (EvaluationLink (Predicate "P") (VariableNode "$X")))
scheme@(guile-user)> (define gl (Get (ForAllLink (Variable "$X") (Variable "$B"))))
scheme@(guile-user)> (cog-satisfying-set gl)
$3 = (SetLink
)
```

ngeiswei:
F###, this variation works:
```
scheme@(guile-user)> (ForAllLink (VariableNode "$X") (EvaluationLink (Predicate "P") (VariableNode "$X")))
scheme@(guile-user)> (define gl (Get (ForAllLink (Variable "$V") (Variable "$B"))))
scheme@(guile-user)> (cog-satisfying-set gl)
$3 = (SetLink
   (ListLink
      (VariableNode "$X")
      (EvaluationLink
         (PredicateNode "P")
         (VariableNode "$X")
      )
   )
)
```
???

It seems the on the fly alpha-conversion is kinda messing stuff. I have to say I'm not sure this is a good idea, it is utterly confusing for the user to enter an atom and see in return its variable renamed, utterly...

ngeiswei:
Well, the first one failing actually makes sense because the Variable "$X" doesn't self-match. However this shows how this plays so badly with on-the-fly alpha-conversion. Imagine that one enters first a pattern using "$X", then enters the same pattern using "$V" to hopefully avoid colliding with "$X" is some grounds to match, then it will silently fail. The user would have to delete first the pattern with "$X" then re-enter the one with "$V".

linas:
I'll take a look at this tomorrow. From your description, there is a bug in the self-matching code: its not actually a self-match, if the thing being matched is alpha-convertible to something else.  Of course, easy to say, harder to do.  The self-matching code is very very old, dating back to the beginning, long before lambda and scope links got introduced. So -- no surprise that it now does the wrong thing.

linas:
[...]

PR #957, opened 2016-10-25
Switch to atom hashes, remove node and link indexes

linas:
This is a major overhaul of the atomspace internals.

It fixes multiple problems:
-- It should satisfy Nil's desires for content-based atom compare.
-- it should fix issue #940 (but currently untested)
-- it should remove another UUID stumbling block.

It should have the following improvements:
-- Reduce memory usage in the atomspace
-- Improve performance in multiple ways.

The changes are:
-- All atoms now store a content-based hash.
-- All atom comparisons use the content-based hash.
-- This allows the removal of the node and link indexes, which 
   should save RAM and improve atom add/delete performance.


ngeiswei:
And it even faster, I say wow!

However I wonder what happens in case of hash collision?

linas:
hash collisions fall are supposed to fall through to direct content compare.

I tried to avoid hash collisions, but I suspect the hashing function is rather poor. The whole thing is a bit trickier than you might guess.

I measured the perf: -- removing atoms from atomsapce is 50% faster ...adding node to the atomspace is 50% faster, adding links got slower by 2x. Not sure why; it might be the cost of hashing the very first time.  Problem is, for a lot of our algos, its always "the very first time" -- we generate and discard lots of atoms.

I'm going to merge this, because all atomspace unit tests pass, but the opencog tests are suffering  -- MOSESPLNSynergyUTest is failing for me, so is PLNRulesUTest -- is this something you did? are these tests supposed to pass?

ngeiswei:
MOSESPLNSynergyUTest and PLNRulesUTest fail for other reasons, mainly due to #950 I believe.

Issue #958, opened 2016-10-26
Pattern matcher should catch exception


ngeiswei:
If an exception is thrown while grounding a solution, the pattern matcher should skip it rather than crashing.

To produce this behavior, run the pattern matcher over the following

``` scheme
;; Failing bind link
(define get-bindlink
(BindLink
  (VariableList
    (TypedVariableLink
      (VariableNode "$A-648d8b51") ; [56][45]
      (TypeChoice
        (TypeNode "LambdaLink") ; [3][1]
        (TypeNode "PredicateNode") ; [4][1]
      ) ; [5][1]
    ) ; [57][45]
    (TypedVariableLink
      (VariableNode "$A-4c2b3a57") ; [49][45]
      (TypeChoice
        (TypeNode "LambdaLink") ; [3][1]
        (TypeNode "PredicateNode") ; [4][1]
      ) ; [5][1]
    ) ; [50][45]
  ) ; [58][45]
  (AndLink
    (ImplicationLink
      (VariableNode "$A-4c2b3a57") ; [49][45]
      (PredicateNode "T") ; [32][1]
    ) ; [51][45]
    (VariableNode "$A-648d8b51") ; [56][45]
    (ImplicationLink
      (VariableNode "$A-648d8b51") ; [56][45]
      (VariableNode "$A-4c2b3a57") ; [49][45]
    ) ; [59][45]
  ) ; [63][45]
  (ExecutionOutputLink
    (GroundedSchemaNode "scm: crisp-modus-ponens-formula") ; [12][1]
    (ListLink
      (ExecutionOutputLink
        (GroundedSchemaNode "scm: crisp-modus-ponens-formula") ; [12][1]
        (ListLink
          (VariableNode "$A-648d8b51") ; [56][45]
          (ImplicationLink
            (VariableNode "$A-648d8b51") ; [56][45]
            (VariableNode "$A-4c2b3a57") ; [49][45]
          ) ; [59][45]
          (VariableNode "$A-4c2b3a57") ; [49][45]
        ) ; [61][45]
      ) ; [62][45]
      (ImplicationLink
        (ExecutionOutputLink
          (GroundedSchemaNode "scm: crisp-modus-ponens-formula") ; [12][1]
          (ListLink
            (VariableNode "$A-648d8b51") ; [56][45]
            (ImplicationLink
              (VariableNode "$A-648d8b51") ; [56][45]
              (VariableNode "$A-4c2b3a57") ; [49][45]
            ) ; [59][45]
[...]

PR #959, opened 2016-10-26
Utest for #958

ngeiswei:


ngeiswei:
It's failing but a fix is coming soon.

PR #962, opened 2016-10-26
Fix issue #958

ngeiswei:
This is a fix for issue #958 

Another option is to ignore this issue and add have that try/catch only in a pattern matcher callback dedicated to the backward chainer.

All unit tests passes and I see no harm in it, but let me know what you think, @linas .

### DIFF
--- a/opencog/query/Implicator.cc
+++ b/opencog/query/Implicator.cc
@@ -49,8 +49,12 @@ bool Implicator::grounding(const HandleMap &var_soln,
 	if (_result_set.size() >= max_results)
 		return true;
 
-	Handle h = inst.instantiate(implicand, var_soln);
-	insert_result(h);
+	// By wrapping this in a try/catch we can ignore the cases when
+	// executions create ill-formed atoms.
+	try {
+		Handle h = inst.instantiate(implicand, var_soln);
+		insert_result(h);
+	} catch(const InvalidParamException& e) {}
 
 	// If we found as many as we want, then stop looking for more.
 	return (_result_set.size() >= max_results);
--- a/tests/query/no-exception.scm
+++ b/tests/query/no-exception.scm
@@ -70,7 +70,6 @@
 
 ;; Schema returning undefined handle
 (define (crisp-modus-ponens-formula A AB B)
-  (cog-logger-debug "crisp-modus-ponens-formula A = ~a AB = ~a B = ~a" A AB B)
     (let (  (sA (cog-stv-strength A))
             (cA (cog-stv-confidence A))
             (sAB (cog-stv-strength AB))


linas:
OK, I thought about this some more, and it seems like an OK thing to do. However, can you change the exception to a non-logging exception type?  There are currently 5 non-logging exceptions -- i.e. exceptions that are "normal" and will not be written to the log file:

SilentException -- base class
DeleteException -- for DeleteLink only
NestingException -- expected problems with quoting/nesting (I cannot avoid these for compicated rasons)
 NotEvaluatableException -- 
NotFoundException -- 
TypeCheckException -- 

For the problem you describe, it seems like the NotFound exception would be best: its as if a match was found, but is then rejected.

Although, even so -- I'm still not convinced that this a good way for rejecting matches. If you want to reject a match, it should be done in the search phase, and not after the search has been completed.  That is,   BindLink X Y is the same as PutLink Y GetLink X   and what you are trying to do is more like MapLink Y GetLink X  where you are doing the rejection in the Y not in the X....

opencog/query/Implicator.cc
```diff
@@ -49,8 +49,12 @@ bool Implicator::grounding(const HandleMap &var_soln,
...
 		return true;
 
-	Handle h = inst.instantiate(implicand, var_soln);
-	insert_result(h);
+	// By wrapping this in a try/catch we can ignore the cases when
+	// executions create ill-formed atoms.
+	try {
+		Handle h = inst.instantiate(implicand, var_soln);
+		insert_result(h);
+	} catch(const InvalidParamException& e) {}
```
linas: To be clear: this should change to NotFound exception,

ngeiswei: But if I change that to NotFound when the ImplicationLink fails line https://github.com/opencog/atomspace/blob/master/opencog/atoms/core/ImplicationLink.cc#L65 it will not catch it.

linas: Well, the code that is throwing this has to throw the right thing...

ngeiswei: Indeed. I put that specific exception because it is currently the only one bothering me, my plan was to extend as needed. BTW, this code has been tested against backward chainer inferences and has been proven correct for now. That is said if you consider that it is too permissive, I can easily move this to the callback.

linas: All I'm saying is that the exception will clutter up the opencog log file with megabytes of crap.  Whatever it is that is throwing that exception should probably be changed to throw a non-logging exception.

I ca do this, but I am still behind, and haven't finished my last project, yet.

ngeiswei: I see. Don't bother for now.

PR #967, opened 2016-11-02
Bug-fix for bug.pull #966

linas:
One more quotation fix, for #966, a spin-off from bug #950


linas:
whoops, this fix is incomplete, RecognizerUTest now fails. Am working on a fix.

PR #972, opened 2016-11-04
4x improvement in pattern matcher performance with this one weird little trick....

linas:
You won't believe how simple it is to improve pattern matcher performance!

Did you know that exception stack unwinding is very expensive?  Most people don't, and this includes people like Linas!  Using exceptions to handle default code paths can be a huge drag on performance, and converting the `std::` container class `.at()` method, which throws when a value is not found, to the `find()` method, which simply returns the iterator `end()`, results in much much faster code!  About 4x faster, for the pattern matcher!

### DIFF
--- a/opencog/benchmark/README.md
+++ b/opencog/benchmark/README.md
@@ -147,6 +147,7 @@ Install:
 apt-get install linux-tools
 ```
 
+oprofile
 
 
 ### Using valgrind ###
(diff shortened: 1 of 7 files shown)


ngeiswei:
Very useful observation, regarding the exception.

I don't know if the DO_LOG is worth it according to your benchmark though.

linas:
Yeah, doesn't make much of a difference.

PR #978, opened 2016-11-07
Change sti,lit and vlti types to float

misgeatgit:
When time-based rent parameters are used, calculated rent values will be floating points. So, we need to make STI, LTI and VLTI of type float so that calculated rent values will not be rounded.


opencog/truthvalue/AttentionValue.h
```diff
@@ -60,10 +60,10 @@ class AttentionValue
...
     static const vlti_t DEFAULTATOMVLTI; //!< very long-term default
 
-    static const sti_t MAXSTI = SHRT_MAX;
-    static const lti_t MAXLTI = SHRT_MAX;
-    static const sti_t MINSTI = SHRT_MIN;
-    static const lti_t MINLTI = SHRT_MIN;
+    static constexpr sti_t MAXSTI = SHRT_MAX;
+    static constexpr lti_t MAXLTI = SHRT_MAX;
+    static constexpr sti_t MINSTI = SHRT_MIN;
+    static constexpr lti_t MINLTI = SHRT_MIN;
```
linas: do you want these to be SHRT_MIN and MAX (+/- 32786) or FLOAT_MIN/MAX  (+/- 1.0e38)  There should be a comment added explaiing this choice.

linas:
Aside from these changes, I believe that additional changes are needed to the AttentionBank, which still uses shorts in its math, and also in the unit tests?  Or are all of these OK?

opencog/truthvalue/AttentionValue.h
misgeatgit: @linas I think SHRT_MAX and MAX are fine for now as the total STI funds being used are less SHRT_MAX.

misgeatgit:
I will have to look that. Thanks for reminding me.

opencog/truthvalue/AttentionValue.h
linas: OK, can you add a not saying something like "In attentionBank.wahtever.cc the total STI funds being used are less SHRT_MAX and so these values are reasonable min/max values" -- or something like that, so that the next person who looks at the code can understand that its OK.

linas:
OK, I assume that everything that needed to be done got done.

linas:
I just reverted this. Merging this breaks opencog; there are build errors in opencog/attention/WARentCollectionAgent.cc and others.  Please fix these.

Issue #977, opened 2016-11-07
Split implication and equivalence links into scoping and non scoping types


ngeiswei:
As discussed in https://groups.google.com/forum/#!searchin/opencog/ImplicationLink$3F%7Csort:relevance/opencog/PI4THQ5etwU/hfMgCpg8AwAJ we want to split `ImplicationLink` and alike, `EquivalenceLink`, `ExtensionalImplicationLink`, etc, into scope and non-scope versions.

I suggest to rename the scope versions

```
ImplicationScopeLink
ExtensionalImplicationScopeLink
IntensionalImplicationScopeLink
EquivalenceScopeLink
...
```

I do understand that according to the English language it gives the `scope` qualifier a stronger position than the `implication` qualifier, on the other hand it follows the type inheritance prefix order which has been followed so far in the atom type inheritance taxonomy.

linas:
OK, yes, reviewed, looks good to me. Assuming teh wiki pages have been updated, this can be closed?

linas:
Hmm. After looking at this more closely, it occurs to me that the following are equivalent:
```
ImplicationScopeLink
     VariableNode  x
     P(x)
     Q(x)
```
and
```
ForAllLink
    VariableNode x
    ImplicationLink
        P(x)
        Q(x)
```

If these two are NOT equivalent, then how are they different?  If they are equivalent, do we really need to have an ImplicationScopeLink?

ngeiswei:
There are NOT equivalent. Looking for an old email I wrote over a year ago explaining why there are not...

ngeiswei:
The wiki is documented well enough I believe http://wiki.opencog.org/w/ImplicationScopeLink

I'm closing.

linas:
Ah, yes. I think Ben said this in an email: that its like "AverageQuantifierLink" or something like that.

PR #979, opened 2016-11-07
Split implication link

ngeiswei:
Implement issue #977 an opencog repo companion is coming.


linas:
OK, reviewed, looks good to me.

PR #981, opened 2016-11-07
Have Implicator::grounding catch all exceptions.

ngeiswei:
I know it's pretty extreme but it turns out there are more exceptions
that need to be accounted for like for instance when a scheme function
doesn't get the right number of arguments (due to ill formed ListLink
that hand the arguments to a rewrite part of a rule).

### DIFF
--- a/opencog/query/Implicator.cc
+++ b/opencog/query/Implicator.cc
@@ -58,7 +58,7 @@ bool Implicator::grounding(const HandleMap &var_soln,
 	try {
 		Handle h = inst.instantiate(implicand, var_soln);
 		insert_result(h);
-	} catch(const InvalidParamException& e) {}
+	} catch(...) {}
 
 	// If we found as many as we want, then stop looking for more.
 	return (_result_set.size() >= max_results);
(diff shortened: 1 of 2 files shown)


linas:
Well, OK, but its a hack, right?  Exceptions are thrown for two reasons:
-- If the user has a bug in their code, one throws an exception to let the user know about it.  Ignoring these exceptions defeats the reason for their existence.

-- The second reason is to provide a return path for "normal operation", with the property that the return path allows the higher performance in the "main" path, by avoiding error checks in the main path. i.e. the main path has less error-checking overhead.

I'm nervous that, by hiding bugs, people will write buggy code, and assume that it's OK, because the system will  mostly work, anyway. It is "well known" that its much much harder, more expensive to fix bugs later, than it is to fix them earlier.

ngeiswei:
Yes it is a hack, I just want to have the BC work as soon as possible, after that I remove this and put as much as possibly pre-conditions in the rule implicant to avoid throwing an exception, then for what I can't I'll select the exception subset which is unavoidable.

linas:
ok

PR #987, opened 2016-11-08
Add local quote

ngeiswei:
Only upgraded the code for PatternTerm and such, but it doesn't hurt the rest of the code at this point, so I'm merging now to not collides with Linas changes.


opencog/atoms/pattern/PatternLink.cc
```diff
@@ -889,7 +890,9 @@ void PatternLink::make_term_tree_recursive(const Handle& root,
 	_pat.connected_terms_map[{h, root}].emplace_back(ptm);
 
 	// Update the PatternTerm quotation level
-	if (QUOTE_LINK == t)
+	if (not parent->isQuoted() and LOCAL_QUOTE_LINK == t)
+		ptm->setLocalQuote(true);
```
linas: Don't introduce more complexity, just increment and decrement the quote level!

opencog/query/PatternTerm.h
```diff
@@ -94,6 +94,9 @@ class PatternTerm
 	// literally.
 	int _quotation_level;
 
+	// True if the atom has been wrapped by a LocalQuoteLink
+	bool _local_quote;
```
linas: Don't introduce more complexity!  Just increment and decrement the quote level!   Keep it simple.

ngeiswei:
The problem is that the UnquoteLink doesn't undo a LocalQuoteLink, besides you need to know that the children would need to have their quotation level automatically go down, and since the LocalQuoteLink is being discarded I don't see how to do that unless I store that somewhere.

ngeiswei:
The problem is that the UnquoteLink doesn't undo a LocalQuoteLink, besides you need to know that the children would need to have their quotation level automatically go down, and since the LocalQuoteLink is being discarded I don't see how to do that unless I store that somewhere.

ngeiswei:
The problem is that the UnquoteLink doesn't undo a LocalQuoteLink, besides you need to know that the children would need to have their quotation level automatically go down, and since the LocalQuoteLink is being discarded I don't see how to do that unless I store that somewhere.

ngeiswei:
The problem is that the UnquoteLink doesn't undo a LocalQuoteLink, besides you need to know that the children would need to have their quotation level automatically go down, and since the LocalQuoteLink is being discarded I don't see how to do that unless I store that somewhere.

ngeiswei:
The problem is that the UnquoteLink doesn't undo a LocalQuoteLink, besides you need to know that the children would need to have their quotation level automatically go down, and since the LocalQuoteLink is being discarded I don't see how to do that unless I store that somewhere.

ngeiswei:
The problem is that the UnquoteLink doesn't undo a LocalQuoteLink, besides you need to know that the children would need to have their quotation level automatically go down, and since the LocalQuoteLink is being discarded I don't see how to do that unless I store that somewhere.

ngeiswei:
The problem is that the UnquoteLink doesn't undo a LocalQuoteLink, besides you need to know that the children would need to have their quotation level automatically go down, and since the LocalQuoteLink is being discarded I don't see how to do that unless I store that somewhere.

ngeiswei:
The problem is that the UnquoteLink doesn't undo a LocalQuoteLink, 
besides you need to know that the children would need to have their 
quotation level automatically go down, and since the LocalQuoteLink is 
being discarded I don't see how to do that unless I store that somewhere.

ngeiswei:
Man, the new Github reviewing interface has some problems...

opencog/atoms/pattern/PatternLink.cc
ngeiswei: Actually there's might be a simpler way indeed! Consider that negative quotation level encodes alocal quotation. This may work because a `LocalQuoteLink` is only active if it is so far unquoted, as explained here http://wiki.opencog.org/w/LocalQuoteLink.

ngeiswei: On second thought I'm not sure one can do that. What if one uses `UnquoteLink` that is not yet preceded by `QuoteLink`, and in that case we might be unable to use the quotation level to store local quotation. Maybe it's still possible but I'd rather go with the certain route for now, and optimize afterward if possible.

linas:
Re this:
> What if one uses UnquoteLink that is not yet preceded by QuoteLink

That is exactly what the PatternTerm class solves. During pattern matching and pattern mining, as one moves upwards from a Node, one might hit an UnquoteLink.  Perhaps the Unquote is part of the pattern one is searching for, and maybe its a part of something totally unrelated, that just happens to be in the incoming set of the atom. In the first case, we can deduce that there should be a balanced quotelink above it.  In the second case, the unquote is from outer space and one cannot deduce anything.  Use of the pattern term helps clarify these two cases.

PR #989, opened 2016-11-09
Disable LocalQuote utest till it gets fixed

ngeiswei:
Since all utests now passes and the automatic build system works as expected I prefer to disable that last failing test.


ngeiswei:
Actually other utest are still failing `BasicSaveUTest` and `PersistPGSQLUTest`, it would be cool if they could be fixed, now that we get notified by slack when it's failing.

linas:
?? For me, `BasicSaveUTest` always passes. I almost never run `PersistPGSQLUTest` because the PG interfaces are broken, they were never finished, and you can disable that test for all I care. but Basic and odbc should work fine.

Issue #991, opened 2016-11-09
Make ListLink optional when there is only one argument


ngeiswei:
## Proposal

It would be convenient if the ListLink was optional in the ExecutionOutput (and I suppose Evaluation too).

## Reproduce

```scheme
scheme@(guile-user)> (define (f X) (List X (Concept "A")))
scheme@(guile-user)> (cog-execute! (ExecutionOutput (GroundedSchema "scm: f") (Concept "B")))
ERROR: In procedure opencog-extension:
ERROR: Throw to key `C++-EXCEPTION' with args `("cog-execute!" "ExecutionOutputLink must have args! Got (ConceptNode \"B\") ; [3][1]\n (/home/nilg/OpenCog/atomspace/opencog/atoms/execution/ExecutionOutputLink.cc:66)")'.

Entering a new prompt.  Type `,bt' for a backtrace or `,q' to continue.
scheme@(guile-user) [1]> 
scheme@(guile-user)> (cog-execute! (ExecutionOutput (GroundedSchema "scm: f") (List (Concept "B"))))
$1 = (ListLink
   (ConceptNode "B")
   (ConceptNode "A")
)
```

Anyone see any reason not to?

linas:
I can't think of a reason not to do this.

I'm somewhat nervous hat PLN is over-using ExecutionOutputLink and the scm: formulas much too much.  But whatever, I guess that can be deferred until some later date.

ngeiswei:
Agreed, it would be better if the formula where composed in atomese, but for now I suppose it is the shortest path.

linas:
Not sure I want the formulas in atomese -- there are performance issues. I'm thinking more like, replacing ExecutionOutputLink by, for example, PLNFormulaLink, which would "magically" know the right formula to use.  Whatever. Not urgent. Just an idea.

another idea: if scm becomes a bottleneck, there is also the `lib:` api, which was invented for haskell, but it should work for calling C++, also. I'm assuming it would be much much faster (sub-microsecond, as opposed to dozens of microseconds to get to scm.)  But probably that is not where the current bottleneck is.

ngeiswei:
I think this is fully supported now. Closing.

PR #993, opened 2016-11-11
Make Listlink optional when single argument

ngeiswei:
Addresses issue #991 

@linas I'm reasonably confidence it's correct, but it's better if you have a look at it and merge.


linas:
Merging, but I'll convert eager back to lazy in just a moment.

PR #994, opened 2016-11-11
Remove UUID's from atoms.

linas:
This pull req removes UUID's from atoms.

The work is mostly done: all unit tests pass, however, some of the interactions with the databse are not yet correct.


linas:
not ready to be merged.

Issue #980, opened 2016-11-07
Implement LocalQuoteLink


ngeiswei:
The idea is to have a quote link that only applies to the direct child but not its descendants, as initially suggested in #132. This has the potential to simplify many PLN rules. For instance the following

```scheme
 (QuoteLink
   (ImplicationScopeLink
     (UnquoteLink (VariableNode "$TyVs"))
     (UnquoteLink (VariableNode "$P"))
     (UnquoteLink (VariableNode "$Q"))))
```

would be replaced by

```scheme
 (LocalQuoteLink
   (ImplicationScopeLink
     (VariableNode "$TyVs")
     (VariableNode "$P")
     (VariableNode "$Q")))
```

It's not very urgent but worth introducing I believe.

linas:
Yes, its a good idea.  From what I can tell, almost all uses of quote/unquote are actually localquote.

Side comment: handling full quote/unquote semantics in the pattern matcher turned out to be very painful and difficult and confusing, I spent a difficult week untangling it.

ngeiswei:
I decided to work on that next as it's gonna simplify my subsequent work.

linas:
OK. Well, please don't invent a brand-new framework to handle it. Just alter PatternTerm as needed. In fact, we should probably use pattern term in more places, not just in the pattern matcher, because the kind of stuff that you are doing is very top-down hierarchical and asymmetric, and for that, PatternTerm seems to be the correct concept.

linas:
Um, I hope this will not collide with your work, but I want to move PatternTerm.h PatternTerm.cc and Pattern.h from the opencog/query directory to the opencog/atoms/pattern directory.

ngeiswei:
Oh, OK, let me push what I've done so far then.

ngeiswei:
I've merged my work so far, if you can be done with you changes by the end today that would be ideal, that way I can resume tomorrow, otherwise I'll switch to another task till your done. Basically if your plan is to use PatternTerm more through out the pattern matcher code that is super welcome, it would definitely make completing LocalQuote implementation much easier.

linas:
See pull req #988 for movement of header files.

[...]

PR #999, opened 2016-11-15
Complete local quote

ngeiswei:
For issue #980 

It introduces a `Quotation` class to keep track of the quotation state and simplify a lot of the existing code with it while supporting `LocalQuoteLink` as well.

Note that it might slow down a bit the atomspace and the pattern matcher, although I didn't measure a substantial slowdown when running the utest before and after.

@linas I'm confident, but I think it's wiser if you review it.

### DIFF
--- a/opencog/atoms/pattern/PatternLink.cc
+++ b/opencog/atoms/pattern/PatternLink.cc
@@ -756,25 +756,20 @@ bool PatternLink::add_dummies()
 /// evaluatable.
 void PatternLink::trace_connectives(const std::set<Type>& connectives,
                                     const HandleSeq& oset,
-                                    int quotation_level)
+                                    Quotation quotation)
 {
 	for (const Handle& term: oset)
 	{
 		Type t = term->getType();
 
-		// Deal with quote
-		if (t == QUOTE_LINK)
-			quotation_level++;
-		else if (t == UNQUOTE_LINK)
-			quotation_level--;
+		quotation.update(t);
 
-		if (quotation_level > 0
-		    or connectives.find(t) == connectives.end()) continue;
+		if (quotation.is_quoted() or connectives.find(t) == connectives.end())
+			continue;
 		_pat.evaluatable_holders.insert(term);
 		add_to_map(_pat.in_evaluatable, term, term);
 		if (term->isLink())
-			trace_connectives(connectives, term->getOutgoingSet(),
-			                  quotation_level);
+			trace_connectives(connectives, term->getOutgoingSet(), quotation);
 	}
 }
 
--- a/opencog/atoms/pattern/PatternLink.cc
+++ b/opencog/atoms/pattern/PatternLink.cc
@@ -867,43 +867,17 @@ void PatternLink::make_term_tree_recursive(const Handle& root,
                                            Handle h,
                                            PatternTermPtr& parent)
 {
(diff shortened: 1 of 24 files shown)


opencog/atoms/pattern/PatternLink.cc
```diff
@@ -867,43 +862,18 @@ void PatternLink::make_term_tree_recursive(const Handle& root,
...
-	//
-	// Ignore quoting and unquoting nodes in the PatternTerm
-	if ((not parent->isQuoted() and (QUOTE_LINK == t or LOCAL_QUOTE_LINK == t))
-	    or (parent->getQuotationLevel() == 1
-	        and not parent->isLocalQuoted() and UNQUOTE_LINK == t)) {
-		if (1 != h->getArity())
-			throw InvalidParamException(TRACE_INFO,
-			                            "QuoteLink/UnquoteLink has "
-			                            "unexpected arity!");
-		h = h->getOutgoingAtom(0);
```
linas: It seems this line of code was cut, and I don't see what replaces it.  I assume the unit tests still pass, so it is surprising that cutting this still allows things to work ...

linas: Never mind. I spoke too soon. I see that it was moved to the PTM. That is.. unexpected, but I guess its OK.

opencog/query/DefaultPatternMatchCB.cc
```diff
@@ -389,33 +391,34 @@ bool DefaultPatternMatchCB::is_self_ground(const Handle& ptrn,
...
 		// valid way to get to here.
-		if (ptype == QUOTE_LINK and 0 == quote_level and ptrn == grnd) return true;
-		if (ptype == UNQUOTE_LINK and 1 == quote_level and ptrn == grnd) return true;
-
-		if (ptype == QUOTE_LINK) quote_level++;
-		else quote_level--;
+		if (quotation.consumable(ptype) and ptrn == grnd) return true;
 
 		const Handle& qpat = ptrn->getOutgoingAtom(0);
-		return is_self_ground(qpat, grnd, term_gnds, varset, quote_level);
```
linas: This is confusing. Up above, the unwrapping of the quote link was done in the pattern term (line 49) -- while here, it is unwrapped a second time?  Or maybe the outgoing set is no longer the outgoing set of the atom, but is instead the outgoing set of the quote?  

That is, it seems that `ptrn->getOutgoingAtom(0)`  has the same value as `ptrn->getHandle()` whenever the ptrn is a QuoteLink. This seems like a kind-of schizo behavior in the code ...

ngeiswei: I would agree, except that it's not a pattern term that is being used here, so I presume the unwrapping didn't apply. As you said pattern terms should be used instead, but that's another, though related, issue.

linas: OK, yes, I see now.  It seems weird to me that the quote-unwrapping happens in the PTM ctor, but perhaps this is dictated by local-quote, and perhaps we've spent enough time on this already.

ngeiswei: It's not specifically due to the local quote, I just felt that this is where it belongs. But this is perhaps more a matter of taste than anything.

PR #1001, opened 2016-11-17
Fix local scope bind

ngeiswei:
I'm merging as I'm fairly confident about it. However the corresponding utest in `LocalScopeUTest` breaks for a reason I do understand but prefer to keep for a subsequent PR.

### DIFF
--- a/opencog/atoms/core/Variables.cc
+++ b/opencog/atoms/core/Variables.cc
@@ -24,6 +24,7 @@
 #include <opencog/atoms/base/Atom.h>
 #include <opencog/atoms/base/Link.h>
 #include <opencog/atoms/base/ClassServer.h>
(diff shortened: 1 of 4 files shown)


ngeiswei:
Damn it, it breaks
```
	 75 - LocalQuoteUTest (Failed)
	 85 - ScopeUTest (Failed)
```
Let me see what is wrong...

ngeiswei:
This is utterly bizarre, all `ScopeUTest` pass when I run `ScopeUTest` in standalone, but it always fails if run with the test suit. I'm clueless.

ngeiswei:
Gees, I think I have some brain damage, I made the confusion between `QuoteUTest` and `ScopeUTest`.

ngeiswei:
It's a not a bug, the utest was wrong. :-)

PR #1002, opened 2016-11-17
Disable variables in clauses check in PatternLink

ngeiswei:
In order to have `LocalQuoteLink` work on `PatternLink` (i.e. `BindLink`) I need to disable checks that clauses need to contains variables.

For instance this allows to define the following BindLink https://github.com/opencog/atomspace/blob/master/tests/query/LocalQuoteUTest.cxxtest#L144 used as rewrite term of meta-rule https://github.com/opencog/atomspace/blob/master/tests/query/LocalQuoteUTest.cxxtest#L139.

@linas What do you think about these checks, are they really necessary? Can we somehow make them optional so that in this context they allow such a `BindLink` to exist?

### DIFF
--- a/opencog/atoms/pattern/PatternLink.cc
+++ b/opencog/atoms/pattern/PatternLink.cc
@@ -75,7 +75,7 @@ void PatternLink::common_init(void)
 	_num_comps = _components.size();
 
 	// Make sure every variable is in some component.
-	check_satisfiability(_varlist.varset, _component_vars);
+	// check_satisfiability(_varlist.varset, _component_vars);
 
 	// If there is only one connected component, then this can be
 	// handled during search by a single PatternLink. The multi-clause
@@ -456,6 +456,9 @@ void PatternLink::validate_clauses(OrderedHandleSet& vars,
 			              __FUNCTION__);
 		}
 	}
+
+	// Ignore the subsequent checks
+	return;
 
 	// Make sure that each declared variable appears in some clause.
 	// We won't (can't) ground variables that don't show up in a


linas:
I don't see how it makes sense.  If a variable never appears in a pattern, then it cannot ever be given a value: its undefined. If you try to use that variable, you (should) get an exception thrown.  If you remove the sanity checks early, then you'll just get some weird, inconsistent, undefined behavior later on. 

However, from what I can tell, it looks like all three variables, -- $body, $rewrite and also $vardecl all do appear, as bound variables, in the pattern. So it seems to me like the test at 133 should pass, just fine, without this pull request.  If the test at line 133 is failing, then its due to some other bug, somewhere else, and not this check.

ngeiswei:
The test fails on the `BindLink` at line https://github.com/opencog/atomspace/blob/master/tests/query/LocalQuoteUTest.cxxtest#L144 because that bind link, before the rewrite occurs, doesn't have variable `$vardecl` in its clauses. Of course in practice it's OK since only its quoted version is used, but the non-quoted version still exists and generate this error.

linas:
Ahh. I see what the bug is.  It seems to be this: the localquote is implemented incorrectly.   Right now, it's spec'ed and implemented so that it works "just like quote, but only one level deep".  Instead, it seems that it should have worked "just like unquote, but one level deep".

That is, the current implementation of localquote allows one to build completely crazy, semantically invalid atoms --  in this case, forces you, by necessity, to build semantically invalid, crazy atoms.  Removing the checks for craziness doesn't seem like a good solution.

I'm thinking that what we need is something like this:   `(BindLink (LocalUnquote (...)))`

ngeiswei:
So `LocalUnquoteLink` would make the incoming atoms local quoted, right? But then you can tell `BindLink` factory that when it sees a `LocalUnquoteLink` it should perform any sanity check, right?

ngeiswei:
The other option is to perform the sanity check at run-time (when calling the pattern matcher on a pattern link). I agree it isn't as cool as performing the check statically, but it allows to use this cool `LocalQuoteLink` syntax.

ngeiswei:
I'm closing this. It's much safer to ban `LocalQuoteLink` over `ScopeLinks` for now anyway.

PR #1018, opened 2016-11-29
Improve bc

ngeiswei:
Various fixes in the BC and elsewhere.


ngeiswei:
Oh, for what concerns others add `tv->bool` and `bool->tv` in scm utilities.

Also @linas you may want to have a look at https://github.com/opencog/atomspace/commit/1afde67692e45cb32c9d52a268889a06555a3cb8 just in case.

BTW, this PR makes catching exceptions by rule apply obsolete and uses instead rule preconditions. I prefer to keep the catch for now though just in case.

PR #1019, opened 2016-11-30
Stochastic diffusion

misgeatgit:
Needed an access to AttentionBank and the only way to accessing it seemed to put the accessor class in this directory. Details about the purpose are documented in the code.


linas:
Can this code be placed in the opencog repository as well? i.e. here:  https://github.com/opencog/opencog/tree/master/opencog/attention 

The long-term goal is to move the attention bank out of the atomspace.

misgeatgit:
@linas it will be possible if  the AttentionBank object is refactored out from the AtomSpace class. Now it is a private member of the AtomSpace class.

PR #1023, opened 2016-12-08
Fix for python import error issue

AmeBel:


linas:
Microsoft invented DLL hell more than 25 years ago. At last, Linux has finally caught up with what Microsoft Windows could do so very easily, oh so long ago.

Issue #1029, opened 2016-12-16
Add Atom::size() method


ngeiswei:
I'm surprised to find no C++ function that calculates the size of an atom (1 if a node, 1 + sizes of its outgoings if a link).

I suggest to add an Atom::size() method.

Knowing the size of an atom can be convenient is some contexts such as biasing search towards smaller atoms, etc.

linas:
Repeat of earlier remark: I don't know of any code anywhere that does anything like this.

The pattern matcher does try to find the "thinnest" atom, which is an atom with the smallest incoming set.  "Thin" atoms are good places to start searches.

ngeiswei:
Cool. Closing then.

PR #1030, opened 2016-12-16
Add Atom::size(). Returns the number of atoms in a hypergraph

ngeiswei:
More specifically

1 if the atom is a node
1 + sizes of the outgoings if the atom is a link

Related to issue #1029


ngeiswei:
@linas I know you're busy, since it doesn't harm anything I'm merging. If it happens to be redundant I'll remove it.

linas:
Its not redundant, we've never had anything like this before.

PR #1037, opened 2016-12-17
XXX hack -- try locking the protection

linas:
untested, I hope this doesn't break anything

### DIFF
--- a/opencog/guile/SchemeEval.cc
+++ b/opencog/guile/SchemeEval.cc
@@ -38,6 +38,7 @@
 using namespace opencog;
 
 std::mutex init_mtx;
+std::mutex prot_mtx; // temporary hack
 
 /**
  * This init is called once for every time that this class
@@ -57,6 +58,7 @@ void SchemeEval::init(void)
 	_in_eval = false;
 	_eval_thread = SCM_EOL;
 
+	std::unique_lock<std::mutex> protlck(prot_mtx);
 	// User error and crash management
 	_error_string = SCM_EOL;
 	_error_string = scm_gc_protect_object(_error_string);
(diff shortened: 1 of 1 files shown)


bgoertzel:
did this work?  what's the current state of the crashing/threading issue?

linas:
Did not do a thing.

Many (maybe most/all?) of the crash stack traces show stack corruption. One hypothesis: stack size is too small: we have 80+ threads, each with a large stack.  Work around: increase the stack size with setrlimit.  Crash seems to be assocaited with high-speed net-catting. Work-around: add delays to saliency, other systems, to produce no more than 10 or 20 messages/sec.  These two, used together, seems to prevent the crash. Not sure which one did the trick.

After this, there is an exception, involving `GreaterThan (Get (State ...)))`   Hack around this to not throw exception.

After this, there is a hang, some sort of locking issue.  working on that, now.

linas:
to summarize: three different bugs:
1) some sort of stack corruption, for unknown reason
2) StateLink can be briefly undefined, while state is being updated.  This can cause issues with the any GetLink that assumes exactly one value will be returned (e.g. GreaterThanLink)
3)some hang, due to some deadlock.

linas:
issue 2) above is tracked in https://github.com/opencog/ros-behavior-scripting/issues/142

linas:
Issue 3 is not a hang -- guile and the cogserver are alive and respond.  the psi-rules to do facial animations continue to work (after hitting the stop/start button on the gui). However, face and saliency tracking stop working -- this suggests that the octree/space-server is hung or is ignoring data or failing to respond.

(also, the left eyelid does not blink, and the left eye is acting weird.)

linas:
Issue 3 is not octree -- that seems to work. Faces do get pushed into the octree, and queries of face positions succeed (e.g. `(get-last-xyz "faces" (NumberNode 42) 8000)` works fine, with 42 being the current face id)

what happens is that psi stops running. `(psi-get-loop-count)` does not change. saying `(psi-halt)` and `(psi-run)` gets it running again. Oddly, it prints a new message it never printed before...

linas:
The new message is from ` (DefinedPredicate "Salient:Curious")`  which suggests that somehow, this rule was not running before... !?

Issue #1038, opened 2016-12-18
StateLink race condition.


linas:
When updating the value of a StateLink, the link can be briefly undefined. This will cause exceptions in any code that expects exactly one value: e.g. `(GreaterThanLink (GetLink (StateLink ...)))`  The thrown exception is described in https://github.com/opencog/ros-behavior-scripting/issues/142 

This bug is being opened, because the fix will need to be here.

linas:
A symptom of the error was `Don't know how to do arithmetic with this: (SetLink)` 

Closing -- fixed in #1053

PR #1053, opened 2017-01-02
atomic StateLink -- fix for bug #1038

linas:
The bug #1038 was one of the major causes for failure leading up to the 20 Dec demos. This pull req fixes this.


linas:
Oh, this fixes (should fix) https://github.com/opencog/ros-behavior-scripting/issues/142

PR #1056, opened 2017-01-04
Keep and update lists of top k atoms by their STI value

misgeatgit:


opencog/atomspace/AtomSpace.h
```diff
@@ -527,7 +527,7 @@ class AtomSpace
     std::string to_string();
 
     /* ----------------------------------------------------------- */
-    // ---- Signals
+   // ---- Signals
```
linas: ??

linas:
looks good to me.

opencog/atomspace/AtomSpace.h
misgeatgit: I've just fixed it.

PR #1063, opened 2017-01-10
Make hash-value returned using python api consistent

AmeBel:


opencog/cython/opencog/atom.pyx
```diff
@@ -10,7 +10,7 @@ cdef class Atom(object):
         del self.handle
 
     def value(self):
-        return void_from_cptr(self.handle)
+        return self.handle.value()
```
linas: The `void_from_cptr(self.handle)` uniquely identifies an atom -- if the values are equal, then we are 100% certain that the atoms are identical.   By contrast, the hash has no such guarantee -- there is a good probability that inequivalent atoms will have different hashes, but its not that high -- the "birthday paradox" guarantees this. https://en.wikipedia.org/wiki/Birthday_problem  

What problem is this pull req trying to solve? The rest of the changes look OK; but the above breaks the semantics of "value" -- its always been guaranteed-unique, in the past.

linas: Oh, I think I get it -- you want it to be consistent with what the c++ code calls the 'value, which, I guess is probably returning the hash, right now.  I'm thinking that it might be best to remove 'value' from both the c++ and python API -- there is not that much use for it, except for printing, and some sanity-checking.

AmeBel: void_from_cptr(self.handle) doesn't uniquely identify an atom, which is the issue I am trying to fix. I should have mentioned it on the description.
```python
>>> from opencog.atomspace import *
>>> a = AtomSpace()
>>> n1 = a.add_node(types.Node, "foo")
>>> n2 = a.add_node(types.Node, "foo")
>>> n2.value()
30170656
>>> n1.value()
30171824
```

AmeBel: I guess what you mean by uniquely idenfity is,
```python
>>> # continuing from above
>>> n3 = Atom(n2.value(), a)
>>> n2 == n3
True
```
which can't be done with the change I made, it segfaults. I didn't consider this when I made the change. And if the plan is to remove value() from the public api, then I have to remove its usage in the concept-blending code. Closing pr.

linas: the chandles are supposed to be identical. If the chandles are not identical, there's a serious bug somwhere.

linas: the point is, if you want to change the TV on the atom, the atom needs to be unique. otherwise, you are changing the TV on two different atoms, and they will get out of sync.   This could be a really serious issue for python -- if it really is tracking two different atoms as if they were the same, (and this is not just some stupid glitch), then the python bindings are seriously broken.  

We need to get to the bottom of this. Maybe opening a new issue for this is worth it.

linas:
OK, I just tried this:
```
$ python
Python 2.7.13 (default, Dec 18 2016, 20:19:42) 
[GCC 6.2.1 20161215] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from opencog.atomspace import *
>>> a = AtomSpace()
>>> n1 = a.add_node(types.Node, "foo")
>>> n2 = a.add_node(types.Node, "foo")
>>> n2.value()
93850896547920
>>> n1.value()
93850896548864
>>> 
```
so its as if there are two atoms. But then:
```
>>> n1.truth_value( 0.5, 0.8)
(Node "foo" (stv 0.500000 0.800000)) ; [4234988883757294181][0]
>>> print n1.tv
(stv 0.500000 0.800000)
>>> print n2.tv
(stv 0.500000 0.800000)
```

so its behaving as if there really is one atom,, and not two.  Argh.

linas:
OK, so #1065 partly fixes this issue, but the unit test now break.  I guess more fixing is needed.

PR #1065, opened 2017-01-11
fix for issue described in #1063

linas:


opencog/cython/opencog/Cast.cc
```diff
@@ -8,3 +8,4 @@ using namespace opencog;
 Handle atom_from_the_void(long p) { return *((Handle*) p); }
 long void_from_candle(const Handle& hp) { return (long) (&hp); }
 long void_from_cptr(Handle* hp) { return (long) (hp); }
+long avoid_from_cptr(Handle* hp) { return (long) (hp->operator->()); }
```
AmeBel: what is `hp->operator->()`?

linas: Its the pointer to the actual atom. Same value as `hp->get()`

Recall that Handles are just smart pointers, they only point to the actual atom.  The raw pointer is quasi-hidden, so that users won't user Atom*  but you can still get Atom* by saying wither hp->get or hp->operator->() -- either one works.

AmeBel: What is the purpose of Handles? Why not use c++ standard smart pointers?

bgoertzel: Amen, there seems clear value for Handles in a distributed system (multiple machines, or multiple processes on the same machine) -- they allow an Atom on one machine to refer to an Atom on another machine... smart pointers don't do that

bgoertzel: (but that comment is really about the value of UUIDs wrapped in Handles, not about Handles as wrappers for pointers)

bgoertzel: As for smart pointers, there are plenty of pluses and minuses there, which Linas can explain better than me...

linas: I think the answer is that Cython cannot work directly with C++ objects; it can only store pointers. Since handles are C++ objects, it can only store pointers to handles.

Next, cython cannot work directly with C++ pointers either, so everything must be cast into long integers. So pointers get converted to long integers, and back.  Its not type-safe, but I don't know of a better way. -- it would take a lot of work to design some better way. (which might not be better in the end)

linas: Re UUID's, they are now handled as a utility in the "libatomspaceutils" directory.  Both the database bindings, and the REST server uses UUID's.

PR #1090, opened 2017-01-18
Overload AtomSpace::add_link for empty outgoing set

ngeiswei:


rTreutlein:
When are Links without an outgoing set useful? Intuitively Links that don't link anything seem kinda pointless.

ngeiswei:
Well, they are not very many obviously, but depending on the context they may be frequently used. For instance an empty `VariableList` can be used in a scope link to indicate that it has no locally bound variables. Or, less practical but still, if you want encode numbers using sets, where the empty `Set` is 0, `Set Set` is 1, etc.

Issue #1092, opened 2017-01-20
Persistence related opencog and scm commands are not working


misgeatgit:
I was going through the [Postgres setup instructions](https://github.com/opencog/atomspace/tree/master/opencog/persist/sql) and I have encountered the following issues:
  - All sql opencog commands are not being found ( i.e sql-open, sql-load, sql-store)
  - (store-atom ATOM) throws ```unbounded variable store-atom``` error message even after prompting ```(use-modules (opencog) (opencog persist-sql))```

linas:
Works great for me; I've been using this heavily for the last 3 weeks, and have fixed many bugs.  Anything more specific?

linas:
oh, try also `(use-modules (opencog persist))`  (without the -sql) you need both  that it
```
(use-modules (opencog persist) (opencog persist-sql))
```

the first one is generic, the second is for sql-only.

linas:
closing, I believe that this should really work just fine.  Please reopen if above suggesttion doesn't fix it.

misgeatgit:
Your suggestion solved the command not found issue error but another error as shown below occurred.
```
guile> (use-modules (opencog persist) (opencog persist-sql))

guile> (sql-open "mycogdata" "opencog_user" "cheese")
Backtrace:
In ice-9/boot-9.scm:
 157: 10 [catch #t #<catch-closure 1511aa0> ...]
In unknown file:
   ?: 9 [apply-smob/1 #<catch-closure 1511aa0>]
In ice-9/boot-9.scm:
 157: 8 [catch #t #<catch-closure 1511980> ...]
In unknown file:
   ?: 7 [apply-smob/1 #<catch-closure 1511980>]
   ?: 6 [call-with-input-string "(sql-open \"mycogdata\" \"opencog_user\" \"cheese\")\n" ...]
In ice-9/boot-9.scm:
2320: 5 [save-module-excursion #<procedure 1041cc0 at ice-9/eval-string.scm:65:9 ()>]
In ice-9/eval-string.scm:
  44: 4 [read-and-eval #<input: string 9f10d0> #:lang ...]
  37: 3 [lp (sql-open "mycogdata" "opencog_user" "cheese")]
In unknown file:
   ?: 2 [opencog-extension sql-open ("mycogdata" "opencog_user" "cheese")]
In ice-9/boot-9.scm:
[...]

PR #1097, opened 2017-01-23
Fixpsql

misgeatgit:
A fix for the failing unit tests which use libpq. ODBC based tests are still failing on my machine though.

### DIFF
--- a/opencog/persist/sql/README.md
+++ b/opencog/persist/sql/README.md
@@ -545,7 +545,7 @@ So here's a super-short recap:
 
  * Create a user 'opencog_tester' with password 'cheese'.
  * Create a new database: e.g.  `$ createdb opencog_test`
- * Create the tables: `$ cat atom.sql | psql `
+ * Create the tables: `$ cat opencog/persist/sql/multi-driver/atom.sql | psql opencog_test  -U opencog_tester -W -h localhost`
  * (Optional) Create an entry in `~/.odbc.ini`, as explained above.
    The entry should be called `opencog_test`, and use `opencog_tester`
    as the user. Password is `cheese`.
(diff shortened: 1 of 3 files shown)


linas:
Thanks. However... before I saw your pull request, I made a different fix (but similar in nature), and have already pushed it; give it a try.

I would rather not add the hostname to the authentication method, things are already too complicted. I would much rather change the API to use a URI, such as the one documented here:

https://www.postgresql.org/docs/9.4/static/libpq-connect.html  section 31.1.1.2  for example:
```
postgresql:///mydb?host=localhost&port=5433
postgresql:///dbname?host=/var/lib/postgresql
postgresql://
postgresql://localhost
postgresql://localhost:5433
postgresql://localhost/mydb
postgresql://user@localhost
postgresql://user:secret@localhost
postgresql://other@localhost/otherdb?connect_timeout=10&application_name=myapp
```

and let postgres do all of the parsing, as it wishes.  That would reduce the paramters to just one, ...

linas:
In fact, at the risk of breaking backward compatibility, I think I should make this change now...

linas:
pull req #1098 changes the API complete.  O have tested the following they all work for me:
```
> (sql-open "odbc://opencog_tester:cheese/opencog_test")
> (sql-open "postgres://opencog_tester@localhost/opencog_test")
> (sql-open "postgres://opencog_tester:cheese@localhost/opencog_test")
> (sql-open "postgres:///opencog_test?user=opencog_tester")
> (sql-open "postgres:///opencog_test?user=opencog_tester&host=localhost")
> (sql-open "postgres:///opencog_test?user=opencog_tester&password=cheese")
```

Note that some don't need an explicit passwd to be given.

linas:
pursuant to #1092  and passing the unit tests.

To pass the unit tests on buildbot, I needed to change the postgres URL to this:
```
postgres:///opencog_test?user=opencog_tester&password=cheese
```
however, I bet that is broken for your config, you probably need to use
```
postgres://opencog_tester@localhost/opencog_test
```
or maybe 
```
postgres://opencog_tester:cheese@localhost/opencog_test
```
You'd need to add that line to pg_hba.conf in order to be consistent with how buildbot tests.

misgeatgit:
OK, I should close this PR then.

Issue #1089, opened 2017-01-18
BindLink bug when only constant clauses


ngeiswei:
The following
```scheme
(cog-bind (BindLink (Concept "A") (Concept "B")))
```
should return
```scheme
(Set (Concept "B"))
```
because `(Concept "A")` is trivially satisfied but instead it returns
```scheme
(Set)
```

Add a virtual clause (who knows it might have worked)
```scheme
(cog-bind (BindLink (And (Concept "A") (Identical (Concept "A") (Concept "A"))) (Concept "B")))
```
yields the same wrong answer.

Ah, this one works! But certainly for the wrong reason...
```scheme
(cog-bind (BindLink (And (Concept "A") (Not (Identical (Concept "A") (Concept "A")))) (Concept "B")))
```

ngeiswei:
@linas I need this because sometimes backward inference chains don't have variables (that may tend to happen when the target is a tv query, but not only). I could skip entirely using a BindLink you may say but I still need to define preconditions that determine whether the inference is valid or not, and so I'm using the BindLink as a kind of `if-then` or `when`.

@linas I know you're busy, I'm taking care of it (already partially fixed on my local branch) but I wanted to let you know about it in case you want to say something about it.

ngeiswei:
Aha, the following
```scheme
(cog-satisfying-set (Get (Inheritance (Concept "A") (Concept "S"))))
```
returns
```scheme
$7 = (SetLink
   (ListLink
   )
)
```
Given that a BindLink is the composition of Put and Get the behavior of the BindLink really is inconsistent. Indeed a BindLink with a trivial constant clause should be seen as **getting** a empty ListLink, such should indeed trigger the rewrite term.

ngeiswei:
Hmm, by the same logic
```scheme
(cog-satisfying-set (Get (Concept "A")))
```
should return `(Set (List))` as above but it returns the empty set instead.

Anyway, I'm reviewing the code, I'll soon understand...

ngeiswei:
The reason `(cog-bind (BindLink (Inheritance (Concept "A") (Concept "B")) (Concept "C"))` returns `(Set (Concept "C"))` is because `InitiateSearchCB::neighbor_search` (wrongly) successfully processes it as if it were a normal clause. It fails on a clause like `(Concept "A")` because its neighbors obviously fall out of the candidate set.

I think the correct way to fix that is to

1. Fix `InitiateSearchCB::neighbor_search` to discard constant clauses altogether
[...]

PR #1099, opened 2017-01-24
Fix PM to support multiple constant and evaluatable clauses

ngeiswei:
Addresses #1089. @linas, I think it's correct but please have a look at it and merge if you think it's OK. Thanks!


ngeiswei:
@linas if you need to review a single commit it would be this one https://github.com/opencog/atomspace/pull/1099/commits/fa7246edbb39eae8e15a135b35ced0ae37c3c1f1

ngeiswei:
@linas it seems I have forgotten a use case, which is not covered in the unit tests, so you don't need to review and merge just yet.

ngeiswei:
I've added one more case (it wasn't the problem but more test is better so). @linas you can review and merge whenever possible.

opencog/atoms/execution/Instantiator.cc
```diff
@@ -135,7 +135,7 @@ Handle Instantiator::walk_tree(const Handle& expr)
 		// If we are here, we found a variable. Look it up. Return a
 		// grounding if it has one, otherwise return the variable
 		// itself.
-		std::map<Handle,Handle>::const_iterator it = _vmap->find(expr);
+		HandleMap::const_iterator it = _vmap->find(expr);
```
linas: these kinds of changes make the code harder to read and harder to understand.  Why make them?

opencog/atoms/pattern/PatternLink.cc
```diff
@@ -23,6 +23,8 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
+#include <boost/range/algorithm/find_if.hpp>
```
linas: I've finally concluded that boost consists of some of the most poorly written, poorly-designed C++ code in the known universe. It would probably be a good idea to try to minimize the amount of boost utilities that we use. Its painful, creaky, bug-prone.

opencog/atoms/execution/Instantiator.cc
linas: oh, never mind. HandleMap seems to be sufficiently generic, that this is worth-while.

linas:
Merged. It looks very reasonable to me; .. in particular, it looks like code that I would write, so I approve :-)

opencog/atoms/pattern/PatternLink.cc
ngeiswei: I agree with boost in general, but the range library is fairly simple and stable and saves a bit of the verbosity of the STL algorithms.

PR #1112, opened 2017-02-13
Improve unify and fix subunification

ngeiswei:
Another major refactoring of unify + fixes.

### DIFF
--- a/opencog/atoms/core/Variables.h
+++ b/opencog/atoms/core/Variables.h
@@ -106,8 +106,8 @@ protected:
 	                         Quotation quotation=Quotation()) const;
 };
 
-typedef std::map<Handle, const std::set<Type> > VariableTypeMap;
-typedef std::map<Handle, const OrderedHandleSet > VariableDeepTypeMap;
+typedef std::map<Handle, const std::set<Type>> VariableTypeMap;
+typedef std::map<Handle, const OrderedHandleSet> VariableDeepTypeMap;
 
 /// The Variables struct defines a list of typed variables "unbundled"
 /// from the hypergraph in which they normally occur. The goal of this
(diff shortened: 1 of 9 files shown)


ngeiswei:
You need to update cogutils as well.

opencog/atoms/core/Variables.h
```diff
@@ -132,6 +132,11 @@ struct Variables : public FreeVariables
 	VariableDeepTypeMap _deep_typemap;
 	VariableDeepTypeMap _fuzzy_typemap;
 
+	/// Return true iff all variables are well typed. For now only
+	/// simple types are supported, specifically if some variable is
+	/// simple typed NOTYPE, then it returns false.
```
linas: what does it mean for something to be "well-typed"? Can you give a definition or a few examples?

ngeiswei: Well-typed, IMO, means the associated set of "well-formed" expressions of that type is non-empty. For instance the intersection between ConceptNodes and PredicateNodes is the empty set, because neither of them inherits from the other, thus it is ill-typed. I don't think we have atomese representation yet for type intersection so I can't write you an atomese example of that. However we do have C++ functions that calculates type intersection, and as you can see one of them deals with "shallow" atom types, the C++ `Type` class, and this one would return NOTYPE when just empty intersection is detected. See https://github.com/opencog/atomspace/blob/master/opencog/atomutils/TypeUtils.cc#L318

linas: OK, Can you add a comment to the code that the definition of a "well-typed variable" is a variable that has an associated type? Otherwise, I am left guessing as to what "well-typed" really means, in this context.

PR #1114, opened 2017-02-14
Fix target unify

ngeiswei:
The problem was that `Handle::operator==` is pointer based.

During substitution, the atoms being created are added to any atomspace yet, as a result some functions, like filter_vardecl, may behave incorrectly.

I suggest we switch to content based comparison by default. I don't expect this would incur any overhead because we can first test pointer address, and only if there are different defer to content based.


ngeiswei:
Oh, so to be more specific, what fixed the problem here is this commit https://github.com/opencog/atomspace/pull/1114/commits/1056b5547258e79afdf758ecd98c390e060b5b1e but obviously this problem should be properly tackled once and for all. This is not the first time I'm having problems due to that.

linas:
The true failure here is the failure to distinguish equality and equivalence. They are just not the same thing.  Pointer-comparison is "syntactic equality" -- content comparison is "semantic equivalence".  As long as the API tries to use a single symbol for both, `operator==()` you will just have errors and bugs. 

My knee-jerk reaction would be to create an `equiv()` method (... and I thought there already was one!?) and just use that, when you want to test for equivalence.

linas:
That is, pretend you are coding in C, using plain pointers and plain C structs.   Then, Handle is a pointer and `operator==` is by definition a pointer compare.  Don't break that, don't redefine that, you will break ordinary textbook C semantics.  

The standard way of doing content compare in C is to deref the pointer.  Thus:
```
struct blah  *p, *q;
if (p == q) // pointer compare
if (*p == *q) // content compare
``` 

Handles are like pointers.  If you want to content-compare, dereference the pointer first.

ngeiswei:
Agreed about deferencing the pointer to get content comparison. The question is what to do with Handle? I suppose Handle should be viewed as a pointer, but it might be a bit more than that. I don't know.

ngeiswei:
BTW, I create a branch to attempt to address that. Basically
```c++
Handle::operator==(const Atom* atom)
```
remains pointer based. However
```
Handle::operator==(const Handle& h)
```
becomes content based.

See https://github.com/ngeiswei/atomspace/commit/f0a520db6617bd8454c43bb624706de54722a77e .

It actually breaks a couple utests, like MultiSpaceUTest.

linas:
The current code has tension between Handle, AtomPtr and ProtoAtomPtr which needs to be resolved. The correct resolution requires finishing the protoatom design, which is almost done, but still has a few tricky/complex decisions left to be made.  One of these is that Handle and AtomPtr could be essentially the same thing, however, they cannot be the ProtoAtomPtr, and doing this wrong has significant performance impacts in casting.  I have this on my list of things to do, but don't have time for it for a the next few weeks or month or so.

PR #1121, opened 2017-02-20
Convert TruthValue to use ProtoAtom as a base type

linas:
Small step, part of the work on issue #513

### DIFF
--- a/opencog/haskell/Utils_CWrapper.cpp
+++ b/opencog/haskell/Utils_CWrapper.cpp
@@ -5,39 +5,41 @@ int Utils_toRawType(TruthValuePtr tv
                     , Type* tv_type
                     , double* parameters)
 {
-switch(tv->getType())
-    {
-        case SIMPLE_TRUTH_VALUE: {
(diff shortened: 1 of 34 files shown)


opencog/haskell/Utils_CWrapper.cpp
```diff
@@ -2,42 +2,44 @@
...
-            parameters[1]=tv->getCount();
-            parameters[2]=tv->getConfidence();
-            break; }
-        default:
-            throw RuntimeException(TRACE_INFO,
-                "Invalid TruthValue Type.");
-            break;
+    Type tvt = tv->getType();
+    if (tvt == SIMPLE_TRUTH_VALUE){
+        parameters[0]=tv->getMean();
```
AmeBel: Why switch from `switch` statements to `if` statements?

ngeiswei: Because for some technical reason (that I forgot) atom types cannot be used in switch.

PR #1129, opened 2017-02-22
Erase and-BIT from bit_as as well

ngeiswei:


misgeatgit:
Compiler is throwing the following error both on the CI server and my machine.
```
BIT.cc:648:31: error: no matching function for call to ‘std::vector<opencog::AndBIT>::erase(std::vector<opencog::AndBIT>::const_iterator&, std::vector<opencog::AndBIT>::const_iterator&)’
  return andbits.erase(from, to);
```

ngeiswei:
Weird, what is yours and the CI's gcc version?

ngeiswei:
If you comment out the declaration and definition of BIT::erase(from, to), do you get other errors?

misgeatgit:
My gcc was set on v 4.8.4, the error vanished after setting gcc to version 6.2.  I don't know which version of gcc the CI server uses, though.

ngeiswei:
That's what I thought, too old gcc don't follow C++11 standard completely.

I suppose upgrading the CI server would be indicated. Otherwise, if that's troublesome, let me know, I can rewrite it to be backward compatible with older gcc. In such case I'd need to know the exact version of gcc on the CI server.

AmeBel:
gcc version is 4.8.4, we are still on ubuntu:14.04

AmeBel:
@ngeiswei misgana suggested I upgrade the gcc version used by the ci containers. If that is ok, by you will do that.

pennachin:
@AmeBel have you or anyone else tried to upgrade the containers to use ubuntu 16.04? If that's possible without too much work, I guess it would be better than upgrading just gcc and keeping a three year old base system, no?

AmeBel:
For the non-robotics repos, upgrading to 16.04 is at most a days work, once the decision is made. Wenwei tells me that there is an issue with running blender on 16.04 while using https://github.com/hansonrobotics/HEAD, though.

ngeiswei:
I think it's fixed, let me know otherwise. I would still recommend upgrading gcc because it's current support for C++11 is incomplete, but it shouldn't be an immediate requirement.

PR #1132, opened 2017-02-26
Implement a "valuation space"

linas:
This continues the work on proto-atoms, as described in issue #513 In particular, it implements the concept of a "valuation" that shows up towards the end of the conversation in  issue #2333

The "Valuation" is kind-of-like an EvaluationLink, except that it is not an atom; its not stored in the atomspace.  Instead, it is kind of like a key-value system, where the key is kind-of-like a PredicateNode, and the value is any value (for example, a truth value, and attention value, or any other kind). The valuation couples together a triple (atom, key, value) so that, for any given atom, you can specify a key, and obtain a value for that (key,atom) pair.   This generalizes the notion of TV or AV, where TV is a key, and AV is a key, and the TV of an atom is the truth value of that atom.

The ValuationSpace is ... the least irritating solution I could find that solves a technical problem with managing smart pointers and memory management.  Its sort of like that AtomSpace, except that it contains Valuations.

Unlike atoms, valuations are mutable.  Unlike atoms, valuations do not store an incoming set.  Unlike atoms, valuations cannot be searched for, except by the (atom,key) pair.  Thus, they are in many ways simpler and lighter-weight than atoms, and are useful for storing/managing different kinds of data.

Remaining work is to provide a more usable API, a scheme and python binding, and a way of storing these in a database.


ngeiswei:
@linas I suspect you've introduce some compile error over the examples. See below when doing `make examples`

```
/home/nilg/OpenCog/atomspace/examples/c++-guile/PrimitiveExample.cc:35:5: error: ‘NodePtr’ was not declared in this scope
     NodePtr n = NodeCast(h);
     ^~~~~~~
```

linas:
It needs #inclide Node.h to fix it.

PR #1134, opened 2017-02-27
Remove multi-iterator version of BIT::erase

ngeiswei:
so that it compiles under older versions of gcc, see #1129


linas:
Its not compiling under build-bot, see for example, this:

http://61.92.69.39:8080/job/ci-atomspace/1064/console

ngeiswei:
@linas I have templatized the method and the error has gone away, however there's now this one http://61.92.69.39:8080/job/ci-atomspace/1065/console

PR #1142, opened 2017-03-03
Ongoing work to implement ProtoAtoms in SQL

linas:
This is a partial implementation for storing protoatoms (generic values) in the SQL backend. It passes the unit tests, and the database layout is still backwards-compatible with the old layout. The next pull req in this series will break this compatbility and will use the new system to store truth values.


linas:
Note that, in order to pass the unit tests, the new database structure has to be loaded.

PR #1145, opened 2017-03-06
Implement restore of values from SQL DB

linas:
This provides the second half of pull-requests #1136 and #1138 -- values can now be restored from the SQL DB.

Part of the ongoing work to finish protoatom, as described in #513.


linas:
Merge even though there are two failing unit tests: FetchUtest, which I will fix later, and ValueUTest, which passes for me; it may need a jenkins postgres sql update

PR #1152, opened 2017-03-08
Simplify FC

ngeiswei:
Massive code simplification (almost a 1000 lines less) using `Rule::unify_source`.


leungmanhin:
Looks like R2L is affected by this update, `nlp-parse` no longer generates any R2L outputs...

ngeiswei:
That's because the R2L rules do not follow the new standard. Sorry I completely forgot about them.

It's all explained here

http://wiki.opencog.org/w/URE_Configuration_Format#Defining_Rules

but there are two very important aspects of that standard

1. PM Clauses and URE premises are disentangled.
2. Conclusions are represented as first argument of the rule's rewrite ExecutionOutput, and premises as following arguments.

So someone, maybe me if you're in a hurry and have no else to do it, has to convert all R2L rules according to that standard.

For your relief this standard has been very carefully thought and is not about to change. I'll document soon better why I made such decisions.

ngeiswei:
Took a look at R2L rules. It clearly will be much less work for me to make the FC back-compatible with the old standard, than rewriting all rules to accommodate the new one. This should be fine as long as the R2L rules aren't used by the Backward Chainer.

Will take care of that tomorrow.

leungmanhin:
I see, if it's not too much work it would be great to be R2L compatible for a little longer :)
Thanks a lot!

ngeiswei:
PR #1158 should have fixed the problem, let me know otherwise.

leungmanhin:
Hmm... R2L is still not working

leungmanhin:
Here is a quick test case:
```
(use-modules (opencog) (opencog nlp) (opencog nlp relex2logic) (opencog nlp chatbot) (opencog nlp fuzzy))

(get-r2l-set-of-sent (car (nlp-parse "Bob eats apples")))
```
It should return the R2L outputs of the input sentence. It's giving an empty `SetLink` at the moment

ngeiswei:
@leungmanhin do you know the R2L rule that is supposed to be triggered and isn't, in this parse? That would help me.

leungmanhin:
Yes, for example [svo](https://github.com/opencog/opencog/blob/master/opencog/nlp/relex2logic/rules/svo.scm), [declarative](https://github.com/opencog/opencog/blob/master/opencog/nlp/relex2logic/rules/declarative.scm), and [definite](https://github.com/opencog/opencog/blob/master/opencog/nlp/relex2logic/rules/definite.scm) should be triggered in the example "Bob eats apples"

ngeiswei:
Hopefully PR #1160 fixes it.

leungmanhin:
Yes it's fixed, thank you @ngeiswei

PR #1153, opened 2017-03-09
Thread safe access and update

misgeatgit:


linas:
yes, this looks fine to me.

PR #1154, opened 2017-03-09
ProtoAtom/Value guile bindings work

linas:
Guile bindings for #513 -- the protoatom work.  This should be enough to deal with the core work for guile bindings for values, although some more work is needed to for spit & polish.

### DIFF
--- a/tests/scm/BasicSCMUTest.cxxtest
+++ b/tests/scm/BasicSCMUTest.cxxtest
@@ -83,6 +83,9 @@ class BasicSCMUTest :  public CxxTest::TestSuite
 	// TODO
 	//void test_evaluation_link(void);
 
+	// Values
+	void test_values(void);
+
 	// TVs
 	void check_truth_value(const TruthValuePtr& tv);
 	void test_truth_value(void);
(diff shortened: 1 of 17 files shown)


tests/scm/BasicSCMUTest.cxxtest
```diff
@@ -287,6 +290,50 @@ void BasicSCMUTest::test_truth_value(void)
...
+
+	ProtoAtomPtr lv = eval->eval_v("(cog-value h k)");
+	TSM_ASSERT("Failed to get value", not eval->eval_error());
+	TSM_ASSERT("Failed to get correct value", l == lv);
+
+	ProtoAtomPtr alv = h->getValue(k);
+	TSM_ASSERT("Failed to get correct value", l == alv);
+
+	eval->eval("(cog-set-value! h k '(0.1 0.2 0.3))");
+	TSM_ASSERT("Failed to set float", not eval->eval_error());
```
AmeBel: Continuing on this, I get the following
```
scheme@(guile-user)> (cog-set-value! h k '("qwe" "rty" "uio"))                                         
$11 = (ConceptNode "foo")

scheme@(guile-user)> (cog-value h k)
$12 = (StringValue "qwe" "rty" "uio")
scheme@(guile-user)> f
$13 = (FloatValue 0.100000 0.200000 0.300000)
scheme@(guile-user)> s
$14 = (StringValue "x" "y" "z" "another one")
scheme@(guile-user)> (cog-set-value! h k '(f s))
ERROR: In procedure cog-set-value!:
ERROR: In procedure cog-set-value!: Wrong type argument in position 3 (expecting opencog value): f

Entering a new prompt.  Type `,bt' for a backtrace or `,q' to continue.
```
Bug or user-error ?

linas: what you want is to say  `(cog-set-value! h k (list f s))`  -- and that will work.

the tick mark ' actually means "quote"  not "list" ... I'm trying to figure out if this is a user-error or a bug...

linas: what you want is to say  `(cog-set-value! h k (list f s))`  -- and that will work.

the tick mark ' actually means "quote"  not "list" ... I'm trying to figure out if this is a user-error or a bug...

linas: I think I just made this work in d9208a77d04a1e0379bec71fb6769a3ddd51912a --

AmeBel: yap it works

PR #1159, opened 2017-03-10
Change importance values type from short to double

misgeatgit:
This will allow us to do floating point importance calculations without losing precision as mentioned  [here] (https://github.com/opencog/opencog/issues/676). Opencog repo seems to be building fine when compiled with these changes. So, I am tempted to make a PR.


linas:
looks good to me. I see no harm in this.

PR #1163, opened 2017-03-12
Fix Build Issue on Arch Linux (incomplete)

rTreutlein:
Same as for cogutils why do we need a custom FindBoost.

I hope requiring the exact python version doesn't break anything but it at least makes cmake realize that I hasn't found the right version on Arch Linux. Not sure how to make it find python2.7 instead of python3.


linas:
Looks good to me. Amen said he'd merge.

AmeBel:
@rTreutlein After you confirmed https://github.com/rTreutlein/atomspace/pull/1 works, please propagate the changes to opencog/opencog as well. Thanks.

Issue #1048, opened 2016-12-28
Move atom factories to classserver


linas:
The assorted linking issues associated with the atom factories would be best and most easily resolved by moving them to the ClassServer.  That is, the ClassServer would act as the factory.

I sat down to try this just now, and realized that the design is impacted by the ProtoAtom work (issue #513)  as well as the AttentionBank refatoring (issue https://github.com/opencog/opencog/issues/2333) This it seems unwise to actually do this work until those issues get resolved.

ngeiswei:
:+1:

linas:
Close -- resolved by #1164

PR #1164, opened 2017-03-13
Move atom factory to the classserver

linas:
This implements design request #1048.  Having all of the atom factories located in the AtomTable caused a number of linking and dependency issues, which have been festering for years. This should solve that.

This was supposed to be a quick, easy lazy-afternoon task, but it turned out to require a two-and-a-half day coding marathon. It boils the ocean, changes about 100 files.  Oh well.  It works, although it appears to expose some bugs in the Unifier; See issue #1165 for  description.

This requires a matching pull request opencog/opencog#2630 to compile.

### DIFF
--- a/opencog/atoms/base/ClassServer.h
+++ b/opencog/atoms/base/ClassServer.h
@@ -253,7 +253,7 @@ public:
 
 ClassServer& classserver();
 
-#define DECLARE_LINK_FACTORY(CNAME,CTYPE)                         \
+#define DEFINE_LINK_FACTORY(CNAME,CTYPE)                          \
                                                                   \
 Handle CNAME::factory(const Handle& base)                         \
 {                                                                 \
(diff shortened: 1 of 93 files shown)


linas:
merging desipte the above-mentioned bug #1165 - it appears to be a valid catch.

opencog/atoms/base/ClassServer.h
```diff
@@ -210,6 +253,21 @@ class ClassServer
 
 ClassServer& classserver();
 
+#define DEFINE_LINK_FACTORY(CNAME,CTYPE)                          \
+                                                                  \
```
AmeBel: @linas searched the repository and build directory for the use of this macro in the creation of atoms such as ConceptNode, couldn't find any. What is the magic here?

linas: CocneptNode doesn't need this. Most atoms don't -- only the atoms that corresponds to C++ classes, and need to have thier constructors run, when the atom is created.   Right now, these are some of the link types, and NumberNode and TypeNode.   The Number& TypeNdes are still done by hand (they don't interact badly with the python bindings)

Issue #1165, opened 2017-03-13
Chainer Unit tests fail after pull req #1164


linas:
Pull req #1164 changes a lot of code, but conceptually does very little: it just moves the atom factories over to the classserver.  This causes, in some cases, constructors for certain atom types to run earlier than before.  These constructors perform some checks to make sure the atom is in a valid format ... and the chainer unit tests seem to be tripping on some invalidly-formated atoms.

I am not sure why this is the case, there is nothing obvious that pops up.  Easiest to understand is BITUTest: the VariableList constructor prints the message below.  The message seems accurate: the variably list contains a TypeChoice that is just not attached to anything, and it also includes an EvaluationLink which clearly cannot belong there.  It would appear that something somewhere is missing a few parenthesis in the right place.
 
```
In BITUTest::test_expand_1:
/home/linas/src/atomspace/tests/rule-engine/BITUTest.cxxtest:44: Error: Test failed: Expected a VariableNode or a TypedVariableLink, got: EvaluationLink
VariableList is (VariableList
  (TypedVariableLink
    (TypedVariableLink
      (VariableNode "$X") ; [6809909406030619949][1]
      (TypeNode "ConceptNode") ; [3788388250666256890][1]
    ) ; [17718894144153964593][18446744073709551615]
    (TypeChoice
      (TypeNode "TypedVariableLink") ; [1359653833156464510][1]
      (TypeNode "VariableNode") ; [5989952939322278073][1]
      (TypeNode "VariableList") ; [7311909679895453518][1]
    ) ; [11822067860997730867][18446744073709551615]
  ) ; [16814188638364779502][18446744073709551615]
  (EvaluationLink
    (PredicateNode "contain") ; [6339226537938665996][1]
    (ListLink
      (ConceptNode "treatment-1") ; [1169226509931710999][1]
      (ConceptNode "compound-A") ; [5571266438078870282][1]
    ) ; [16887940857902133716][18446744073709551615]
  ) ; [15649912025792571919][18446744073709551615]
) ; [17958762067029919729][18446744073709551615]
 (/home/linas/src/atomspace/opencog/atoms/core/VariableList.cc:62)
.
Failed 1 and Skipped 0 of 3 tests
Success rate: 66%
```

ngeiswei:
@linas I suspect the scope link no longer detects its variables correctly, still investigating...

ngeiswei:
Hmm, apparently it does detect its variables, I still don't understand what's going on.

[...]

PR #1167, opened 2017-03-13
Add unit test for ScopeLink

ngeiswei:
in particular to test ScopeLink variable extraction and access.


ngeiswei:
Small unit test to reproduce #1165

Issue #1176, opened 2017-03-20
bindlink value clash


ngeiswei:
I'm getting the same hash value for 2 different non alpha-equivalent bindlinks. I know collision is impossible to avoid but since I'm getting that now I suspect the hash value algo isn't terribly good (either that or I'm incredibly unlucky).

The 2 clashing bindlinks are

```scheme
(BindLink
  (VariableList
    (TypedVariableLink
      (VariableNode "$B-a083600") ; [1075803195405159200][3]
      (TypeChoice
        (TypeNode "OrLink") ; [1641200931040822982][1]
        (TypeNode "AndLink") ; [5150993344963829271][1]
        (TypeNode "LambdaLink") ; [6352458681478507967][1]
        (TypeNode "NotLink") ; [7029353635928836065][1]
        (TypeNode "PredicateNode") ; [7305492803481351454][1]
      ) ; [12553683925648135849][1]
    ) ; [12502753568736379187][3]
    (TypedVariableLink
      (VariableNode "$B-5560bed3") ; [954956885300767879][3]
      (TypeChoice
        (TypeNode "OrLink") ; [1641200931040822982][1]
        (TypeNode "AndLink") ; [5150993344963829271][1]
        (TypeNode "LambdaLink") ; [6352458681478507967][1]
        (TypeNode "NotLink") ; [7029353635928836065][1]
        (TypeNode "PredicateNode") ; [7305492803481351454][1]
      ) ; [12553683925648135849][1]
    ) ; [17738197372146241402][3]
    (TypedVariableLink
      (VariableNode "$Q-5fa8b92b") ; [8930023734698229664][3]
      (TypeChoice
        (TypeNode "LambdaLink") ; [6352458681478507967][1]
        (TypeNode "PredicateNode") ; [7305492803481351454][1]
      ) ; [15286865261196305675][1]
    ) ; [16170795669022151701][3]
  ) ; [13184149590889680758][3]
  (AndLink
    (VariableNode "$B-a083600") ; [1075803195405159200][3]
    (VariableNode "$Q-5fa8b92b") ; [8930023734698229664][3]
    (NotLink
      (EqualLink
        (VariableNode "$B-a083600") ; [1075803195405159200][3]
        (VariableNode "$Q-5fa8b92b") ; [8930023734698229664][3]
      ) ; [9530461658253061016][3]
    ) ; [9550780633133926061][3]
    (NotLink
      (IdenticalLink
[...]

PR #1177, opened 2017-03-20
Fix term hash

ngeiswei:
Fix for issue #1176

### DIFF
--- a/opencog/atoms/core/ScopeLink.cc
+++ b/opencog/atoms/core/ScopeLink.cc
@@ -301,7 +301,7 @@ ContentHash ScopeLink::term_hash(const Handle& h,
 	// Quotation
 	quotation.update(t);
 
-	// Other embedded ScopeLinks might be hiding some of our varialbes...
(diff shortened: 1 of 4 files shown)


ngeiswei:
@linas, looks good to me but please just in case have a look at https://github.com/opencog/atomspace/pull/1177/commits/b75c18272b688a43686328f6ee20062e3d3cceea and merge if you think it's OK.

opencog/atoms/core/ScopeLink.cc
```diff
@@ -327,7 +327,7 @@ ContentHash ScopeLink::term_hash(const Handle& h,
 	ContentHash hsh = ((1UL<<8) - 59) * t;
 	for (const Handle& ho: h->getOutgoingSet())
 	{
-		hsh += mixer * (1UL<<5) + term_hash(ho, bound_vars, quotation);
+		hsh += mixer * (hsh<<5) + term_hash(ho, bound_vars, quotation);
```
linas: This is the most surprising change, but it seems to be correct. Huh.

PR #1181, opened 2017-03-23
Remove constant clauses from fcs

ngeiswei:
+ improve pattern matcher constant clause detection. Indeed some ScopeLinks and EvaluationLinks where not correctly classified as constant.


opencog/atomutils/Unify.cc
```diff
@@ -348,14 +355,126 @@ bool Unify::is_bound_to_ancestor(const Variables& variables,
...
-	// non variables as well
-	//
 	// TODO: make sure that ts.second contains the declaration of all
 	// variables
-	Handle h = bl->alpha_conversion(strip_context(ts.first), ts.second);
+	return substitute(bl, strip_context(ts.first), ts.second);
+}
+
+Handle Unify::substitute(BindLinkPtr bl, const HandleMap& var2val,
+                         Handle vardecl)
```
linas: I'm unclear about why this method is here, and not elsewhere. Naively, I would think that since the first argument must be a BindLinkPtr, then perhaps this should be a method on the BindLink C++ class, itself?

linas:
Also, FYI, there is some quasi-generic "substitute" code lurking inside of both PutLink (you can think of PutLink as being a kind-of SubstituteLink)  Merging that code into a generic substitute system might be nice, but its low priority.

Note also that the Instantiator class in the atoms/execution directory also does a kind-of substitution. Similar but different.  (I think it even does "lazy substitution" of not actually doing it unless its actually required)

Note also that MapLink is kind-of-ish kind-of-like BindLink, except that it only "searches" its second argument, instead of searching all of the AtomSpace.   Currently, MapLink is a kind of BindLink-lite, missing many of the features and capabilities that BindLink has. I sometimes think about adding these, or making these two share more common code, but that would actually be a big task, so I loose interest in doing that work.

opencog/atomutils/Unify.cc
ngeiswei: I initially put it in `BindLink`, but then I realized that it slightly collides in name with `LambdaLink::substitute` which is doing something a bit different. The former is a more specific kind of substitution that also removes constant clauses and useless/ill quotations and the unifier is the only consumer of it, so I thought it's better to move it to `Unify`. But maybe it could be in `BindLink`.

ngeiswei:
Yeah, I'm aware of all these substitution codes, and I tried to re-use them at first but I eventually realized the only piece that I can re-use is `Variables::substitute_*`. Refactoring the code to enable more re-usability could be possible, but I'm not even so sure.

linas:
OK, no action required, just wanted to be sure that you knew.

opencog/atomutils/Unify.cc
linas: Perhaps you could call the method "rewrite()" or "reduce()" or "beta_reduce()" or something like that?

ngeiswei: I thought of `beta_reduce()` but again was afraid of conflating with the existing methods of that name. I though of `partial_substitute()` as well. Maybe `partial_beta_reduce()` could be appropriate.

PR #1189, opened 2017-03-28
Update CWrapper and HaskellBindings

rTreutlein:
This fixes the HaskellBindings that broke after the move of TruthValues to ProtoAtoms.
Also includes various other fixes.


opencog/haskell/TruthValue_CWrapper.cpp
```diff
@@ -0,0 +1,107 @@
...
+                        , double* parameters)
+{
+    const std::string & tvt = classserver().getTypeName(tv->getType());
+
+    *tv_type = (char*) malloc(sizeof(char) * (tvt.length()+1));
+    if(! *tv_type)
+        throw RuntimeException(TRACE_INFO,"Failed malloc.");
+    std::strcpy(*tv_type, tvt.c_str());
+
+    if (strcmp(*tv_type,"SimpleTruthValue") == 0){
```
linas: A simpler solution would have been to just cast these to a FloatValue, which already defines an array of floats that occur in exactly the correct order.  Here, you are confusingly defining an array that just might happen to be in a different order from the "natural" order of the truth value.

Also, given that all truth values are now arrays, it would simplify the code -- you would not need any kind of special-casing here, for each of the different types.

linas: Also, you might want to think about eventually supporting other kinds of values: not just truth values, but generically, string values, and link values (which are just lists of other values ... i.e. the values are now fully hierarchical).

If this is unclear, its because I am writing the wiki documentation for this now.  See issue #513 for other status and details

rTreutlein: Yea I thought a bit about supporting all kinds of values but wasn't quite sure on how to best do that right now. So I decide to just change the minimal amount needed to get it to work.
But I am a bit more comfortable with that now so will implement that next.

PR #1190, opened 2017-03-29
Type defintions update

linas:
Update the type definitions so that they are in line with what the wiki says.  This includes several new proposed types: these seemed like a "good enough idea" that they can be added.

### DIFF
--- a/opencog/atoms/TypeNode.h
+++ b/opencog/atoms/TypeNode.h
@@ -61,7 +61,8 @@ public:
 		: Node(n),
 		  value(classserver().getType(n.getName()))
 	{
-		OC_ASSERT(TYPE_NODE == n.getType(), "Bad TypeNode constructor!");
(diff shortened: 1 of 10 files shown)


ngeiswei:
@linas if the backward chainer utest fails on `test_criminal_trace()` only then it's likely not serious, as this test is sensitive to the atoms values, basically I have to re-adjust that test whenever something changes the atom values. Maybe I should modify it to be more resilient to changes.

linas:
Yep, that's the one.

linas:
Nil, note that undoing commit d8280b5 causes RuleUtest to fail, and I don't quite see why (I can imagine, but...)  That diff really is not important, at all, but it just seemed that it was "the right thing to do".

ngeiswei:
I think this is due to the fact that it slightly changes the definition of what is a constant clause, but I'm not sure. But if this commit doesn't matter to you then fine.

linas:
OK thanks

PR #1193, opened 2017-03-30
Fix comments

ngeiswei:


opencog/rule-engine/backwardchainer/BackwardChainer.cc
```diff
@@ -197,7 +197,17 @@ void BackwardChainer::fulfill_fcs(const Handle& fcs)
...
+	// Warning: since tmp_as is a child of _as, TVs of existing atoms
+	// in _as, that are modified by running fcs will be modified on
+	// _as as well. This can create involontary TVs changes, hopefully
+	// mitigated by the merging the TVs properly (for now the one with
+	// the highest confidence wins). To avoid that side effect, we
+	// could operate on a copy the atomspace, of its zone of focus. Or
+	// alternatively modify some HypotheticalLink wrapping the atoms
+	// of concerns instead of the atoms themselves, and only modify
+	// the atoms if there are existing results to copy back to _as.
 	Handle hresult = bindlink(&tmp_as, fcs);
```
linas: This is a very interesting comment, and it's what I was talking about a year or two ago about "kripke frames". So, as the chainer is running, you want to have some way of making it look as if there were multiple atomspaces, each one holding possibly different TV's for the same atom.  The TV's in the "main" atomspace would only be updated (maybe) at the very end of the chaining.

I was calling each of these different atomspaces "parallel universes". 

The comment above seems like you are struggling with exactly this idea. 

The actual implementation of this idea does not actually have to use different atomspaces .. any technique where you can have one atom with different TV's on them will be enough.

ngeiswei: Yes, that is exactly the incarnation of the "kripke frames" comments you made over the years. I'm not sure yet how to approach that, we might want to keep all TVs/proofs in different atomspaces or have "superposed" atoms/TVs in the same AtomSpace, or merge them as they come. The merger TV function will need a way to estimate how much evidence 2 TVs/proofs have in common, that's gonna the hard part.

ngeiswei: BTW, the reason I'm storing the proof and its results in a different atomspace is solely to not pollute the existing atomspace. The kripke-frames/merging problem is crudely solved by merging the results as they come, according to the most conservative highest-confidence-wins merging scheme.

linas: OK.  Perhpas "merging as we go" might sort-of work for simple cases, but breaks down when you have a proof-by-contradiction.  Such proofs start with "Assume X" and so you raise the truth/confidence of X and then, at the end, with the RAA step, do you find out that X is false, and now its hard to undo the damage caused by all the intermediate steps.

This is one of those "important but not urgent" problems.

ngeiswei: I think to do such reasoning we would wrap the "Assume X" in a context, implication or inheritance link, to avoid being temporarily inconsistent with the atomspace. So the context links or such would be used to essentially describe kripke frames.

linas: OK, that would work, but then everything derived from assuming X has to be wrapped in the same context as well.  I like the idea of using contexts; just that this now reifies the problem: we would need a context-management subsystem in opencog.

ngeiswei: A context-management subsystem is not absolutely required, the PLN rules alone (once completed) would be enough. However it would be very handy, that way knowledge within the same context can easily be chained together. We made such experiments with Ari's old C++ implementation but never with the new one.

PR #1194, opened 2017-03-31
Allow VariableNode as type in variable declaration

ngeiswei:
@linas this is the first PR of a series of potentially controversial PRs. Look at it carefully. You may observe that `BackwardChainerUTest::test_conditional_partial_instantiation()` does not pass without it. Let me know if that's acceptable, and if not what alternative would you suggest.

### DIFF
--- a/opencog/atoms/core/VariableList.cc
+++ b/opencog/atoms/core/VariableList.cc
@@ -278,6 +278,12 @@ void VariableList::get_vartype(const Handle& htypelink)
 		ts.insert(vartype);
 		_varlist._fuzzy_typemap.insert({varname, ts});
 	}
+	else if (VARIABLE_NODE == t)
+	{
+		// This occurs when the variable type is a variable to be
+		// matched by the pattern matcher. There's nothing to do
+		// except not throwing an exception.
+	}
 	else
 	{
 		throw SyntaxException(TRACE_INFO,
--- a/tests/rule-engine/BackwardChainerUTest.cxxtest
+++ b/tests/rule-engine/BackwardChainerUTest.cxxtest
@@ -160,7 +160,6 @@ void BackwardChainerUTest::test_deduction()
 	_eval.eval("(load-from-path \"tests/rule-engine/bc-transitive-closure.scm\")");
 	randGen().seed(0);
 
-	// load 1 deduction rule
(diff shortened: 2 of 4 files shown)


ngeiswei:
Oh, fuck, I've mechanically merged. Well, feel free to revert if you think that's not acceptable.

linas:
I can't make any reasonable comments based on just reading this diff. Whatever this new thing is, you will need to create a wiki page to explain how it works ... might be easier to just explain it in an email first? 

As to the unit test, I don't see at all anywhere where this new strange typing is required -- the ImplicationScope in the unit test itself looks completely normal and doesn't use any new typing style...

ngeiswei:
The `conditional-partial-instantiation-meta-rule.scm` is complicated and hard to read, but put simply it requires to pattern match what is going on inside a variable declaration, to extract the type of some variable. For instance if you have the grounded data
```
ScopeLink
  VariableList
    TypedVariable
      Variable "$X"
      Type "ConceptNode"
    TypedVariable
      Variable "$Y"
      Type "PredicateNode"
  <body>
```

the `conditional-partial-instantiation-meta-rule.scm` will want to look inside the scope link's vardecl and know that `$X` is of type `ConceptNode` and `$Y` is of type `PredicateNode`. In order to do that it requires to write patterns describing variable declarations, thus having the type itself be a variable, like
```
BindLink
  VariableList
    TypedVariable
      Variable "$XVar"
      Type "VariableNode"
    TypedVariable
      Variable "$XType"
      Type "TypeNode"
  Quote
    ScopeLink
      Unquote
        VariableList
          TypedVariable
            Variable "$XVar"
            Variable "$XType"
        ...
```

Do you see what I mean?

Agreed, this should be explained somewhere, perhaps in the `TypedVariableLink` wiki page. And probably add a simpler utest under `test/query`.

linas:
OK yes, I see. I suspected it might be something like that, hadn't quite thought it through.

OK, right, documenting this is pretty low priority.  A super-simple utest would be good, as there is a small but finite risk that this might get broken over the next few years, and if no one understands the utest, then it might stay broken.

linas:
Actually, the existing unit test is simple enough, so there is nothing to be done.

I didn't understand it until your comment above, so perhaps that comment should be copied into the unit test...

PR #1196, opened 2017-04-03
Add comment to BC test_conditional_partial_instantiation

ngeiswei:
Explains why a type VariableNode may be useful and thus accepted as
valid atom type in a TypedVariable link.


ngeiswei:
Add comment as suggested in #1194

PR #1197, opened 2017-04-04
Improve URE API and scheme binding

ngeiswei:
- Expose the source/target variable declaration to cog-fc and cog-bc
- Have the forward chainer support source variable declaration


ngeiswei:
Please note that the `cog-fc` and `cog-bc` scheme API has changed a bit. The rule-base comes first, then the source/target, its vardecl, and finally focus-set. See https://github.com/opencog/atomspace/tree/master//opencog/rule-engine#how-to-call-the-forward-chainer-from-a-scheme-interface

PR #1200, opened 2017-04-05
Add silent exception in do execute

ngeiswei:
@linas you need to review that one carefully.

It adds a silent flag in ExecutionOutputLink::execute and related callers and callees. The silent flag is set to false by default. However some callers like `Implicator::grounding` call it (via instantiate) with silent set to true, which might be considered too silent.

### DIFF
--- a/opencog/atoms/core/Variables.h
+++ b/opencog/atoms/core/Variables.h
@@ -181,7 +181,7 @@ struct Variables : public FreeVariables
 	// filtering, where type mis-checks are expected and normal.
 	Handle substitute(const Handle& tree,
 	                  const HandleSeq& vals,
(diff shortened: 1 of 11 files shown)


ngeiswei:
I think the only contentious commit is https://github.com/opencog/atomspace/pull/1200/commits/e10057c852dfdd92298bb7c0f77471fbebeef969 all the others consists mostly of defining and passing silent as it should.

linas:
I think its fine; merge whenever you're ready

ngeiswei:
The BackwardChainerUTest fails, but it passes on both my laptop and desktop (which have 2 different versions of gcc and generate different random sequences). I'm merging, will debug it based on someone else output if the problem persists.

PR #1205, opened 2017-04-21
Relax ExecutionOutputLink checks

ngeiswei:
Consider `ExecutionOutputLink` as executable iff its schema is grounded.

@linas, I don't think there's anything harmful here, but you probably ought to review it.

Ultimately we'll want to execute ungrounded schemata as well, but it seems subtle enough that we'd rather have that taken care of by another component (probably the URE with dedicated rules I assume).

### DIFF
--- a/opencog/atoms/execution/ExecutionOutputLink.cc
+++ b/opencog/atoms/execution/ExecutionOutputLink.cc
@@ -45,9 +45,8 @@ public:
 
 void ExecutionOutputLink::check_schema(const Handle& schema) const
 {
-	if (DEFINED_SCHEMA_NODE != schema->getType() and
-	    LAMBDA_LINK != schema->getType() and
-	    GROUNDED_SCHEMA_NODE != schema->getType())
+	if (not classserver().isA(schema->getType(), SCHEMA_NODE) and
+	    LAMBDA_LINK != schema->getType())
 	{
 		throw SyntaxException(TRACE_INFO,
 		                      "ExecutionOutputLink must have schema! Got %s",
--- a/opencog/atoms/execution/ExecutionOutputLink.cc
+++ b/opencog/atoms/execution/ExecutionOutputLink.cc
@@ -101,14 +101,19 @@ ExecutionOutputLink::ExecutionOutputLink(const Link& l)
 ///
 Handle ExecutionOutputLink::execute(AtomSpace* as, bool silent) const
 {
+	if (_outgoing[0]->getType() != GROUNDED_SCHEMA_NODE) {
+		LAZY_LOG_FINE << "Not a grounded schema. Do not execute it";
+		return getHandle();
+	}
+
 	return do_execute(as, _outgoing[0], _outgoing[1], silent);
 }
 
 /// do_execute -- execute the SchemaNode of the ExecutionOutputLink
 ///
 /// Expects "gsn" to be a GroundedSchemaNode or a DefinedSchemaNode
-/// Expects "args" to be a ListLink
-/// Executes the GroundedSchemaNode, supplying the args as argument
+/// Expects "cargs" to be a ListLink unless there is only one argument
+/// Executes the GroundedSchemaNode, supplying cargs as arguments
 ///
 Handle ExecutionOutputLink::do_execute(AtomSpace* as,
                                        const Handle& gsn,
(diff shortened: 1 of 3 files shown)


opencog/atoms/execution/ExecutionOutputLink.cc
```diff
@@ -105,14 +101,19 @@ ExecutionOutputLink::ExecutionOutputLink(const Link& l)
 ///
 Handle ExecutionOutputLink::execute(AtomSpace* as, bool silent) const
 {
+	if (_outgoing[0]->getType() != GROUNDED_SCHEMA_NODE) {
+		LAZY_LOG_FINE << "Not a grounded schema. Do not execute it";
```
linas: ?? I'm pretty sure you need to execute the lambda's and the defined schemas too.  Right?  Why wouldn't you want to do this?

opencog/atoms/pattern/PatternLink.cc
```diff
@@ -642,16 +642,25 @@ void PatternLink::unbundle_virtual(const OrderedHandleSet& vars,
...
-			// can be evaluated on the spot.
-			if (2 <= num_unquoted_in_tree(sh, vars))
+			// There is an exception with ExecutionOutputLink that
+			// needs to have a grounded schema node in order to be
+			// executable. If they have non grounded schema node then
+			// their execution is themselves (i.e. they are not
+			// executable).
+			if (sh->getType() != EXECUTION_OUTPUT_LINK or
+			    sh->getOutgoingAtom(0)->getType() == GROUNDED_SCHEMA_NODE)
 			{
```
linas: The logic here seems to be backward. Any function link is executable, as is any EO links that has a grounded or defined schema or a lambda link in it.

linas:
From what I can tell, this is not "relaxing" the checks at all; it's making the checks more stringent. You are allowing fewer things to be executable than before. Why do you need to tighten this?

opencog/atoms/execution/ExecutionOutputLink.cc
ngeiswei: I think it's because "unfolding" defined schemata occurs earlier in the process, that's the idea I was left with after reviewing `do_execute`. Since all unit tests pass I assumed it was OK.

opencog/atoms/pattern/PatternLink.cc
ngeiswei: Do you mean

1. that the code shouldn't dismissing defined schema and the like?
2. or that any EO *shouldn't* accept `SchemaNode` because they are not executable by nature?

if it's 1. then I'm as surprised as you, but since all unit tests pass I have to assume that define schemata are taken care of earlier. For 2. I think that they are, it's just that executing them can be more subtle (they might not be fully determined, etc).

opencog/atoms/execution/ExecutionOutputLink.cc
linas: Oh right. OK.  In some future, it would make more sense to do "lazy evaluation" and unfold, only if the execution link was actually hit. But that's for later.

ngeiswei:
I mean that it allows `ExecutionOutputLink` to hold non grounded schema nodes, in that sense it relaxes it's domain.

The reason I need that is because in SUMO (http://www.adampease.org/OP/ an large ontology I'm using to experiment with the URE) schemata are used in such a way in the knowledge base, such as

```
(=>
   (contraryAttribute @ROW)
   (=>
      (inList ?ELEMENT (ListFn @ROW))
      (instance ?ELEMENT Attribute)))
```

in SUMO `ListFn` is a function, a schema in atomese. The behavior of that function is described in the SUMO KB, as axioms. Thus I think it is appropriate to translate that as `SchemaNode`, as opposed to `GroundedSchemaNode`, or even `DefinedSchemaNode`, then use the URE to perform the "execution". At that point I'm not experimenting with that, but I want to be able to import the whole SUMO KB to stress test the URE.

opencog/atoms/pattern/PatternLink.cc
linas: never mind, I misread the if-statement.

linas:
OK.  I thought I spotted two errors in the code, but in fact, I just misread it. :-) Looks good to me, then.

PR #1207, opened 2017-04-22
Add scheme utility for working with protoatom/values

linas:
The new `cog-value->list` function converts an opencog value to a scheme list.

This commit also remove some obsolete functions.


linas:
This was outstanding work noted in bug #531 and discussed in pull reqs  #1154 #1147

PR #1213, opened 2017-04-25
Added new scm function types.

shujingke:


linas:
I strongly object to these changes. These changes will almost surely result in an badly designed scheme API.  Whatever API it is that you are trying to create, it should be using only atom handles, and not int's booleans and strings!  Please redesign your API so that it works with handles.

Anyway, the code is badly indented!

ngeiswei:
@linas I never understood why it should only use Handles. Do you mean that our scheme functions should take say NumberNodes instead of int and bool, etc. If so why? I do see an advantage though, that way can it directly be called as GroundedSchema or GroundedPredicate, is that the reason?

linas:
I'm imagining something that can be invoked from a GroundPredicateNode  How does one invoke it, if the API doesn't use handles?

Perhaps it is not yet time for this; its not yet ready for this. But as a general design philosophy, I'd like to have everything work with atoms (although I clearly break this philosophy on a daily basis -- it depends on what the actual subsystem is, what it does).

Issue #1215, opened 2017-04-27
"make test" fails under Ubuntu 16.04


cambazz:
Hello,

I have followed the instructions and managed to compile atomspace, but when I issued make test command I got the following error:

`[ 30%] Building CXX object tests/truthvalue/CMakeFiles/EvidenceCountTruthValueUTest.dir/EvidenceCountTruthValueUTest.cpp.o
In file included from /usr/include/cxxtest/RealDescriptions.h:20:0,
                 from /usr/include/cxxtest/TestRunner.h:22,
                 from /home/can/atomspace/build/tests/truthvalue/EvidenceCountTruthValueUTest.cpp:10:
/home/can/atomspace/build/tests/truthvalue/../../../tests/truthvalue/EvidenceCountTruthValueUTest.cxxtest: In member function ‘void EvidenceCountTruthValueUTest::test_mean()’:
/home/can/atomspace/build/tests/truthvalue/../../../tests/truthvalue/EvidenceCountTruthValueUTest.cxxtest:82:5: error: ‘isnan’ was not declared in this scope
     TS_ASSERT_IS_NAN(ectvs[i]->getMean());
     ^
/home/can/atomspace/build/tests/truthvalue/../../../tests/truthvalue/EvidenceCountTruthValueUTest.cxxtest:82:5: note: suggested alternative:
In file included from /usr/include/c++/5/complex:44:0,
                 from /usr/include/cxxtest/StdHeaders.h:24,
                 from /usr/include/cxxtest/StdValueTraits.h:22,
                 from /usr/include/cxxtest/TestSuite.h:25,
                 from /usr/include/cxxtest/RealDescriptions.h:20,
                 from /usr/include/cxxtest/TestRunner.h:22,
                 from /home/can/atomspace/build/tests/truthvalue/EvidenceCountTruthValueUTest.cpp:10:
/usr/include/c++/5/cmath:641:5: note:   ‘std::isnan’
     isnan(_Tp __x)
     ^
tests/truthvalue/CMakeFiles/EvidenceCountTruthValueUTest.dir/build.make:66: recipe for target 'tests/truthvalue/CMakeFiles/EvidenceCountTruthValueUTest.dir/EvidenceCountTruthValueUTest.cpp.o' failed
make[3]: *** [tests/truthvalue/CMakeFiles/EvidenceCountTruthValueUTest.dir/EvidenceCountTruthValueUTest.cpp.o] Error 1
[...]

PR #1218, opened 2017-05-03
Workaround for isnan namespace problem (#1215)

jswiergo:


linas:
well, except now the unit test doesn't compile on old versions of ubuntu. See the cricle-ci status above.

A better fix is to replace `TS_ASSERT_IS_NAN()` by `TS_ASSERT(std::isnan())`

linas:
never mind, I pushed this in 236c0ab817a20794240c3ebf5c27046be30853f8

Issue #1226, opened 2017-05-10
Bug in multiatomspace managing alpha equivalent links


ngeiswei:
See unit test https://github.com/opencog/atomspace/blob/master/tests/atomspace/MultiSpaceUTest.cxxtest#L470

@linas I think the right thing to do is to have the alpha equivalent scope in the parent atomspace being returned when adding the scope in the child atomspace (which is what seems to happen for non scope atoms). Please correct me if I've got the wrong idea. Thanks.

ngeiswei:
The problem comes from `AtomTable::getLinkHandle` that fails to identify that the scope is equivalent to the one in the parent atomspace. `getLinkHandle` recurses till it stumble on `Y` which is not in the parent atomspace, thus failing to fetch the whole scope.

I'm not sure yet what is the right way to fix that but I'm pretty sure that's the problem.

linas:
The child atomspace is supposed to be as it has all of the atoms of the parent atomspace, and also some more.  This mostly makes sense, but there are some pathological cases, that seem hard to avoid, and I don't know how to solve them.

Say you add (Variable $X) to the child space, and then later on, add the same one to the parent.  The one in the child now hides the one in the parent, leading possibly to unexpected behavior ... and this is what your test case does:
```
		Handle as_A = as.add_node(CONCEPT_NODE, "A");
		Handle as_X = ex.add_node(VARIABLE_NODE, "X");
```
above creates (Variable X) in the child. The below creates a (variable X) in the parent, that is not the same as the one in the child: you now have two different atoms with the same type and name.  After that, its just confusing...
```
Handle as_XA = as.add_link(MEMBER_LINK, as_X, as_A);
```

linas:
anyway, its not clear to me if the above is a unit-test bug, or if you meant to do that.

ngeiswei:
I meant
```
Handle as_X = as.add_node(VARIABLE_NODE, "X");
```
this has been corrected in the fix proposal #1230

PR #1229, opened 2017-05-10
Fix alpha equivalence multispace

ngeiswei:
Possible fix for #1226 by excluding unquoted scope link from resolution, that way it manages to reach the _atom_store look up a few lines below and find the alpha equivalent scope in the parent space.

@linas I suppose that rather than having a negative conditional to exclude cases from resolution, we should have a positive conditional with all cases that require resolution (like NumberNode as the comment suggest). Anyway, it fixes the problem and doesn't break any unit test so I assume it's good enough for now but I let it on you to decide whether it should be merged or not.

### DIFF
--- a/opencog/atomspace/AtomTable.cc
+++ b/opencog/atomspace/AtomTable.cc
@@ -279,17 +279,20 @@ Handle AtomTable::getLinkHandle(const AtomPtr& orig, Quotation quotation) const
     quotation.update(t);
 
     // Make sure all the atoms in the outgoing set are in a valid
-    // format. One of the troublemakers here is the NumberNode,
-    // which will hash incorrectly, unless its in proper format.
-    // The other troublemaker is any ScopeLink.
-    HandleSeq resolved_seq;
-    for (const Handle& ho : seq) {
-        Handle rh(getHandle(ho, quotation));
-        if (not rh) return Handle::UNDEFINED;
-        resolved_seq.emplace_back(rh);
-    }
+    // format. One of the troublemakers here is the NumberNode, which
+    // will hash incorrectly, unless its in proper format. We exclude
+    // unquoted scope links from it, otherwise it will prematurely
+    // abort and possibly miss alpha equivalent atom in _atom_store.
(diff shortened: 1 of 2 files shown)


linas:
This fix does not seem right.  When placing a link in an atomspace, all of the atoms in the link *MUST* be in that atomspace, or in one of its parents.  That's the whole point of trying to resolve the handles. By ifdefing it out, you are allowing links to be added that are not in any atomspace at all.

This might be due to the way you wrote the test case.  Try this instead:
```
Handle A = createNode(CONCEPT_NODE, "A");
Handle X = createNode(VARIABLE_NODE, "X");
 Handle XA = createLink(MEMBER_LINK, X, A);
 Handle S = createLink(SCOPE_LINK, XA);
// so far, the scope link is not in any atomspace...
Handle as_S = as.addAtom(S);  // now add to as
Handle ex_S = ex.addAtom(S)  // now add to ex
// Now add XA .. you can do this earlier if you wish.
Handle as_XA = as.addAtom(XA)
```

If you do this, with your changes, then, from what I can tell, you'll have a handle S that is not in any atomspace (and that's just fine), you'll have `as_S==ex_S` point to the same thing, which will be a copy of S, and the outgoing set of `as_S` will not be in any atomspace at all .. just dangling in mid-air.

so for example `as_S->outgoing[0] == XA`  and  `as_S->outgoing[0] != as_XA`  which is bad, because `XA != as_XA`.

ngeiswei:
@linas I tried but it doesn't behave the way you describe, as soon as
```
Handle as_S = as.add_atom(S);  // now add to as
```
is executed all outgoings are inserted to the atomspace. That is because the code in charge of doing that is executed before resolution, see https://github.com/opencog/atomspace/blob/master/opencog/atomspace/AtomTable.cc#L440 

The resolution code is only here apparently to facilitate the _atom_store look-up.

linas:
Oh, OK. I misread where the patch was being applied, I thought it was in the main `add()`, and not the side.

PR #1231, opened 2017-05-11
Add method to clear database stats

linas:
This will not compile until cogutils is also updated!


linas:
I merged this, but all sers need to update cogutils before you pull this!

PR #1241, opened 2017-05-27
Copy over the correlation-matrix code over from the opencog repo.

linas:
As mentioned in pull req https://github.com/opencog/opencog/pull/2743 I've managed to implement a generic correlation-matrix API on top of the atomspace. Basically, by specifying a few basic attoms types and value-getters, you can specify the location of a correlation matrix in the atomspace.  The prototypical example consists of word-pairs e.g
```
    EvaluationLink   (count=6789)
        PredicateNode "word-pair"
        ListLink
            WordNode "foo"
            WordNode "bar"
```
which can be though of as a matrix element N(x,y) where x is the atom "foo", y is the atom "bar" and the value of N is 6789.   You can designate pretty much anything to be the matrix, not just word pairs. The rows and columns do not even have to be the same type!  Anyway, once done, you can then do assorted statistical analysis. 

Since this provides a generic matrix-shaped "window" into the atomspace, I figured the code belonged here.  See the README for more info.


AmeBel:
First woow, NICE!!! Reading this object-api was educational. On Goops, it seems define-generic was what you are looking for ??? https://www.gnu.org/software/guile/manual/guile.html#Generic-Function-and-Method-Examples

PR #1242, opened 2017-05-29
Control the number of matches being made for a GlobNode in pattern matching

leungmanhin:
This can be done by using `IntervalLink` e.g.
```
(TypedVariable
  (Glob "$star")
  (IntervalLink (Number 0) (Number 1))
```
We can also have type restriction as well by using `TypeSetLink` e.g.
```
(TypedVariable
  (Glob "$star")
  (TypeSetLink
    (IntervalLink (Number 0) (Number -1))
    (Type "ConceptNode")))
```

I have also updated the Pattern Matcher to adapt the changes, seems to work for various simple cases.

### DIFF
--- a/examples/pattern-matcher/glob.scm
+++ b/examples/pattern-matcher/glob.scm
@@ -88,3 +88,102 @@
 
 ; This will find the match (Number 42); it will NOT match (Concept "you")
 ; (cog-execute! love-type-glob)
+
+; -----------------------------------------------------------------
+; Globs can have interval restriction
+
+(define love-interval-glob
+    (BindLink
+        (TypedVariable (Glob "$star") (IntervalLink (Number 0) (Number 1)))
+        (ListLink
+            (Concept "I")
+            (Concept "love")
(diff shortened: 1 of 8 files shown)


leungmanhin:
@linas and/or @ngeiswei please review if you have time

ngeiswei:
Wow! That looks fine to me but I think it's good if @linas takes a look at it as he is the author of the modified code.

bgoertzel:
hmm, but @linas seems heads-down focused on language learning work right now, and we may need this code approved/merged to proceed with Ghost (chatbot authoring) work... maybe you should merge it now if it looks OK and let Linas check it over later...

leungmanhin:
Thanks!
Indeed part of these changes are required for Ghost, and these changes will/should not affect the usual pattern matching that uses only VariableNodes, so I guess it's harmless and is ok to merge it now

ngeiswei:
Sure. Great to see more people willing to go down the pattern matcher rabbit hole, well done!

PR #1245, opened 2017-05-31
ECAN changes

misgeatgit:
This PR changes the definition of the Attentional focus as a set of top K STI valued atoms at a certain instance of time instead of its current definition as a set of atoms above some arbitrary STI value aka attentional focus boundary. The need for these changes has been discussed F2F with @bgoertzel and @mattew Ikle, the main motivation being reducing the ramifications of dynamically updating the focus boundary.


rTreutlein:
I haven't kept up to date with that stuff but just out of curiosity is that K going to be fixed or will it change over time?

misgeatgit:
@rTreutlein currently it is a fixed value but we can make it runtime adjustable.

misgeatgit:
@rTreutlein I have created c++ and scheme functions for setting the size at runtime.

PR #1246, opened 2017-06-01
Update the recognizer to read scoped glob-restrictions

leungmanhin:
So take a simple example, if we have:
```
(Satisfaction
  (TypedVariable
    (Glob "$x")
    (Type "PredicateNode"))
  (List (Concept "A") (Glob "$x") (Concept "C")))

(Satisfaction
  (TypedVariable
    (Glob "$y")
    (Type "ConceptNode"))
  (List (Concept "A") (Glob "$y") (Concept "C")))
```

and if we do:
```
(cog-execute! (Dual (List (Concept "A") (Concept "B") (Concept "C"))))
```

It will get only the second one as it satisfies the type restriction.

### DIFF
--- a/opencog/query/Recognizer.cc
+++ b/opencog/query/Recognizer.cc
@@ -57,6 +57,7 @@ class Recognizer :
    public virtual DefaultPatternMatchCB
 {
 	private:
+		bool bypass_node_match = false;
 		std::vector<Variables> _soln_vars;
 		void get_glob_decl(const Handle&);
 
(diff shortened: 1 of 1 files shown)


linas:
I don't quite get it. This seems like a very complex solution. All you really had to do was to check if the type of the atom, and the type of the glob agreed, and this would have been a one or two-line check, in the original code, no?

linas:
pull req #1242 was the first part to this.

leungmanhin:
Yes I understand, the check can be done in 1-2 lines, the extra complexity added here is to try to deal with a potential issue related to scoping. For example if the query is this again:
```
(cog-execute! (Dual (List (Concept "A") (Concept "B") (Concept "C"))))
```
and at some point we find this as a potential solution:
```
(List (Concept "A") (Glob "$x") (Concept "C"))
```
The recognizer then try to see if it is under any scope, and it finds these two:
```
(Satisfaction
  (TypedVariable
    (Glob "$x")
    (Type "PredicateNode"))
  (List (Concept "A") (Glob "$x") (Concept "C")))

(Satisfaction
  (TypedVariable
    (Glob "$x")
    (Type "ConceptNode"))
  (List (Concept "A") (Glob "$x") (Concept "C")))
```
Note both of them have the same `Glob $x`. In principle, the `Glob "$x"` should be treated as two different GlobNodes as they are bounded in two different scopes, and the recognizer should reject the first one and accept the second one because of the type restriction. On the other hand they happen to have a glob with the same name, should the recognizer returns `(List (Concept "A") (Glob "$x") (Concept "C"))` as the solution in this case? Or do something else?
Amen and I had a discussion, and we think maybe the recognizer should return the whole SatisfactionLink for patterns that are scoped (not implemented yet), what do you think?

linas:
Ah. Hmm. The recognizer was never really meant to do that -- it was never supposed to provide a perfect inversion of a problem, and it was not supposed to do scoping, and it was not supposed to check for types, either. That would add a lot of complexity.  I'm not sure of what the point of that would be.

The point of the recognizer was to provide a short list of possible rules; you would still have to run the rules to see if they actually apply to the given situation.  It was not meant to be a perfect inverter that can guarantee that what it finds will always work. That kind of a guarantee is hard to provide, and providing it would end up being just as complicated as actually running the rule. 

So my gut instinct is to say that the type-checking should be removed from the recognizer; its the wrong place to perform type-checking, or any kind of forward-processing. Its really meant to do reverse-processing only, that's why its "dual".

If you want to do a side-effect free reverse followed by forward processing, you could use the old, unenhanced DualLink to find rule candidates, then run them in the forward direction, in a temporary atomspace, see which ones succeed, and then return only those.

Without thinking too hard about it, my gut instinct is that type-checking in the Dual link is fundamentally wrong: it adds a forward direction to the reversed arrow: that arrow was supposed to point backwards, and if you now try to make it also point forwards, you (a) limit it's usefulness (b) force a lot of complexity into the system.  I mean, its known that isomorphisms are just more complex than homomorphisms, precisely because you have to perform the quotienting, and keep only the remainders (the cosets)  There's all sorts of well-known pathologies and craziness that happens when you do that; wikipedia has a horse-stable of articles on the resulting paradoxes. So don't do that.  Keep the reversed arrow pointing backwards, and use only the forward arrow to go forwards.  Don't try to do both at the same time.

Or rather: let me put it this way: you seem to want a combined BindLink+DualLink.  So, invent a new IsoLink. which works like the old DualLink, but then adds the new typechecking, and returns the full, scoped match.  Keep the old DualLink simple, lightweight.

leungmanhin:
ok I see your point, I will revert these changes in the recognizer.

There is one tricky bit, right now by default it's assumed that a GlobNode is "one-to-many", but there are also rules containing GlobNodes that should be considered as "zero-to-many". For example if we do:
```
(cog-execute! (Dual (List (Concept "A"))))
```
The old DualLink won't be able to find this:
```
(List (Glob "$x") (Concept "A") (Glob "$y"))
```
where `(Glob "$x")` and `(Glob "$y")` are meant to be "zero-to-many".

So do you think it's ok to change the default GlobNode to "zero-to-many"?

PR #1257, opened 2017-06-06
Get the GlobNode interval restrictions when constructing the PatternLink

leungmanhin:
It ignored the interval restrictions for `GlobNodes` in a `BindLink`/`SatisfactionLink` if there were more than one `TypedVariableLink` in the `VariableList`. This should fix the problem.

### DIFF
--- a/opencog/atoms/pattern/PatternLink.cc
+++ b/opencog/atoms/pattern/PatternLink.cc
@@ -124,7 +124,8 @@ void PatternLink::setup_components(void)
 	for (size_t i = 0; i < _num_comps; i++)
 	{
 		Handle h(createPatternLink(_component_vars[i], _varlist._simple_typemap,
-		                           _components[i], _pat.optionals));
+		                           _varlist._glob_intervalmap, _components[i],
+		                           _pat.optionals));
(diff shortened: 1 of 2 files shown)


leungmanhin:
@linas I think it's ok to merge, but it would be better if you can have a look at this first

linas:
The GlobIntervalMap should be a pair of ints or shorts, not a pair of doubles.  The problem with doubles is that sometimes, rarely they might get set to values like 3.99999 which get truncated to 3 when you are counting how many matches there are, because 3.999 < 4.  So, in general, its safer to use ints.  

actually unsigned ints, since values less than zero don't make sense.

leungmanhin:
Or round the number before doing the comparison? Because the value of a NumberNode is a double.
On the other hand I can also change them to ints, the sign is needed here since there is a use case where -1 in the upper bound means "as many as possible".

linas:
round them here, before storing.That way you can avoid rounding later.   Also if you make them unsigned, then -1 is the same as max-unsigned-value, because of two's complement.

leungmanhin:
Oh right, ok I'll update it soon

PR #1261, opened 2017-06-07
Incoming Set - second attempt

linas:
The first "fix" in pull req #1259 actually made things worse. So here's a second attempt.


linas:
never mind, this doesn't work either.

PR #1263, opened 2017-06-08
Update the recognizer to treat GlobNodes as zero-to-many by default

leungmanhin:
Changing from "one to many" to "zero to many", so a larger set of rules can be found

### DIFF
--- a/opencog/query/Recognizer.cc
+++ b/opencog/query/Recognizer.cc
@@ -195,9 +195,13 @@ bool Recognizer::link_match(const PatternTermPtr& ptm, const Handle& lsoln)
 	// match by grounding $x to nothing and $y to Concept
 	// "B" and "C". But a side-by-side comparison here
 	// only compares their nodes at the same position
-	// (i.e. A-$x, B-A, C-$y) once, and decide whether
-	// or not to accept it. As a result it skips a lot
-	// of candidates that we are expecting...
+	// (i.e. A-$x, B-A, C-$y), and decide whether to
+	// reject it when there is a mis-match. As a result
+	// a lot of candidates that we are expecting are
+	// rejected...
+	// And the reason of going to fuzzy_match is that
+	// all the glob-matching logic is there, so it
+	// should be able to handle this better.
 	if (contains_atomtype(lsoln, GLOB_NODE) and
 	    lpat->getArity() == lsoln->getArity())
 	{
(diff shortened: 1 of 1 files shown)


leungmanhin:
@linas This is needed in Ghost (chatlang) for finding the rules, I would like to merge it now in order to proceed, please let me know if you see anything wrong, thanks.

opencog/query/Recognizer.cc
```diff
@@ -166,10 +174,45 @@ bool Recognizer::link_match(const PatternTermPtr& ptm, const Handle& lsoln)
...
+	// === lpat ===      === lsoln ===
+	// Concept "A"       Glob $x
+	// Concept "B"       Concept "A"
+	// Concept "C"       Glob $y
+	// and both of the globs $x and $y have an interval
+	// restriction of zero to infinity, it should be a
+	// match by grounding $x to nothing and $y to Concept
+	// "B" and "C". But a side-by-side comparison here
+	// only compares their nodes at the same position
+	// (i.e. A-$x, B-A, C-$y), and decide whether to
```
linas: well, but the side-by side code has all this logic for advancing pointers to the next item to compare.  It sounds to me that it needs to be updated, not that zero matches are allowed.

opencog/query/Recognizer.cc
```diff
@@ -166,10 +174,45 @@ bool Recognizer::link_match(const PatternTermPtr& ptm, const Handle& lsoln)
...
+	// reject it when there is a mis-match. As a result
+	// a lot of candidates that we are expecting are
+	// rejected...
+	// And the reason of going to fuzzy_match is that
+	// all the glob-matching logic is there, so it
+	// should be able to handle this better.
+	if (contains_atomtype(lsoln, GLOB_NODE) and
+	    lpat->getArity() == lsoln->getArity())
+	{
+		if (fuzzy_match(lpat, lsoln))
```
linas: what does fuzzy matching got to do with it? why do you want a fuzzy match here ?  I'm confused ... why not do strict matching?

linas:
looks good to me. I'm confused about why fuzzy matching is needed.

opencog/query/Recognizer.cc
linas: oh, I see, this is a different fuzzy match than the fuzzy matcher.  Since we already have a fuzzy matcher that does something completely different, could you change the name of this to something else? Maybe approx_match or imprecise_match or some other synonym?

leungmanhin: Sorry for the confusion, the "fuzzy_match" here is actually one of the pattern matcher callbacks that we have been using for the glob-matching, I am only reusing it for comparing lpat and lsoln with the same arity as well, I'll update the comment to clarity a bit more.

linas: oh, right. I forgot. Misgana originally added this callback.  But then later it turned out the fuzzy matcher didn't actually need the pattern matcher. However, the callback was useful, so it stayed.  I guess its OK to keep that callback as is.  Maybe someday in the future I'll change its name when I think of a better name.

PR #1266, opened 2017-06-08
Scheme utility for copying atoms into independent atomspace

AmeBel:
This also works for copying from child to ancestor or  ancestor to child.

### DIFF
--- a/opencog/scm/opencog/base/utilities.scm
+++ b/opencog/scm/opencog/base/utilities.scm
@@ -1202,6 +1202,26 @@
 )
 
 ; -----------------------------------------------------------------------
+(define-public (cog-cp LST AS)
+"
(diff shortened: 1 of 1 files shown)


opencog/scm/opencog/base/utilities.scm
```diff
@@ -1202,6 +1202,26 @@
...
+
+  (if (equal? AS initial-as)
+    (error "Destination atomspace is the same as the current atomspace\n"))
+
+  ; Switch to destination atomspace.
+  (cog-set-atomspace! AS)
+
+  ; The creation of a SetLink or any other link would result in the atoms
+  ; being inserted in the current atomspace.
+  (Set LST)
```
linas: well, you probably want to remove this set, after creating it,  This can be done by saying
```
(cog-delete! (Set LST))
```
which will still cause the Set to be created, and the atoms copied, but then deletes the Set and only the set.

PR #1277, opened 2017-06-15
Simplify SchemePrimitive using variadic templates

ngeiswei:


ngeiswei:
Should fix that https://github.com/opencog/opencog/issues/2772

ruiting:
It seems it has caused another error

/home/ruiting/hansonrobotics/opencog/opencog/opencog/nlp/fuzzy/FuzzySCM.cc:139:1:
  required from here
/usr/local/include/opencog/guile/SchemePrimitive.h:183:16: error: no
matching function for call to ‘opencog::SchemePrimitiveBase<opencog::Handle,
opencog::nlp::FuzzySCM, opencog::Handle, short unsigned int, const
std::vector<opencog::Handle, std::allocator<opencog::Handle> >&,
bool>::scm_to(scm_unused_struct*&, long unsigned int,
std::__tuple_element_t<2ul, std::tuple<opencog::Handle, short unsigned int,
const std::vector<opencog::Handle, std::allocator<opencog::Handle> >, bool>
>&)’
   return scm_to(args, I, std::get<I>(t));
make[2]: *** [opencog/nlp/fuzzy/CMakeFiles/nlpfz.dir/FuzzySCM.cc.o] Error 1



Ruiting Lian

ngeiswei:
Should work now.

linas:
Nil,  I guess this is a nice change but the core pproblem is that we have huge numbers of users who are wildly abusing the API. Soe of the primary abusers are the dim-embed module, and shujinke patten miner.  I asked her to clean this up, and I'm very concerned that your changes remove the pressure for doing so.

The real point being that variadic would not even be needed if people did not feel entitlled to totally abuse the API.  You've kind of opened the gates, and made it eve easier to abuse, when in fact we should be making it more painful, more punishing for the abuse

 To be clear: I want to make shujing's life painful enough that eventually she fixes all of the bugs and bauses, instead of continuing to ignore them.  So this change is kind of a move in the wrong direction.

opencog/guile/SchemePrimitive.h
```diff
@@ -30,640 +78,231 @@ class PrimitiveEnviron
...
-// V_TI   -- DimEmbedModule::embedAtomSpace()
-// V_TIDI -- DimEmbedModule::addKMeansClusters
-//
-// This API needs to be re-thought, from scratch. Its offensive.
-// S_AS   -- PatternMiner
-// S_B    -- PatternMiner
-// S_I    -- PatternMiner
-// S_V    -- PatternMiner
-// V_SI   -- PatternMiner
-
```
linas: This list should be restored. It lists all the big users and abusers of the API: its the master "naughty and nice" list, and it took a LOT of work to asemble.  I don't want to loose this vital info.

opencog/guile/SchemePrimitive.h
```diff
@@ -30,640 +78,231 @@ class PrimitiveEnviron
...
+	int scm_to(SCM args, size_t idx, int) const
+	{
+		SCM arg = scm_list_ref(args, scm_from_size_t(idx));
+		return SchemeSmob::verify_int(arg, scheme_name, idx);
+	}
+	double scm_to(SCM args, size_t idx, double) const
+	{
+		SCM arg = scm_list_ref(args, scm_from_size_t(idx));
+		return SchemeSmob::verify_real(arg, scheme_name, idx);
+	}
```
linas: thi9s really needs to be removed. No one should be passing doubles.  That';s just fundamentally wrong.

opencog/guile/SchemePrimitive.h
```diff
@@ -30,640 +78,231 @@ class PrimitiveEnviron
...
+	size_t scm_to(SCM args, size_t idx, size_t) const
+	{
+		SCM arg = scm_list_ref(args, scm_from_size_t(idx));
+		return SchemeSmob::verify_size(arg, scheme_name, idx);
+	}
+	int scm_to(SCM args, size_t idx, int) const
+	{
+		SCM arg = scm_list_ref(args, scm_from_size_t(idx));
+		return SchemeSmob::verify_int(arg, scheme_name, idx);
+	}
```
linas: these three should be removed, no one should be passing bools, ints or size_t types through the API, that's just fundamentally wrong.

opencog/guile/SchemePrimitive.h
```diff
@@ -30,640 +78,231 @@ class PrimitiveEnviron
...
+	Handle scm_to(SCM args, size_t idx, const Handle&) const
+	{
+		SCM arg = scm_list_ref(args, scm_from_size_t(idx));
+		return SchemeSmob::verify_handle(arg, scheme_name, idx);
+	}
+	AtomSpace* scm_to(SCM args, size_t idx, const AtomSpace*) const
+	{
+		SCM arg = scm_list_ref(args, scm_from_size_t(idx));
+		return SchemeSmob::verify_atomspace(arg, scheme_name, idx);
```
linas: this should probably be removed, no one should be passing atomspaces.  Again, another abuse of the API.

opencog/guile/SchemePrimitive.h
```diff
@@ -30,640 +78,231 @@ class PrimitiveEnviron
...
+	std::string scm_to(SCM args, size_t idx, const std::string&) const
+	{
+		SCM arg = scm_list_ref(args, scm_from_size_t(idx));
+		return SchemeSmob::verify_string(arg, scheme_name, idx);
+	}
+	Type scm_to(SCM args, size_t idx, Type) const
+	{
+		SCM arg = scm_list_ref(args, scm_from_size_t(idx));
+		return SchemeSmob::verify_atom_type(arg, scheme_name, idx);
+	}
```
linas: this too is wrong, no one should be passing types.  This is an abuse of the API.

linas:
the important task here is to go through the "naugty and nice" list, and fix all of the broken API's that we have.  Pretty much everything should be using handles and only handles, and never, ever passing ints, bools, doubles, strings.   The fact that things got so bad is due to poor management: we let too many people commit code, without any kind of review or approval process.

ngeiswei:
I really don't get why we should be passing only handles.

opencog/guile/SchemePrimitive.h
ngeiswei: Why?

ngeiswei:
@linas, over the dozen of modules using this API, exactly only one doesn't abuse it, `InferenceSCM`. All the others do, even `PatternSCM` (it has functions returning `bool`). But why is that bad?

I don't think the Python or Haskell bindings have made that commitment either.

linas:
Err ... but the API is fundamental to the opencog, !?   How can you store a bool, int, double in the atomspace? ... you can't!   I mean, you can create a ConceptNode "True"  or NumberNode 42 but that's it.  You can use atom values to store int's or floats or strings  (or poke them into TV's or into AV's) but that's it, there are no other options at all.  The only API we have are atoms, TV's, AV's values. Anything that does not use this API is just-plain broken. No other way to explain it.

ngeiswei:
But not everything that you want to interact with in opencog needs to be stored in the atomspace. Typically the logger falls in that category. So I don't think the API should force the developer to use solely Handles. Instead the function of the module itself should determine how the scheme bindings should be.

For instance the URE bindings have no abuse but that is because reflexivity, so that opencog/AGIprime can eventually have full control over it, has been thought from its conception. However the Logger bindings have many abuses but it's not gonna hurt in the mid, probably not even in the long term, because here reflexivity doesn't matter much.

As for the pattern miner, I don't know, but if there are abuses, the pressure to fix them will come from the usages, less so from the API (as the numerous abuses have already shown). You see what I mean?

linas:
Yes, there are exceptions. I'd like to minimize the exceptions, and one way to do that is to make it harder tor the exceptional branch to be taken.

PR #1278, opened 2017-06-15
Reduce contention for multi-threaded AtomTable adds

inflector:
While working on code for ObserveText in C++, I found that multi-threading actually decreased the speed by about 30%. Most of the threads were blocking on mutex locks in the code for adding atoms to the atom table.

I noticed a couple of potential improvements. The changes here are an implementation of those improvements:

1) I reduced the contention between atom index adds and reads and the atom inserts by separating the single mutex `_mtx` into separate mutexes for changes to the atom store and the atom indexes:  `_store_mtx` and `_index_mtx`. This way reads of the index don't block adds of the atom except for the brief period where the actual index is added. And vice versa. 

2) I moved the locking code further down in `AtomTable::add`. In particular, it now doesn't lock while cloning the atom. It only locks while actually updating the counts and inserting the atom into the store.

3) I moved the check to see if the atom is already in the atom table higher in the code since we are no longer holding the lock while this check is made the first time. So while before, we wanted to minimize the work done while the lock is held which argued for doing more work first, now, we want to do the check as soon as possible since this covers a very common case, i.e. the atom is already in the table.

4) I added another search to check again to see if the atom is not in the store to avoid the double insert case in the event that another thread just happened to be attempting to insert the atom. This check doesn't add much additional time since we were already hashing in the old code, so I moved the hash above the search and use that same hash for the insert. See: `_atom_store.insert` on old line 507 and new line 523 of AtomTable.cc. Thus, the lookup code for the additional search does not require an additional hash.

5) I moved the unlock code for  _store_mtx above the call to insert the atom in the index since `put_atom_into_index` call already locks the index-specific mutex` _index_mtx` internally. So the `_store_mtx` lock is now released just after the atom is stored into `_atom_store`.

The combination of these changes increased multi-threaded CPU performance in my tests to be about 30% faster on a 6 cpu machine, and it is now slightly faster than for a single-thread test whereas before these changes it was about 30% slower multi-threaded than single-threaded, now it is about 5% to 10% faster.


linas:
OK looking at it.

linas:
Well, UseCountUTest fails after your changes. This is a unit test that adds atoms to the atomspace, and tries to make sure that all adds succeed. Its one of two? or three? multi-threading unit tests for the atomspace.

my guess that the rason its failing is the fist change: by using two different mutexes, you add a window where atom adds can race with one-another, resulting in an atom being added twice.  Future references to that atom get confused, because its no longer clear which copy is which.  That's just a wild guess, based on the symptoms, and your description.

opencog/atomspace/AtomTable.cc
```diff
@@ -498,23 +495,45 @@ Handle AtomTable::add(AtomPtr atom, bool async)
     atom->keep_incoming_set();
     atom->setAtomSpace(_as);
 
+    ContentHash ch = atom->get_hash();
+
+    // Lock, to prevent two different threads from trying to add exactly
+    // the same atom.
+    std::unique_lock<std::recursive_mutex> lck(_store_mtx);
```
linas: well, this lock has to include the state-link check, as otherwise the statelink will get garbled.

I do not beleive we have a multi-threaded safety test for the statelink, just right now.  I'd have to think about that,

opencog/atomspace/AtomTable.cc
```diff
@@ -498,23 +495,45 @@ Handle AtomTable::add(AtomPtr atom, bool async)
     atom->keep_incoming_set();
     atom->setAtomSpace(_as);
```
linas: I think you have to set the atomspace under the lock, as otherwise, if the insert fails, you have this atom that advertises that its in an atomspace, but isn't actually in it.  Right?

opencog/atomspace/AtomTable.cc
```diff
@@ -498,23 +495,45 @@ Handle AtomTable::add(AtomPtr atom, bool async)
...
     lck.unlock();
 
-    // Update the indexes asynchronously
-    if (not _transient and async)
-        _index_queue.enqueue(atom);
+    // Note: we don't need to hold the lock for the
+    // index since it locks an index-specific mutex.
+    if (not _transient)
+        {
+        if (async)
```
linas: FYI, totally unrelated to your changes --  the "async" design here is broken, and that entire code branch should just be removed.

opencog/atomspace/AtomTable.cc
linas: And here's the bug: doing this set: `atom->setAtomSpace(_as);` is advertising that the "atom is now in the atomspace".  The test `Handle hcheck(getHandle(orig));    if (hcheck) return hcheck;` is testing this same location.  This was intended to be an atomic test-n-set, but now that it isn't, callers can get back two different copies of the same atom, both copies claiming to be in the atomspace.  Since they are two different copies, updating truth-values, counts will get scrambled, as the TV's on these two atoms will diverge over time.  All sorts of strange things from there: when its time to save, the one that is not in the atomspace won't get saved.

opencog/atomspace/AtomTable.cc
```diff
@@ -455,13 +459,6 @@ Handle AtomTable::add(AtomPtr atom, bool async)
...
         atom = clone_factory(atom_type, atom);
 
-    // Lock before checking to see if this kind of atom is already in
-    // the atomspace.  Lock, to prevent two different threads from
-    // trying to add exactly the same atom.
-    std::unique_lock<std::recursive_mutex> lck(_mtx);
-    Handle hcheck(getHandle(orig));
-    if (hcheck) return hcheck;
-
     atom->copyValues(Handle(orig));
```
linas: This, the `copyValues`, can be safely moved out from under the lock.  Not sure about the rest.

opencog/atomspace/AtomTable.cc
```diff
@@ -498,23 +495,45 @@ Handle AtomTable::add(AtomPtr atom, bool async)
...
+
+    // Search again, now that we hold the lock. This is necessary for
+    // the case where another atom was just inserted by another
+    // thread while we were cloning and copying values in this thread.
+    auto range = _atom_store.equal_range(ch);
+    auto bkt = range.first;
+    auto end = range.second;
+    for (; bkt != end; bkt++) {
+        if (*((AtomPtr) bkt->second) == *atom) {
+            return bkt->second;
```
linas: well, this return is no longer legal, after you've built the incoming set for the atom.  Once you've built the incoming set, that's it, you can no longer abandon the transaction.  Or rather, if you did abandon, you would have to unbuild the incoming set here, before returning.  But that opens a race condition, if anyone happens to be traversing the incoming set at just that instant: they'll find atoms that are not in the atomspace

The core problem is that e.g. pattern matching, but also many other algos traverse the graph by following the incoming set of an atom. But this return means that there are now atoms in the incoming set, that are not in the atomsapce ... so the graph walk will walk out of the atomspace and start following junk -- viz atoms that are copies, and thus have bad TV's on them.  This will break the rule engine, which is driven by TV's.

linas:
OK, so overall, I see multiple issues.  The problem is that two of them are not tested by any unit test at all.  So maybe the right thing to do would be to write unit tests for those, first, and then see you can make locking changes that pass the tests.  Specifically:

racey traversals of the incoming set: can one ever traverse the incoming set, and find atoms that are not in the atomspace? This should not be possible, but without mutex protection, this is what your chanes allow.

racey StateLink updates: is it possible that if two threads are racing to update the statelink, that one might win, even though the state is wrong?  This would be a multi-threaded version of the green-light/red-light unit test, where e.g. the two threads are two cars at an intersection: only one should get a green light, and never both.  This is coupled to the incoming set problem; since some of the statelink checks are done with the pattern matcher, which traverses the incoming set.  The risk is that it finds some "other" statelink, which happens to be green, when the "real" statelink, the one in the atomspace, is actually red.

fwiw,  the statelink is used a lot in the robot, to record what the robot sees/hears/is-currently -doing.

linas:
Still have this test failure:  `14 - UseCountUTest (Failed)` which is explicitly testing the code that you are changing here. You also have not addressed the various issues I raised about the design.  Do you really want this to be an open pull request? I don't want it being merged, as long is it still has design flaws, and still doesn't actually work...

linas:
Maybe we could close this pull request, until such time that it's actually ready to be merged?

linas:
Maybe you should put the benchmark code into some different pull request?  It seems to be unrelated...

inflector:
Closing this request for now... too little gain for the potential problem at the moment.

inflector:
I moved the benchmark code to another branch and submitted a separate pull request.

PR #1284, opened 2017-06-19
Scheme bindings support multiple loggers

ngeiswei:
Now one can use multiple loggers. The scheme functions haven't changed, instead they have been completed with an optional first argument representing the logger. If no argument is provided then the default logger is considered. For instance
```
(cog-logger-set-level! (cog-ure-logger) "fine")
```
change the URE logger level to fine, while
```
(cog-logger-set-level! "fine")
```
change the default logger level to debug.

### DIFF
--- a/opencog/guile/CMakeLists.txt
+++ b/opencog/guile/CMakeLists.txt
@@ -15,6 +15,7 @@ ADD_LIBRARY (smob
 	SchemeSmobNew.cc
 	SchemeSmobTV.cc
 	SchemeSmobValue.cc
+	SchemeSmobLogger.cc
 )
 
 ADD_LIBRARY (logger
--- a/opencog/guile/CMakeLists.txt
+++ b/opencog/guile/CMakeLists.txt
@@ -36,6 +36,7 @@ TARGET_LINK_LIBRARIES(logger
 	smob
 	${GUILE_LIBRARIES}
 	${COGUTIL_LIBRARY}
+    ruleengine # for the ure logger
 )
 
 TARGET_LINK_LIBRARIES(randgen
--- a/opencog/guile/LoggerSCM.cc
+++ b/opencog/guile/LoggerSCM.cc
@@ -22,6 +22,7 @@
  */
 
 #include <opencog/util/Logger.h>
+#include <opencog/rule-engine/URELogger.h>
 #include <opencog/guile/SchemeModule.h>
 #include "SchemePrimitive.h"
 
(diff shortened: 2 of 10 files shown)


ngeiswei:
Note, it's still not possible to instantiate a new logger within scheme, but it shouldn't be too hard to add now, let me know when you need that and I'll add it.

linas:
Well, optional arguments tend to come after the the required arguments, at least if you use the optargs mechanism.  Otherwise, you can use named arguments.  See https://www.gnu.org/software/guile/manual/html_node/lambda_002a-and-define_002a.html#lambda_002a-and-define_002a  and https://www.gnu.org/software/guile/manual/html_node/ice_002d9-optargs.html#ice_002d9-optargs

linas:
FWIW those documentation pages are a bit optuse and hard to comprehend.  The actual use is easy; for example:
```
(define*-public (my-func required-arg-a required-arg-b #:optional  (opt-arg-c default-value))
   ... do stuff...)
```
The idea is that you can handle all the optional args in scheme, and then pass only a fixed number of args into the C code, which makes the C code simpler.

opencog/guile/CMakeLists.txt
```diff
@@ -36,6 +37,7 @@ TARGET_LINK_LIBRARIES(logger
 	smob
 	${GUILE_LIBRARIES}
 	${COGUTIL_LIBRARY}
+    ruleengine # for the ure logger
```
linas: Oof. This sure smells like a circular dependency, to me.  I think it would be a lot better if this got loaded in the scheme module, rather than being linked in the C code.

ngeiswei: The reason I need that is to be able to recognize a Logger* type, with COG_LOGGER. I'd be more than happy to move that in the module if there's a way.

ngeiswei:
@linas I agree about optional arguments, the reason the logger argument comes first is because the logging function also incorporate scheme string `format`, which may take a arbitrary number of arguments that the end, so I'm having 2 different type of optional arguments, I thought I'd rather set them apart. Maybe using named arguments would be the way to go, I'll look into it.

opencog/guile/LoggerSCM.cc
```diff
@@ -140,29 +170,42 @@ LoggerSCM::LoggerSCM() : ModuleWrap("opencog logger") {}
 /// Thus, all the definitions below happen in that module.
 void LoggerSCM::init(void)
 {
-	define_scheme_primitive("cog-logger-set-level!",
+	define_scheme_primitive("cog-default-logger",
+		&LoggerSCM::do_default_logger, this, "logger");
+	define_scheme_primitive("cog-ure-logger",
+		&LoggerSCM::do_ure_logger, this, "logger");
```
linas: this doesn't make sense to me.  Why is this in C++? Why not just create something like this:
```
(define my-fav-logger (create-new-logger "/path/to/log/file"))
(cog-logger-set-loglevel my-fav-logger  "fine")
```
 and if you really want to get whizzy, stick a bunch of methods on it:
```
(define my-fav-logger (create-new-logger "/path/to/log/file"))
(my-fav-logger 'set-loglevel "fine")
```
FWIW, I'm very tempted to make the entire scheme API to opencog be object-oriented in this way, viz
```
(define my-atom (ConceptNode "foo"))
(my-atom 'set-tv (stv 0.42 0.69))
```
I don't know why I didn't do this from the very beginning.  What was I thinking?

opencog/guile/CMakeLists.txt
linas: I guess I'm suggesting something like this: editing `opencog.scm` and adding a line like
```
(load-extension "libure" "opencog_ure_init")
```
or something like that. I'd have to dive into the guile documentation to figure this out, precisely.  I'm sort of skimming your code, and I'm guessing that it the only thing you want is the logger, that it would be enough to load the shared library, like this, and you'd get it ... !?  Maybe there's a gotcha. Dunno.

opencog/guile/LoggerSCM.cc
ngeiswei: Regarding doing
```scheme
(define my-fav-logger (create-new-logger "/path/to/log/file"))
(cog-logger-set-loglevel my-fav-logger  "fine")
```
yes, that's the idea, it's just I didn't implement `create-new-logger` yet, and also it doesn't change the fact that I need to be able to reference the existing loggers created on the C++ side. So the plan is to have both.

linas:
OK, I see what you did now. I guess its OK to leave the opt arg as it is.  However, switching to a more OO-sih api would be cool, though.  That's also optional, but its my current infatuation. Like I say, not sure why I didn't do the whole API that way.

linas:
Sorry, Clearly, I'm writing comments first, and thinking about it later. My apologies, some or many of my suggestions are foolish.

PR #1291, opened 2017-06-27
Remove TV merge functions from Atom.h

linas:
This does not remove any capability -- one can still merge TV's directly.  This just removes some the C++ wrappers from class Atom, where they provided almost no value-add, and mostly clogged up the API.


linas:
all tests pass; and the usual tests pass on opencog.

PR #1296, opened 2017-06-29
Make glob matching greedy by default

leungmanhin:


linas:
If you have the time, the energy, the desire ... while your memory of this code is still fresh in you mind, perhaps you could add a flag for greedy vs. lazy matching?  I just now added a section to the wiki: http://wiki.opencog.org/w/GlobNode#Greedy_Matching that describe two differnt ways this might be done. I think I like the second one better.

leungmanhin:
Sure I'll do that as well

linas:
OK, thanks!

linas:
Also, I describe two different proposals on how to do that. I think I like the second one better.

leungmanhin:
Nice, and how about using only one `TypeStyleNode` to indicate what matching style should be used, i.e. change it from:
```
TypedVariableLink
    VariableNode   "$foo"
    TypeSetLink
        TypeStyleNode "greedy"
        TypeStyleNode "glob"
```
to:
```
TypedVariableLink
    VariableNode   "$foo"
    TypeStyleNode "greedy"
```
this assumes that the variable is a glob if there is a `TypeStyleNode`?

btw there are more work needs to be done in chatlang, let me do this update afterwards.

linas:
Yes, I guess that could work.  The only reason to use TypeSetLink is if one needed to specify several constraints at the same time (e.g. length and also type).

I'm guessing that somewhere, deep in the bowels of chatlang, there will probably be a need for lazy matching.  I'm making the suggestion now, because it can sometimes be easiest to write the code while it's still fresh in you mind, instead of trying to remember how it worked weeks or months later.  But really that is up to you; this is an optional feature at the moment.

linas:
Oh, "the second one" meant "using TypeStyleNode instead of using LazyGlobNode".

Eliminating all uses of GlobNode would be hard, and maybe unwise. You probably should not do that, unless you are convinced that its some great idea.

leungmanhin:
I think using VariableNode for all kinds of matching is neat, but this definitely requires more work, I'll go for the TypeStyleNode approach before I forget about the code :)

PR #1316, opened 2017-07-25
Atom prototype benchmark

inflector:
Added an experimental benchmark to prototype and time an alternate paged-based storage for atoms in order to determine the potential speed improvements to be gained by improving locality of reference for atoms by storing them in adjacent memory.

### DIFF
--- a/opencog/benchmark/CMakeLists.txt
+++ b/opencog/benchmark/CMakeLists.txt
@@ -53,3 +53,15 @@ IF (HAVE_GUILE)
 		dl
 	)
 ENDIF (HAVE_GUILE)
+
+ADD_EXECUTABLE (atom_bench
+	PagedAtoms.cc
+	atom_bench.cc
+)
+
+TARGET_LINK_LIBRARIES (atom_bench
+	atombase
+	atomspace
+	${ATOMSPACE_LIBRARIES}
+	${COGUTIL_LIBRARY}
+)
--- a/opencog/benchmark/PagedAtoms.cc
+++ b/opencog/benchmark/PagedAtoms.cc
@@ -89,7 +89,25 @@ void EdgePage::add_edge(AtomHandle in, AtomHandle out)
     edge_ptr->inbound = (uint64_t) in;
     edge_ptr->outbound = (uint64_t) out;
     
-    // Move some bytes to emulate the performance of inserts.
+    // Move some bytes to emulate the performance of inserts. This code is
+    // intended to reflect a move of random bytes. It moves bytes past the
+    // end of the used area so these moves will not affect stored atoms. The
+    // work is also reversed in that full pages have less work while empty pages
+    // have more work, unlike a real page where full pages will have inserts
+    // which move half the entries, on average. In an actual implementation,
+    // these pages will likely be stored using an arracy of atom offsets at
+    // the beginning of the page, with the actual edges/atoms/index entries
+    // stored from the back of the page. See: 
+    //
+    // http://rachbelaid.com/introduction-to-postgres-physical-storage/
+    // 
+    // for how this is done with Postgres. This approach means that only the 2-byte
+    // offsets are moved when inserting new rows/edges/atoms/index entries. The actual
+    // rows are kept in place and only the index entries at the start of the page are
+    // kept in sorted order. This will result in less work than this emulation, but
+    // there will be more work binary searching the sorted rows to find the new
+    // insert positions. So the net effect will be that the actual code will 
+    // take more time than this implementation.
     if (free > 2)
     {
         size_t slots_to_move = randomGenerator->randint(MAX_EDGES_PER_PAGE - next);
@@ -141,7 +159,25 @@ void IndexPage::add_page(uint64_t key, void* page)
     slot_ptr->key = (uint64_t) key;
     slot_ptr->page = (void*) page;
     
-    // Move some bytes to emulate the performance of inserts.
+    // Move some bytes to emulate the performance of inserts. This code is
+    // intended to reflect a move of random bytes. It moves bytes past the
+    // end of the used area so these moves will not affect stored atoms. The
+    // work is also reversed in that full pages have less work while empty pages
+    // have more work, unlike a real page where full pages will have inserts
+    // which move half the entries, on average. In an actual implementation,
+    // these pages will likely be stored using an arracy of atom offsets at
+    // the beginning of the page, with the actual edges/atoms/index entries
+    // stored from the back of the page. See: 
+    //
+    // http://rachbelaid.com/introduction-to-postgres-physical-storage/
+    // 
+    // for how this is done with Postgres. This approach means that only the 2-byte
+    // offsets are moved when inserting new rows/edges/atoms/index entries. The actual
+    // rows are kept in place and only the index entries at the start of the page are
+    // kept in sorted order. This will result in less work than this emulation, but
+    // there will be more work binary searching the sorted rows to find the new
+    // insert positions. So the net effect will be that the actual code will 
+    // take more time than this implementation.
     if (free > 2)
     {
         size_t slots_to_move = randomGenerator->randint(MAX_PAGES_PER_PAGE - next);
@@ -190,7 +226,15 @@ AtomHandle AtomPage::add_node(Type type, const std::string& name)
 
     std::lock_guard<std::mutex> lock(page_lock);
 
-    // Save the header information - type and not a node.
+    // This code fills pages 100%. In an actual implementation, the fill factor would
+    // likely be an adjustable setting, especially for bulk inserts. For rarely
+    // changing data, with few new inserts and deletes, one will want a high
+    // fill-factor. Since these pages will be sorted by type and name or outgoing
+    // set, this means that inserts may cause pages to split. Since the current
+    // logic does not implement sorting, it does not implement splitting either. 
+    // An actual implementation will likely have fill factors between 66% and 90%.
+
+    // Save the header information - the type and that this is a node.
     AtomHandle atom = &atom_slots[next];
     atom->as_type = type;
     atom->as_node = true;
@@ -205,8 +249,26 @@ AtomHandle AtomPage::add_node(Type type, const std::string& name)
     count++;
     next += 2;
     free -= 2;
-
-    // Move some bytes to emulate the performance of inserts.
+    
+    // Move some bytes to emulate the performance of inserts. This code is
+    // intended to reflect a move of random bytes. It moves bytes past the
+    // end of the used area so these moves will not affect stored atoms. The
+    // work is also reversed in that full pages have less work while empty pages
+    // have more work, unlike a real page where full pages will have inserts
+    // which move half the entries, on average. In an actual implementation,
+    // these pages will likely be stored using an arracy of atom offsets at
+    // the beginning of the page, with the actual edges/atoms/index entries
+    // stored from the back of the page. See: 
+    //
+    // http://rachbelaid.com/introduction-to-postgres-physical-storage/
+    // 
+    // for how this is done with Postgres. This approach means that only the 2-byte
+    // offsets are moved when inserting new rows/edges/atoms/index entries. The actual
+    // rows are kept in place and only the index entries at the start of the page are
+    // kept in sorted order. This will result in less work than this emulation, but
+    // there will be more work binary searching the sorted rows to find the new
+    // insert positions. So the net effect will be that the actual code will 
+    // take more time than this implementation.
     if (free > 2)
     {
         size_t slots_to_move = randomGenerator->randint(ATOM_SLOTS_PER_PAGE - next);
@@ -231,7 +293,15 @@ AtomHandle AtomPage::add_link(Type type, const AtomVector& outgoing)
 
     std::lock_guard<std::mutex> lock(page_lock);
 
-    // Save the header information - type and not a node.
+    // This code fills pages 100%. In an actual implementation, the fill factor would
+    // likely be an adjustable setting, especially for bulk inserts. For rarely
+    // changing data, with few new inserts and deletes, one will want a high
+    // fill-factor. Since these pages will be sorted by type and name or outgoing
+    // set, this means that inserts may cause pages to split. Since the current
+    // logic does not implement sorting, it does not implement splitting either. 
+    // An actual implementation will likely have fill factors between 66% and 90%.
+
+    // Save the header information - the type, that this is a link, and the arity.
     AtomHandle atom = &atom_slots[next];
     atom->as_type = type;
     atom->as_node = false;
@@ -253,7 +323,25 @@ AtomHandle AtomPage::add_link(Type type, const AtomVector& outgoing)
     next += slots_used;
     free -= slots_used;
 
-    // Move some bytes to emulate the performance of inserts.
+    // Move some bytes to emulate the performance of inserts. This code is
+    // intended to reflect a move of random bytes. It moves bytes past the
+    // end of the used area so these moves will not affect stored atoms. The
+    // work is also reversed in that full pages have less work while empty pages
+    // have more work, unlike a real page where full pages will have inserts
+    // which move half the entries, on average. In an actual implementation,
+    // these pages will likely be stored using an arracy of atom offsets at
+    // the beginning of the page, with the actual edges/atoms/index entries
+    // stored from the back of the page. See: 
+    //
+    // http://rachbelaid.com/introduction-to-postgres-physical-storage/
+    // 
+    // for how this is done with Postgres. This approach means that only the 2-byte
+    // offsets are moved when inserting new rows/edges/atoms/index entries. The actual
+    // rows are kept in place and only the index entries at the start of the page are
+    // kept in sorted order. This will result in less work than this emulation, but
+    // there will be more work binary searching the sorted rows to find the new
+    // insert positions. So the net effect will be that the actual code will 
+    // take more time than this implementation.
     if (free > 2)
     {
         size_t slots_to_move = randomGenerator->randint(ATOM_SLOTS_PER_PAGE - next);
--- a/opencog/benchmark/PagedAtoms.cc
+++ b/opencog/benchmark/PagedAtoms.cc
@@ -1,0 +1,347 @@
+/*
+ * atomspace/opencog/benchmark/PagedAtoms.cc
+ *
+ * Copyright (C) 2017 by OpenCog Foundation
+ * All Rights Reserved
+ *
+ * Written by Curtis Faith
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License v3 as
+ * published by the Free Software Foundation and including the exceptions
+ * at http://opencog.org/wiki/Licenses
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program; if not, write to:
+ * Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include <iostream>
+#include <fstream>
+#include <string>
+/*
+#include <list>
+#include <ctime>
+*/
+#include <sys/sysinfo.h>
+
+#include "PagedAtoms.h"
+
+using namespace opencog;
+
+
+#define USE_ATOMSPACE   false
+
+#define RAND_SEED       1505051
+
+
+MT19937RandGen* opencog::randomGenerator = new MT19937RandGen(RAND_SEED);
+
+
+AtomSlot::AtomSlot()
+{
+    as_64 = 0;
+}
+AtomSlot::~AtomSlot()
+{
+    as_64 = 0;
+}
+
+
+uint64_t EdgePage::page_count = 0;
+
+EdgePage::EdgePage()
+{
+    flags = 0;
+    count = 0;
+    next = 0;
+    free = MAX_EDGES_PER_PAGE;
+    page_count++;
+}
+
+bool EdgePage::can_add_edge()
+{
+    if (free >= 1)
+        return true;
+    else
+        return false;
+}
+
+void EdgePage::add_edge(AtomHandle in, AtomHandle out)
+{
+#if ! USE_ATOMSPACE
+    if (!can_add_edge())
+    {
+        std::cerr << "add_edge called for full page" << std::endl;
+        exit(1);
+    }
+
+    std::lock_guard<std::mutex> lock(page_lock);
+
+    // Save the edge pair.
+    Edge* edge_ptr = &edge_slots[next];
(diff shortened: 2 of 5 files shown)


linas:
Sounds interesting.  There are at least five different types of locality problems:
1) data aliases to the same cache lines, and thus cache lines get ejected too often.
2) data sits on many many 4K pages, thus overwhelms the number of available TLB entries.  The linux hugepage support can help with this.
3) data sits in different pages in RAM, causing inefficient swapin/swapout behavior.
4) overuse of pointers, pointer-chasing when in-place indexes in contiguous arrays would be better.
5) Cache lines filled with useless junk, wasted space, instead of being dense with needed data.

I don't know that we can do anything about 1 and solving 3 seems pointless.  That leaves 2 & 4,5.

For 2, a quickie experiment might be to use linux hugepages for opencog. My datasets are dozens of GBytes or more, so TLB thrashing is presumably extremely common.  So, experiments include:
a) does linux hugepage help with performance?
b) you mentioned some sort of whizzy malloc/free library on the mailing list ... it sounded cool .. have you tried this? What happened?

For 4 and 5, I did try to use `std::emplace_back` and similar ctors when building the outgoing set of an atom. The idea was that all of the outgoing set would be in an array, on the same cache line. This made no difference for small datasets. Did  not try for big ones.

My typical data is this:
```
EvaluationLink
      PredicateNode "pair"
      ListLink
            WordNode "foo"
            WordNode "bar"
``` 
It would be nice if we could get both the ListLink and the EvaluationLink located on the same cache line. Note, however: what I need to repeatedly access are the various values hanging on the EvaluationLink: the CountTruthValue, the FloatValue holding the frequency (probability) the FloatValue holding the mutual information, the FloatValue holding the cosine angle.   So, its not so much the locality of the atoms that matters, but the locality of the values attached to the atoms.

The problem with locality is "wasted airspace", Suppose you did manage to put both links, and all four values (count, freq, MI, cosine) on the same cache line. My algos usually examine ONLY the mi, or ONLY the cosine, so putting all these on the same cacheline is a waste: it just uses dead space.

It would be much much better if all   CountTruthValue's were on the same hugepage, if all cosines were on the same hugepage.... if we could alloc atom values based in their key. !! That might be awesome..

Also: need to ask "what is an atom, really?"  In my datasets, I see about 2KBytes RAM per atom. Given that `sizeof(Atom)` is maybe 150 or 200 bytes, where's the rest?  The answer seems to be:

* 100-200 bytes in outgoing set
* 500+ bytes in the incoming set
* 500+bytes in the various values attached to atom.
* 400 bytes in the atomspace indexes.

or something like that.  Not sure.  What this means is that a typical atom is actually 10x bigger than the `sizeof (class Atom)`, and that typical acces to one atom touches many, or most, or maybe *all* of these different places in RAM.   I don't know ho to even measure the non-locality of this access, and its really hard to guess which of these are important, and which are not.

linas:
So, I'm skimming your code, and so perhaps I totally misunderstand it, but ... it seems you are trying to store the string-name of a node in contiguous memory with the node itself.  This seems to me like not just a waste, but a performance-deaccelerator, since it packs useless data (the node name) with the node, thus wasting cache-line space!

The point is that there are almost never any accesses to the node-name, its almost a set-and-forget value. What does happen a lot are atomspace queries, to obtain a node of a given name and type. That works by creating an atom, computing the hash for it, and throwing away the atom.  There is a content compare only on hash collisions.  All of the rest of the time, the atoms are always access by their handles, and their handles are used mostly to find values on them - truth values or whatever, and that is where the CPU time all goes. Trying to drag along the string name with the node seems to mostly clutter RAM with boring data that is never accessed. See what i mean?

inflector:
In general, I have found that custom allocators for common and high-volume C++ objects is often much much faster than the built-ins. This appears to be true for Atoms as well.

I just added some more details from my notes and some benchmark results to the ReadMe, as well as more comments to explain the code. Also note the diagrams in the ReadMe which show the current design memory layout. I also performed a fairly detailed analysis of the current Atom memory requirements which can be found in that document.

> seems you are trying to store the string-name of a node in contiguous memory with the node itself

The code stores the std::string object there but std::string doesn't store the actual string as that is implemented via a malloc or equivalent somewhere else, std::string is a fixed 8-bytes.

> It would be much much better if all CountTruthValue's were on the same hugepage, if all cosines were on the same hugepage.... if we could alloc atom values based in their key. !! That might be awesome..

The prototype did not implement any values storage but what you outline above, i.e. allocating atom values based on their key, would be fairly straightforward. If we allocated values out of memory-pages ordered by key, then atom, then the values for a given predicate would be contiguous. This should exhibit similar speed improvements as indicated in the test notes. 

The other possibility, and one that would be higher performance, and less memory in most use scenarios, would be to have separate pages for each different key, that way each page would contain closely-packed typed values for the key. 

This would be a bit more work as you'd have to maintain a directory of pages for each different key, but it would save you 64-bits per entry. If the pages were fairly small, like 1K each, this might not be too much overhead and the number of values stored per page would increase by 50% for this approach, assuming 8-byte Atom pointers and 8-bytes for value data.

inflector:
The images are not visible until this Pull Request is merged as they are MarkDown references. So I am adding them here so you can seem them:

image

and

image

inflector:
Time to traverse 1 million EvaluationLinks looking for a match on a particular node:

```
2.54s - current AtomSpace
0.11s - paged-Atoms prototype
```

of this time, for our current AtomSpace, 1.05s is spent in get_handles_by_type building the HandleSeq list.

linas:
> I have found that custom allocators for common and high-volume C++ objects is often much much faster than the built-ins. This appears to be true for Atoms as well.

Then perhaps we should switch to one of these.  

> std::string is a fixed 8-bytes

Yeah. Since the name and the outgoing set are immutable, it makes sense to ditch `std::string` for storage and just use fixed-length arrays for this, for the outgoing set, and for the lists of values.  The only real problem with this suggestion is that our current API returns `const std::string&` which makes such a change ugly/unworkable :-(

> separate pages for each different key
Separate malloc pools for each key, yes, exactly.

linas:
I'm not sure what to say about the rest of the post.  You seem to be discarding or discounting various design points.   The reason we have `(Evaluation (Predicate  "breath") (List (Concept "birds") (Concept "air")))` instead of the flatter structure `(Evaluation (Predicate  "breath") (Concept "birds") (Concept "air"))` is partly historical, and partly to minimize RAM usage: there will also be `(Evaluation (Predicate  "fly_through") (List (Concept "birds") (Concept "air")))` and `(Evaluation (Predicate  "inhale") (List (Concept "birds") (Concept "air")))` and `(Evaluation (Predicate  "use") (List (Concept "birds") (Concept "air")))`  and so on -- so having a ListLink indirection makes the structures smaller, overall.

Note that incoming sets can be huge: my current incoming sets for English and Chinese range from a one to about a million atoms. The actual distribution is Zipfian, I can show you charts. I'm concerned that you are optimizing for a situation where you have tiny incoming sets. I would like you to do your prototyping and measuring on some of my actual datasets. 

Your diagram shows "set-by-type" -- this was added about 1/2 year ago, maybe more recently, and gave an instant 10x performance improvement in incoming set traversal. This was an important change, yet its not visible in your altered diagram. The issue here is that this solved a major performance bottleneck for me, yet that bottleneck is not documented anywhere, not measured in any unit test or benchmark. 

My experience with performance tuning is that it can be like squeezing a balloon: you squeeze hard in one spot, and it pops out somewhere else. We don't have good tools to measure this stuff.

linas:
FYI: during the  "set-by-type"  design: I tried 4 or 5 alternate designs, including vectors, arrays, hash tables, multi-maps, before settling on the current design. All of them sucked in different ways. the hash tables blew up ram usage two-fold, because most hash slots were vacant. The vector/array variants were too hard to edit in-place, because atoms were coming and going all the time. The also required computing an index, which was CPU-expensive. The std::multi-maps were almost perfect, except that the std:: api for them did not provide a way of iterating over just the types that were desired: one ended up examining/touching many or most atoms, just to grab the one or two that were actually wanted.

The reason I'm saying this is because I look at the "inbound link index" in your second diagram, and it looks like an array, and I already tried something like that, and it didn't work. Of course, what I tried is probably very different than what you've got, but ... my benchmark that day was also very different from your benchmark today.

linas:
Also: you wrote:
> Current AtomSpace          25.0 μs
> Page-based                  1.3 μs

How does this reconcile with the current AtomSpaceBenchmark results, on your machine?  I'm certain that your machine is 2x or 3x faster than mine (mine is about 5-6 years old, now) and I'm getting addNode times of about 3μs and addLink times of 10μs so don't understand why you would be getting 25 μs on what must surely be a much faster machine.

inflector:
Linas wrote:
> I'm concerned that you are optimizing for a situation where you have tiny incoming sets. I would like you to do your prototyping and measuring on some of my actual datasets.

and:
> Your diagram shows "set-by-type" -- this was added about 1/2 year ago, maybe more recently, and gave an instant 10x performance improvement in incoming set traversal. This was an important change, yet its not visible in your altered diagram.

The code in this prototype is designed to test the storage for atoms that uses B*-tree like storage. It is like a b-tree except the pages are in-memory and not paged to disk.  The first phase of this test indicated that the potential gains are worth exploring further because they offer an order of magnitude better performance.

So, I wanted to model performance to see if it was likely to result in a gain before doing a full implementation which might be a two or three month effort to get working well under all circumstances, documented, optimized for multi-threading etc. 

Steps in my process:

1) Model performance of iteration and creation
2) Share results and discuss
3) Model performance of indexed lookups and inbound set iteration

We're on step 2, and the next step will be to address this issue to see if this approach will also help inbound set searching. To do this will require implementing the on-page searching for index pages. In the case of the EdgePages, this will be used to quickly find the page which contains the inbound links for a given type, or type and atom. The way this works in a typical b-tree is to start at the root page, search that page for the sub-page containing the target type and atom, then search the sub-page, recursing until you get to a leaf page. 

From that point, since the leaf edge pages have a linear array layout, and these leaf pages are connected with pointers to adjacent pages, iterations over incoming sets will be linear traversals of memory which is the fastest iteration access pattern for hardware. Since the pages are relatively small, the shuffling around required to maintain sorted order for insertions and deletions is relatively minor, especially when compared with a large vector which might contain all the inbound atoms, for example. In the case of a 1K page, you'd be moving a few hundred bytes in a few already cached lines in the typical case. Only testing will show for certain.

Another approach worth testing and possible for the edge case is to add a vector with entries with the pointer to the leaf edge page for every type, that way one could simply look up the page in the vector and then search only that leaf page and not have to do a binary search. It may be that this approach could totally replace the b-tree itself for edge pages.

In the current prototype code, there are calls to memmove that are there to perform work that is approximately the same as what will be done in phase II when the actual search and sorting is implemented. What is not currently implemented is the actual comparison functions.

In the case of EdgePages where inbound links are stored, since each edge-pair is 16 bytes, and cache lines are 64 bytes, it may be that a straight linear search is faster than the more traditional binary search. For a 1K page, there are 16 64-byte cache lines, each containing 4 16-byte pairs. It may be that linear searches that check every fourth element at first and then check the skipped items on the first comparison that is greater than the target will be faster than a pure binary search because of the cache line pre-fetching hardware works well for linear searches at regular offsets in either ascending or descending order.

I expect this approach to exceed the performance of the current Map to Set standard containers, especially for larger incoming sets because there will be much less pointer chasing.

> How does this reconcile with the current AtomSpaceBenchmark results, on your machine? I'm certain that your machine is 2x or 3x faster than mine (mine is about 5-6 years old, now) and I'm getting addNode times of about 3μs and addLink times of 10μs so don't understand why you would be getting 25 μs on what must surely be a much faster machine.

I have been testing using a fairly old and slow machine from Ben's AI hedge-fund days which is a Dell PowerEdge T420 with a 6-core Xeon E5-2420 @ 1.90GHz and 32G of RAM. It is about 5 or 6 years old, as well.

Also, remember that the per-op time is for the EvaluationLink which is actually a creation of two link atoms, the EvaluationLink and the ListLink. There are no nodes created during the timed benchmarks. This corresponds to creating observations on word pairs, the predicate atom and the word nodes will already have been created so the additions will be the EvaluationLink and (perhaps) the ListLink for pairs that have not been observed before. In the benchmark, there is a pool of 100,000 nodes created in the test setup that are randomly paired for the ListLink, so most of the pairs will be new and result in a ListLink being created.

I reran some of these tests and am now getting numbers between 21 to 22 μs fairly consistently. This seems pretty close to your measurements, as it corresponds to a 10.5 to 11.0 μs per link.

inflector:
I can see how the less flat structure that includes an additional ListLink saves memory, however, it also loses semantics and makes it possible to introduce subtle errors if you are not very careful.

For instance, in digging into the NLP processing of pairs, I noticed this from [object-api.scm](https://github.com/opencog/atomspace/blob/master/opencog/matrix/object-api.scm#L327-L338):

```
	(define (get-right-stars ITEM)
		(define want-type (LLOBJ 'right-type))
		(define pair-type (LLOBJ 'pair-type))
		(filter
			(lambda (lnk)
				(define oset (cog-outgoing-set lnk))
				(and
					(equal? 2 (cog-arity lnk))
					(equal? ITEM (first oset))
					(equal? want-type (cog-type (second oset)))
				))
			(cog-incoming-by-type ITEM pair-type)))
```

When called from [the inner loop of compute-mi.scm](https://github.com/opencog/atomspace/blob/master/opencog/matrix/compute-mi.scm#L472-L475):

```
    ; Run the inner loop
    (for-each
         do-one-pair
        (star-obj 'right-stars left-item))
```

this will get all incoming ListLinks where the `left-item` is the first element.

It looks like this will match ListLinks for other predicates, and not just the ones you might be looking for. For instance, if the atomspace contained:

```
(Evaluation
    (Predicate "*-Sentence Word Pair-*")
    (List
        (Word "bird")
        (Word "air")))
```
and 

```
(Evaluation
    (Predicate "*-Idiom Pairs-*")
    (List
        (Word "bird")
        (Word "brain")))
```

The code will look at the incoming set for (Word "bird") and find all arity 2 ListLinks where "bird" is the first element and the second element is type Word. So it will match the \*-Idiom Pairs-\* predicate ListLink for <bird, brain>, even when those are not actually part of the observed words set.

If you had the more compressed forms:

```
(Evaluation
    (Predicate "*-Sentence Word Pair-*")
    (Word "bird")
    (Word "air")))
```
and 

```
(Evaluation
    (Predicate "*-Idiom Pairs-*")
    (Word "bird")
    (Word "brain")))
```

then the filter could look for Evaluations where the desired predicate was the first element, the left item is the second, and the right item is the third. With the additional level of indirection, if you want to avoid this problem, you end up having to look at the incoming set of the ListLink to see if it contained an EvaluationLink that had the desired \*-Sentence Word Pair-\* predicate in first position and the ListLink being filtered in second position. This is considerable additional work.

When doing counts, because a ListLink could appear in many places, a count on the ListLink will end up incrementing the pair count for all Evaluation (and other) links that contain that ListLink pair. If different outer-level links interpret those counts differently, that might result in subtle bugs which will be hard to find or even notice.

If you count at the outer EvaluationLink level, this problem does not occur.

linas:
> Steps in my process:

Usually there is a step 0, which defines the collection of datasets and algorithms to be optimized. Up until now, this has been ad-hoc and sort-of haphazard. In the past, my performance work was driven by performance-annoyance-of-the-day and occasional spot checks with AtomSpaceBenchmark. I never permanently stored specific benchmark datasets cause this is a lot of work. I suggest that right about now is the time to start doing that.  

The problem is that AtomSpaceBenchmark is kind-of hokey, buggy, badly-designed. it also measures only some bare minimum, and not any real-world scenarios. It would be better if we had some collection of shell scripts and datasets that timed some "real-world" applications (in addition to the current or new/improved AtomSpaceBenchmark).  Because datasets might be large ad-hoc ugly tangles, we should probably devote a new repo to that.

linas:
> EdgePages

I did not entirely understand what you wrote.  However, two concerns arise: 

1) the system ends up with lots of half-full pages.  This would be terrible. The reason I mentioned Zipf (or scale-free networks, in general) is that a Zipf-style distribution of 1 million atoms will have 500K atoms with an incoming sets of size 1, and then 250K atoms with incoming set of size 2,  then 120K of size 4, 60K of size 8, and .. and (because the Zipf edges tend to be rounded) a few dozen or a few hundred atoms with an incoming set size of 1 million.  Notice its log-log linear. 30K size 16, then 15K of size 32, then 8K of size 64 -- these are ballpark numbers, I might be off by a factor of 2-4, and sometimes the power law is not 1/f  pink noise, but 1/f^2 brown noise. etc.  Some of my data is not power law, too.  But its never "white noise" and AtomSpaceBenchmark measures white noise or something like that.

2) it sounds like there will be lots of copying/shuffling/movement of data on each page. !?? We do NOT want to accidentally re-invent an equivalent of the MS-DOS  file system, with all the fragmentation and locality issues it had. (Recall the details: it was like a restaurant filled with 4-seat tables. A party of 3 walks in, gets seated at the first table. Another party of 3 walks in; one person is seated at the first table, the other 2 at the second table. This continues until defrag, when everyone is asked to pick up their plates and  knives and napkins, and move to a different table. Ugh. Real-life restaurants don't do this cause its stupid.)  The way you talked about the pages makes it seem like there would be "holes" whenever atoms are deleted, and lots of fragmentation.

3) FWIW, its almost more important to get this right for Values, instead of Atoms. The atoms are fairly static, mostly get laid down once, and that's it. By contrast, values are being constantly recomputed, so there's gonna be a huge churn of RAM as a result.   The good news: values do not have incoming sets, by design, precisely in order to avoid the above issues. 

This is a super-important, critical point for users: the performance profile of atoms, and the performance profile of values are meant to be very very different, with atoms being used for quasi-static, unchanging  data, while values get used for high-frequency rapidly changing data. The performance tuning that we do needs to reflect this design point.

linas:
Re: The actual pairs not not always EvaluationLinks and PreidcateNodes. There are four types of matrixes that I currently use:
```
(Evaluation
    (Predicate "*-Sentence Word Pair-*")
    (List
        (Word "bird")
        (Word "air")))
```
while, for cosine similarity:
```
    (SimilarityLink
        (Word "bird")
        (Word "air")))
```
What's paired up is different.  The pair datasets almost don't overlap.  Then there's
```
    (Section
         (Word "bird")
         (ConnectorSeq ; aka "disjunct"
                  (Connector (Word "is") (ConnDir "+"))
                  (Connector (Word "the") (ConnDir "+"))
                  (Connector (Word "word") (ConnDir "+"))
       ))
```
The above is asymetric: the left and right sides are very different. My datasets (English and Chinese) have approx 100K words and approx 6M ConnectorSeqs. forming maybe 10M word-disjunct pairs.

To study distance correlations, I use the structures:
```
     ExecutionLink
         SchemaNode "*-Pair Distance-*"
         ListLink
             WordNode "lefty"
             WordNode "righty"
         NumberNode 3
```
which is only a "pair" when you look at the middle bit.  

I am planning a more complex structure, "real soon now", to handle English idioms, Chinese words and indoeuropean morphology; it will look like the `SectionLink` up above, except that some of the connectors will get connected, while other connectors remain free.  And there may be 2 or 3 or 4 words (aka morphemes) in these kinds of sections.

The point is: only the simplest things can be jammed into EvaluationLinks+PredicateNodes. Not everything fits.

linas:
p.s. the reason for using `SimilarityLink` is because it is an `UnorderedLink`  i.e. is left-right symmetric.  So: 
```
(SimilarityLink
        (Word "bird")
        (Word "air"))
```
is exactly the same atom as
```
(SimilarityLink
        (Word "air")
        (Word "bird"))
```
and so this cuts by half the number of atoms that have to be stored.   The atomspace handles the re-ordering automatically, under-the-covers.

inflector:
Linas wrote:
> However, two concerns arise:  1) the system ends up with lots of half-full pages. This would be terrible
It's not so bad in actual practice. You can set things up so that pages are at least 2/3 full or 90% full, it all depends what you do when you split pages. Since the RAM requirements for this approach are much less, you could add a 50% penalty for 2/3 full pages, and still save 2/3 off the current memory footprint.

> it sounds like there will be lots of copying/shuffling/movement of data on each page
The work for this is already modeled in the prototype benchmark. On delete, one create in-place linked lists for empty slots. Inserts first fill empty slots. All nodes will the the same size, any most links will be of fairly common sizes so there won't be real problems with fragmentation.

Or you can have a separate-process cleanup deleted item memory, potentially even moving around atoms and changing inbound references. Postgres does this with their disk-based pages. No reason you can't do the same thing with an in-memory page-based system.

> FWIW, its almost more important to get this right for Values, instead of Atoms
Yes, I can see why this would be important.

For batch processes that compute values and then consume them later as inputs, we should be able to get very high performance and locality of reference.

linas:
So: here's one more thing to think about. 

When one first learns what a graph is, one gets the most-basic definition, which is this: a graph is a set of vertexes V, and a set of edges E which are ordered pairs that connect vertexes, possibly carrying some label.  That is, a single edge is the ordered triple (label, v1, v2) -- this is exactly the form that your redesigned evaluation link is like -- its (predicate, concept1, concept2).

When one learns some more graph theory that goes beyond the bare-bones basics,  one encounters "frames and locales", which are much better for describing the local structure of a graph.  The problem with the above naive definition is that if I give you a vertex, you have no clue what might be connected to it.  You have to search the whole set of edges to find out.  The solution is to turn the naive definition "upside-down", and instead describe a graph as a set of sections (v, e1, e2, e3, ...) where, if I match up with the naive definition, then the first element v can be thought of as being kind-of-like a vertex, and the e1,... en can each thought to be kind-of-like an edge, so that the whole section  (v, e1, e2, e3, ...) is kind--of-like a vertex, plus all of the locally-connected edges on that vertex.  Visually, you can think of it as a spider with n legs on it.  The graph as a whole is a gluing together of these spiders. 

Both ways of describing the graph are equivalent (when the graph has finite number of vertexes/edges) The section-based way is nicer, because it instantly tells you about the local structure of the graph, e.g. the degree of a vertex is instantly available. Also, when you glue together two sections, you still have a section. By contrast, when you glue together two edges, you no longer have an edge, you just have some arbitrary subgraph, and you don't know anything about what kind of subgraph it might be. By contrast, the gluing axioms for sections result in all sorts of nice properties (and are always given in chapter one of any book on algebraic topology). 

So I'm trying to move to, work with the atomspace in a more section-like way, treating sections as the basic objects, rather than vertexes/edges. The vertex-edge model for a graph is better than nothing, but its awkward in many ways.  I suspect that if we were somehow able to organize the data storage in the atomspace to be more section-like, then graph traversal would be better, faster, because the locality and connectedness information would be "right there" at your fingertips, instead of being spread out over many different pages.

That is, the risk is that graph traversal, with the pattern matcher,  would be doing a lot of page-hopping, bouncing from page to page.  Which suggests that one or several of the benchmarks should be pattern-matcher crawls over some graph.

linas:
Is this ready to be merged?

inflector:
It can be merged at any point, but I do intend to model inbound link traversal in the near future.

Your comments about sections do seem like a better way of thinking about things. In many ways edges are the real stuff of graphs and a vertex-centric approach masks this.

linas:
ok, merge whenever you want to.

ngeiswei:
What is blocking this PR from being processed?

linas:
If this is a dead project, should it be merged at all?  Someday in the future, someone will look at this say "WTF, this is dead", and will have to remove this code.  If its dead code, better not to merge it at all. Right?

linas:
Closing for now.

PR #1325, opened 2017-08-02
Support backtracking to a glob-containing clause in pattern matcher

leungmanhin:
For a term that has globs in it, there may be more than one valid way to ground those globs to the same candidate. If at some point the pattern matcher discovers that the grounding of that term conflicts with some other term in the same pattern, it should backtrack and try to ground the term in another way, this is what this PR attempts to do.
But this is incomplete, I need more time to understand different parts of the pattern matcher code

### DIFF
--- a/opencog/query/InitiateSearchCB.cc
+++ b/opencog/query/InitiateSearchCB.cc
@@ -398,9 +398,9 @@ bool InitiateSearchCB::neighbor_search(PatternMatchEngine *pme)
 			              << h->toShortString();})
 			bool found = pme->explore_neighborhood(_root, _starter_term, h);
 
-			// If the starter term has globs in it and it's possible to
-			// ground it to the same candidate in a different way, go for
-			// it, before moving on to the next candidates.
+			// If the starter term (or any other term we've explored) has globs
+			// in it and it's possible to ground it to the same candidate
+			// differently, go for it, before moving on to the next candidates.
 			if (not found and pme->has_more_to_explore())
 				pme->explore_neighborhood(_root, _starter_term, h);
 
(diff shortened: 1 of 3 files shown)


leungmanhin:
@linas / @ngeiswei , what are the usual places in the pattern matcher engine that I should probably pay more attention to when it comes to doing this kind of backtracking? Any advice would be much appreciated...

bgoertzel:
@linas @ngeiswei I know you are both busy, but getting Ghost working is quite urgent for some of our near-term robotics goals... so if one of you could take a day or two (or an hour or two, whatever it takes) to help Man Hin with this it would be much appreciated...

linas:
The code looks plausible. There's lots and lots of comments in the code explaining what it does and why. So I'm not sure what to add.  Backtracking  is done via a stack: current state is saved on a stack, so backtracking is done by popping the stack. 

So, you're at point A on a graph, you have an incoming set of size N, so you have N choices to try out. Pick one push the stack, get to point B, where you have M choices, push the stack get to point C where you have K choices.  After trying all K choices, if none are accepted, pop the stack back to point B, and try the next choice there.  For plain-old ordinary OrderedLinks, that's it: all branches/choices are in the incoming set. Its easy. There are push-pop locations in the code that should be easy to find.

Its harder for UnorderedLinks, because you have to do the above, AND ALSO deal different permutations of the UnorderedLinks. Handl;ing globs in unordered links will be a bit painful, I imagine.  Anyway, there is also a very clear-cut location in the code where this push/pop is done.

There are two more minor twists. One is the AbsentLink, where you explicitly check that something is NOT present.  Finally, there is ChoiceLink, where you are given several radically different graphs, and you only have to match one of them. These are also wrapped up in their own modular blocks of code.

All of them use the same push-pop idea, all use the same "API" -- four different routines, but taking the same arguments, handling the above four cases.

linas:
I edited above answer. Also: I should mention that some (many) of the pushes are "automatic", in that the C stack handles everything: you push when you call a subroutine; current state is saved on the stack, and when the subroutine call returns, then obviously, the C stack popped.  There are three explicit stacks that get used, because its impractical to use the C call stack for that particular state. The explicit stacks are for the solution as a whole, the individual clauses, and for the UnordredLink permutations.  All other data that needs push/pop lives in the C stack, i.e. in local variables in the calling function.

linas:
edited again for clarity; read the comments in github, instead of the emails.  The emails were a bit garbled.

leungmanhin:
Yea I am basically reusing those existing push/pop mechanisms, I've been using it to test some of the ghost (chatlang) rules and now it seems to work with those that were failing earlier. Some more tweaking in the glob_compare to go back and forth between states will be needed before this can be merged, thanks @linas

leungmanhin:
Let me do it in a different PR with unit test covered, thanks

Issue #1318, opened 2017-07-25
BackwardChainerUTest failing in Jenkins


linas:
The BackwardChainerUTest has been consistently failing, 100% of the time, in Jenkins, for maybe a month, now.  It passes, 100% of the time, on my machine.

This is starting to get annoying, github repo is always reporting as "failed" -- can we fix this?

ngeiswei:
It'd be really cool if I can have access to a machine where it fails.

linas:
```
Masayuki Hatta <mhatta@gmail.com> | Masayuki Hatta <mhatta@gmail.com> | 4:13 AM (19 hours ago)
-- | -- | --
Masayuki Hatta <mhatta@gmail.com> writes that: 

Hi,

I'm now trying to build atomspace on Debian sid, but one of tests (BackwardChainerUTest) fails.

I think the following is the culprit, but can't figure out what and why.  Any help appreciated.

Best regards,
MH

(snip)

/home/mhatta/work/Debian/atomspace/atomspace-0.0~git20170802.d79fd34/tests/rule-                                                                                        [94/1850]
engine/BackwardChainerUTest.cxxtest:57: Error: Test failed: Backtrace:                                                                                                          
In ice-9/boot-9.scm:                                                                                                                                                            
 160: 8 [catch #t #<catch-closure f1aa730ac0> ...]                                                                                                                              
In unknown file:                                                                                                                                                                
   ?: 7 [apply-smob/1 #<catch-closure f1aa730ac0>]                                                                                                                              
In ice-9/boot-9.scm:                                                                                                                                                            
 160: 6 [catch #t #<catch-closure f1aa730980> ...]                                     
In unknown file:                                                                       
   ?: 5 [apply-smob/1 #<catch-closure f1aa730980>]                                     
In ice-9/eval.scm:                                                                     
 386: 4 [eval # ()]                                                                    
 393: 3 [eval #<memoized gt-zero-confidence> ()]                                       
In unknown file:                                                                       
   ?: 2 [memoize-variable-access! # #]                                                 
In ice-9/boot-9.scm:                                                                   
 105: 1 [#<procedure f1aa4eb480 at ice-9/boot-9.scm:100:6 (thrown-k . args)> unb       
ound-variable ...]                                                                     
In unknown file:                                                                       
   ?: 0 [apply-smob/1 #<catch-closure f1aa730940> unbound-variable ...]                
                                                                                       
ERROR: In procedure apply-smob/1:                                                      
[...]

PR #1327, opened 2017-08-03
Experimental fix for bug #1318

linas:
I'm guessing its the old environment bug, again.  Lets see if this fixes it.


linas:
this fix doesn't work, based on jenkins

PR #1334, opened 2017-08-13
Clear the groundings of variables that are grounded in glob_compare when backtracking

leungmanhin:


opencog/query/PatternMatchEngine.cc
```diff
@@ -693,6 +693,17 @@ bool PatternMatchEngine::glob_compare(const PatternTermSeq& osp,
 	GlobGrd glob_grd;
 	GlobPosStack glob_pos_stack;
 
+	// A variable or glob in osp may have been grounded before
+	// reaching here if it exists in some other clause also.
+	// Clear only those that are grounded here in glob_compare().
+	// solution_push/pop is not used as it seems to become a bit
+	// tricky when we are resuming form a previous match...
+	auto clear_grounding = [&](const PatternTermPtr& p)
```
linas: Yes, the solution_push/pop  is not enough.  You absolutely do need to push/pop the glob groundings.

Example:  pattern has two globs, one variable, say blob1, blob2 and var.    Say that there are two solutions: (blob1=g1 blob2=g2 var=g3) and  (blob1=g1 blob2=g1 var=g4)   So:  you have to save things on a stack, because if the pattern matcher finds (blob1=g1 blob2=g1 var=g1)  you have to pop (var=g1) but keep the two blobs on the stack. Then later on, after you've exhausted all possibilities (blob1=g1 blob2=g1 var=anything) you have to pop (blob2=g1) but you do need to still keep (blob1=g1) on the stack.    Then, after exhausting (blob1=g1 blob2=anything var=anything), you have to pop  (blob1=g1) and start trying (blob1=g2) and so on.

So it seems to me that you don't need a "clear". you just need save the glob groundings in the same stack where variable groundings are saved.

leungmanhin: How about, say, after grounding it to (blob1=g1 blob2=g2 var=g3), it turns out that it's not a valid grounding because the var happens to exist in some other term as well and it's expected to be g4, so we will have to go back to glob_compare for the same candidate and try to ground var to something else (g4), would it also work if I save the glob groundings in the same stack where variable groundings?

linas: sorry for late reply. Yes, glob and variable groundings should go on the same stack. It should be safe to think of globs as being a "special kind of variable" (or the other way around - a variable is  a special case of glob)  so they should be handled the same way.

The only reason to handle them sperately might be performance: if a pattern has no globs, then don't bother pushing/poping a stack for them,

In fact, that is really the only reason why there are multiple stacks: to save some cpu time, because not all of them need to be pushed every time something minor happens.   Conceptually, one stack is enough.  For performance, having several is faster.

PR #1341, opened 2017-08-17
Scheme should not print extracted atoms.

linas:
This is a fix for issue #127.   Atoms that are not in any atomspace are possible; the C++ API explicitly allows this. However, the scheme API was designed with the assumption that all atoms are always in an atomspace, and so printing atoms that have been removed is ... confusing.


linas:
Very unexpectedly, this fix, which is essentially a guile printing fix, also makes the `BackwardsChainerUTest` fail! This implies that the backwards chainer is somehow working with atoms that have been removed from any atomspace, yet is still manipulating them as being valid atoms.  I'm pretty sure this is not a desired behvaior.

Issue #1347, opened 2017-08-18
The path to core_types.scm is confusing and gives error


mhatta:
I'm using my Debian packages and got the following error.

```
scheme@(guile-user) [1]> (use-modules (opencog))
While compiling expression:
ERROR: In procedure primitive-load-path: Unable to find file "opencog/atoms/base/core_types.scm" in load path
```
core_types.scm is actually installed under /usr/share/opencog/scm/opencog/base/.  Which one is correct?  With "atoms" or not?

mhatta:
See PR #1349

AmeBel:
The error is caused because you didn't run `(add-to-load-path "/usr/share/opencog/scm")` before trying to load the module.

mhatta:
@AmeBel Actually I did run it in ~/.guile.  It worked until recently.  I think the path is recently changed.

AmeBel:
no it isn't https://github.com/opencog/atomspace/blob/fc07287a2f55356622b292e448399d0a8595f257/opencog/scm/opencog.scm#L96

linas:
@AmeBel didn't you recently create a file-system link from the default guile module path, to `/usr/share/opencog/scm` ?  That means that ` (add-to-load-path "/usr/share/opencog/scm") ` should no longer be necessary.  Anyway, yes, that PR is wrong

mhatta:
Hmm, ok, so what should I do?  Is that a guile-2.0 bug or something?

AmeBel:
@linas yes, but @mhatta is trying out  a debian package thus the issue

AmeBel:
@mhatta it is isn't a guile bug. It seems it is an issue of using compiled guile code. 

I installed opencog-atomspace and run the following,
```scheme
scheme@(guile-user)> 
root@8c6846938629:/host# guile
GNU Guile 2.0.13
Copyright (C) 1995-2016 Free Software Foundation, Inc.

Guile comes with ABSOLUTELY NO WARRANTY; for details type `,show w'.
This program is free software, and you are welcome to redistribute it
under certain conditions; type `,show c' for details.

Enter `,help' for help.
scheme@(guile-user)> %load-path
$1 = ("/usr/share/guile/2.0" "/usr/share/guile/site/2.0" "/usr/share/guile/site" "/usr/share/guile")
scheme@(guile-user)>  (add-to-load-path "/usr/share/opencog/scm")
scheme@(guile-user)> %load-path
[...]

PR #1349, opened 2017-08-19
Seems core_types.scm need to be in ${DATADIR}/scm/opencog/atoms/base

mhatta:
Currently core_types.scm is installed in ${DATADIR}/scm/opencog/base.  Then Guile can't find it.  I guess the correct path is ${DATADIR}/scm/opencog/atoms/base.


linas:
per discussion in #1347 this does not fix any issues, and creates new ones.  Sorry.

Issue #1336, opened 2017-08-16
Bug in AtomSpace::get_handles_by_type


ngeiswei:
`AtomSpace::get_handles_by_type` manages to fetch the same atom twice, moreover one of them is orphan while the other has incoming link. By same atom I can confirm that there are identical, by content, not just by hash value.

I don't have a simple way to reproduce the bug yet, will edit this issue as soon as I can.

ngeiswei:
Found more, there are 2 atoms, identical by content, stored in the TypeIndex. I can tell since they have 2 different memory addresses.

I have no idea how that can happen. The atoms are all created via a scheme script. I should mention that the script manages multiple atomspaces and copy atoms from one atomspace to another, in case that's relevant.

ngeiswei:
This is the first offer
```
[2017-08-16 07:39:25:901] [DEBUG] AtomSpace::add atom = (QuoteLink
  (ImplicationScopeLink
    (UnquoteLink
      (VariableNode "$V-4fe729a2") ; [8925185760973662502][76]
    ) ; [18175843278382742799][76]
    (UnquoteLink
      (VariableNode "$P-42099628") ; [1261471462976964581][76]
    ) ; [10512128980386044878][76]
    (UnquoteLink
      (VariableNode "$Q-209c03b2") ; [1012044293022762668][76]
    ) ; [10262701810431842965][76]
  ) ; [12268298862788601796][76]
) ; [12295003801203452998][76]

[2017-08-16 07:39:25:903] [DEBUG] TypeIndex::insertAtom [0x339e220] = [12295003801203452998][76]
[2017-08-16 07:39:25:905] [DEBUG] AtomSpace::add atom = (QuoteLink
  (ImplicationScopeLink
    (UnquoteLink
      (VariableNode "$V-4fe729a2") ; [8925185760973662502][76]
    ) ; [18175843278382742799][76]
    (UnquoteLink
      (VariableNode "$P-42099628") ; [1261471462976964581][76]
    ) ; [10512128980386044878][76]
    (UnquoteLink
      (VariableNode "$Q-209c03b2") ; [1012044293022762668][76]
    ) ; [10262701810431842965][76]
  ) ; [12268298862788601796][76]
) ; [12295003801203452998][76]

[2017-08-16 07:39:25:906] [DEBUG] TypeIndex::insertAtom [0x3371460] = [12295003801203452998][76]
```

happening here continuously.

linas:
[...]

PR #1356, opened 2017-08-23
Improve utest

ngeiswei:
1. Simplify a bit BackwardChainer utest
2. Add utest for scm function `(clear)`. It attempts to delete neasted Lists and Sets, failed to do so as pointed in issue #1336


ngeiswei:
@linas feedfack on that would be very welcome, as I still don't know what would be a proper fix.

linas:
Feedback?  The addition to the unit test looks reasonable. The fix to #1357 looks great.  Not sure what feedback to give, the commentary in #1336 is pretty long.

PR #1359, opened 2017-08-30
Fix type index add atom

ngeiswei:
Here's my f#### fix for the problem described in #1342 . Please @linas review and merge only if it makes sense (well the unit test should be merged however, maybe I should have splitted the PR in two).

### DIFF
--- a/opencog/atomspace/AtomTable.cc
+++ b/opencog/atomspace/AtomTable.cc
@@ -286,11 +286,7 @@ Handle AtomTable::getLinkHandle(const AtomPtr& orig, Quotation quotation) const
     // Currently, ScopeLinks use a custom hash, and, in order
     // for it to work, we must have an actual instance of the
     // class, so that the correct virtual method can be called.
-    //
-    // However, bad quotation nesting means that some things
-    // that look like ScopeLinks are just invalid fragments
-    // of search patterns. Ignore those.
-    if (unquoted and classserver().isA(t, SCOPE_LINK)) {
+    if (classserver().isA(t, SCOPE_LINK)) {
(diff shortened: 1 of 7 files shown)


opencog/atomspace/AtomTable.cc
```diff
@@ -286,11 +286,7 @@ Handle AtomTable::getLinkHandle(const AtomPtr& orig, Quotation quotation) const
     // Currently, ScopeLinks use a custom hash, and, in order
     // for it to work, we must have an actual instance of the
     // class, so that the correct virtual method can be called.
-    //
-    // However, bad quotation nesting means that some things
-    // that look like ScopeLinks are just invalid fragments
-    // of search patterns. Ignore those.
-    if (unquoted and classserver().isA(t, SCOPE_LINK)) {
+    if (classserver().isA(t, SCOPE_LINK)) {
```
ngeiswei: @linas this is the meat of the fix.

linas:
I am merging now, so that I can play with this. I was never quite convinced that quotation worked correctly, so maybe your fix is right. Lets see...

linas:
So ... I'm working on this code, and if I simplify it, all that seems to happen is that the whole quotation infrastructure becomes un-neeed. So whatever quotation used to do, it isn't doing it any more.  Apparently, there aren;'t any unit tests to test quotation.

ngeiswei:
I had suspected that too, that quotation isn't needed here. If it doesn't break unit tests then fine. It might break something on the opencog repo but whatever it breaks should be unit-tested in the atomspace repo so...

linas:
Well, you are the one who added the quotation device, so I'm lobbing this back to you. Currently, its just an extra argument passed around in the methods. I left it there, for now.

PR #1361, opened 2017-08-31
Remove seemingly useless quotation in AtomTable::getHandle

ngeiswei:


ngeiswei:
As per #1359

linas:
I don't think quotatino and context belong in atoms/base because they are not atoms. Maybe they belong in atomspaceuitls.  I might move them now.

linas:
Ugh. The more closely I look at quotation, the more thoroughly I hate it.   There seems to be something deeply and fundamentally flawed with it. Its just plain wrong.   

I'm not quite sure, but my gut sense is that: we really should be using Kripke frames instead. That is the correct way to do contexts. Hacking contexts and quotations into atoms just seems wrong, a bad design.   We need to engineer kripke frames.

linas:
I wish great harm and damnation in hell to QuoteLink and LocalQuoteLink.  These are probably the #1 worst bad ideas we've ever implemented in opencog.

PR #1371, opened 2017-09-07
improve control policy

ngeiswei:
re-organize rule-engine test directory as well.


opencog/atoms/execution/EvaluationLink.cc
```diff
@@ -475,7 +475,11 @@ TruthValuePtr EvaluationLink::do_eval_scratch(AtomSpace* as,
 	{
 		return do_eval_scratch(as, DefineLink::get_definition(evelnk), scratch);
 	}
-	else if (INHERITANCE_LINK == t or IMPLICATION_LINK == t)
+	else if (// Links that evaluate to themselves
+		INHERITANCE_LINK == t or
+		IMPLICATION_LINK == t or
+		EXECUTION_LINK == t
```
linas: This is incorrect. Execution links are executable. They are not evaluatable.  If you want evaluation, ue an EvaluationLink. If you want execution, use an ExecutionLink!

linas: FWIW, I am maybe over-stating my case.  Now that we have generic values, it is perhaps the right time to merge ExecutionLink and EvaluationLink into one.    Note that both TV's and Atoms are special cases of values.

Specifically, a ValuationLink could be any link that returns a value, EvaluationLink could be any ValuationLlink that returns a truth value, and an ExecutionLink is any kind of ValuationLink that returns an atom.

opencog/rule-engine/backwardchainer/ActionSelection.cc
```diff
@@ -40,12 +40,12 @@ HandleCounter ActionSelection::distribution()
 	// Generate all beta distributions of all TVs
 	std::vector<BetaDistribution> betas;
```
linas: FWIW, a FloatValue is any vector for floats.  So perhaps you want to use that to implement distributional TV's.

linas: Also, if you need two different vectors of floats, you can have that too -- a value can be an n-ary tree of strings and floats. Its built in, its got unit tests, it can be saved in the sql database, it works.

opencog/atoms/execution/EvaluationLink.cc
ngeiswei: @linas an `ExecutionLink` doesn't execute anything, it keeps a record what has been executed, or is expected to be executed, as explained here http://wiki.opencog.org/w/ExecutionLink

I agree that the terminology can be confusing, I'm used to it, but maybe others would prefer something like
`ExecutionOutputLink` -> `ApplicationLink`
`ExecutionLink` -> `ApplicationRecordLink`

PR #1372, opened 2017-09-11
Fix handle comparision

linas:
The handle-comparison operators were broken.  Thus, std::set<Handle> misbehaved and incorrectly merged non-identical atoms.  This is a serious bug, which potentially affects a lot of subsystems in subtle ways, typically by causing them to fail to explore all possibilities.

### DIFF
--- a/opencog/atoms/base/Handle.cc
+++ b/opencog/atoms/base/Handle.cc
@@ -92,7 +92,7 @@ int Handle::compare(const Handle& h1, const Handle& h2)
 	return 0;
 }
 
-bool Handle::operator< (const Handle& h) const noexcept
+bool Handle::operator<(const Handle& h) const noexcept
 {
 	if (get() == h.get()) return false;
 	if (get() == nullptr) return true;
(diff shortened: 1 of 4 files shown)


ngeiswei:
@linas what about having `operator==` content based as well?

ngeiswei:
@linas what about having `operator==` content based as well?

linas:
> what about having `operator==` content based as well?

I dunno. You'd have to look closely at who the users are, and how they use it.  The most important user being the atomspace itself. 

The faulty `operator<()` wasn't just something I changed on a whim - it was a real bug that was breaking usage of std::set<Handle> In particular, it was screwing up my datasets -- data was being thrown out that should not have been.  Since std::set<Handle> is used all over the place, lord knows how many other subtle strange bugs it was causing.

I don't want to just change operator==() without either some good reason, or a close audit of everywhere it's used, and I don't have time for an audit or bid experiments right now.  I've been hit by major bugs ever two weeks, and its getting exhausting trying to beat them back.

PR #1375, opened 2017-09-19
Fix an issue related to saving the glob groundings

leungmanhin:
Revert https://github.com/opencog/atomspace/pull/1334 and use solution_push/pop for managing the glob states


linas:
Thanks!!! Much better!

PR #1379, opened 2017-09-25
SatisfactionLink evaluation

AmeBel:


linas:
Looks reasonable. I assume that callback gets used whenever `SatisfactionLink`'s get used?

AmeBel:
Yes it is used with [SatisfactionLink](https://github.com/opencog/atomspace/blob/6d2d85d318ee439adfe77808f52f0b3903b731eb/opencog/atoms/execution/EvaluationLink.cc#L433). 

In addition, the logic used in opencog::satisfaction_link, the only place the Satisfier callback is used, implies that the value of `_result` is what matters. Also, `Patternlink::satisfy` used to always return false, even after `Satisfier::grounding` has been called, which made no sense, and there seems no need to explore more if `_result` isn't changed. Of course I only have a feeling of the horn and trunk of the elephant, haven't gotten to the gut yet :smile:

linas:
The fix sounds right. when I wrote this code, I was probably in a hurry or distracted and just failed to notice the return value.  Juggling lots of lines of code at once, this one was dropped.

linas:
It would be nice to unit test this, but I can't think of any easy way to do this, as it only affects performance, and not the answer.

AmeBel:
I think  `Satisfier::search_finished` could be tested when there are no groundings and the pattern has `SEQUENTIAL_AND_LINK` or `SEQUENTIAL_OR_LINK` but even that seems to not make sense b/c such cases seem to be handled in `InitiateSearchCB::no_search`.  Maybe, I will explore that more once I dig deeper.

PR #1380, opened 2017-09-25
Add unit test for multi atomspace scheme bug

ngeiswei:
Utest for issue https://github.com/opencog/opencog/issues/2895

### DIFF
--- a/opencog/guile/SchemeEval.cc
+++ b/opencog/guile/SchemeEval.cc
@@ -547,14 +547,14 @@ void SchemeEval::do_eval(const std::string &expr)
 	per_thread_init();
 
 	// Set global _atomspace variable in the execution environment.
-	AtomSpace* saved_as = NULL;
+	AtomSpace* saved_as = nullptr;
 	if (_atomspace)
 	{
 		saved_as = SchemeSmob::ss_get_env_as("do_eval");
 		if (saved_as != _atomspace)
 			SchemeSmob::ss_set_env_as(_atomspace);
 		else
-			saved_as = NULL;
+			saved_as = nullptr;
 	}
 
 	_input_line += expr;
--- a/opencog/guile/SchemeSmobAS.cc
+++ b/opencog/guile/SchemeSmobAS.cc
@@ -18,7 +18,7 @@ using namespace opencog;
 // Need a lock to protect the map, since multiple threads may be trying
 // to update this map.  The map contains a use-count for the number of
 // threads that are currently using this atomspace as the current
-// attomspace.  When the count drops to zero, the atomspace will be
+// atomspace.  When the count drops to zero, the atomspace will be
 // reaped if the number of SCM references also drops to zero (and the
 // GC runs).
 std::mutex SchemeSmob::as_mtx;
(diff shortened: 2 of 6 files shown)


ngeiswei:
I still don't know how to fix it, other than commenting out

https://github.com/opencog/atomspace/blob/master/opencog/guile/SchemeSmobAS.cc#L76

I observe that guile is calling release_as excessively too often but I don't know how to control that.

linas:
well, `deletable_as` is trying to be a use-count, and if the use-count drops to zero, line 76 deletes it. So either the use-count is not being incremented when it should be, or its getting decremented too often.

There are some strange interactions with guile threads -- its possible that is why the use-count is not incremented.  In guile, thread-local storage is called a "fluid", and fluids do not need to be one-to-one with threads, they can be one-to-many. 

Let me know if you want me to look at it -- otherwise I will continue to ignore this problem.

linas:
According to https://github.com/opencog/atomspace/blob/master/opencog/guile/SchemeSmobAS.cc#L53-L63 the garbage collector is calling the release-atomspace function.  So, I'm guessing that an atomspace gets created in guile, then passed into PLN or the pattern matcher, or somewhere, where it gets used. Meanwhile, the guile variable that held the atomspace goes out of scope, and so the garbage collector thinks that it is not being used any more.  That's just a guess.

If this guess is correct, then there are two solutions: one is to put the atomspace into some guile variable that will not be removed until after PLN is done.  The other (ugly) solution is to add a new special call that says "this atomspace is being used, reserve it and do not free it".   I really really want to avoid this "ugly" solution, because it will leak memory.

ngeiswei:
What puzzles me is that the variable holding the created atomspace never gets out of scope, see https://github.com/opencog/atomspace/blob/master/tests/scm/pm.scm#L10

Maybe `deletable_as` isn't incremented as it should, I'm looking into that...

I agree about avoid memory leaks, I hope the problem doesn't come from guile itself, otherwise, we might be forced to go with the ugly solution.

linas:
Yes but no. I think I see the issue. You think that you declared `post-init-as` and `bug-as` to have global scope. But I think that's not true. Guile looks at it, sees them as variables that are set but never used, and deletes them.

To see if this is true, do this: after the for-loop, at the very end of the program, add the line:
```
(format #t "These are ~A and ~A\n" post-init-as bug-as)
```
This will force both atomspaces to stay in scope until after the print, and so I bet the bug will go away...

linas:
...unless guile is also making assumptions about side effects, and does the print before the loop...

ngeiswei:
Adding the format call at the end doesn't change anything. I don't think guile assumes anything about side effects, so I'm gonna start debugging guile.

linas:
I doubt the bug is in guile per-se, as everything would be broken, because memory is constantly being reclaimed. Rather, we are somehow not using it correctly. Although I am out of ideas, right now.

ngeiswei:
Indeed. According to my understand guile essentially passes `SchemeSmob::free_misc` `GC_register_finalizer_no_order` from http://www.hboehm.info/gc/. I tried to install the last version of boehmgc and recompiled guile but the problem persists, as we'd expect.

I'm out of idea too. Should I go with the ugly memory leak?

ngeiswei:
Well before surrendering I'll try to simplify the unit test, who knows what that might reveal.

Issue #1382, opened 2017-09-25
New atomspace prematurely released by guile when combined with pattern matcher


ngeiswei:
This is a follow up of issue https://github.com/opencog/opencog/issues/2895 now that I've gathered more information.

To reproduce the issue you may run `MultiAtomSpaceUTest::test_pm_scm` under `tests/scm`. Basically it occurs when
1. creating a new atomspace with `(cog-new-atomspace)`,
2. calling the pattern matcher on this new atomspace,
3. calling `(gc)` thereafter.

According to my (possibly flawed) understanding the problem is *not* with how `deleteable_as` counts are kept tracked, but rather how guile prematurely calls `SchemeSmob::free_misc` on the current atomspace.

Maybe the fact that the pattern matcher is involved indicates that guile is somehow confused w.r.t. variable scopes or something. I could observe that calling the gc before the pattern matcher does not produce the bug.

I should also note that setting the initial `deleteable_as` count to a value higher than 1, does not fix the problem, it only postpones it. Whenever the gc is called (after a pattern matcher call), a burst of dozens of calls of `release_as` follow, decreasing the count of `deleteable_as` till it reaches zero and shortly crashes after.

I don't know what to do next with the exception of debugging guile itself to understand why its garbage collector triggers these calls.

ngeiswei:
I ran valgrind on the unit test (disabling all tests but `test_pm_scm`) but I'm only getting notices about Boehm GC. Never, it seems, about opencog. Find the file attached
[valgrind-track-origins.txt](https://github.com/opencog/atomspace/files/1333210/valgrind-track-origins.txt)

ngeiswei:
Well this one mentions opencog
[valgrind-full-track-origins.txt](https://github.com/opencog/atomspace/files/1333238/valgrind-full-track-origins.txt)

ngeiswei:
I can't explain these ones, any idea?
```
==32313== 7,168 bytes in 1 blocks are still reachable in loss record 220 of 222
==32313==    at 0x4C2E19F: operator new(unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==32313==    by 0x6EC9D0B: allocate (new_allocator.h:104)
==32313==    by 0x6EC9D0B: allocate (alloc_traits.h:416)
[...]

PR #1384, opened 2017-09-27
Encrypt/decrypt atomspace smob

ngeiswei:
Possible fix for bug #1382

### DIFF
--- a/opencog/guile/SchemeSmob.h
+++ b/opencog/guile/SchemeSmob.h
@@ -54,6 +54,11 @@ private:
 		COG_EXTEND      // callbacks into C++ code.
 	};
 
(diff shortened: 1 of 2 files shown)


linas:
did it help at all? Based on my last comment, I suspect it did not.

linas:
Oh I see. OK, let me review quickly and commit

linas:
Can you do me a favor? comment out the `WORK_AROUND_GUILE_20_GC_BUG` define in SchemeSmobNew.cc, and see if the encrypt trick fixes that.  I'm guessing that it does, but cannot easily test right now

ngeiswei:
Yes, the encrypt trick fixes `WORK_AROUND_GUILE_20_GC_BUG`.

PR #1385, opened 2017-09-27
Handle TruthValue and ProtoAtoms in define_scheme_primitive

AmeBel:
For those who are adding scheme bindings using define_scheme_primitive,  and want to pass TruthValue or ProtoAtoms as arguments.


linas:
looks reasonable to me.  I'm sort-of surprised that wasn't in there already.

Issue #1387, opened 2017-09-27
Pattern matcher fails if query is in another atomspace than kb


ngeiswei:
The following script highlights the problem
```
(use-modules (opencog query))

;; Define query in the initial atomspace
(define query
(BindLink
  (EvaluationLink
    (PredicateNode "P")
    (VariableNode "$X"))
  (Concept "A"))
)

;; Switch to kb-as
(define kb-as (cog-new-atomspace))
(cog-set-atomspace! kb-as) ;; <--- bug

(EvaluationLink (stv 1 1)
   (PredicateNode "P")
   (Concept "my-arg"))             

(cog-bind query)
```
It returns the empty Set, only returns the solution `(Concept "A")` if the current atomspace isn't set to `kb-as`.

ngeiswei:
@linas I think I know what the problem is, `InitiateSearchCB::neighbor_search` considers the incoming set of `(PredicateNode "P")` in the atomspace of the query instead of `kb-as`. But just to be sure, it's a bug, not a feature, right?

linas:
Its a feature, more-or-less. By declaring `kb-as` to be a new atomspace, instead of a derived one, you've created two distinct atoms that are `(PredicateNode "P")`  These two atoms have two distinct incoming sets, and so the exploration of the connected graph doesn't find them to be attached.

Think of it like this: the atomspace is an implicit part of the name of the atom, so you really have  `(PredicateNode "P in primary-atomspace")`   and  `(PredicateNode "P in other atomspace")`  and now its clear that these really are two different atoms. So of course the patten matcher rejects the second "P" -- its clearly a mismatch.

The above should make clear that this has nothing to do with the pattern matcher - its really about how multiple atomspaces are defined. The pattern miner, and any other block of code will all trip over this same issue.

ngeiswei:
I'm still feeling a bit ambivalent about it, I suppose time will tell if it is a bug or a feature.

However, in the immediate term we want to add some explanation somewhere (if it's not already present), log a warning, etc, right?

ngeiswei:
Suppose atom `C` is in 2 disconnected atomspaces `A1` and `A2`, let `h1` and `h2` be the handles of `C` in `A1` and `A2` respectively. Observe that `A2.get_atom(h1)` returns `h2`.

To be consistent with that, I think it makes sense to allow mining atomspace `A2` with a query from atomspace `A1`.

ngeiswei:
[...]

PR #1389, opened 2017-09-28
Support mining with query from another atomspace

ngeiswei:
Fix for issue #1387

### DIFF
--- a/opencog/atoms/base/Atom.cc
+++ b/opencog/atoms/base/Atom.cc
@@ -84,7 +84,7 @@ namespace opencog {
 
 Atom::~Atom()
 {
-    _atom_space = NULL;
+    _atom_space = nullptr;
     if (0 < getIncomingSetSize()) {
         // This can't ever possibly happen. If it does, then there is
         // some very sick bug with the reference counting that the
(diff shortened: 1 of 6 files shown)


ngeiswei:
@linas in case we come to an agreement that this is a bug, here's a fix.

ngeiswei:
The only relevant commit is https://github.com/opencog/atomspace/pull/1389/commits/4ba17df45494f275dd18d5c6fca6c880461cf291

linas:
Wow. That's such a small change, that it seems "almost harmless". But...

1) As I wrote elsewhere, this is a "feature" not a bug: it is not that "atoms" are universally unique, its that (atomspace, atom) pairs are universally unique.  So although this patch "fixes" the pattern matcher, it does not fix the pattern miner, or any other subsystems that access atoms directly.  Those other subsystems would still see the old behavior.

2) This patch does have a non-trivial overall-negative performance implication.  And it impacts everyone, even people who don't care about this.

3) there is another fix to this problem, that does NOT impact everyone else, but only impacts you:  Create a new temporary child subspace of your search space, and copy your query into it.  This will have the desired result of fixing your bug and of not degrading performance for everyone.  

I think/hope it should be really pretty fast to create such a temp atomspace for performing the query, and in fact, I'm starting to think that we should always do this, automatically, in all cases...

So I'm thinking I like solution 3 above much much better, although it is a bit more work ...

linas:
Let me know if you agree w/ above assessment; I think this is important to get right in the long term; I'm sure this will come back in the future...

ngeiswei:
I'm still feeling ambivalent about it. But you're right about putting the query in a sub-atomspace, it would fix my problem. I'm OK with closing the bug and PR for now.

Issue #1374, opened 2017-09-15
Too low outgoing arity?


ngeiswei:
Outgoing arity is a 16-bit

https://github.com/opencog/atomspace/blob/master/opencog/atoms/base/Atom.h#L86

meaning that if say my pattern matcher query has more than 65K answers, this call

https://github.com/opencog/atomspace/blob/master/opencog/atoms/base/Link.h#L137

will silently get it wrong. There doesn't seem to be any max-arity check at construction either

https://github.com/opencog/atomspace/blob/master/opencog/atoms/base/Link.cc#L58

I know we want to avoid having links with very large outgoing sets, but if permissible it shouldn't break the system.

linas:
OK

PR #1391, opened 2017-10-02
Set Arity type to matches actual outcoming set arity type

ngeiswei:


ngeiswei:
To address issue #1374

Issue #1303, opened 2017-07-12
Consume DontExecLink


ngeiswei:
It seems to me we want `DontExecLink` to be consumed by an execution, just like a `QuoteLink`. For instance, given the atomspace populated with
```
(EvaluationLink (stv 1 1)
   (PredicateNode "URE:BC:and-BIT")
   (BindLink
      (AndLink)
      (Concept "Hello World")))
```
Running
```
(cog-bind
  (BindLink
    (TypedVariableLink
      (VariableNode "$A")
      (TypeNode "BindLink")
    )
    (EvaluationLink
      (PredicateNode "URE:BC:and-BIT")
      (VariableNode "$A")
    )
    (DontExecLink
      (VariableNode "$A")
    )
  )
)
```
produces
```
(SetLink
   (DontExecLink
      (BindLink
         (AndLink
         )
         (ConceptNode "Hello World")
      )
   )
)
```
but, I believe, should produce instead
```
(SetLink
   (BindLink
      (AndLink
      )
      (ConceptNode "Hello World")
   )
)
```

linas:
Yes, that sounds right. I think that was the original intent, but I do not have a clear memory of that any more, and I don't know if it got carefully documented.

linas:
See also #992

ngeiswei:
I'll look into the syntax-case thing. Though probably merely fix `DontExecLink` for now.

ngeiswei:
Actually, after experimenting a bit I would like to remove this feature. The reason being that I need to pass these `BindLinks` around from function calls to function calls and keep treating them as objects. After consuming the first `DontExecLink` the function becomes vulnerable to executions. I tried various tricks like inserting `DontExecLink` in multiple places or adding `QuoteLink` but it doesn't work, so the best way I was is to revert this change (though I won't revert the commit because I want to keep the unit test).

I assume it's OK since it was this in the first place.

linas:
Yeah, I guess so. This is one of those kinds of problems where some short-term arbitrary quick-fix is good enough, but the correct, best long-term answer requires a fair amount of thinking and design.

linas:
Is it still the case that you do not want to consume the DontExecLink ?  Or did this fall by the wayside?

ngeiswei:
Yes I still want to have DontExecLink, or else something else, that do not get consumed.

PR #1392, opened 2017-10-03
Work on the DontExecLink

linas:
PutLink seemed to be mis-handling it. See issue #1303 for discussion.


linas:
Nil, I merged this because it passes unit tests. However, I was wondering if the issue described in #1303 still affects the chainer. It seems to me like DontExec should be consumed, but if it breaks the chainer code, I don't quite care that much, right now.

ngeiswei:
Cool, yes it would definitely break my work on inference control learning (which is probably badly unit test covered ATM).

PR #1402, opened 2017-10-08
Cache results of pattern matching.

linas:
This provides a more general, more formal solution for Aemn's recent openpsi enhancements.

Specifically, the most recent result of running any PatternLink (BindLink, GetLink, etc.) can be cached with the PatternLink itself, and be recalled later.  This is particularly useful with the SatisfactionLink, which returns only a truth value; but the variable groundings that lead to the result were previously tossed.  Now, they're saved.... 

The caching is generic -- the values are stored as Values anchored to the atoms (and thus can be saved in the database).  It might make sense to store more of the various bits and pieces of some of the C++ structs in this way -- e.g. the bound and free variables in a LambdaLink could be stored as values on the LambdaLink.


leungmanhin:
thanks a lot!

linas:
welcome!  See also example at /examples/pattern-matcher/satisfaction.scm to see how it works.

leungmanhin:
yea I tried the example too, though there was one unexpected problem when I turned the `(Concept "funny")` into a variable... seems that it has something to do with whether the results (the `ListLink` created if there are more than one variables) are in the atomspace or not, please review https://github.com/opencog/atomspace/pull/1403

PR #1403, opened 2017-10-09
Place the SatisfactionLink results in the atomspace

leungmanhin:


linas:
Yeah, I'm finding it kind-of annoying that things don't work if we don't place all atoms into the atomspace.  It would be nice if "temporary" atoms, temporary results, did not have to be put into the atomspace.  Fixing this might require some fairly complex work on the guts of the evaluator.

linas:
Yeah, I'm finding it kind-of annoying that things don't work if we don't place all atoms into the atomspace.  It would be nice if "temporary" atoms, temporary results, did not have to be put into the atomspace.  Fixing this might require some fairly complex work on the guts of the evaluator.

PR #1409, opened 2017-10-19
Old Cleanup

rTreutlein:


rTreutlein:
Well Jenkins seems to fail due to some unrelated reason so ill just merge it as it is.

PR #1416, opened 2017-10-24
Centralize the atom factories into one location.

linas:
The atom factories were sprinkled haphazardly throughout the code. Move them all into one central location.


ngeiswei:
:+1:

PR #1418, opened 2017-10-25
Caml to snake

ngeiswei:
Yet another futile nazy syntax nonsense of mine.

Sorry I can't help but being bothered by the arbitrary mixture of caml and snake case. The standard does indicate a preference for snake case for methods (and caml case for types). It is known that snake case is easier to read than caml case so it makes sense to be the predominant style. And I needed some mindless task to contrast with my deep thinking tasks from the past days.


linas:
And here I thought you were adding CamL bindings, like Haskell was not enough.

PR #1419, opened 2017-10-26
Update value example

AmeBel:
Also update the string representation of values. Didn't get why so much padding is required for FloatValues so not touched.

### DIFF
--- a/examples/guile/values.scm
+++ b/examples/guile/values.scm
@@ -11,35 +11,33 @@
 (use-modules (opencog))
 
 ; Values can store vectors of floats ...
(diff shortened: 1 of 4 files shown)


linas:
Thanks. I just expanded that example to include TV's and AV's: 16bd50604deb78f5534cf9e0cd02591b4a3651b0

linas:
oh, and attention values too:  e1d1319ee6b1c9f6540ecad2a213283c561d927c

AmeBel:
More simplifications cool, thanks for pointing that out.

PR #1434, opened 2017-11-11
fix fail build on reference error

zhangyan612:
the reference error caused build fail, I corrected the path


ngeiswei:
Oh I see, according to https://github.com/opencog/atomspace/blob/master/opencog/atoms/execution/CMakeLists.txt#L47 I was able to compile because I wrote that code after installing.

Thanks for the catch!

PR #1435, opened 2017-11-11
build tweak: missing hashbang

WillemMali:
I fixed the MapLink #include and a warning related to Debian running the Haskell build.sh script with the sh interpreter while the scrip expected bash (due to the use of the [[ program), but it seems like the include was fixed while I was preparing this PR.


linas:
thanks!

PR #1438, opened 2017-11-17
Miscs

ngeiswei:


opencog/atoms/base/Handle.h
```diff
@@ -180,7 +180,7 @@ typedef std::set<Handle> HandleSet;
 typedef std::unordered_set<Handle> UnorderedHandleSet;
 
 //! an ordered map from Handle to Handle set
-typedef std::map<Handle, UnorderedHandleSet> HandleMultimap;
+typedef std::map<Handle, HandleSet> HandleMultimap;
```
linas: Hye Nil, be super-duper careful changing this kind of stuff.  There are parts of the code where are very very very senstive to the structure used -- RAM usage doubles with some of these structures, CPU usage triples with others.  The wrong choices for some of these make the atomspace totally unusable, and a complete performance disaster.

I *think* I carefully protected those parts from this kind of damage, but I'm not sure.  So this change might be harmless, but I'm not sure. Just be aware that parts like this are delicate.

ngeiswei: Note taken, Linas. Just so you know, this data type wasn't used anywhere at all, so that is why I allowed myself to make the change (and it more consistent with the surrounding naming conventions).

linas: OK, thanks.  Long-term someday, there should be performance acceptance tests ("I just loaded 1 million atoms, how long did that take, and how much RAM was needed?")  but that's for some future day.

Issue #1428, opened 2017-11-08
The Pattern Matcher checks for atoms in a child atomspace when queried from parent atomspace


Yidnekachew:
Constant clauses in a child atomspace are expected not to be accessible from the parent atomspace. But, when tested with this scenario, a query run in a parent atomspace checks for constant clauses in the child atomspace. 

```
(define parent-as (cog-atomspace))

(define derived-as (cog-new-atomspace parent-as))

(cog-set-atomspace! derived-as)

(define rule
    (Bind
        (Inheritance (Concept "a") (Concept "b"))

        (Concept "b")))

(cog-set-atomspace! parent-as)

(cog-bind rule)
```
evaluates to
```
$1 = (SetLink
   (ConceptNode "b")
)
```
when the expected result is an empty `(SetLink )`

linas:
You've defined the rule itself in the child atomspace!!

I agree that the behavior here is unexpected.    There are several options:

1) Make `cog-bind` check to see if the rule is defined in the current atomspace, and, if it is not, refuse to work - throw an error message.  (By the way, you should be using `cog-execute`, not `cog-bind`)

2) Make `cog-bind` check to see if the rule is defined in the current atomspace, and, if it is not, copy the rule into the current atomspace.

3) don't change the current behavior.

4) verify that the test-pattern exists in the current atomspace.

I don't like option 2 at all. It's just asking for trouble, it seems.  Option 1 is .. flawed in complicated ways: consistent error checking has some fairly deep side-effects.  You don't like option 3.   I guess you want option 4.

The biggest problem with option 4 is that it forces an additional check to be performed for *every* match, which penalizes everyone, always, for what seems to be a rare and unusual use case.

What is this for? Does this really affect anything? Is there a "real world" use case for this?

linas:
There is also another strange, interesting effect:
```

(use-modules (opencog) (opencog exec))

(define parent-as (cog-atomspace))

(define derived-as (cog-new-atomspace parent-as))

(cog-set-atomspace! derived-as)

(define rule
    (Bind
        (Inheritance (Variable "x") (Concept "b"))

        (Concept "b")))

(cog-set-atomspace! parent-as)

(cog-execute! rule)

(Inheritance (Concept "a") (Concept "b"))

(cog-execute! rule)
```

[...]

PR #1444, opened 2017-11-24
Remove constants right before pattern matching

Yidnekachew:
Implemented as a fix to issue #1428

### DIFF
--- a/opencog/atoms/pattern/BindLink.h
+++ b/opencog/atoms/pattern/BindLink.h
@@ -52,7 +52,8 @@ public:
 	BindLink(const Handle& body, const Handle& rewrite);
 	explicit BindLink(const Link &l);
 
-	bool imply(PatternMatchCallback&, bool check_connectivity=true);
+	bool imply(PatternMatchCallback&, AtomSpace* as,
+	           bool check_connectivity=false);
 	const Handle& get_implicand(void) { return _implicand; }
 
 	// Return the cached implication results
--- a/opencog/atoms/pattern/PatternUtils.cc
+++ b/opencog/atoms/pattern/PatternUtils.cc
@@ -22,6 +22,8 @@
  */
 
 #include <opencog/atomutils/FindUtils.h>
+#include <opencog/atomspace/AtomSpace.h>
+#include <boost/range/algorithm/find.hpp>
 #include "PatternUtils.h"
 
 using namespace opencog;
@@ -29,7 +31,8 @@ using namespace opencog;
 namespace opencog {
 
 /**
- * Remove constant clauses from the list of clauses.
+ * Remove constant clauses from the list of clauses if they are in
+ * the queried atomspace.
  *
  * Make sure that every clause contains at least one variable;
  * if not, remove the clause from the list of clauses.
@@ -59,20 +62,38 @@ namespace opencog {
  * Returns true if the list of clauses was modified, else returns false.
  */
 bool remove_constants(const HandleSet &vars,
-                      HandleSeq &clauses,
-                      HandleSeq &constants)
+                      Pattern &pat,
+                      HandleSeqSeq &components,
+                      const AtomSpace &queried_as)
 {
 	bool modified = false;
 
 	// Caution: this loop modifies the clauses list!
 	HandleSeq::iterator i;
-	for (i = clauses.begin(); i != clauses.end(); )
+	for (i = pat.clauses.begin(); i != pat.clauses.end();)
 	{
 		Handle clause(*i);
-		if (is_constant(vars, clause))
+
+		if (is_constant(vars, clause) && is_in_atomspace(clause, queried_as))
 		{
-			constants.emplace_back(clause);
-			i = clauses.erase(i);
+			pat.constants.emplace_back(clause);
+			i = pat.clauses.erase(i);
+
+			// remove the clause from _components.
+			auto j = boost::find(components, HandleSeq{clause});
+			if (j != components.end())
+				components.erase(j);
+
+			// remove the clause from _pattern_mandatory.
+			auto m = boost::find(pat.mandatory, clause);
+			if (m != pat.mandatory.end())
+				pat.mandatory.erase(m);
+
+			// remove the clause from _cnf_clauses.
+			auto c = boost::find(pat.cnf_clauses, clause);
+			if (c != pat.cnf_clauses.end())
+				pat.cnf_clauses.erase(c);
+
 			modified = true;
 		}
 		else
@@ -82,6 +103,11 @@ bool remove_constants(const HandleSet &vars,
 	}
 
 	return modified;
+}
+
+bool is_in_atomspace(const Handle& handle, const AtomSpace& atomspace)
+{
+	return (bool)atomspace.get_atom(handle);
 }
 
 bool is_constant(const HandleSet& vars, const Handle& clause)
--- a/opencog/query/PatternMatch.cc
+++ b/opencog/query/PatternMatch.cc
@@ -24,15 +24,12 @@
 #include <opencog/util/Logger.h>
 
 #include <opencog/atoms/pattern/BindLink.h>
-#include <opencog/atoms/pattern/PatternLink.h>
 
 #include <opencog/atomspace/AtomSpace.h>
-#include <opencog/atoms/base/ClassServer.h>
 #include <opencog/atoms/pattern/PatternUtils.h>
 
 #include "PatternMatch.h"
 #include "PatternMatchEngine.h"
-#include "PatternMatchCallback.h"
 #include "DefaultPatternMatchCB.h"
 
 using namespace opencog;
@@ -354,6 +351,7 @@ bool BindLink::imply(PatternMatchCallback& pmc, AtomSpace* as, bool check_conn)
 		}
 
 		_num_comps = _components.size();
+		make_connectivity_map(_pat.cnf_clauses);
 	}
 
 	return PatternLink::satisfy(pmc);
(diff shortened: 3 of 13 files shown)


opencog/atoms/pattern/BindLink.h
```diff
@@ -52,7 +52,8 @@ class BindLink : public PatternLink
 	BindLink(const Handle& body, const Handle& rewrite);
 	explicit BindLink(const Link &l);
 
-	bool imply(PatternMatchCallback&, bool check_connectivity=true);
+	bool imply(PatternMatchCallback&, AtomSpace* as,
+	           bool check_connectivity=false);
 	const Handle& get_implicand(void) { return _implicand; }
```
linas: why change the default connectivity check?  The connectivity check is intended to avoid combinatorial explosions - its a performance issue.

opencog/atoms/pattern/PatternUtils.cc
```diff
@@ -22,14 +22,17 @@
  */
 
 #include <opencog/atomutils/FindUtils.h>
+#include <opencog/atomspace/AtomSpace.h>
+#include <boost/range/algorithm/find.hpp>
```
linas: Gahh, can we avoid using boost, if possible? Its a bunch of nasty, poorly designed code, I'm tired of having to debug the issues that it causes.

opencog/query/PatternMatch.cc
```diff
@@ -328,13 +329,33 @@ bool PatternMatch::recursive_virtual(PatternMatchCallback& cb,
...
 			"BindLink consists of multiple disconnected components!");
 			// PatternLink::check_connectivity(_comps);
 
+	// Make sure that the user did not pass in bogus clauses
+	// in the queried atomspace.
+	// Make sure that every clause contains at least one variable.
+	// The presence of constant clauses will mess up the current
+	// pattern matcher.  Constant clauses are "trivial" to match,
+	// and so its pointless to even send them through the system.
+	bool bogus = remove_constants(_varlist.varset, _pat, _components, as);
```
linas: well, you've moved this code out of the compiled code path, and into the executed code path.  Now, instead of checking for constants just once, when the pattern is created, you are checking them every time that the pattern is executed -- and since a pattern can be re-used thousands of times, that's a thousand-fold impact on performance.

linas:
This code looks very clean and nice, except for one primary issue: it moves the constant-clause removal step from the compile code path and into the execute code path, and I don't quite see why that is needed, or how that avoids the issues in #1428.  Wasn't removing the constant clauses sufficient?

opencog/atoms/pattern/BindLink.h
ngeiswei: It was already set to false, `bindlink` the main entry point for using the pattern matcher sets it to false https://github.com/opencog/atomspace/blob/master/opencog/query/Implicator.cc#L90

the only places it wasn't set to false were a couple unit tests (failing as a result of these changes). If you prefer we can revert that and set to false only for the couple of failing unit tests, we just thought it makes more sense to set this one to false since again, for all practical purpose, it was already set to false.

opencog/query/PatternMatch.cc
ngeiswei: Yes, I know, but you *can't* check whether a constant is actually present in the to-be-queried atomspace at compile time. Besides, seriously, how much performance overhead is it gonna have, +0.001%? it's still insignificant compared to the search itself.

opencog/atoms/pattern/PatternUtils.cc
ngeiswei: We can, but so you know boost.range is extremely mature, I'm been using it for years, never had a problem once (unlike boost.graph for instance). BTW, range is coming to the standard https://en.wikipedia.org/wiki/C%2B%2B20 BTW but it's still 3 to 4 years away.

ngeiswei:
Having the URE not remove the constant clause is not enough since the pattern matcher would then remove it. One way to solve our problem without this would be to basically move this check to the URE and not allow it to call the pattern matcher if it fails. I can't help but to think that it belongs to the pattern matcher though.

Perhaps we need to benchmark, maybe it's not negligible, if so, maybe we want to keep this compile time checking optional, for instance the scheme interface could do compile-time constant checking, and the URE run-time constant checking?

Actually, the various PLN examples can provide good benchmark as the URE often generates queries with constant clauses, if it has no impact there, then I think it would be settled.

opencog/query/PatternMatch.cc
linas: Eh? when I last measured, the search was fast, and the building of the bindlinks were slow. There's a lot of overhead and calculation to construct a bind-link (by design).  Of course, it's data-dependent, but the actual execution of a bindlink should be fast.  Normally, there's not much to do.

linas:
> the various PLN examples can provide good benchmark as the URE often generates queries with constant clauses, if it has no impact there, then I think it would be settled.

OK.  It would be nice to have a generic benchmark that measures bindlink creation time and bindlink execution time. and have that in the benchmark directory, where it could be re-run every now and then, without fear of loosing reulsts because someone changed the example code.

linas:
So what's the upshot here? Merge or not merge?

ngeiswei:
Don't merge yet, it introduces other bugs somewhere else.

Yidnekachew:
I've tried to run `MOSESPLNSynergyTest` twice with 100 runs in each and I found out the following result.

First Run:
Mean = `17.48 sec`, SD = `1.03` with out this change and
Mean = `17.06 sec`, SD = `1.93` with this change.

Second Run:
Mean = `17.30 sec`, SD = `1.08` with out this change and
Mean = `17.49 sec`, SD = `1.01` with this change .

ngeiswei:
@linas I think using MOSESPLNSynergyUTest as benchmark is not bad because it defines its rules once and re-use them multiple times (see https://github.com/opencog/opencog/blob/master/examples/pln/moses-pln-synergy/scm/moses-pln-synergy-pm.scm), unlike examples using the URE that most queries generated by the URE are run only once (thus having the constant clause check shouldn't make a difference).

Ideally we would manufacture a benchmark that can exhibit a possible bottleneck, on the other hand I think it shows that in practice it doesn't have any measurable impact.

All that to say that I think it can be merged.

linas:
@ngeiswei  8 days ago, you wrote:

> Don't merge yet, it introduces other bugs elsewhere.

But you don't indicate if these problems were fixed yet, or not?

ngeiswei:
Yes, they have been fixed.

ngeiswei:
Apologies, this PR is breaking a few unit tests, I'm working on it ATM.

linas:
Can someone explain, again, why removing constant clauses at compile time is wrong?  This code path is triggered in bug #1598 and I'm trying to figure out why.

I scanned the comments above: there is one claim about constant clauses and subspaces that cannot possibly be correct. There is also a suggestion that the URE frequently generates patterns with constant clauses.  But so what? Sooner or later those constant clauses get removed, anyway, so why not remove them earlier? Why wait?

ngeiswei:
This unit test explains it all I think https://github.com/opencog/atomspace/pull/1444/commits/424a24517108f1624bb851e0ac52c417515e7c22#diff-0e30d4a40eda8351af942b35f1f2db77R380

To restate, if the query is in a sub-atomspace and contains constant clauses not present in the queried atomspace, we then we don't want to consider these constant clauses as matches.

linas:
Wouldn't the removal of the constant clauses solve this problem?  Remove them from the query entirely ... then they can't affect any atomspace, right?

ngeiswei:
I suppose what you mean is: not building queries with constant terms that are not in the query atomspace to begin with.

Yes, it would solve the problem. But when we were at this cross road, I felt it's better to fool prove the pattern matcher than the pattern matcher users because users, in particular humans, would expect the pattern matcher to correctly handle whatever they throw at it.

linas:
I just don't get it. In the beginning, the pattern matcher removed constant clauses. Then this was undone for some reason I could not understand, but it seemed harmless so I let it slide. But then this bug showed up as a result, and so now we keep constant clauses, but later on remove them, so that we can keep them elsewhere, and finally remove them at runtime -- all of this for reasons I still don't understand.  

I'm all for the pattern matcher doing "the right thing", but how are users supposed to be able to understand this, keep/remove/keep/remove dance? What does it accomplish, what is the point?

You understand that C++ compilers do similar things: move constants out of loops, and replace all occurrences of (for example) 2*3 by 6 -- they perform this multiplication at compile time, not at run time.  A constant clause is just like a constant inside of a loop -- it does nothing other than to slow down the loop.

ngeiswei:
1. Let `S` be some atomspace,
2. let `C` be some atomspace child of `S`,
3. let `A` be some closed atom in `C`,
4. let `Q` be the following query in `C`,
```
(Bind (And (Inheritance (Variable "$X") (Concept "cat")) A) (Concept "Eureka"))
```
5. let `S` contain `(Inheritance (Concept "kitty") (Concept "cat"))`,
6. assume that `S` is the current atomspace.

Should running `Q` return `(Set (Concept "Eureka"))` or `(Set)`?

Since `A` is not in `S` I argue that it should return the empty set, not Eureka. Do you agree?

ngeiswei:
I've updated the message above to have a simpler query.

misgeatgit:
@ngeiswei  if Q is in C and S is current atomspace, Shouldn't Q be inaccessible from S?  Otherwise, if Q is accessible, A should too.

ngeiswei:
@misgeatgit good question. I guess it comes down to what we allow the pattern matcher to see or not. I find it very convenient to be able to query an atomspace without having to pollute it with the query. In fact I think one should be able to have the query in no atomspace at all, Linas disagree on the ground that 2 atoms in 2 separate atomspaces should not be considered to match.

PR #1447, opened 2017-11-29
Fix misc

ngeiswei:
Some misc fixes for the experimental pattern miner

### DIFF
--- a/opencog/atoms/base/Handle.cc
+++ b/opencog/atoms/base/Handle.cc
@@ -215,6 +215,30 @@ std::string hps_to_string(const HandlePairSeq& hps)
 	}
 	return ss.str();
 }
+std::string hc_to_string(const HandleCounter& hc)
+{
+	std::stringstream ss;
+	ss << "size = " << hc.size() << std::endl;
+	size_t i = 0;
+	for (const auto el : hc) {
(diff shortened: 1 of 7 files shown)


opencog/atoms/base/Handle.cc
```diff
@@ -215,6 +215,30 @@ std::string hps_to_string(const HandlePairSeq& hps)
...
+	ss << "size = " << hc.size() << std::endl;
+	size_t i = 0;
+	for (const auto el : hc) {
+		ss << "atom[" << i << "]:" << std::endl << oc_to_string(el.first)
+		   << "num[" << i << "]:" << el.second << std::endl;
+		i++;
+	}
+	return ss.str();
+}
+std::string huc_to_string(const HandleUCounter& huc)
```
linas: do we really need both signed and unsigned versions? That strikes me as cheezy laziness. Pick one, pick the other, don't do both.

linas: anyway, we have count truth values for counting.  And other kinds of numeric values.  I bet that it probably makes more sense to use those, instead of using this.

ngeiswei: Regarding int vs unsigned int, I tend to use unsigned when I want to indicate to the developer the domain of the integer type more than anything else. I agree about count truth value, I might use that.

Issue #1446, opened 2017-11-29
content_based_handle_less isn't correct over scope links not in atomspace


ngeiswei:
See unit test added in https://github.com/opencog/atomspace/commit/8c7ca0f7d2f92bf5941b2b7e3f93226cbb5cbb65

@linas could you confirm that you see this as a bug as well?

mjsduncan:
did you update the unit test yet?
```
...
[2017-11-29 16:54:38:136] [INFO] BEGIN TEST: test_content_less

In ScopeLinkUTest::test_content_less:
/home/mjsd/oc/atomspace/build/tests/atoms/../../../tests/atoms/ScopeLinkUTest.cxxtest:188: Error: Assertion failed: not lt(scY, scX)
Failed 1 and Skipped 0 of 6 tests
```

ngeiswei:
I have updated the unit test with what I believe to be a bug. I have committed a fix #1448 but I'm waiting for Linas feedback.

linas:
Yes, the unit test addition in 8c7ca0f seems like it should pass.

linas:
I think this bug can be closed.

PR #1448, opened 2017-11-29
Fix atom content less

ngeiswei:
Possible fix for #1446 

@linas I don't know if that the best way to fix it, it certainly seems like the simplest, but instead I could have overloaded operator< for ScopeLink. I think this fix is more generic but might have some unwanted performance overhead. Let me know what you think.


linas:
Yes, this looks like exactly the right fix.

linas:
I don't know if content_less is performance critical or not.  I guess it is used to build c++ sets and maps...

PR #1449, opened 2017-12-01
Add ScopeLink::partial_subtitute

ngeiswei:
Similar to `LambdaLink::substitute` but produce a partially substituted scope link instead of a fully substited body.

I have decided to have this method in `ScopeLink` instead of `LambdaLink` because not all scope links inherits `LambdaLink`.

### DIFF
--- a/opencog/atoms/base/Handle.cc
+++ b/opencog/atoms/base/Handle.cc
@@ -100,6 +100,19 @@ bool Handle::operator<(const Handle& h) const noexcept
 	return get()->operator<(*h);
 }
 
+bool content_eq(const HandleSeq& lhs, const HandleSeq& rhs)
+{
+	if (lhs.size() != rhs.size())
+		return false;
+
+	auto lit = lhs.begin(), rit = rhs.begin();
+	for (; rit != rhs.end(); ++lit, ++rit)
+		if (not content_eq(*lit, *rit))
+			return false;
+	return true;
+
+}
+
 bool content_eq(const HandleSet& lhs, const HandleSet& rhs)
 {
 	if (lhs.size() != rhs.size())
--- a/opencog/atoms/base/Handle.h
+++ b/opencog/atoms/base/Handle.h
@@ -203,6 +203,9 @@ typedef Counter<Handle, unsigned> HandleUCounter;
 //! a handle iterator
 typedef std::iterator<std::forward_iterator_tag, Handle> HandleIterator;
 
+bool content_eq(const opencog::HandleSeq& lhs,
+                const opencog::HandleSeq& rhs);
+
 bool content_eq(const opencog::HandleSet& lhs,
                 const opencog::HandleSet& rhs);
 
--- a/opencog/atoms/core/ScopeLink.cc
+++ b/opencog/atoms/core/ScopeLink.cc
@@ -29,6 +29,7 @@
 #include <opencog/atoms/base/ClassServer.h>
 #include <opencog/atoms/core/LambdaLink.h>
 #include <opencog/atoms/core/TypeNode.h>
+#include <opencog/atomutils/TypeUtils.h>
 
 #include "ScopeLink.h"
 
@@ -364,12 +365,14 @@ inline HandleSeq append_rand_str(const HandleSeq& vars)
 	return new_vars;
 }
 
-Handle ScopeLink::alpha_conversion(HandleSeq vars) const
+Handle ScopeLink::alpha_conversion() const
 {
-	// If hs is empty then generate new variable names
-	if (vars.empty())
-		vars = append_rand_str(_varlist.varseq);
+	HandleSeq vars = append_rand_str(_varlist.varseq);
+	return alpha_conversion(vars);
+}
 
+Handle ScopeLink::alpha_conversion(const HandleSeq& vars) const
+{
 	// Perform alpha conversion
 	HandleSeq hs;
 	for (size_t i = 0; i < get_arity(); ++i)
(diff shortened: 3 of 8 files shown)


linas:
I'm not entirely thrilled by this.   For many kinds of scope links, beta reduction is a fundamentally illegal act: you are not allowed to substitute in a for-all, there-exists, or any integral, which do inherit from scope.

I have two suggestions:

1) make IMPLICATION_SCOPE_LINK inherit from LambdaLink, and not from ScopeLink, since apparently, you want lambda and not scopiing.

2) get rid of IMPLICATION_SCOPE_LINK, and merge all of the function it supports back into the ordinary ImplicationLink (and make ImplicationLink inherit from LambdaLink).  I like this best, because I see no reason at all for why an ImplicationLink with one or more variables should be treated differently from an ImplicationLink with zero variables.

ngeiswei:
On 12/01/2017 10:00 PM, Linas Vepštas wrote:
> I'm not entirely thrilled by this. For many kinds of scope links, beta 
> reduction is a fundamentally illegal act: you are not allowed to 
> substitute in a for-all, there-exists, or any integral, which do inherit 
> from scope.

Fundamentally illegal? Can you justify that?

Illegal substitution in ForAll? What about universal instantiation? Etc.

> 
> I have two suggestions:
> 
>  1.
> 
>     make IMPLICATION_SCOPE_LINK inherit from LambdaLink, and not from
>     ScopeLink, since apparently, you want lambda and not scopiing.

I have no problem with that. Then ScopeLink will simply be useless, and 
we can remove it. Which fundamental difference do you think exist 
between Lambda and Scope? Except for the fact that substitution would be 
forbidden on a scope and not forbidden on a lambda?

>  2.
> 
>     get rid of IMPLICATION_SCOPE_LINK, and merge all of the function it
>     supports back into the ordinary ImplicationLink (and make
>     ImplicationLink inherit from LambdaLink). I like this best, because
>     I see no reason at all for why an ImplicationLink with one or more
>     variables should be treated differently from an ImplicationLink with
>     zero variables.

ImplicationScopeLink
   X
   P(X)
   Q(X)

is merely sugar for

ImplicationLink
   Lambda X P(X)
   Lambda X Q(X)

as explained here

https://wiki.opencog.org/w/ImplicationScopeLink

linas:
On Mon, Dec 4, 2017 at 12:12 AM, ngeiswei <notifications@github.com> wrote:

> On 12/01/2017 10:00 PM, Linas Vepštas wrote:
> > I'm not entirely thrilled by this. For many kinds of scope links, beta
> > reduction is a fundamentally illegal act: you are not allowed to
> > substitute in a for-all, there-exists, or any integral, which do inherit
> > from scope.
>
> Fundamentally illegal? Can you justify that?
>
> Illegal substitution in ForAll? What about universal instantiation? Etc.
>

I don't know how to respond to that.  Lets try beta reduction on an
integral.
```
\int dx \sin(x)
```
substitute x=52 to get
```
\int  d52 sin(52)
```
 That's not even syntactically valid.

Let's try another.

```
there exists x such that x<10
```
substitute x=52 to get

```
there exists 52 such that 52<10
```
 I don't know of any kind of "52" that is less than 10.

Both of these are nonsense, syntactically invalid expressions.  Bta
reduction is not defined in integral calculus, in set theory, or in
first-order logic. The only place where beta reduction is defined is in
lambda calculus, and the assorted comp-sci systems that have lambda-like
properties.




>
> >
> > I have two suggestions:
> >
> > 1.
> >
> > make IMPLICATION_SCOPE_LINK inherit from LambdaLink, and not from
> > ScopeLink, since apparently, you want lambda and not scopiing.
>
> I have no problem with that. Then ScopeLink will simply be useless, and
> we can remove it. Which fundamental difference do you think exist
> between Lambda and Scope? Except for the fact that substitution would be
> forbidden on a scope and not forbidden on a lambda?
>

Scoping is not the same as lambda.

https://en.wikipedia.org/wiki/Scope_(computer_science)

Pick any example whatsoever from that article: you cannot legally
beta-substitute in any of the examples.


> > 2.
> >
> > get rid of IMPLICATION_SCOPE_LINK, and merge all of the function it
> > supports back into the ordinary ImplicationLink (and make
> > ImplicationLink inherit from LambdaLink). I like this best, because
> > I see no reason at all for why an ImplicationLink with one or more
> > variables should be treated differently from an ImplicationLink with
> > zero variables.
>
> ImplicationScopeLink
> X
> P(X)
> Q(X)
>
> is merely sugar for
>
> ImplicationLink
> Lambda X P(X)
> Lambda X Q(X)
>
> as explained here
>
> https://wiki.opencog.org/w/ImplicationScopeLink


Right. I am suggesting that we get rid of ImplicationScopeLink, and just
add that as the default behavior for ImplicationLink.

There should not be a difference between having zero arguments and having
one argument.

That is, writing

ImplicationScopeLink
  (empty-set)
   P(empty-set)
   Q(empty-set)

is merely anti-sugar for writing

Implication
  P
  Q

ngeiswei:
OK, I agree, not all scopes are lambdas. Though after reading that I'm let thinking that the scoped implication should still inherit `ScopeLink` rather than `LambdaLink`, because it's not a straight lambda, there are extra subtleties (like the fact that substitution is conditioned by the antecedent).

Regarding
```
ImplicationScopeLink
  (empty-set)
  P(empty-set)
  Q(empty-set) 
```
I am not sure it should be allowed, I guess it could be if the empty-set is considered to be its domain, but whether it is allowed or, it is definitely **not** equal to
```
Implication
  P
  Q
```
instead it is equal to
```
Implication
  Lambda (empty-set) P(empty-set)
  Lambda (empty-set) Q(empty-set)
```
So how you define `Lambda (empty-set) P(empty-set)` is perhaps up to debate. We could say that it corresponds to a constant function. If so, assuming `P` is true for this empty-set, and `Q` is false for this empty-set (BTW, let's use empty-list, `List`, instead, as this would actually be valid, since arguments are ordered), then
```
ImplicationScope (stv 1 <tbd>)
  (empty-list)
  P(empty-list)
  Q(empty-list)
```
will be true. On the other hand `Q` might be true half of the time when `P` is, over the remaining inputs, and so
```
Implication (stv 0.5 <tbd>)
  P
  Q
```
See what I mean?

linas:
> there are extra subtleties (like the fact that substitution is conditioned by the antecedent)

I don't know what that means.

>  it is definitely not equal to

How could it not be? Isn't it just tautologically exactly the same thing? What else could it possibly be? 

> over the remaining inputs 

What remaining inputs??  When one writes
```
Implication 
  P
  Q
```
there are no inputs, and no inputs means ... no inputs. There are no variables. 

Actually, now that I think about it, I have absolutely no clue at all what 
```
ImplicationLink 
     Lambda X P(X) 
    Lambda X Q(X)
```
means. We've been around this block before, but I still don't know, formally, what that collection of symbols means.   Thanks to alpha conversion, the only possible way I can read the above is that it means

```
ImplicationLink 
     Lambda X P(X) 
    Lambda Y Q(Y)
```

which is vapid to the point of being meaningless.  So, basically, I just have no clue what you mean whenever you say ImplicationScopeLink.

ngeiswei:
Given 2 predicates `P` and `Q`, with identical domain `D`
```
Implication <TV>
  P
  Q
```
means `prob(Q|P)` where `TV` represents the second order probability of that probability (i.e. taking into account uncertainty).

So for instance if `D` is the integer interval [1..4], `P` is valid (has truth value equal to 1) for of all numbers in `D` and `Q` is valid for all even numbers in `D`, then, in the implication above, TV equals to
```
(stv 0.5 <some low confidence>)
```

If you don't have `P` and `Q` already predefined, you may want to build one using `Lambda` (or other high level constructs), so for instance you may write

```
Implication
  P
  Lambda X (is_divisible X 2)
```

re-using the `is_divisible` predicate to define on the fly the predicate `even`.

Now suppose that you want to define on the fly (anonymously that is) both `P` and `Q`, you'll write something like

```
Implication
  Lambda X blahblah
  Lambda Y blueblue
```
Of course I used X in the antecedent and Y in the consequent, but it doesn't matter, it's also equivalent to the following
```
Implication
  Lambda X blahblah
  Lambda X blueblue
```
right?

So now I can sugar rewrite that into
```
ImplicationScope
   X
   blahblah
   blueblue
```

ngeiswei:
```
Nil: there are extra subtleties (like the fact that substitution is conditioned by the antecedent)
Linas: I don't know what that means.
```
well, as you point out you can directly beta-reduce `Exists x P(x)`, it's not logically valid, you need first to for instance convert it into a forall like `Not ForAll x Not P(x)`, then you can "blindly" beta-reduce the for all part. So that one subtlety. For `ImplicationScope` you need conditional instantiation, that's another subtlety. You can't just substitute `ImplicationScope X P(X) Q(X)` no matter what, you can only if P is true for that particular term to be substituted, for instance is `P(a)` is true, then yes you can produce `Q(a)` (with a certain TV determined by the TV on the `ImplicationScope` and `P(a)`).

ngeiswei:
sorry, forgot a **not** in my post above, I've edited it.

linas:
> So now I can sugar rewrite that into

Wow. OK, I understood that.  For the first time ever.  Can you add that to the wiki page for `ImplicationScope` ?  Yes, I now agree, Implication and ImplicationScope are really very different things.

linas:
> then you can "blindly" beta-reduce the for all part. 
Only if you assume a certain rule of inference that allows this. 
https://en.wikipedia.org/wiki/Universal_instantiation
https://en.wikipedia.org/wiki/List_of_rules_of_inference

I'd rather like to avoid assuming any kinds of rules of inference in the C++ code, as often as possible.

linas:
I don't understand why you didn't make ImplicationScope inherit from Lambda ... and why you just didn't call it ImplicationLambda.

Another way to slice this is that there could be a new SubstituteLink which inherits from ScopeLink.  The Lambda inherits from Substitute, and ImplicationScope inherits from Substitute.   That would make it clear that ScopeLink only designates variables. Substitution is distinct from designation, and can only be used on the substitutable kinds of things.

ngeiswei:
I'm OK with `ImplicationLambdaLink`, I like that it reflects part of its sugar syntax nature.

Issue #1458, opened 2017-12-06
Rename ImplicationScope to ImplicationLambda, etc


ngeiswei:
As discussed in #1449 
1. `ImplicationScopeLink` and such should inherit from `LambdaLink`.
2. `ImplicationScopeLink` and such should be renamed `ImplicationLambdaLink` and such.
3. `ScopeLink::partial_substitute` and such should be moved to `LambdaLink`.

ngeiswei:
@linas woudn't `PutLink` inherit from `LambdaLink`? https://github.com/opencog/atomspace/blob/master/opencog/atoms/base/atom_types.script#L263

ngeiswei:
By the same logic I think `ForAllLink` and `ExistsLink` should inherit from `LambdaLink`. Think of it that way, `ForAllLink` and `ExistsLink` are high level functions over predicates, basically one could rewrite this
```
ForAll
  <variables>
  <body>
```
as this
```
ForAllHighLevelLink
  Lambda
    <variables>
    <body>
```
where `ForAllHighLevelLink` is the equivalent of `std::all_of`in C++ http://en.cppreference.com/w/cpp/algorithm/all_any_none_of.

Although `ForAll` and `Exists` are not equal to `Lambda` they certainly use its beta-reduction operation.

linas:
> woudn't `PutLink` inherit from `LambdaLink`

Um, maybe-ish.  PutLink does type-checking and filtering when it performs the substitution.  The LambdaLink does not, the LambdaLink just blindly substitutues, without checking to see if the result is syntactically valid.

linas:
> `ForAll`  and `Exists`

The conversation is going in circles. In principle, it is illegal to perform substitutions in `ForAll` and `Exists`. It becomes legal only in certain forms of predicate logic, where you have a specific rule of inference that allows this: namely the rule of https://en.wikipedia.org/wiki/Universal_instantiation 

So really, the question is: do you want to be neat and clean, and put all rules in the rule engine, or do you want to be sloppy and fast, and put some rules into C++ methods on atoms?   Right now, we do a little bit of both, depending on... some arbitrary historical decisions and mistakes.

If you have an urgent need to squeeze more performance out of PLN, and the only way to do this is to implement "universal instantiation" by making for-all inherit from lambda ... then OK, do the sloppy hack, and acknowledge that its a sloppy hack.  Otherwise, don't do it. 

[...]

PR #1460, opened 2017-12-07
More scope link fixes

ngeiswei:
Pushing previous fixes before addressing issue #1458

### DIFF
--- a/opencog/atoms/core/ScopeLink.cc
+++ b/opencog/atoms/core/ScopeLink.cc
@@ -30,6 +30,7 @@
 #include <opencog/atoms/core/LambdaLink.h>
 #include <opencog/atoms/core/TypeNode.h>
 #include <opencog/atomutils/TypeUtils.h>
+#include <opencog/atomutils/FindUtils.h>
 
 #include "ScopeLink.h"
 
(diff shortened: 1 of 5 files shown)


opencog/atoms/core/ScopeLink.cc
```diff
@@ -473,6 +474,28 @@ Handle ScopeLink::substitute_vardecl(const Handle& vardecl,
...
+{
+	Handle vardecl = get_vardecl();
+	const Variables& variables = get_variables();
+	HandleSeq nouts;
+	for (size_t i = (get_vardecl() ? 1 : 0); i < get_arity(); ++i) {
+		Handle nbody = consume_ill_quotations(variables, getOutgoingAtom(i));
+		nouts.push_back(nbody);
+		// If the new body has terms with free variables but no
+		// vardecl it means that some quotations are missing. Rather
+		// than adding them we set vardecl to an empty VariableList.
```
linas: This comment does not make sense to me. Free variables are never declared. That is what makes them free. It's the definition of "free variable".

ngeiswei: At this point in the algorithm they really are free while they should be quoted. This part is largely a hack, it wrongly removes quotations, and instead putting them back I remove the variables from the final variable declaration. It's a hack and should be replaced as soon as it is not longer good enough.

linas: OK. I don't quite get what's happening here, but that OK. Thanks.

PR #1463, opened 2017-12-08
Additonal attempts to get the unit test to pass

linas:
I'm really stumped.  Something is messed up somwhere; its either the logger the rule-engine or both.  I can't figure it out.


mjsduncan:
fwiw the the test is passing for me now, thanks!

Issue #1451, opened 2017-12-01
Crash guile on (ExecutionOutputLink (Variable "$X") (Variable "$Y"))


ngeiswei:
I know it is more or less ill formed, one should use instead
```
(Quote (ExecutiontOutput (Unquote (Variable "$X")) (Unquote (Variable "$Y"))))
```
and the latter works, but still the former shouldn't crash guile.

ngeiswei:
Wait! This crashes on Debian 8.7.1 but not on Ubuntu 17.10!!!

ngeiswei:
@linas that constructor

https://github.com/opencog/atomspace/blob/master/opencog/atoms/execution/ExecutionOutputLink.cc#L79

misses the call of `check_schema`. Anyway reason for that?

I'm asking because I noticed a discrepancy between adding `ExecutionOutputLink` with proper quotes with the schema interface and with the C++ interface (the C++ interface raise the exception in `check_schema`, but the scheme interface doesn't [weirdly enough in spite of using the C++ interface, I'm really confused]).

linas:
Can you provide a fuller example? This resmbles #1450 but is somehow diferent, so I'm unclear on the question/problem.

linas:
> misses the call of check_schema. Anyway reason for that?

No. Its a bug, should have been checked.

ngeiswei:
Full example
```
(use-modules (opencog exec)
(define el (ExecutionOutputLink (Variable "$X") (Variable "$Y")))
(cog-execute! el)
```
results in
```
Segmentation fault (core dumped)
```
Still crashes with lattest update a319aa8c8

ngeiswei:
Just to be clear, it's a not urgent as far as I am concerned since I wrap that in quotations.

linas:
What do you mean "crashes"?  I get this:  
```
scheme@(guile-user)> (use-modules (opencog) (opencog exec)
)
scheme@(guile-user)> 
scheme@(guile-user)> (define el (ExecutionOutputLink (Variable "$X") (Variable "$Y")))
ERROR: In procedure cog-new-link:
[...]

PR #1464, opened 2017-12-11
Fix execution output pattern

ngeiswei:
Relax https://github.com/opencog/atomspace/blob/master/opencog/atoms/execution/ExecutionOutputLink.cc#L46 to allows `ExecutionOutputLink` in patterns like
```
(Get (Quote (ExecutionOutput (Unquote (Variable "$schema")) (Unquote (Variable "$arg")))))
```


ngeiswei:
Also semi-adresses issue #1451

PR #1473, opened 2017-12-14
Implement CompositionLink and ProjectionLink

ngeiswei:
Higher order construct to compose functions (patterns, etc). See https://wiki.opencog.org/w/CompositionLink https://wiki.opencog.org/w/ProjectionLink.


linas:
"CompositionLink" sounds strange in English. "ComposeLink" sounds better to me.

linas:
Also it does not seem to work, for me: see #1475

ngeiswei:
I don't mind to rename it to `ComposeLink`, it's shorter. What about `ProjectionLink` vs `ProjectLink`?

linas:
`ProjectionLink` vs `ProjectLink` -- both are about the same, to me.  I don't care.

Issue #1426, opened 2017-11-03
Forward Chainer doesn't check for complete satisfiability of premises 


Yidnekachew:
`(define (return-value x) x)`

```
(define rule
    (BindLink
        (And
            (Inheritance
                (Variable "$a")
                (Variable "$b"))
            (Member
                (Variable "$c")
                (Variable "$d")))

        (ExecutionOutput
                (GroundedSchemaNode "scm: return-value")
                (Inheritance (Variable "$a") (Variable "$b")))))
```

`(cog-fc rule-base (Inheritance (Concept "table") (Concept "furniture")))`

Evaluates to

```
$1 = (SetLink
     (InheritanceLink
        (ConceptNode "table")
        (ConceptNode "furniture")
    )
 )
```

In this case, it will go to the rewrite term without evaluating one of the premises `(Member c d)`. i.e it doesn't check whether membership atom is found in the atomspace.

ngeiswei:
In the absence of type constraints the premise
```
(Member
  (Variable "$c")
  (Variable "$d"))
```
can be itself. You need to specify the types of `$a`, `$b`, `$c` and `$d` in your rule. It is generally very advisable to do that, even without this problem because it can speed up the URE tremendously, due to the type constraints restricting the search space, and atoms being indexed by types in the AtmSpace.

See

https://wiki.opencog.org/w/Category:Type_constructors
https://wiki.opencog.org/w/Type_checker

for more information about type specification and checking.

PR #1474, opened 2017-12-14
create forward chainer unit test for unsatisfied premises

Yidnekachew:
a unit test for issue #1426

### DIFF
--- a/tests/rule-engine/forwardchainer/fc-unsatisfied-premise.scm
+++ b/tests/rule-engine/forwardchainer/fc-unsatisfied-premise.scm
@@ -1,0 +1,30 @@
+(define (return-value x) x)
+
+(define rule
+ (BindLink
+
+  (VariableList
+   (TypedVariable (Variable "$a") (Type "ConceptNode"))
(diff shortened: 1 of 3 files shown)


tests/rule-engine/forwardchainer/fc-unsatisfied-premise.scm
```diff
@@ -0,0 +1,32 @@
...
+   (TypedVariable (Variable "$c") (Type "ConceptNode"))
+   (TypedVariable (Variable "$d") (Type "ConceptNode")))
+
+  (And
+   (Inheritance
+    (Variable "$a")
+    (Variable "$b"))
+   (Member
+    (Variable "$c")
+    (Variable "$d")))
```
ngeiswei: @Yidnekachew to test that well you'd need to have all clauses turned into constants. You may for instance replace
```scheme
 (Member
    (Variable "$c")
    (Variable "$d"))
```
by
```scheme
(Member
    (Variable "$a")
    (Variable "$b"))
```
because I suspect this would pass even before your fix (commit 9547797d3).

PR #1483, opened 2017-12-18
Rewrite documentation for the RewriteLink

linas:
Also, mark many of the methods as being protected.

### DIFF
--- a/opencog/atoms/core/RewriteLink.h
+++ b/opencog/atoms/core/RewriteLink.h
@@ -44,8 +44,46 @@ class RewriteLink : public ScopeLink
 protected:
 	RewriteLink(Type, const Handle&);
 
-protected:
 	void init(void);
+
+	/**
+	 * Helper for partial_substitute. Given a mapping from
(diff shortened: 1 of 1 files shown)


opencog/atoms/core/RewriteLink.h
```diff
@@ -44,86 +44,100 @@ class RewriteLink : public ScopeLink
...
+	                                 const HandleMap& vm);
+
+	/**
+	 * Helper for partial_substitute. Given the variable
+	 * declaration and a mapping from variables to values,
+	 * this performs substitution over all bodies(??),
+	 * consuming ill quotations if necessary.
+	 *
+	 * XXX why does it say "all bodies"? How can there be more
+	 * one body?
```
ngeiswei: I mean terms being scoped, for instance ImplicationScope have 2 "bodies", the antecedent and the consequent.

linas: OK, thanks, I  think I get it now.

PR #1488, opened 2017-12-19
Lay groundwork for the PrenexLink

linas:
Disentangle and clarify the relationships between LambdaLink and RewriteLink, and lay the skeleton for PrenexLink.  Alter PutLink to use RewriteLink for beta reduction.

### DIFF
--- a/opencog/atoms/core/RewriteLink.cc
+++ b/opencog/atoms/core/RewriteLink.cc
@@ -138,20 +138,19 @@ Handle RewriteLink::substitute(const HandleMap& vm) const
 
 Handle RewriteLink::substitute(const HandleSeq& vals) const
 {
-	// XXX this implementation is probably wrong.
-	// Its copied from Lambda.
+	// XXX this implementation is wrong. Its a hack for now.
 	return get_variables().substitute(_body, vals);
 }
 
 HandleSeq RewriteLink::substitute_bodies(const Handle& nvardecl,
-                                               const HandleMap& vm) const
+                                         const HandleMap& vm) const
 {
 	const Variables& variables = get_variables();
 	return substitute_bodies(nvardecl, variables.make_values(vm));
 }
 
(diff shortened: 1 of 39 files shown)


opencog/atoms/core/RewriteLink.cc
```diff
@@ -88,82 +80,109 @@ inline HandleSeq append_rand_str(const HandleSeq& vars)
...
+	// Insert vardecl in the outgoing set, if defined
 	if (nvardecl)
+	{
 		hs.insert(hs.begin(), nvardecl);
+	}
+	else
+	{
+		// Its illegal to create a PutLink that is not in the
+		// form of a redex, i.e. doesn't have variable declarations
+		// in it. So we must not call createLink(), below.
```
ngeiswei: I don't think it's correct, that is because no vardecl doesn't mean no variable. As you can see https://github.com/opencog/atomspace/blob/master/opencog/atoms/core/RewriteLink.cc#L187 if the RewriteLink didn't have any vardecl to begin with then the resulting vardecl will be `Handle::UNDEFINED`.

I realize that this conflates with https://github.com/opencog/atomspace/blob/master/opencog/atoms/core/RewriteLink.cc#L224 i.e. it shouldn't return `Handle::UNDEFINED` when no variable remains after substitution, it should instead return something like an empty `VariableList`. I'll try to create unit tests for that...

linas: Well, but I can guarantee that PutLink will throw. PutLink is a "beta-redex" (google it) and after reducing a beta-redex, it can't continue to be a beta-redex any more, its "gone". 

A lambda is a little more arguable. In standard lambda-calc, after reduction, the lambda is gone.  You can define a non-standard one, which looks like a "function taking zero arguments"  but then the only thing one can do with that is to delta-reduce it.

A pattern with zero variables is ... a toss-up.I guess its well-defined.

ngeiswei: Got it.

PR #1491, opened 2017-12-19
automatic eta reduction in PutLink

linas:
Automate the eta reduction. This makes PutLink a little more usabel for function composition.

### DIFF
--- a/opencog/atoms/core/PutLink.cc
+++ b/opencog/atoms/core/PutLink.cc
@@ -123,11 +123,28 @@ void PutLink::static_typecheck_values(void)
 		return;
 
 	size_t sz = _varlist.varseq.size();
-	Type vtype = _values->get_type();
+
+	Handle valley = _values;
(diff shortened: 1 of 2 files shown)


opencog/atoms/core/PutLink.cc
```diff
@@ -123,11 +123,28 @@ void PutLink::static_typecheck_values(void)
 		return;
 
 	size_t sz = _varlist.varseq.size();
-	Type vtype = _values->get_type();
+
+	Handle valley = _values;
```
ngeiswei: Just curious (especially as English isn't my mother tong), why variable name "valley"?

linas: It sounds like "value" but is spelled differently, making it easy to search for. I use search a lot, so having unique strings is important.  Also, having maximally-different strings is desirable. When everything is vthis and vthat and var vars vars it gets hard to read.

PR #1516, opened 2018-01-16
Implement cog-get-all-subtypes

ngeiswei:
like cog-get-subtypes but is called recursively to include indirect
subtypes as well.


opencog/scm/opencog/base/utilities.scm
```diff
@@ -1205,6 +1206,11 @@
   (cog-cp (apply append (map cog-get-atoms (cog-get-types))) AS)
 )
 
+(define-public (cog-get-all-subtypes type)
+  (let* ((subtypes (cog-get-subtypes type))
```
linas: A few lines documentation would be nice.   Reminder:  at the guile prompt:
```
,a types      ;; apropos - lists all functions relating to "types"
,d cog-get-all-subtypes  ;; describe - print subroutine documentation
```

Issue #501, opened 2015-11-25
PythonEval doesn't build with python3


ebadawy:
I've a problem with atomespace build, got this error with `make`

```
/home/ebadawy/git/OpenCog/atomspace/opencog/cython/PythonEval.cc: In function ‘bool try_to_load_modules(const char**)’:
/home/ebadawy/git/OpenCog/atomspace/opencog/cython/PythonEval.cc:227:73: error: ‘PyString_AsString’ was not declared in this scope
             const char* sysPathCString = PyString_AsString(pySysPathLine);
                                                                         ^
/home/ebadawy/git/OpenCog/atomspace/opencog/cython/PythonEval.cc: In function ‘void opencog::global_python_initialize()’:
/home/ebadawy/git/OpenCog/atomspace/opencog/cython/PythonEval.cc:302:42: error: cannot convert ‘char**’ to ‘wchar_t**’ for argument ‘2’ to ‘void PySys_SetArgv(int, wchar_t**)’
         PySys_SetArgv(1, (char **) &argv0);
                                          ^
/home/ebadawy/git/OpenCog/atomspace/opencog/cython/PythonEval.cc: In member function ‘void opencog::PythonEval::build_python_error_message(const char*, std::string&)’:
/home/ebadawy/git/OpenCog/atomspace/opencog/cython/PythonEval.cc:570:66: error: ‘PyString_AsString’ was not declared in this scope
         char* pythonErrorString = PyString_AsString(pyErrorString);
                                                                  ^
/home/ebadawy/git/OpenCog/atomspace/opencog/cython/PythonEval.cc: In member function ‘void opencog::PythonEval::execute_string(const char*)’:
/home/ebadawy/git/OpenCog/atomspace/opencog/cython/PythonEval.cc:607:18: error: ‘Py_FlushLine’ was not declared in this scope
     Py_FlushLine();
                  ^
/home/ebadawy/git/OpenCog/atomspace/opencog/cython/PythonEval.cc: In member function ‘int opencog::PythonEval::argument_count(PyObject*)’:
/home/ebadawy/git/OpenCog/atomspace/opencog/cython/PythonEval.cc:621:57: error: ‘PyInt_AsLong’ was not declared in this scope
             argumentCount = PyInt_AsLong(pyArgumentCount);
                                                         ^
[...]

PR #1519, opened 2018-01-17
Port atomspace to python3.

linas:
This should fix issue #501 . The work was simply not that hard, in the end -- almost entirely a matter of using UTF-8 for strings, instead of raw ASCII bytes. 

This also adds some debugging and cleans up some bogus error messages.

One important work-item remains: apparently (per stack exchange and my personal experience), CMake itself does not yet support python3, so there is no way to build the python3 bindings without brute-force hackery on the CMakefie.


edyirdaw:
Thanks Linas. I didnt test it out yet  but the following line might do it

cmake -DPYTHON_DEFAULT_EXECUTABLE=/usr/bin/python3

according to https://github.com/opencv/opencv/issues/8425

edyirdaw:
https://github.com/opencv/opencv/issues/8425

linas:
Well, but cmake isn't magic. You'd have to write a bunch of scripts to tell cmake what to do with  `-DPYTHON_DEFAULT_EXECUTABLE` .. you'd also have to write scripts to find the header files, the libraries, and the packages path.

The easiest thing would be to take one of the `lib/FindStuff` files and modify it for python.

linas:
@edyirdaw the pull request #1567 performs the cmake checks that you want.  No -D flag is needed.

Issue #1523, opened 2018-01-21
Build fails on Ubuntu 14.04 / Cythonizing scheme_wrapper.pyx


edguy3:
On a fresh Ubuntu 14.04 install, with Cython version 0.20.1post0, I get the following build error: 
```
Error compiling Cython file:
------------------------------------------------------------
...

cdef extern from "opencog/cython/load-file.h" namespace "opencog":
    int load_scm_file_relative (cAtomSpace& as, char* filename) except +

def load_scm(AtomSpace a, str fname):
    status = load_scm_file_relative(deref(a.atomspace), fname.encode('UTF-8'))
                                                                   ^
------------------------------------------------------------

/home/opencog/atomspace/opencog/cython/opencog/scheme_wrapper.pyx:69:68: Obtaining 'char *' from temporary Python value

```

Builds fine on Ubuntu 16.04 with Cython version 0.23.4

( Patch coming. )

linas:
closing. I assume #1524 fixed this

PR #1524, opened 2018-01-21
fix for cython build error on ubuntu 14.04 as described in issue #1523

edguy3:
Assigning filename to temporary value allows it to compile on Ubuntu 14.04 using Cython version 0.20.1post0. Does not compile otherwise.   Confirmed on  Ubuntu 16.04 as well.  No impact on unit tests.


linas:
Thanks!

Issue #1525, opened 2018-01-22
build.sh file requires Bash rather than default Bourne shell


edguy3:
Encountered this issue while building on Ubuntu 16.04:

```
build.sh: 6: build.sh: [[: not found
Wrong GHC version installed. Running stack setup.
stack will use a sandboxed GHC it installed
For more information on paths, see 'stack path' and 'stack exec env'
To use this GHC and packages outside of a project, consider using:
stack ghc, stack ghci, stack runghc, or stack exec

```
Source is 
../opencog/haskell/CMakeLists.txt:        COMMAND sh build.sh ${CMAKE_CURRENT_BINARY_DIR}

( Pull request with fix in process)

linas:
I merged  #1426 which is correct; I am confused about why there is also a revert that references this... ?

linas:
closing, I assume #1426 fixed this.

PR #1526, opened 2018-01-22
use bash shell instead of bourne on haskell target build issue #1525

edguy3:
Corrects issue checking haskell target where a shell script 
that requires bash is invoked with sh


linas:
Thanks!

Issue #1528, opened 2018-01-23
Crazy pattern matcher bug


ngeiswei:
This one is weird, and I mean WEEEEIIIIIIRRRRRD, like blair witch weird.

Consider the following
```
(LambdaLink
  (VariableList
    (VariableNode "$X")
    (VariableNode "$Y")
  )
  (InheritanceLink
    (VariableNode "$X")
    (VariableNode "$Y")
  )
)

;; Query
(define query
(BindLink
  (AndLink
    (VariableNode "$f-lamb-e84bdd8")
    (NumberNode "2.000000")
    (EvaluationLink
      (GroundedPredicateNode "scm: dummy")
      (NumberNode "2.000000")
    )
  )
  (Concept "A"))
)

(define (dummy X) (stv 1 1))
```
The result I'm getting is `(Set)`, while it is expected to return `(Set (Concept "A"))`.

Now the crazy parts, if I remove any one of the clauses `(VariableNode "$f-lamb-e84bdd8")`, `(NumberNode "2.000000")` or `(Evaluation ...` it works!

But there's more, if I rename `(VariableNode "$f-lamb-e84bdd8")` to `(VariableNode "$f-lamb")` it works too!!!

ngeiswei:
OK, it seems that the name change craziness `"$f-lamb"` vs `"$f-lamb-e84bdd8"` comes from the fact that it changes the canonical order used to turn unordered links (here `AndLink`) to ordered internal representations. Then this causes the pattern matcher to choose a different initial search strategy that happens to be buggy.

PR #1530, opened 2018-01-24
Fix issue #1528

ngeiswei:
There was a discrepancy between `PatternLink::_components` and `PatternLink::_component_patterns`. Fix it by removing constant component patterns from `PatternLink::_component_patterns` as well.


linas:
blair witch

PR #1535, opened 2018-01-26
modify return value on timepair_t AtomSpaceBenchmark::bm_getHandlesBy

edguy3:
Was returning a clock_t when a timepair_t is needed. 

While has no impact on my unit tests, this PR requires a bit of review to ensure
this is the proper fix. 

/ed


linas:
thanks

PR #1537, opened 2018-01-29
Fix nested put links

ngeiswei:
Fix beta-reducing nested put links like
```
(Put
  (Put
    (Lambda
      (Variable "$X")
      (Variable "$X"))
    (Lambda
      (VariableList
        (Variable "$Y")
        (Variable "$Z"))
      (Inheritance
        (Variable "$Y")
        (Variable "$Z"))))
  (List
    (Concept "A")
    (Variable "$X")))
```
that should output
```
(Lambda
  (Variable "X")
  (Inheritance
    (Concept "A")
    (Variable "$X")))
```
instead of crashing.

### DIFF
--- a/opencog/atoms/core/PrenexLink.h
+++ b/opencog/atoms/core/PrenexLink.h
@@ -38,7 +38,7 @@ namespace opencog
 ///
 /// This is used primarily to ensure that PatternLinks remain in
 /// prenex form, even when being rewritten. PatternLinks must have
-/// all avraible declarations out in front, in order to work.
+/// all available declarations out in front, in order to work.
 ///
 class PrenexLink;
 typedef std::shared_ptr<PrenexLink> PrenexLinkPtr;
--- a/opencog/atoms/core/PutLink.cc
+++ b/opencog/atoms/core/PutLink.cc
@@ -117,6 +117,8 @@ void PutLink::static_typecheck_values(void)
 		return;
 	if (DEFINED_PREDICATE_NODE == btype)
 		return;
+	if (PUT_LINK == btype)
+		return;
 
 	// If its part of a signature, there is nothing to do.
 	if (classserver().isA(btype, TYPE_NODE) or TYPE_CHOICE == btype)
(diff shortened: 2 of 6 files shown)


ngeiswei:
@linas, I'm reasonably confident to merge by myself, you may want to take a look at it nonetheless.

linas:
I'm having a huge amount of trouble parsing what this is supposed to do. To me, it looks like two completely malformed PutLinks nested one in the other.  That it crashed in he first place is not surprising.   That it now does something other than throw an exception is hugely surprising.

The inner Put makes sense:
```
(define inner
 (Put
    (Lambda
      (Variable "$X")
      (Variable "$X"))
    (Lambda
      (VariableList
        (Variable "$Y")
        (Variable "$Z"))
      (Inheritance
        (Variable "$Y")
        (Variable "$Z")))))
```

and  `(cog-execute! inner)`  seems to return the expected structure:
```
(LambdaLink
   (VariableList
      (VariableNode "$Y")
      (VariableNode "$Z")
   )
   (InheritanceLink
      (VariableNode "$Y")
      (VariableNode "$Z")
   )
)
```

However, the outer Put seems ill-formed, and should throw an error.

linas:
So continuing that  example:  
```
(define ineval (cog-execute! inner))

(define outer
   (Put
      (VariableList
         (VariableNode "$Y")
         (VariableNode "$Z")
      )
      ineval
      (List
         (Concept "A")
         (Variable "$b"))))
```
The `(cog-execute! outer)` gives the desired result.

linas:
Hmm. I'm now thinking that this is evaluation-order dependent.  Are you evaluating the innerr Put first, and then the outer one? or the outer one first, then the inner one?

It is well-known that reduction is order-dependent whenever the system is not strongly normalizing, and it seems clear that PutLink is not strongly normalizing.  So if you are trying to do the outer one first ... ... I duno. I'm very confused about why you think this should have worked.

linas:
oh never mind.  I see what you are doing. You are evaluating the inner one first, and yes, it should give something that can be used with the outer one.  OK.

Its a confusing example, but yes it seems valid.

linas:
And I reviewed the source code changes, and they seem correct also.

PR #1538, opened 2018-01-29
fix BasicSCMUTest by setting en_US.utf8

edguy3:
BasicSCMUTest requires a UTF-8 character set to complete its tests. 
This change explicitly changes locale to the en_US.utf8 LC_CTYPE within the test code.
This approach requires that en_US.utf8 be installed on the machine. 
( Test was failing on Ubuntu 16.04 and 17.04) 

There are various alternatives to approaching this fix and it seems reasonable to require a given character set for unit tests.   ( NB. I am essentially monolingual, so those of you that live beyond EBCDIC may have a better way to solve this problem.  - I will have a pull request for ocpkg to compliment this approach. )

### DIFF
--- a/tests/scm/BasicSCMUTest.cxxtest
+++ b/tests/scm/BasicSCMUTest.cxxtest
@@ -118,7 +118,7 @@ void BasicSCMUTest::setUp(void)
 	// in the majority of European locales. Unfortunately, guile number
 	// parsing is not locale dependent...
 	// eval->eval("(setlocale LC_ALL \"\")");
-	eval->eval("(setlocale LC_CTYPE \"\")");
+	eval->eval("(setlocale LC_CTYPE \"en_US.utf8\")");
 	eval->clear_pending();
 }
 


DeborggraeveR:
Found a solution for the same issue. See: #1540 
This forces the numeric formatting to use ISO standard, so should be a dot on all systems, while leaving the other locale setting intact.
This seem to have solved multiple unit tests on my system.

linas:
Ed can you verify that #1540 fixes the issue for you?  You pull req seems entirely reasonable, but I can't tell if the original bug was due to the decimal-point problem, or something else.  

The long-term correct fix would probably be to check (in `SchemeEval.C`) if the local is already utf8, and if not, only then set it to en_us.utf8.  

I'm sort-of surprised that your locale is not already utf8 -- which distro is it that does not already set this by default?

edguy3:
My utf8 issue was on ubuntu 16.04.  IIRC, the default from the docker image was POSIX (and the 'locale' package was not installed.)   Will let you know if latest merge works later this evening.  

/ed

linas:
If this is the docker container set up by the opencog docker scripts, it should be fixed to install `locale`

edguy3:
I'm using docker for testing native installs.  The 14.04 opencog/docker script does have a locale install. ( I made changes to octool/ocpkg yesterday  - already merged) 

/ed

edguy3:
With this commit:
```
$ git log | head -30
commit cf3c8516e93bf480a899a1e330d12e46a2454f38
Merge: 88b8dee 0f22173
Author: Linas Vepštas <linasvepstas@gmail.com>
Date:   Mon Jan 29 15:09:01 2018 -0600
    Merge pull request #1540 from DeborggraeveR/master    
    Force iso standard number formatting in scheme
...
```
With LANG and other Character set related environment variable unset, on debian 16.04

```
make test
```
Fails BasicSCMUTest and 
```
LANG=en_US.UTF-8 make test
```
BasicSCMUTest passes.

DeborggraeveR:
What do u get when calling setlocale(LC_CTYPE, NULL) when lang is not set?
Just wondering if setting en_US.UTF8 won't give issues for people with languages like chinese.

Maybe like linas said, if locale is not set, then hard set to en_US.UTF8?
or even get the value from a config file, if not set in config, check if system locale is defined, if not hard set to en_US.UTF8


p.s: #1540 was about fixing decimal number formatting, hard set to ISO standard to make sure it always use a dot. Your issue seem to be something totally different, just on the same unit test :)

linas:
You know, I hit the merge button, and then immediately reconsidered so I undid it in pull req #1542  

The answer is:
-- no config files, please,
-- SchemeEval.C should check for utf8, and if not force locale to En_US, and also print a warning that advises user to set LANG to something reasonable.

then this particular patch would not be needed.

edguy3:
calling setlocale(LC_CTYPE, NULL)  gives an error: 
```
check node (LemmaNode "When life gives you lemmas, make conjugations" (stv 0.476500 0.611540))
check_node(PartOfSpeechNode,はにがうりだそうであってるのかはちょっと不安ですが翻訳ソフトでやくし てみました。,0.347650,0.336110)
check node (PartOfSpeechNode "はにがうりだそうであってるのかはちょっと不安ですが翻訳ソフトでやくし てみました。" (stv 0.347650 0.336110))
In BasicSCMUTest::test_single_atom:
/home/opencog/atomspace/build/tests/scm/../../../tests/scm/BasicSCMUTest.cxxtest:169: Error: Test failed: Failed to create a node
/home/opencog/atomspace/build/tests/scm/../../../tests/scm/BasicSCMUTest.cxxtest:169: Error: Assertion failed: !eval_err
/home/opencog/atomspace/build/tests/scm/../../../tests/scm/BasicSCMUTest.cxxtest:173: Error: Test failed: Failed to find handle
/home/opencog/atomspace/build/tests/scm/../../../tests/scm/BasicSCMUTest.cxxtest:173: Error: Assertion failed: h != nullptr
/home/opencog/atomspace/build/tests/scm/../../../tests/scm/BasicSCMUTest.cxxtest:176: Error: Test failed: Failed to find node
/home/opencog/atomspace/build/tests/scm/../../../tests/scm/BasicSCMUTest.cxxtest:176: Error: Assertion failed: n != nullptr
[2018-01-29 22:55:12:014] [DEBUG] BEGIN TEST: test_single_atom

```

DeborggraeveR:
What kind of error? can u try in single standalone app?

```
#include <stdio.h>
#include <locale.h>
#include <iostream>

int main()
{
        std::cout << "Current locale: " << setlocale(LC_CTYPE, NULL) << std::endl;
        setlocale(LC_ALL, "");
        std::cout << "Current locale: " << setlocale(LC_CTYPE, NULL) << std::endl;

        return 0;
}
```

For me this returns: 
Current locale: C
Current locale: en_US.UTF-8

edguy3:
On all my docker clients ( Ubuntu 14.04, 16.04, 17.10, Debian Stable):
```
Current locale: C
Current locale: C
```
After reading the set locale man page, this result is the expected behavior.

PR #1548, opened 2018-01-31
Misc fix + make PutLink more robust

ngeiswei:


opencog/atoms/core/PutLink.cc
```diff
@@ -316,6 +316,8 @@ Handle PutLink::do_reduce(void) const
 	{
 		PutLinkPtr nested_put = PutLinkCast(bods);
 		bods = nested_put->reduce();
+		if (not bods)
+			return Handle::UNDEFINED;
 		btype = bods->get_type();
```
linas: In all other cases in this code, we throw an exception -- we should throw one here as well. The point of throwing exceptions it to avoid the use of Handle::UNDEFINED, and therefore to avoid having to explicitly check for Handle::undefined. 

In other words, reduce should never return `Handle::UNDEFINED`; it should throw. Then this particular if-test would not be needed.

ngeiswei: OK, but it's a bit inconsistent with what that try/catch is doing https://github.com/opencog/atomspace/blob/master/opencog/atoms/core/PutLink.cc#L393

Anyway, I get the idea, I'll try to get the big picture and follow this guideline.

linas: Ahh, yes, some of the try-catches might be in the wrong places.  Its not really urgent; but mostly I want to avoid having to put an if-test for null handles in the ctor for  Link.cc

ngeiswei: PR #1549 partially fixes that (partially because returning `UNDEFINED` should still be replaced with throwing exceptions elsewhere).

PR #1549, opened 2018-02-01
Wrap evaluation of virtual clause in try/catch

ngeiswei:
and add a unit test for it.

The Instantiator::reset_halt() method is used not to falsely detect
infinit recursion when an exception aborts an evaluation before
Instantiator::_halt gets switched back to false.

### DIFF
--- a/tests/query/PutUTest.cxxtest
+++ b/tests/query/PutUTest.cxxtest
@@ -1,0 +1,97 @@
+/*
+ * tests/query/PutUTest.cxxtest
+ *
+ * Copyright (C) 2018 OpenCog Foundation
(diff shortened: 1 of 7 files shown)


tests/query/PutUTest.cxxtest
```diff
@@ -0,0 +1,97 @@
+/*
+ * tests/query/PutUTest.cxxtest
+ *
```
linas: There already is an `tests/atoms/PutLinkUTest` -- its dangerous to have a second test, in a different directory, with a similar name

ngeiswei:
That's wise, I'll rename `NestedPutUTest`.

PR #1555, opened 2018-02-02
More put fixes

ngeiswei:
Raise exceptions instead of returning `Handle::UNDEFINED` in some places in PutLink. Some exceptions should be probably be silenced but that can be done later.

### DIFF
--- a/opencog/atoms/core/PutLink.cc
+++ b/opencog/atoms/core/PutLink.cc
@@ -20,6 +20,7 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
+#include <opencog/util/Logger.h>
 #include <opencog/atoms/base/atom_types.h>
 #include <opencog/atoms/base/ClassServer.h>
 #include "DefineLink.h"
@@ -389,19 +390,7 @@ Handle PutLink::do_reduce(void) const
 		{
 			HandleSeq oset;
 			oset.emplace_back(_values);
-			try
-			{
-				// return vars.substitute(bods, oset, /* silent */ true);
-				return reddy(subs, oset);
-			}
-			catch (const TypeCheckException& ex)
-			{
-				return Handle::UNDEFINED;
-			}
-			catch (const SyntaxException& ex)
-			{
-				return Handle::UNDEFINED;
-			}
+			return reddy(subs, oset);
 		}
 
 		// If the values are given in a set, then iterate over the set...
@@ -412,7 +401,6 @@ Handle PutLink::do_reduce(void) const
 			oset.emplace_back(h);
 			try
 			{
-				// bset.emplace_back(vars.substitute(bods, oset, /* silent */ true));
 				bset.emplace_back(reddy(subs, oset));
 			}
 			catch (const TypeCheckException& ex) {}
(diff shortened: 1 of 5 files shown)


opencog/atoms/core/PutLink.cc
```diff
@@ -444,22 +424,15 @@ Handle PutLink::do_reduce(void) const
...
+		return reddy(subs, oset);
 	}
 
 	// If we are here, then there are multiple values.
 	// These MUST be given to us as a SetLink.
 	if (SET_LINK != vtype)
 	{
 		if (_silent)
-			throw NotEvaluatableException();
+			throw TypeCheckException();
```
linas: Um, well, a type-checking exception is supposed to be different from a cant-be-velauated exception.  In the putlink context, a type-check exception is totally normal -- it just means the put is being used as a filter, and that particular element was filtered out.  By contrast not-evaluatable means ...

well, inthe above line, I'm not even sure that typecheckexception is correct -- it sounds to me like its a user-error exception, not a type-check exception: the user simply created a PutLink with a bad syntax (so... SyntaxException, if you wish?)  Bad syntax is different than type checking...

ngeiswei: To be honest I didn't think much that through, I just tried to monkey the whole thing. For instance I don't understand why the silent exception type is different than the non silent one. It seems to me they should be the same, just the silent one would not log anything, or am I missing something?

linas: The "silent" ones don't log, and also are not supposed to be presented to the user -- because they are neither user-errors, nor system errors.    They are supposed to be for those cases where you might (on rare occasions) want to return a null handle. So, rather than returning a null handle  and then having to check 100 times for a null handle only to discover that its OK 99% of the time ... instead, just throw a silent exception.  Higher up, catch it, and do the right thing.   

The nice things about exceptions is that they skip over everything in between.  So all that code in between doesn't have to worry about handling corner cases. It can just pretend everything is normal, perfect, always works, and for the 1% of the cases where something special is done, the excepetion does it. It helps avoid spaghetti code, by simplifying error handling.

ngeiswei: I sorta know that (although it doesn't hurt to be reminded), but what I was wondering is
1. how to choose what exception to throw
2. why silent exceptions are different than non-silent ones, see for instance https://github.com/opencog/atomspace/blob/master/opencog/atoms/core/RewriteLink.cc#L165 the silent exception is `TypeCheckException` while the non-silent one is `SyntaxtException`. Shouldn't they both be the same?

ngeiswei: Oh I get it! All silent exceptions derive from the SilentException class! Yet why not all non-silent exception don't have a silent counterpart, I don't really understand...

linas: The non-silent exceptions are due to user errors -- they should result in error messages printed to the terminal.

The silent exceptions are merely fast-paths through the code -- short-cuts to get from point a to point b; there's no need to log them, there's no need to tell the user about them, because they are not actually errors -- they are just "non-local returns".

The tricky part is that sometimes, a user error can result in an execution path that ends with a silent exception -- in which case, it should not really be silent - it should be "loud", the user needs to be told that there's a bug in his code.  This "tricky part" is why there is this stupid "silent" flag in code. I don't like that flag, but I also don't want to hide user errors (because hiding an error makes it hard to debug)

We need a wiki page on this.

ngeiswei: So it seems the silent catches should only catch the `SilentException` class. That would make the code intend very obvious.

linas: Yes.

PR #1556, opened 2018-02-02
Relax another catch in DefaultPatternMatchCB::eval_term

ngeiswei:


opencog/query/DefaultPatternMatchCB.cc
```diff
@@ -655,7 +655,7 @@ bool DefaultPatternMatchCB::eval_term(const Handle& virt,
 		{
 			tvp = EvaluationLink::do_eval_scratch(_as, gvirt, _temp_aspace, true);
 		}
-		catch (const NotEvaluatableException& ex)
+		catch (...)
```
linas: well, but this isn't right.  Some exceptions are "normal" and are encountered during ordinary processing. These can be caught and silenced. Other exceptions are due to user error, these should not be silenced, but should be broadcast back up to the user.   I've been trying to keep these two very different kinds of exceptions clearly distinct from one-another.  Without this distinction, debugging gets even harder than it is.

PR #1558, opened 2018-02-06
Select atoms outside the AF and rename function

misgeatgit:


linas:
Misgana please revert or fix your code.   The git bisect states that one of these commits caused the breakage:
```
There are only 'skip'ped commits left to test.
The first bad commit could be any of:
2f31ca40ca046b6b188dbabee22c1f817d4c5620
9aeae9347c98d1558fbabb0dd2ccaa7bfd2343ae
72f6f53651408fd51cf1603390ba154eed1390d0
ef6027a411b4049399efc741ad5a0606b458c52f
a5b50ae8e8cda1be3e98179626498ad94a634e51
We cannot bisect more!
```

Please do not check in code that does not compile!

Please run unit tests and make sure that they pass before creating pull requests!

Please do not merge pull requests that are not passing unit tests!

Also: please make sure that your new code is properly indented, and treats whitespace in a normal manner.

Issue #1561, opened 2018-02-07
Wrongly consuming quotation while executing


ngeiswei:
Here's the following pattern
```
(LambdaLink
  (VariableList
    (VariableNode "$sha-arg-0") ; [949102443632971434][1]
    (VariableNode "$sha-arg-1") ; [4859387568654154175][1]
  ) ; [9448388737473849152][1]
  (LocalQuoteLink
    (AndLink
      (VariableNode "$sha-arg-0") ; [949102443632971434][1]
      (VariableNode "$sha-arg-1") ; [4859387568654154175][1]
    ) ; [9237651940852305971][1]
  ) ; [9268420674243330182][1]
) ; [9607689798623181459][1]
```
executing this pattern (via cog-execute!) outputs
```
(LambdaLink
   (VariableList
      (VariableNode "$sha-arg-0")
      (VariableNode "$sha-arg-1")
   )
   (AndLink
      (VariableNode "$sha-arg-0")
      (VariableNode "$sha-arg-1")
   )
)
```
which changes the meaning of that pattern.

The former means: *any binary AndLink*.

The latter means: *2 independent clauses*.

That is obviously wrong, but besides that, why quotation should be systematically consumed by execution? I can understand that patterns processed by the pattern matcher should have their quotations consumed so that matching patterns can be turned into proper instances (during the rewriting process of the `BindLink`), but besides that, I see no reason they should be.

ngeiswei:
Yeah the problem is really running deeper, the following
```
(QuoteLink
   (LambdaLink
      (UnquoteLink
         (VariableNode "$vardecl")
      )
      (UnquoteLink
         (VariableNode "$body")
      )
   )
)
```
executes into
```
(LambdaLink
   (VariableNode "$vardecl")
   (VariableNode "$body")
)
```
another non-sense.

So really `Instantiator::walk_expr` shouldn't consume quotations. I'm thinking of disabling it and relying instead on `RewriteLink::consume_ill_quotations()` when substitution occurs (during term rewriting by `PutLink` or `BindLink`).

linas:
?? But I thought that was the whole point of quotations: that they would be consumed by execution.

If they are not consumed, then what's the point of having them?  If they are not consumed, how does it ever become possible to remove them?  And if you never need to remove them, then why do you need them at all?

ngeiswei:
[...]

PR #1563, opened 2018-02-07
More over quotation consumption unit test

ngeiswei:
Related to issue #1561


linas:
At this time, the PutLinkUTest if failing. Is that intentional?

ngeiswei:
Yes, it is intentional. Hopefully I'll have that fixed by tomorrow.

PR #1568, opened 2018-02-08
One Definition Rule, again.

linas:
The python unit tests continue to violate the one-defintion-rule,  by attempting to load shared libraries both from system directories, and from the build directory. The linker-loader cannot tell that these two different libraries are supposed to be the "same library", and then unit-test madness ensues.

Work around this, as best as possible.


linas:
This requires a corresponding update of the opencog code, i.e. pull req opencog/opencog#3001 is needed.

The build will probably break if a `make clean` is not done first.

PR #1570, opened 2018-02-08
Devxx

misgeatgit:


opencog/attentionbank/AttentionBank.h
```diff
@@ -85,8 +87,8 @@ class AttentionBank
      * The amount importance funds available in the AttentionBank.
      * Atomic, so that updates don't need a lock.
```
linas: See the line above, which says "Atomic, so that updates don't need a lock." ?  Well, you just now mde thme non-atomic, and added a lock.  This means that you should also remove this comment, because its out of sync with the code.

opencog/attentionbank/AttentionBank.h
```diff
@@ -100,6 +102,14 @@ class AttentionBank
     AttentionValue::sti_t STIAtomWage;
     AttentionValue::lti_t LTIAtomWage;
 
+    inline AttentionValue::sti_t updateSTIFunds(AttentionValue::sti_t diff) {
+        return fundsSTI += diff;
```
linas: also -- since you changed fundsSTI to not actually be atomic, you now actually DO  need a lock here! 

It is probably safer and easier to make your funds atomic, and NOT use locks -- its usually much faster that way.  But I am not sure of what you are doing. But its obvious that here, you need to either make these atomic, ore use a lock.

linas: By the way, to make them atomic, you would say something like
```
std::atomic<AttentionValue::sti_t> fundsSTI;
```
or something like that. Read the man page:  http://en.cppreference.com/w/cpp/atomic/atomic

opencog/attentionbank/AttentionBank.h
misgeatgit: I wanted double atomic and after skimming through [this thread](https://stackoverflow.com/questions/30048533/why-isnt-atomic-double-fully-implemented), I opted for the mutex version. I should have removed the comment though.

opencog/attentionbank/AttentionBank.h
misgeatgit: I made the write functions private and are accessed only in one method, the AVChanged method.

linas: Ahh OK, if they are private, then that will work. But if they are only called in ino place, it would be even better if you just remove them completely --(it is only a single line of code! And its just an add!) -- this will avoid future mistakes, where someone makes them public again, but forgets the lock.

misgeatgit: OK, I will make the changes you suggested.

PR #1575, opened 2018-02-12
Update Scheme AV bindings

misgeatgit:


opencog/guile/SchemeSmobAV.cc
```diff
@@ -125,9 +125,16 @@ SCM SchemeSmob::ss_av_get_value (SCM s)
...
 
-	SCM sti = scm_from_short(av->getSTI());
-	SCM lti = scm_from_short(av->getLTI());
-	SCM vlti = scm_from_ushort(av->getVLTI());
+	// Format AV values to two decimal pts.
+	auto format = [](double n){
+	char number_str[20];
+	sprintf(number_str, "%.2f", n);
+	return atof(number_str);
+	};
```
linas: Several problems: 

1) bad indentation.

2) computationally, it would be much much faster to use the math library: `0.01*round(100.0*x)`  would truncate to two decimal places.

3) Anyway, are you sure you want to discard decimal places? why two? why not 3 or 6?

misgeatgit: I just wanted to see a formatted output. Maybe I should remove the rounding as it can be optionally done from scheme.

misgeatgit: I just made a [PR](https://github.com/opencog/atomspace/pull/1577) for this.

PR #1579, opened 2018-02-14
fix discrepancy between comment and code

Yidnekachew:


linas:
looks good to me @ngeiswei notice the s-> 1-s change...

PR #1588, opened 2018-02-20
fix forward chainer rule selection to use full rule TV

Yidnekachew:


ngeiswei:
I can accept that, although this part is gonna be re-implemented anyway, to reuse the backward chainer decision code.

ngeiswei:
Just to clarify, this sort of heuristics is good for exploitation but bad for exploration.

PR #1595, opened 2018-02-23
Improve Unify

ngeiswei:
In particular reconstructed variable declarations capture more type restrictions when available.

Also include some minor code simplification, `std::vector<Type>` renamed into `TypeSet`.

### DIFF
--- a/opencog/atoms/base/Handle.h
+++ b/opencog/atoms/base/Handle.h
@@ -300,6 +300,8 @@ std::string oc_to_string(const HandleUCounter& huc, const std::string& indent);
 std::string oc_to_string(const HandleUCounter& huc);
 std::string oc_to_string(Type type, const std::string& indent);
 std::string oc_to_string(Type type);
+std::string oc_to_string(const TypeSet& types, const std::string& indent);
+std::string oc_to_string(const TypeSet& types);
 std::string oc_to_string(const AtomPtr& aptr, const std::string& indent);
 std::string oc_to_string(const AtomPtr& aptr);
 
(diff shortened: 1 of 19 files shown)


opencog/atoms/base/Handle.h
```diff
@@ -300,6 +300,8 @@ std::string oc_to_string(const HandleUCounter& huc, const std::string& indent);
 std::string oc_to_string(const HandleUCounter& huc);
 std::string oc_to_string(Type type, const std::string& indent);
 std::string oc_to_string(Type type);
+std::string oc_to_string(const TypeSet& types, const std::string& indent);
```
linas: Is it important to have the indent-version in the public API? Can these just be moved to the .cc file, so they are not in the .h?

opencog/atoms/core/VariableList.cc
```diff
@@ -228,29 +228,29 @@ void VariableList::get_vartype(const Handle& htypelink)
...
 		}
 	}
 	else if (TYPE_INH_NODE == t)
 	{
 		Type vt = TypeNodeCast(vartype)->get_value();
-		std::set<Type> ts;
-		std::set<Type>::iterator it = ts.begin();
+		TypeSet ts;
+		TypeSet::iterator it = ts.begin();
 		classserver().getChildren(vt, std::inserter(ts, it));
```
linas: These changes make the code hard to read, hard to understand ... is there a positive reason for these changes?

opencog/atoms/base/Handle.h
ngeiswei: I'm afraid we do need them in the .h because other oc_to_string overloading not living within that file (like the onces in the URE or the pattern miner) want to have access to the indent versions. And we also need the non-indent version in the .h for gdb...

opencog/atoms/core/VariableList.cc
ngeiswei: No reason other than the fact that `std::set<Type>` appeared often throughout the code. Since we know (and I assume love) `HandleSet`, I thought introducing `TypeSet` would come naturally.

I tend to feel that C++ code contains too much punctuations, so when I get a chance to remove some (without impairing, in my opinion, understandability) I tend to go for it.

PR #1601, opened 2018-02-27
Added to ure examples.

shujingke:


ngeiswei:
@shujingke It would be much better if these examples were moved to unit tests as that is their intended purpose. So specifically within that file https://github.com/opencog/atomspace/blob/master/tests/rule-engine/backwardchainer/BackwardChainerUTest.cxxtest

PR #1603, opened 2018-02-27
Ported two examples to UTest.

shujingke:
Test 1: "Fact to rule" test in BackwardChainerUTest.
Test 2: "Negation conflict" test in ForwardChainerUTest.

### DIFF
--- a/examples/rule-engine/NegConflict.scm
+++ b/examples/rule-engine/NegConflict.scm
@@ -196,8 +196,15 @@
   (TypedVariable (VariableNode "$who") (TypeNode "ConceptNode"))
 )
 
+(define source
+  (Inheritance (stv 1.0 1.0)
+   (Concept "American")
+   (Concept "person"))
+)
+
 ;; Forward chainer:
-;; (cog-fc Einstein-rbs (Variable "$x"))
+;; (cog-fc Einstein-rbs (Variable "$x")) 
+;; (cog-fc Einstein-rbs source)
 ;; The expected result is:
 ;(SetLink
 ;   (EvaluationLink (stv 1 1)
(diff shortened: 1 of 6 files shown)


linas:
Hi,

Instead of making these unit tests ever fatter, larger and slower, could you consider creating new unit tests, instead?  In fact, the current Forweward/backward chainer unit tests are already so incredibly large and complex that they are essentially undebuggable: when they break, its essentially impossible to figure out why, and provide a fix.  Making them even more complex, more unmaintainable is not good programming practice.

Also: since you have not contributed to this project in a while, can you not instantly merge your updates? I would strongly prefer it if Nil were to review these changes first, so that problems and issues can be corrected, before they are merged.

ngeiswei:
I have no problem having very large unit tests because I can rapidly isolate the failing test by uncommenting the rest in Emacs. But not everybody uses Emacs. Certainly it doesn't hurt to chop them up into smaller pieces.

ngeiswei:
Yeah, maybe better wait a bit for my feedback before merging for the near future, @shujingke.

linas:
> I have no problem having very large unit tests because I can rapidly
> isolate the failing test by uncommenting the rest in Emacs.
>
Before you can comment or uncomment anything at all, you have to figure out
what is failing.  And that particular test case spews tens of thousands of
lines of output, so grepping for errors, and then locating where it starts
is non-trivial.

ngeiswei:
Alt-n in Emacs :)

ngeiswei:
To be a bit more specific
1. Compile and test in the *compile* buffer,
2. Emacs automatically interprets the cxxtest errors as compiling error,
3. I can cycle through the errors using Alt-n, then press enter to open directly the file at the location of the failing test.

If you're editor can do that for compiling errors, I suspect it can do it for cxxtest errors.

ngeiswei:
Linas, spacemacs offers vim key bindings, so it gives you the best of both worlds, you should give it a try! I've never tried it, BTW, but I might soon...

Issue #1606, opened 2018-02-28
Pattern matcher fails on some query


ngeiswei:
The following
```
(EvaluationLink
  (PredicateNode "minsup")
  (ListLink
    (LambdaLink
      (VariableNode "$spe-arg-1")
      (InheritanceLink
        (ConceptNode "A")
        (VariableNode "$spe-arg-1")
      )
    )
    (ConceptNode "texts")
    (NumberNode "2")
  )
)

(EvaluationLink
  (PredicateNode "minsup")
  (ListLink
    (LambdaLink
      (VariableNode "$spe-arg-0")
      (InheritanceLink
        (VariableNode "$spe-arg-0")
        (ConceptNode "C")
      )
    )
    (ConceptNode "texts")
    (NumberNode "2")
  )
)

(define gl
(GetLink
  (TypedVariableLink
    (VariableNode "$f")
    (TypeNode "LambdaLink")
  )
  (AndLink
    (VariableNode "$f")
    (EvaluationLink
      (PredicateNode "minsup")
      (ListLink
        (LambdaLink
          (VariableNode "$top-arg")
          (VariableNode "$top-arg")
        )
        (ConceptNode "texts")
        (NumberNode "2")
      )
    )
  )
)
)
(cog-execute! gl)
```
raises the following error
```
ERROR: In procedure opencog-extension:
ERROR: Throw to key `C++-EXCEPTION' with args `("cog-execute!" "Not implemented! Need to implement a stack, here. (/home/nilg/OpenCog/atomspace/opencog/query/DefaultPatternMatchCB.cc:299)\nFunction args:\n((GetLink\n   (TypedVariableLink\n      (VariableNode \"$f\")\n      (TypeNode \"LambdaLink\")\n   )\n   (AndLink\n      (VariableNode \"$f\")\n      (EvaluationLink\n         (PredicateNode \"minsup\")\n         (ListLink\n            (LambdaLink\n               (VariableNode \"$top-arg\")\n               (VariableNode \"$top-arg\")\n            )\n            (ConceptNode \"texts\")\n            (NumberNode \"2.000000\")\n         )\n      )\n   )\n)\n)")'.
```

ngeiswei:
I'm re-opening as my last fix didn't do it. I have an elegant fix in mind but it's too for me now so I'll implement it tomorrow.

PR #1608, opened 2018-02-28
Pattern Matcher Fix

ngeiswei:
Fix issue #1606

### DIFF
--- a/opencog/atoms/pattern/PatternLink.h
+++ b/opencog/atoms/pattern/PatternLink.h
@@ -174,7 +174,7 @@ public:
 	const HandleSeq& get_fixed(void) const { return _fixed; }
 	const HandleSeq& get_virtual(void) const { return _virtual; }
 
-	bool satisfy(PatternMatchCallback&) const;
+	bool satisfy(PatternMatchCallback&, AtomSpace* as);
 
 	// Return the cached variable groundings.
 	Handle get_groundings() const;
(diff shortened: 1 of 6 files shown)


ngeiswei:
I feel sufficiently confident to merge...

opencog/atoms/pattern/PatternLink.h
```diff
@@ -174,7 +174,7 @@ class PatternLink : public PrenexLink
 	const HandleSeq& get_fixed(void) const { return _fixed; }
 	const HandleSeq& get_virtual(void) const { return _virtual; }
 
-	bool satisfy(PatternMatchCallback&) const;
+	bool satisfy(PatternMatchCallback&, AtomSpace* as);
```
ngeiswei: Actually, I'm not confident about the order of arguments. Should `as` come first or second?

@linas I know that you've commented about argument orders for instance in https://github.com/opencog/atomspace/blob/master/opencog/scm/opencog/base/utilities.scm#L1177 but I don't understand the rationals.

The only rational I have for argument order is from Haskell, put the most frequently changing arguments last, because that way you can simplify a function call by partially calling it on the first arguments (function currying).

ngeiswei:
Oops, I've been overconfident, it breaks 2 other tests, I'm working on it...

ngeiswei:
Yep, it breaks `RecognizerUTest`, removing the constants is obviously wrong since the queries for the recognizer are typically constant terms.

I think the correct place for removing constants is in `set_pattern`, except of course `Recognizer::set_pattern`.

I'm gonna revert https://github.com/opencog/atomspace/pull/1608/commits/3fb11c337a0946266463e09871d2b6e0d1b8866f, sorry about that.

PR #1611, opened 2018-02-28
Changed fact to rule example into dummy executionoutput example

shujingke:
Delete the fact to rule test in UTest as well.


linas:
Why are you deleting this? was something wrong with it?

ngeiswei:
@shujingke please avoid merging the master back into your branch before committing your PR, it tends to make the git history more convoluted than it is.

If you rebase your branch on the master it will look pretty clean, however be sure you fully understand what rebasing is otherwise this may turn into nightmares. In any case merging the master to your branch, or rebasing it on the master should be a rare operation and should only be done if you really need it, for instance if the master contains features that you need.

shujingke:
Sorry .. I knew... I just haven't worked with github for a long time and forgot to new a local branch..

ngeiswei:
BTW, Emacs has a very powerful git front-end called magit, I think it's the most powerful git front-end out there (more powerful than Atom's or VSCode's as far as I read).

shujingke:
@linas Because this test was non-sense... This “fact to rule” issue has already been discussed with Nil, and it turned out to be my misunderstanding. Because I was careless that I commented out a dummy  ExecutionOutput.

Here is the original discussion in a Email thread if you are interested:

Shujing:
```
To make the backward chainer in URE work, the constant truth in
knowledge base needs to be turned into explicit rules. But somehow it still
doesn't work.  Like the below simple example:

;;Only two truth in AtomSpace:
American is a person;
Cat is a pet;

;; One rule: person-keep-pet-rule
If ( A is person ) and (X is pet ), then (A keep X).

Run this rule BindLink alone or run the forward chainer, it will both
output that the (American keep cat). But if I set a target that ($who keep
cat), the backward chainer will output empty result.
If we simulate the backward reasoning process of this problem:
Target 0:  ($who keep cat)
STEP 1: Search for a rule to achieve target 0:
               "person-keep-pet-rule" is found. Variable bound:
               If ( $who is person ) and ( cat is pet ), then ( $who keep
cat ).
Two preconditions of this rule become the next level targets:
Target 1.1:   ( $who is person )
Target 1.2:   ( cat is pet )   (already achieved)

So to achieve Target 1.1, basically the system needs an explicit rule like:
if ($A == American ), then ( $A is person)

Therefore the fact that "American is a person" needs to be converted into such
a rule to make the backward chainer work.
```
Nil:
```
I don't think that's the reason, the following inference tree should output the target

(BindLink
  (TypedVariableLink
    (VariableNode "$who") ; [7437563460386647115][1]
    (TypeNode "ConceptNode") ; [3788696113922021255][1]
  ) ; [11489669533468719910][1]
  (AndLink
    (InheritanceLink
      (VariableNode "$who") ; [7437563460386647115][1]
      (ConceptNode "person") ; [4664786838289451278][1]
    ) ; [10871444790883389195][3]
    (InheritanceLink (stv 1.000000 1.000000)
      (ConceptNode "cat") ; [4498201032894765211][1]
      (ConceptNode "pet") ; [3840025400633503973][1]
    ) ; [14504815651397872946][1]
  ) ; [14278984156087881767][3]
  (ExecutionOutputLink
    (GroundedSchemaNode "scm: dommyoutput") ; [6225619069995320054][1]
    (EvaluationLink
      (PredicateNode "keep-pet") ; [6600004461605315654][1]
      (ListLink
        (VariableNode "$who") ; [7437563460386647115][1]
        (ConceptNode "cat") ; [4498201032894765211][1]
      ) ; [10647385313682827715][1]
    ) ; [18015027962572704632][1]
  ) ; [13583531515789522408][3]
) ; [16225970567237422245][3]

```
```
It passes for me, but I had to uncomment the formula

(define (dummyoutput k)
   (cog-set-tv! k (stv 1 1))
)
I think that you do not need these rule facts. I assume that you're problem is that the formula is commented.
```

linas:
oh ok

PR #1616, opened 2018-03-12
Port a bunch of ure logger interface to guile.

shujingke:
cog-ure-logger-get-level
cog-ure-logger-set-level!
cog-ure-logger-get-filename
cog-ure-logger-set-filename!
cog-ure-logger-set-stdout!
cog-ure-logger-set-sync!
cog-ure-logger-set-timestamp!

### DIFF
--- a/opencog/rule-engine/InferenceSCM.cc
+++ b/opencog/rule-engine/InferenceSCM.cc
@@ -78,15 +78,14 @@ protected:
 
 	Handle get_rulebase_rules(Handle rbs);
 
-    /**
-     *
-     * Set ure logger level
-     *
-     * @param level        The string of the new ure logger level want to set to.
-     *
-     * @return             A string of previous ure logger level.
-     */
+    std::string get_ure_logger_level();
     std::string set_ure_logger_level(const std::string& level);
+    std::string get_ure_logger_filename();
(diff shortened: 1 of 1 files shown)


ngeiswei:
Actually these are rather redundant, you can already access the ure logger https://wiki.opencog.org/wikihome/index.php/Scheme#Using_a_different_logger_than_the_default_one

And if you want precisely these functions it is better to define them as shorthand scheme functions, such as in https://github.com/opencog/opencog/blob/master/examples/pln/inference-control-learning/icl-utilities.scm#L11

linas:
Also, if you still wanted a "convenience" tool, you could, for example, set:
```
(define (cog-ure-logger-set-level! x) (cog-logger-set-level! (cog-ure-logger) x)
```

which provide a single-line implementation.

shujingke:
I think these interfaces are not really unnecessary...  As a relatively old user/developer of Opencog, I couldn't find how to change the logger of ure until I went into its c++ codes, so I ported it from c++. The convenient interfaces could be useful for new users. Because one wouldn't expect that the ure logger can be set from its "mother class" - Logger.  
As you guys pointed out, there were already defined Logger interfaces in scheme, so that I changed to define ure logger setting shortcuts from scheme as you mentioned, see the new pull request #1618

PR #1618, opened 2018-03-15
Port ure logger interface to guile by defining scm functions.

shujingke:


shujingke:
It seems there is not a list of documentation of ure scheme interfaces in wiki. Maybe we should add the ure section into this page: https://wiki.opencog.org/wikihome/index.php/Scheme

opencog/scm/opencog/rule-engine/rule-engine-utils.scm
```diff
@@ -313,6 +314,34 @@
 "
   (map (lambda (x) (gen-rand-variable prefix base length)) (iota n)))
 
+(define (set-ure-logger-level! level)
```
ngeiswei: I think function names like `ure-logger-set-level!` would be better as they would be more consistent with the existing function names. What do you think?

AmeBel:
@ngeiswei what do you mean by "subroutings of the URE "

ngeiswei:
@AmeBel I mean functions for
1. Unifying a rule with a conclusion or premise
2. Expanding an inference tree backward or forward
3. Selecting an inference tree from the inference tree pool
4. Selecting a premise from an inference tree
5. Selecting a valid rule to expand an inference tree

probably other functions as well. This way a user should be able to implement it's own URE in just a few lines of Scheme code.

shujingke:
OK I changed the names. See the new pull request.

PR #1619, opened 2018-03-16
URE logger interfaces with new names.

shujingke:


ngeiswei:
I've moved that in a separate PR and fixed a few things.

PR #1620, opened 2018-03-17
Verifying rule format.

shujingke:
First, assert if a user defined rule is defined as a DefinedSchemaNode with an alias in UREConfig::fetch_common_parameters. Otherwise it will crash (or Assert fail) later when fetching the BindLink of the rule and user has no idea that it is because the rule is not defined as a DefinedSchemaNode with an alias.
Then a verify_rule function is added to Rule class and is called during init(), currently check if the conclusions are wrapped with an ExecutionOutputLink when there are multiple conclusions and give warnings. This function should not be added into Rule::is_valid(), because is_valid() check if the rule has been assigned a valid Handle, which has been used in many places. verify_rule() give warning, should only be used when init a rule.

### DIFF
--- a/opencog/rule-engine/Rule.h
+++ b/opencog/rule-engine/Rule.h
@@ -5,6 +5,7 @@
  *
  * Authors: Misgana Bayetta <misgana.bayetta@gmail.com>  2015
  *          Nil Geisweiller 2015-2016
+ *          Shujing Ke 2018
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU Affero General Public License v3 as
@@ -106,6 +107,16 @@ public:
 	void init(const Handle& rule_member);
 	void init(const Handle& rule_alias, const Handle& rbs);
 	void init(const Handle& rule_alias, const Handle& rule, const Handle& rbs);
+
+    /**
+     * Verify if this rule is defined in the required format.
+     * The main purpose is to give user the corresponding warnings and info.
+     * Only verify a normal rule, do not use this function to verify a meta rule.
+     * Current verify items:
+     * 1.If there are multiple conslusions, they also need to be wrapped with one
+     *   single ExecutionOutPutLink, otherwise it won't work for bc.
+     */
+    bool verify_rule();
 	
 	// Comparison
 	bool operator==(const Rule& r) const;
--- a/opencog/rule-engine/UREConfig.cc
+++ b/opencog/rule-engine/UREConfig.cc
@@ -137,7 +137,15 @@ void UREConfig::fetch_common_parameters(const Handle& rbs)
 {
 	// Retrieve the rules (MemberLinks) and instantiate them
 	for (const Handle& rule_name : fetch_rule_names(rbs))
+    {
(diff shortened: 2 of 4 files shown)


opencog/rule-engine/Rule.h
```diff
@@ -106,6 +107,16 @@ class Rule : public boost::totally_ordered<Rule>
...
+
+    /**
+     * Verify if this rule is defined in the required format.
+     * The main purpose is to give user the corresponding warnings and info.
+     * Only verify a normal rule, do not use this function to verify a meta rule.
+     * Current verify items:
+     * 1.If there are multiple conslusions, they also need to be wrapped with one
+     *   single ExecutionOutPutLink, otherwise it won't work for bc.
+     */
+    bool verify_rule();
```
ngeiswei: Note that this file uses smart tabs, not whitespaces for indentation.

I really like smart tabs, it allows the user to render its code with its chosen tab size, yet while preserving the alignments no matter what, and in the end makes browsing the code slightly faster. The inconvenience is that it requires the developer have an editor that supports it. Maybe the following Stackoverflow issue can help you add smart tabs to VSCode https://stackoverflow.com/questions/2728658/are-there-any-visual-studio-add-ins-for-true-smart-tabs otherwise just use plain tabs.

shujingke: Are you sure want to make "using mart tables" as an Opencog coding rule even in c++? 
I personally fine, but everyone has its own favorite tools. I use Qt Creator. This rule may be an unattractive rules for new people to join the development of Opencog.

shujingke: What about we agree on "using smart tabs" in Scheme, and free to use whatever everyone gets used to in c++?

ngeiswei: Sure, I don't want to force people to use smart tabs, I've edited the dev standard to make it clearer

https://wiki.opencog.org/w/Development_standards#Formatting

basically the most important is to respect the indentation style of the file, and if the style is smart tabs, then regular tabs can be used, no big deal.

shujingke: Ok, understood.

linas: smart tabs or not, this is just badly indented.    The indentation used should, at a minimum, be consistent with the indentation in the rest of the file. This is not.

opencog/rule-engine/UREConfig.cc
```diff
@@ -137,7 +137,15 @@ void UREConfig::fetch_common_parameters(const Handle& rbs)
...
 	for (const Handle& rule_name : fetch_rule_names(rbs))
+    {
+        OC_ASSERT((rule_name->get_type() == DEFINED_SCHEMA_NODE),
+              "The rule: \n%s \n is not a DefinedSchemaNode."
+              "A rule needs an alias and to be defined a DefinedSchemaNode.\n"
+              "Please check rules in /atomspace/examples/rule-engine for example.\n\n",
+              rule_name->to_short_string().c_str());
+
 		_common_params.rules.emplace(rule_name, rbs);
+    }
```
linas: This also is incorrectly indented.   Please fix this.

PR #1624, opened 2018-03-29
fixed README and code comments to match calculation of MI as +log2() …

glicerico:
…and save future readers from the confusion @akolonin and I went through...


glicerico:
Noted, thanks!

PR #1627, opened 2018-04-05
remove benchmark after moving under opencog/benchmark/atomspace

vsbogd:
To be merged after https://github.com/opencog/benchmark/pull/3 is merged


ngeiswei:
```
CMake Error at opencog/CMakeLists.txt:15 (ADD_SUBDIRECTORY):
  ADD_SUBDIRECTORY given source "benchmark" which is not an existing
  directory.
```

vsbogd:
Fixed, thanks. BTW did you try to make it locally or look at CI results?

ngeiswei:
@vsbogd I built locally. To be precise I did
```
git remote add vsbogd git@github.com:vsbogd/atomspace.git
```
or rather `M-x magit-remote-add` under emacs, then check out `move-benchmakrs-into-new-repo`, remove the existing `build` directory and rebuild.

ngeiswei:
Regarding CI, it used to work and should be resurrected soon, including benchmarks.

vsbogd:
Thanks, @mjsduncan, yes it is the thing I was looking for. I also have two failed tests after update from master and I was guessing whether it is reproducible somewhere else.

PR #1632, opened 2018-04-12
Use PROJECT_SOURCE_DIR in unit tests.

vsbogd:
Unit tests use relative paths to load scheme definitions. It makes
impossible running unit tests in any folder except subfolder of source
root.


linas:
Specifically, `make test` in the build directory?

vsbogd:
Sure, @linas, I have checked it.

Issue #1625, opened 2018-03-31
Hash collision between atoms


ngeiswei:
The following, rather similar but different, scope links have the same hash
```
      (LambdaLink
        (VariableList
          (VariableNode "$PM-565f9848") ; [3051023176053525644][1]
          (VariableNode "$PM-164e1b09") ; [5471808729358182586][1]
        ) ; [14860589810072735837][1]
        (ImplicationLink
          (InheritanceLink
            (VariableNode "$PM-565f9848") ; [3051023176053525644][1]
            (VariableNode "$PM-164e1b09") ; [5471808729358182586][1]
          ) ; [14496589888635524888][1]
          (InheritanceLink
            (VariableNode "$PM-565f9848") ; [3051023176053525644][1]
            (VariableNode "$PM-164e1b09") ; [5471808729358182586][1]
          ) ; [14496589888635524888][1]
        ) ; [14820499435918140327][1]
      ) ; [9607689818773494246][1]
```
```
      (LambdaLink
        (VariableList
          (VariableNode "$PM-164e1b09") ; [5471808729358182586][1]
          (VariableNode "$PM-164e1b09-47189056") ; [7316959371767173953][1]
        ) ; [13581315379842424018][1]
        (ImplicationLink
          (InheritanceLink
            (VariableNode "$PM-164e1b09") ; [5471808729358182586][1]
            (VariableNode "$PM-164e1b09") ; [5471808729358182586][1]
          ) ; [11372164815996221702][1]
          (InheritanceLink
            (VariableNode "$PM-164e1b09-47189056") ; [7316959371767173953][1]
            (VariableNode "$PM-164e1b09-47189056") ; [7316959371767173953][1]
          ) ; [9543682399918497524][1]
        ) ; [17442028992361417521][1]
      ) ; [9607689818773494246][1]
```

ngeiswei:
Some more background about that issue in case someone not familiar with OpenCog decides to take it on.

Hash collision is unavoidable in principle and, when it happens, is well handled by OpenCog, still we would like to minimize it in practice. Here's why
1. Collision is costly to manage, for instance when comparing 2 different atoms, if they happen to have different hashes, comparison is instantaneous. Otherwise it must recursively compare their outgoings, etc.
2. Since collision is rare, seeing the same hashes on 2 different atoms tends to catch the attention of the user/dev and raises a false alarm, thus wasting his/her time.

Here are a some guidelines to help to fix this issue
[...]

PR #1636, opened 2018-04-19
use fnv1a hash function for ScopeLink (issue #1625)

noskill:


ngeiswei:
That looks cool @noskill! Could you move your commit to a branch of its own. If you are inexperienced with git you find that page useful https://wiki.opencog.org/w/Development_standards#Git_HOWTO

ngeiswei:
BTW, to really tell github that this PR is related to issue #1625 you need to have it appear in the conversation (like I just did) or in a git commit message. It won't be registered if it is just in the github title of the PR.

The other thing that could be improved is some more comment explaining why and how your change solves the problem (that could be as simple as linking to the wikipage of the hash encoding you've chosen, etc).

PR #1637, opened 2018-04-21
cog-af get top N atoms

misgeatgit:
A function that returns top N (based on STI value) atoms subset of the AF; a feature requested by the ghost-script team.


linas:
A better API would be to just add an (optional) argument to `cog-af`.  If the argument is missing (or if it is -1) then return all atoms; else return the top n.

misgeatgit:
@linas I have figured it out now with the help I got from @AmeBel. Please see PR #1642.

PR #1638, opened 2018-04-21
use fnv1a hash function for ScopeLink(issue #1625)

noskill:


ngeiswei:
@noskill thanks a lot for your contribution, the code looks nice and clean. However, I still managed to stumble on cases that it doesn't handle, like:
```
      (LambdaLink
        (VariableList
          (VariableNode "$PM-41da0db7-3475e0e1") ; [436842134846457090][1]
          (VariableNode "$PM-61bb02a1") ; [3581365595255232229][1]
          (VariableNode "$PM-61bb02a1-5779870f") ; [3194784765217401525][1]
        ) ; [10113316890528704563][1]
        (ImplicationLink
          (InheritanceLink
            (VariableNode "$PM-41da0db7-3475e0e1") ; [436842134846457090][1]
            (VariableNode "$PM-61bb02a1") ; [3581365595255232229][1]
          ) ; [9348520726392294521][1]
          (InheritanceLink
            (VariableNode "$PM-41da0db7-3475e0e1") ; [436842134846457090][1]
            (VariableNode "$PM-61bb02a1-5779870f") ; [3194784765217401525][1]
          ) ; [18185311933209239625][1]
        ) ; [14643635789851217497][1]
      ) ; [10019169672645051183][1]
```
and
```
      (LambdaLink
        (VariableList
          (VariableNode "$PM-61bb02a1-5779870f") ; [3194784765217401525][1]
          (VariableNode "$PM-61bb02a1") ; [3581365595255232229][1]
          (VariableNode "$PM-41da0db7-3475e0e1") ; [436842134846457090][1]
        ) ; [13158986756314112243][1]
        (ImplicationLink
          (InheritanceLink
            (VariableNode "$PM-61bb02a1-5779870f") ; [3194784765217401525][1]
            (VariableNode "$PM-61bb02a1") ; [3581365595255232229][1]
          ) ; [17350279196940478604][1]
          (InheritanceLink
            (VariableNode "$PM-41da0db7-3475e0e1") ; [436842134846457090][1]
            (VariableNode "$PM-61bb02a1-5779870f") ; [3194784765217401525][1]
          ) ; [18185311933209239625][1]
        ) ; [11223876249152793804][1]
      ) ; [10019169672645051183][1]
```
Although these 2 scopes are very similar looking are different, one means
```
Lambda X1 X2 X3: X1->X2 and X1->X3
```
while the other means
```
Lambda X1 X2 X3: X1->X2 and X3->X1
```

Do you think you could fix your PR to encompass that case (with the corresponding unit test)?

If so that would be awesome as these sort of things can be tricky to get right.

tests/atoms/scopelinks.scm
```diff
@@ -732,3 +732,40 @@
...
+             (VariableNode "$PM-565f9848")
+             (VariableNode "$PM-164e1b09")
+           )
+           (InheritanceLink
+             (VariableNode "$PM-565f9848")
+             (VariableNode "$PM-164e1b09")
+           )
+         )
+    )
+)
```
ngeiswei: Also, could you fix the indentation here to be more regular.

tests/atoms/scopelinks.scm
```diff
@@ -732,3 +732,40 @@
...
+            (VariableNode "$PM-164e1b09")
+            (VariableNode "$PM-164e1b09")
+          )
+          (InheritanceLink
+            (VariableNode "$PM-164e1b09-47189056")
+            (VariableNode "$PM-164e1b09-47189056")
+          )
+        )
+      )
+)
```
ngeiswei: Save comment about the indentation.

opencog/atoms/core/ScopeLink.cc
```diff
@@ -313,29 +338,21 @@ ContentHash ScopeLink::term_hash(const Handle& h,
...
-	ContentHash mixer = (ContentHash) is_ordered;
+	// have good mixing/avalanching properties.
 	ContentHash hsh = ((1UL<<8) - 59) * t;
+	std::vector<ContentHash> hash_vec;
 	for (const Handle& ho: h->getOutgoingSet())
 	{
-		hsh += mixer * (hsh<<5) + term_hash(ho, bound_vars, quotation);
+		hash_vec.push_back(term_hash(ho, bound_vars, quotation));
+	}
+	std::sort(hash_vec.begin(), hash_vec.end());
```
ngeiswei: I suspect `hash_vec` should be sorted only for unordered links.

ngeiswei:
Here's another collision
```
      (LambdaLink
        (VariableList
          (VariableNode "$PM-61bb02a1-5779870f") ; [3194784765217401525][1]
          (VariableNode "$PM-61bb02a1") ; [3581365595255232229][1]
        ) ; [17714279118377689553][1]
        (ImplicationLink
          (InheritanceLink
            (VariableNode "$PM-61bb02a1-5779870f") ; [3194784765217401525][1]
            (VariableNode "$PM-61bb02a1") ; [3581365595255232229][1]
          ) ; [17350279196940478604][1]
          (InheritanceLink
            (VariableNode "$PM-61bb02a1-5779870f") ; [3194784765217401525][1]
            (VariableNode "$PM-61bb02a1-5779870f") ; [3194784765217401525][1]
          ) ; [16963698366902647900][1]
        ) ; [10002262682846202079][1]
      ) ; [10698402097715356724][1]
```
```
      (LambdaLink
        (VariableList
          (VariableNode "$PM-61bb02a1") ; [3581365595255232229][1]
          (VariableNode "$PM-61bb02a1-5779870f") ; [3194784765217401525][1]
        ) ; [11638121605878720465][1]
        (ImplicationLink
          (InheritanceLink
            (VariableNode "$PM-61bb02a1") ; [3581365595255232229][1]
            (VariableNode "$PM-61bb02a1") ; [3581365595255232229][1]
          ) ; [11660702514479340220][1]
          (InheritanceLink
            (VariableNode "$PM-61bb02a1") ; [3581365595255232229][1]
            (VariableNode "$PM-61bb02a1-5779870f") ; [3194784765217401525][1]
          ) ; [11274121684441509516][1]
        ) ; [10247468253117788991][1]
      ) ; [10698402097715356724][1]
```
If you can add a unit test for that one as well that'd be awesome.

noskill:
I added tests + 64bit hashing constants

opencog/atoms/core/ScopeLink.cc
```diff
@@ -313,29 +358,24 @@ ContentHash ScopeLink::term_hash(const Handle& h,
...
-
-	// Prevent mixing for UnorderedLinks. The `mixer` var will be zero
-	// for UnorderedLinks. The problem is that two UnorderdLinks might
-	// be alpha-equivalent, but have their atoms presented in a
-	// different order. Thus, the hash must be computed in a purely
-	// commutative fashion: using only addition, so as to never create
-	// any entropy, until the end.
-	//
-	// XXX As discussed in issue #1176, a better fix would be to
+	// XXX As discussed in issue #1176,
```
ngeiswei: @noskill You could now remove the XXX and reformat the text to form a nice paragraph (Alt-q in Emacs).

ngeiswei:
That looks a bit like voodoo to me but I was able to generate thousands of looking alike scopes without a single hash collision, that's good enough for me.

@linas, do you want to have a look at it before I merge?

Anyway, thank you so much @noskill !

ngeiswei:
Actually, before merging we want to have a look at impact on performance.

opencog/atoms/core/ScopeLink.cc
```diff
@@ -313,29 +358,23 @@ ContentHash ScopeLink::term_hash(const Handle& h,
...
-	// XXX As discussed in issue #1176, a better fix would be to
-	// compute the individual term_hashes first, then sort them,
-	// and then mix them!  This provides the desired qualities:
-	// different unordered links can be directly compared, and also
-	// have good mixing/avalanching properties. The code below
-	// only allows for compare; it fails to mix.
-	//
-	bool is_ordered = not classserver().isA(t, UNORDERED_LINK);
-	ContentHash mixer = (ContentHash) is_ordered;
+	// As discussed in issue #1176, compute the individual term_hashes
```
linas: Don't use github issues as replacements for actual documentation. Make the documentation sufficiently clear first, and then, for more info, refer to actual perma-URL's. Don't refer to issue #'s because its not clear which repo that issue is from  (we've got maybe 50 or 100 repos now).

linas: Also, deletion of the comment above is... dangerous: it says something important about how unordered links need to be handled: they behave very differently than ordered links.  Loosing track of this will create subtle, confusing bugs.

linas:
Yeah, a before-n-after benchmark would be nice.  Does a slower hash outweigh the cost of hash bucket collisions?

I'm kind-of surprised that there are collisions, and that they impact performance...

opencog/atoms/core/ScopeLink.cc
```diff
@@ -313,29 +358,23 @@ ContentHash ScopeLink::term_hash(const Handle& h,
...
 	ContentHash hsh = ((1UL<<8) - 59) * t;
+	std::vector<ContentHash> hash_vec;
 	for (const Handle& ho: h->getOutgoingSet())
 	{
-		hsh += mixer * (hsh<<5) + term_hash(ho, bound_vars, quotation);
+		hash_vec.push_back(term_hash(ho, bound_vars, quotation));
+	}
+	// hash_vec should be sorted only for unordered links
+	if (classserver().isA(t, UNORDERED_LINK)) {
+		std::sort(hash_vec.begin(), hash_vec.end());
```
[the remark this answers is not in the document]
linas: Specifically, the actual sort order used should be identical to that in the constructors in `atoms/core/UnorderedLink.cc`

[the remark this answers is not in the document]
linas: See also the `unorder()` method in `atoms/core/FreeLink.cc`. 

*If* you can guarantee the the ctors for these various link types have already run, by the time that you compute the hash, then you are home-free.  However, I am not completely sure if they have.  In the past, many of these links types did not have ctors, and so the sorting was done ad-hoc, in the atomspace.   However, I think this has now been fixed. I think.  I just did a `grep -r sort *` and do not see any problem areas.  Still, caution is urged...

opencog/atoms/core/ScopeLink.cc
linas: Specifically, its not clear to me that you understood the comment about alpha-equivalence and commutativity.  Specifically,
```
    Lambda A B
          Set A B
```
is alpha-equivalent to 
```
Lambda  P Q
     Set P Q
```
but the UnordererdLink for `(Set A B)` and the unorderedLink for `(Set P Q)` might place the contents in a different order.  Thus, if you use mixing during the hash computations, you will get two different hashes, when, in fact, you should have gotten exactly the same hash for both cases, viz, independent of the order in the unordered link.

I mean, I would need to look at this much more closely, to be convinced that you are doing the right thing.  Are you SURE you are doing the right thing?

opencog/atoms/core/ScopeLink.cc
ngeiswei: @linas said
> Specifically, the actual sort order used should be identical to that in the constructors in `atoms/core/UnorderedLink.cc`

I don't see why it would need to be identical, at the very least it only needs to be compatible, when the hashes are different, and obviously it is, since `Link::operator<` and `Node::operator<` rely on the hashes, when they are different. But again, even that doesn't seem necessary, it's just to compute the hash of the scope, nothing else.

opencog/atoms/core/ScopeLink.cc
ngeiswei: I tested and it works (using different atomspaces to be sure not to enable automatic alpha-conversion). This is consistent with my understanding of the code since only considers the variable indexes as opposed to their names (you wrote that code).

opencog/atoms/core/ScopeLink.cc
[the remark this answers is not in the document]
ngeiswei: @linas said
> Specifically, the actual sort order used should be identical to that in the constructors in 
atoms/core/UnorderedLink.cc

I don't see why it could need to be the same
Are you sure? Why would it need to be? It's just to calculate the hash. And if variables are renamed to account for alpha-equivalence, it likely isn't the same already, right?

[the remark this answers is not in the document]
ngeiswei: OK, I see what how the `UnorderedLink` constructor affects `Link::compute_hash`. However, I don't think it has any impact on how this code work, because the hash obtained by `ScopeLink::term_hash` is simply different from the hash obtained from `Link::compute_hash`, and the later isn't used at all.

That is said, it is indeed good for performances if both return the same order as then it makes the sorting in `ScopeLink::term_hash` cheaper (cause it's already sorted), but that should be all.

linas:
OK. Then I see two issues blocking this:

1. Lack of any measurement of performance impact.

2. Verification that nothing got broken, viz that `ScopeLinkUTest` is performing a sufficient amount of test coverage.

I'm looking at 2. now. As to point 1 -- at this point its a coin toss. @ngeiswei do you care, or should this be merged?

linas:
I went ahead and merged. Performance-wise, the new hash functions might be maybe slightly slower, but this depends a lot on what the compiler decides to do, so I figure its a wash, performance-wise.  Whatever  pipeline bubbles may have been created or eliminated, its not worth fiddling with.

As to correctness, I think I have to retract my earlier concerns about sorting: the new code sorts the hash values, and since these are known-correct, then, an unordered link is sortable according to the hashes, so my imagined bogey-man goes away.  Or, at least, I'm pretty sure about that. The unit tests pass.

linas:
Hmm. BackwardChainerUTest now fails.

linas:
Crap. Before merging this, 5 out of 5 passes of the BackwardChainerUTest pass. After merging this, 5 out of 5 fail.  So, in fact, this pull req does ... change something.   It is possible that this pull req is buggy; it's possible that it is exposing some bug in the backwards chainer.  I cannot tell which.

@ngeiswei this is now back in your lap: do you want to revert this change? Do you want to investigate the backward chainer? Do you want to revert this, until such time  that you have time to investigate further?

ngeiswei:
@linas No need to revert the merge, I'll look into the BC utests.

PR #1646, opened 2018-04-26
Use const auto& in loops to prevent copying

vsbogd:
Originally it was introduced to improve performance of
getIncomingSetSize() method. I have visited all such places in code
then.

Benchmark results:
$ ./atomspace/atomspace/atomspace_bm -m getIncomingSetSize -n 10000000

Before optimization:
Benchmarking AtomSpace's getIncomingSetSize method 10000000 times ..........
Sum clock() time for all requests: 1873069 (1.87307 seconds, 5.33883e+06 requests per second)

After optimization:
Benchmarking AtomSpace's getIncomingSetSize method 10000000 times ..........
Sum clock() time for all requests: 1062651 (1.06265 seconds, 9.41043e+06 requests per second)

### DIFF
--- a/opencog/atoms/base/Atom.cc
+++ b/opencog/atoms/base/Atom.cc
@@ -363,7 +363,7 @@ size_t Atom::getIncomingSetSize() const
     std::lock_guard<std::mutex> lck (_mtx);
 
     size_t cnt = 0;
-    for (const auto pr : _incoming_set->_iset)
+    for (const auto& pr : _incoming_set->_iset)
         cnt += pr.second.size();
     return cnt;
 }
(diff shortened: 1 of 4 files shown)


ngeiswei:
Good catch. This reminded me that cppcheck is supposed to detect that sort of things, unfortunately it didn't (maybe it doesn't support c++11 that well). The following command
```
cppcheck atomspace/ --enable=performance --std=c++11 --output-file=cppcheck.out
```
didn't include any of these.

vsbogd:
Looks like auto is not well supported by cppcheck: https://trac.cppcheck.net/ticket/4345

linas:
Yeah thanks for catching this! That;'s core code there.

Issue #1599, opened 2018-02-26
AttentionUTest fails


linas:
As of right now, AttentionUTest fails.

linas:
Confirm that #1652 fixes this.

PR #1652, opened 2018-04-30
Fix AttentionUTest

vsbogd:
Fix for https://github.com/opencog/atomspace/issues/1599

### DIFF
--- a/opencog/attentionbank/AttentionBank.cc
+++ b/opencog/attentionbank/AttentionBank.cc
@@ -323,15 +323,15 @@ Handle AttentionBank::getRandomAtomNotInAF(void)
     std::lock_guard<std::mutex> lock(AFMutex);
     auto find_in_af = [this](const Handle& h){
        auto it = std::find_if(attentionalFocus.begin(), attentionalFocus.end(),
-                [h](std::pair<Handle, AttentionValuePtr> hsp) { return hsp.first == h; });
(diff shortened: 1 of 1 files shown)


opencog/attentionbank/AttentionBank.cc
```diff
@@ -323,15 +323,15 @@ Handle AttentionBank::getRandomAtomNotInAF(void)
...
     auto find_in_af = [this](const Handle& h){
        auto it = std::find_if(attentionalFocus.begin(), attentionalFocus.end(),
-                [h](std::pair<Handle, AttentionValuePtr> hsp) { return hsp.first == h; });
+                [h](const std::pair<Handle, AttentionValuePtr>& hsp) { return hsp.first == h; });
        return it;
     };
 
     int count = 50; // We might get stuck in the while loop if there are no atoms outside the AF
     Handle h = Handle::UNDEFINED;
-    while(find_in_af(h) != attentionalFocus.end() and --count > 0){
```
vsbogd: Handl::UNDEFINED doesn't present in attentionalFocus, so while conditions is false even on first iteration.

linas: FYI, Just saying `Handle h;` already initializes h to `UNDEFINED`, so the extra assignment is un-needed.

linas:
Thanks!

Issue #1507, opened 2017-12-26
Implement NextLink


linas:
The NextLink is an idea to alter the pattern matcher to return one result at a time, rather than waiting for it to find all results.  It is meant to replace the `cog-bind-first-n` scheme function.  It should work something like this:
```
NextLink
    anchor-atom
    PatternLink
```

Here, `PatternLink` might be `GetLink` or `BindLink` or `DualLink` - any of the pattern matching links.  The anchor-atom can be any atom, it will be used to distinguish different pattern searches running in parallel. (Its not obvious that the anchor is needed, so maybe it should be optional?)

Each time `NextLink` is called, another result atom is returned. If there are no results, then simple `(Atom)` is returned, to denote the empty result.

A reasonable implementation would be to cause the `PatternLink` to run in a distinct thread. The purpose of this thread is not to speed up or parallelize, but simply to make it easier to manage the internal state of the pattern matcher.  That is, the `NextLink` is a de-facto continuation, and continuations need to store state somewhere. For this particular case, it seems easiest to store the pattern matcher state in a distinct thread.

Bonus points: since `NextLink` is essentially a "call with current continuation" (or call/cc) it might be usable with other link types, e.g. `PutLink` or with `FilterLink` or with `MapLink`, all of which accept sets as input and return sets as output. That is, NextLink could be used to iterate over the elements of a set. Whether or not this is a good idea is unclear; the bonus-points are to clarify this.

linas:
There are several ways to think of the above proposal.

1) launch a thread to find all pattern bindings, and use the C++ NextLink::execute() method to return answers one-by-one

2) do lazy evaluation: i.e. search for and return only one answer at a time; defer computation until it is needed.

linas:
See also #1645 "Parallelize the pattern matcher" for discussion of what a scatter-gather/map-reduce type API might be appropriate for pattern matching.

linas:
See also #1750 as maybe a much better way of publishing search results.

Issue #1502, opened 2017-12-22
Stop using SetLink for search results!


linas:
`SetLink` has a number of technical issues with respect to the core concept of atoms and how atoms should be used.  Basically, SetLink is a bad citizen.  The only answer I can find is to simply excommunicate, banish SetLink from the Kingdom.  The correct solution is to use `MemberLink` instead. (See however, other ideas below.)

Unfortunately, `BindLink` and `GetLink`, when executed, return `SetLink`'s containing the search results.  This needs to be replaced by `MemberLink`, and specifically, I think the following form might work best:
```
MemberLink
    SomeSearchResultAtom
    AnchorNode "pattern-9a30bd4cbdcd6"
```
and then have `cog-execute!` return the `AnchorNode` -- all search results can then be found, attached to this anchor. 

Q: Why is SetLink bad?
A: See the wiki page for it.  https://wiki.opencog.org/w/SetLink

Q: Why AnchorNode?
A: It seems like the best kind of node for this job.

Q: What other benefits (besides avoiding SetLink)?
A1: Search results can be reported incrementally, as the search proceeds, instead of waiting for one big glob to appear at the end.

A2: The SetLinks that get returned glom up the atomspace, and people usually forget to delete them.  They are hard to delete - you have to actually have them in your hand to delete them - searching for them later on is really hard or impossible.  By contrast, finding and deleting the AnchoreNodes is easy, and can be done at any time.

A3: Its a baby-step to converting the AtomSpace itself to a special kind of MemberLink and/or ContextLink.  Having to pass the AtomSpace around in various methods is a real pain, and causes a lot of programming issues. It would be nice to just get rid of the atomspace entirely, and instead declare membership as:
```
AtomSpaceLink
    AtomSpaceNode "primary atomspace"
    SomeAtom
```
This or something like this could solve a lot of technical problems, and nuking `SetLink` is a natural first step.

~~I believe that this will simplify creating a distributed atomspace.  We could layer everything on top of (for example Redis or Riak and have that do all the work of distributing atoms.~~ I don't believe this any longer. The `StorageNode` is a superior solution for distributed atomspaces.

A4: If we do the above or something like it, we could merge MapLink/BindLink, since they both do almost the same thing, just that MapLink does it for its arguments, while BindLink does it for the entire atomspace.  Similarly GetLink/FilterLink...

Q: Is this a lot of work?
A: You bet. BindLinks are used all over the place.

ngeiswei:
[...]

PR #1664, opened 2018-05-09
Consolidate most attentionbank code to just one location

linas:
The attentionbank code was starting to get invasive, causing dependency problems.

This isolates most of that code to just one directory.


linas:
@misgeatgit please look this over.  This defines a new guile module, which will require you to say  `(use-modules (opencog attention-bank))`  to get the usual scheme `cog-av`, and `cog-af` and etc. functions.  I imagine that there is scheme code in the opencog repo that will need to change to add the `use-modules`  but I am not sure where that code is.

linas:
After grepping, I see that the following need to say `(use-modules (opencog attention-bank))`:
```
ghost/matcher.scm
ghost/procedures/schemas.scm
ghost/stimulation.scm
ghost/test.scm
openpsi/action-selector.scm
nlp/chatbot/chat-utils.scm
```
I think this is the full, exhaustive list.

misgeatgit:
@linas correct. those are the places that need these change. What was the reason for changing the name though?

linas:
changing what name?  I don't think I changed the name of anything ...

linas:
I'll merge in a few hours, unless you have comments ...

opencog/attentionbank/AttentionSCM.cc
```diff
@@ -0,0 +1,214 @@
...
+/**
+ * Return the list of top n atoms in the AttentionalFocus or
+ * return all atoms in the AF if n is unspecified or is larger
+ * than the AF size.
+ */
+Handle AttentionSCM::update_af(int n)
+{
+	AtomSpace* atomspace = SchemeSmob::ss_get_env_as("cog-af");
+
+	Handle af_anchor = atomspace->add_node(ANCHOR_NODE,
```
AmeBel: @linas Why would you want to record the list of atoms in the attention-focus, explicitly in the atomspace?

opencog/attentionbank/AttentionSCM.cc
```diff
@@ -0,0 +1,214 @@
...
+	/* Add all the atoms in the current attentional focus that are not
+	 * already there. */
+	HandleSet curr_af;
+	for (size_t i = isz - N; i < isz; i++)
+	{
+		Handle hi = attentionalFocus[i];
+		auto gone = prev_af.find(hi);
+		if (prev_af.end() == gone)
+		{
+			atomspace->add_link(MEMBER_LINK, hi, af_anchor);
```
AmeBel: This could create a nested list of `MemberLinks` if `prev_af` has `MemberLink`s used to represent the set of atoms in the attentional-focus(af). The logic = the representative sets are likely to be in the af because they are connected to atoms that are in the af, and `sti` will diffuse to them as a result of which they are likely to enter the af.

linas:
Hi @AmeBel to answer both your questions: the API was designed to return an (unordered) list. There are two ways to represent a list in the atomspace: with a SetLink, or a MemberLink. I picked MemberLink, because SetLink has a variety of technical drawbacks.  The technical reasons for this are detailed in #1502.

A different API, that might be better, would be to use the NextLink construct proposed in #1507 - however, that is not yet implemented, and it's not yet clear how such an API should be used.

As to diffusing STI to anchors: the answer is real simple: don't do that!  Diffuse the STI to things that actually matter, for whatever problem you are trying to solve. Don't diffuse STI to core infrastructure. Perhaps it would be better to have an AnchorLink, with a rule that says "never diffuse to AnchorLink".  There's a long, long list of link types that you never want to diffuse to: BindLink, or and subtype of PatternLink, VariableNode, or any of the VariableTypes, LambdaLink, and probably not the function links -- roughly, you never want to diffuse to any link type that is executable or evaluatable.

linas:
@ngeiswei @AmeBel I just opened #3107 to track this as an attention-allocation agent issue.  That way we can be on-target for that discussion.

PR #1668, opened 2018-05-10
Fix link error for examples

ngeiswei:
This PR addresses linking errors when compiling examples `make examples`.

### DIFF
--- a/examples/c++-api/CMakeLists.txt
+++ b/examples/c++-api/CMakeLists.txt
@@ -6,6 +6,7 @@ ADD_EXECUTABLE(crudexample
 	)
 TARGET_LINK_LIBRARIES(crudexample
 	atomspace
+	attentionbank
 	)
 
 ADD_EXECUTABLE(ureexample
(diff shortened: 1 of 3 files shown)


ngeiswei:
@linas I'm almost confident it is right but please have a look at it anyway, the most contention commit would be this one https://github.com/opencog/atomspace/pull/1668/commits/30a0c99e6b75e688834b2e391aa2bd0a16fc03b4

linas:
Yeah, looks good. There are two nasty (viz, hard-to-fix) circular dependencies in the atsomapce repo. One is libexecution (EvalluationLink) which needs python and guile so that it can run the py: and scm: GroundedPredicateNodes -- but also guile, which needs libexecution for some or another reason.

There's a different circular dependency with libatomutils - the quotation stuff there needs various atoms, but the various atoms need quotation.

The circular dependencies are deeply annoying and sometimes the source of compile-time misery; but seem to be otherwise harmless, once one finds the right cmake file contents

Issue #1669, opened 2018-05-10
Can't find atomspace_api.h


ngeiswei:
After cleaning my build folder and removing installed opencog includes and libs from my system I can no longer compile the atomspace. Getting the following error
```
[ 48%] Building CXX object opencog/cython/CMakeFiles/PythonEval.dir/PythonEval.cc.o
/home/nilg/OpenCog/atomspace/opencog/cython/PythonEval.cc:42:10: fatal error: opencog/atomspace_api.h: No such file or directory
 #include "opencog/atomspace_api.h"
          ^~~~~~~~~~~~~~~~~~~~~~~~~
```

ngeiswei:
The bug has been introduced in 46783ef72

PR #1670, opened 2018-05-10
Fix build issue

AmeBel:
Resolves https://github.com/opencog/atomspace/issues/1669


linas:
Thanks!  Good catch, good fix!

Issue #1640, opened 2018-04-23
Dismiss displaying outgoins across atomspaces


AmeBel:
### Description
When a link is created in a child atomspace between an atom only found in it and another atom only found in its parent, the atom in the parent disappears from the parent atomspace.

### Steps to reproduce
``` scheme
>> guile
GNU Guile 2.2.3
Copyright (C) 1995-2017 Free Software Foundation, Inc.

Guile comes with ABSOLUTELY NO WARRANTY; for details type `,show w'.
This program is free software, and you are welcome to redistribute it
under certain conditions; type `,show c' for details.

Enter `,help' for help.
scheme@(guile-user)> (use-modules (opencog))
scheme@(guile-user)> (define parent-as (cog-atomspace))
scheme@(guile-user)> (define child-as (cog-new-atomspace parent-as))
scheme@(guile-user)> (define p (Node "p"))
scheme@(guile-user)> (cog-prt-atomspace)
(Node "p")
scheme@(guile-user)> (cog-set-atomspace! child-as )
$1 = #<atomspace 0x55555555541f12b5>
scheme@(guile-user)> (cog-prt-atomspace )
(Node "p")
scheme@(guile-user)> (define c (Node "c"))
scheme@(guile-user)> (cog-prt-atomspace )
(Node "p")
(Node "c")
scheme@(guile-user)> (cog-set-atomspace! parent-as )
$2 = #<atomspace 0x5555555554198ac5>
scheme@(guile-user)> (cog-prt-atomspace )
(Node "p")
scheme@(guile-user)> (cog-set-atomspace! child-as )
$3 = #<atomspace 0x55555555541f12b5>
scheme@(guile-user)> (cog-prt-atomspace )
(Node "p")
(Node "c")
scheme@(guile-user)> (Link p c)
$4 = (Link
   (Node "p")
   (Node "c")
)

scheme@(guile-user)> (cog-prt-atomspace )
(Link
   (Node "p")
   (Node "c")
)
scheme@(guile-user)> (cog-set-atomspace! parent-as )
$5 = #<atomspace 0x5555555554198ac5>
scheme@(guile-user)> (cog-prt-atomspace )
scheme@(guile-user)> (cog-set-atomspace! child-as )
$6 = #<atomspace 0x55555555541f12b5>
scheme@(guile-user)> (cog-prt-atomspace )
(Link
   (Node "p")
   (Node "c")
)
```

### Repository details
[...]

PR #1674, opened 2018-05-12
#1640 - some atoms might not be printed by cog-prt-atomspace

konstantin-t:
This fix addresses the symptoms of the problem only.
But it seems like the root cause lies deeper - the nested atomspaces are not enough isolated from each other.

I suppose that the right behavior is that children atomspaces must see all atoms from its parents, but not vice-versa. Otherwise the idea of using children atomspaces for temporary computations is compromised, as in fact they might pollute scopes of their parents.

Even if it is so, I do not think that fixing this behavior is a top priority task. @linas, @ngeiswei, what is your opinion?

### DIFF
--- a/opencog/scm/opencog/base/utilities.scm
+++ b/opencog/scm/opencog/base/utilities.scm
@@ -324,10 +324,6 @@
   those atoms that have no incoming set in the atomspace or its ancestors,
   and thus are at the top of a tree.  All other atoms (those which do
   have an incoming set) will appear somewhere underneath these top-most atoms.
-
-  N.B. That means, if the atomspace contains a link, refering an atom in a
-  child or independent atomspace, the printed result will include this
-  alien atom too.
 "
 	(traverse-roots display)
 )
--- a/opencog/scm/opencog/base/utilities.scm
+++ b/opencog/scm/opencog/base/utilities.scm
@@ -288,22 +288,48 @@
 )
 
 ; -----------------------------------------------------------------------
+(define (traverse-roots func)
+"
+  traverse-roots -- Applies func to every root atom in the atomspace.
+
+  The root atoms are those, which have no incoming atoms,
+  located in the atomspace or its ancestors (i.e. visible from the atomspace).
+"
+	(define (is-visible? atom)
+		(member
+			(cog-as atom)
+			(get-atomspace-and-parents
+				(cog-atomspace)
+				`())))
+	(define (get-atomspace-and-parents atomspace res)
+		(if (null? atomspace)
+			res
+			(get-atomspace-and-parents
+				(cog-atomspace-env atomspace)
+				(cons atomspace res))))
+
+	(define (apply-if-root h)
+		(if (null? (filter is-visible? (cog-incoming-set h)))
+			(func h))
+		#f)
+
+	(for-each (lambda (ty) (cog-map-type apply-if-root ty)) (cog-get-types))
+)
+; -----------------------------------------------------------------------
 (define-public (cog-prt-atomspace)
 "
   cog-prt-atomspace -- Prints all atoms in the atomspace
 
   This will print all of the atoms in the atomspace: specifically, only
-  those atoms that have no incoming set, and thus are at the top of a
-  tree.  All other atoms (those which do have an incoming set) will
-  appear somewhere underneath these top-most atoms.
-"
-	(define (prt-atom h)
-		; Print only the top-level atoms.
-		(if (null? (cog-incoming-set h))
-			(display h))
-		#f)
+  those atoms that have no incoming set in the atomspace or its ancestors,
+  and thus are at the top of a tree.  All other atoms (those which do
+  have an incoming set) will appear somewhere underneath these top-most atoms.
 
-	(for-each (lambda (ty) (cog-map-type prt-atom ty)) (cog-get-types))
+  N.B. That means, if the atomspace contains a link, refering an atom in a
+  child or independent atomspace, the printed result will include this
+  alien atom too.
+"
+	(traverse-roots display)
 )
 
 ; -----------------------------------------------------------------------
--- a/opencog/guile/SchemeSmobAS.cc
+++ b/opencog/guile/SchemeSmobAS.cc
@@ -232,7 +232,7 @@ SCM SchemeSmob::ss_as_uuid(SCM sas)
 
 /* ============================================================== */
 /**
- * Return parent of the atomspace
+ * Return parent of the atomspace, or an empty list if there is no parent
  */
 SCM SchemeSmob::ss_as_env(SCM sas)
 {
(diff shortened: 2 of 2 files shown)


opencog/scm/opencog/base/utilities.scm
```diff
@@ -287,23 +287,49 @@
...
-		(if (null? (cog-incoming-set h))
-			(display h))
-		#f)
+  those atoms that have no incoming set in the atomspace or its ancestors,
+  and thus are at the top of a tree.  All other atoms (those which do
+  have an incoming set) will appear somewhere underneath these top-most atoms.
 
-	(for-each (lambda (ty) (cog-map-type prt-atom ty)) (cog-get-types))
+  N.B. That means, if the atomspace contains a link, refering an atom in a
+  child or independent atomspace, the printed result will include this
```
linas: This statement is misleading; its impossible for an atomspace to contain a link that refers to an atom that is in a child atomspace, or to some atom in some independent atomspace.  (If you are able to construct such beasts, then there is also a bug somewhere else)

linas:
The current (and intended) design is such that children atomspaces do see all atoms from its parents, and never-ever vice-versa. I believe that it is working as designed.

Using pull requests to ask questions or to raise doubts is poor programming practice: those messages will get cryptographically hashed into the history of git, and will leave future readers confused as to what's going on. Can you redo this pull request, without the additional commentary?  You should open an issue if you have doubts, questions, bugs or need general design discussions to take place.

opencog/scm/opencog/base/utilities.scm
konstantin-t: I think so also.

Please try to run this

```
(use-modules (opencog))
(define parent-as (cog-atomspace))
(define child-as (cog-new-atomspace parent-as))
(define p (Node "p"))
(cog-set-atomspace! child-as )
(define c (Node "c"))
(cog-set-atomspace! parent-as )
(Link p c)
(cog-set-atomspace! child-as )
(cog-prt-atomspace )
```

I did not check the internals, but it works. With the updated print function it otputs

```
(Node "c")    
(Link         
   (Node "p") 
   (Node "c") 
)
```

konstantin-t:
Sure.

linas:
thanks. Otherwise, the changes looked good. 

One minor scheme coding remark: a common convention (which I sometimes use; maybe half the code does not follow it) is to write function arguments in UPPERCASE. This makes it easier to read the code, as  it makes it easier to figure out where some variable came from. For some reason, scheme is confusing in that way, and so distinguishing local variables from arguments is useful.  This appercase thing, though, is usually used only for the main function, and not for any internal functions.  I  don't know where this tradition comes from; I did not invent it.

opencog/scm/opencog/base/utilities.scm
linas: Ah. That's interesting. But what you see is perhaps not what you think you see. There are now not one, but two different `(Node "c")` in the system; one that is in the parent atomspace, one that is in the child atomspace.  Unfortunately, the current print statement does not show this, and so its confusing. 
Try this. After the above,
```
(Node "c" (stv 0.5 0.5))
(cog-prt-atomspace)
(cog-set-atomspace! parent-as )
(cog-prt-atomspace)
(Node "c" (stv 0.6 0.6))
(cog-prt-atomspace)
(cog-set-atomspace! child-as )
(cog-prt-atomspace)
``` 
You now see two different truth values, that's because they are in fact distinct atoms, each living in a different atomspace.  When you created `(Link p c)` in the parent AS, the parent AS looked at the c, saw that it was not yet a member of the atomspace, and so cloned it, and kept the clone (every atomspace alwways clones what it is given, so that the atoms it keeps are always thier own private copies, owned by that atomspace).

I suppose that we could change the behavior so that, if `c` belongs to some other atomspace, then the insertion and cloning would be rejected, and an error thrown.  That could reduce confusion and mistakes, I suppose.

linas: Hmm. Altering this behavior is not so easy, it turns out. There seem to be a variety of cases where atoms are created in a temporary child atomspace, during pattern matching or during other kinds of speculative execution, and then, once they are confirmed as "valid" results, are copied from the child into the main, parent atomspace.  A quick check caused twelve unit tests to fail.

It still might be worth fixing this; but it will require some work.  This is what I tested:
```
--- a/opencog/atomspace/AtomTable.cc
+++ b/opencog/atomspace/AtomTable.cc
@@ -229,6 +229,9 @@ Handle AtomTable::getHandle(Type t, const std::string& n) const
 
 Handle AtomTable::getNodeHandle(const AtomPtr& orig) const
 {
+    // If the atomspace is not null, then there's some bug in the calling code.
+    OC_ASSERT(nullptr == orig->_atom_space, "Incorrect call sequence!");
+
     AtomPtr a(orig);
 
     ContentHash ch = a->get_hash();
@@ -256,6 +259,9 @@ Handle AtomTable::getHandle(Type t, const HandleSeq& seq) const
 
 Handle AtomTable::getLinkHandle(const AtomPtr& orig) const
 {
+    // If the atomspace is not null, then there's some bug in the calling code.
+    OC_ASSERT(nullptr == orig->_atom_space, "Incorrect call sequence!");
+
     AtomPtr a(orig);
     Type t = a->get_type();
     const HandleSeq &seq = a->getOutgoingSet();
```

konstantin-t:
Understood. Here I just adhered to the common style.

opencog/scm/opencog/base/utilities.scm
konstantin-t: Seems like such behavior was there by design.

Check row 376
```
// If this atom is in some other atomspace or not in any atomspace,
// then we need to clone it...
```

linas: Yea, I wrote essentially 100% of the code in that file. Maybe the old authors names should be removed -- actually I've written maybe 200% or 300% of the code in that file -- re-written everything maybe 2-3 times, now. It will be re-written again -- one of the open design issues is to rework it so that the atomspace becomes a link type, needed to port everything to apache ignite, which would alter some of the parent-child atomspace relationships. Have not fully thought it through.

opencog/scm/opencog/base/utilities.scm
```diff
@@ -287,23 +287,45 @@
...
+			(cog-as atom)
+			(get-atomspace-and-parents
+				(cog-atomspace)
+				`())))
+	(define (get-atomspace-and-parents atomspace res)
+		(if (null? atomspace)
+			res
+			(get-atomspace-and-parents
+				(cog-atomspace-env atomspace)
+				(cons atomspace res))))
```
ngeiswei: I think the implementation of `get-atomspace-and-parents` can be simplified using `fold` https://srfi.schemers.org/srfi-1/srfi-1.html#FoldUnfoldMap

It's a minor remark, the code looks good anyway.

konstantin-t: I will check, thanks.

konstantin-t: It was a good idea. Check #1692 1692

PR #1678, opened 2018-05-14
Improve QuantityLink comment

ngeiswei:
I've also added wiki page https://wiki.opencog.org/w/QuantityLink

I chose to have the unit comes before the number, let me know if you have a problem with that.

### DIFF
--- a/opencog/atoms/base/atom_types.script
+++ b/opencog/atoms/base/atom_types.script
@@ -488,8 +488,8 @@ ANCHOR_NODE <- NODE
 // as time intervals, or the lower and upper bounds of some range.
 INTERVAL_LINK <- ORDERED_LINK
 
-// Link for mathematics. (???) Huh? What is this? What does it do?
-// Is this actually used anywhere? Is this documented anywhere?
+// Link to represent the magnitude or the multitude of something, such
+// as 10kg, 10s, $10, etc.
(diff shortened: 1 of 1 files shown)


linas:
Several remarks:
1) Can we get a wiki page for this?
2) can we change the name to UnitsLink?  Clearly, its not at all about quantity, magnitude or multiude, but rather about the units of measurement, so calling it "quantity" is misleading. The concepts of magnitude / quantity are unrelated to the concept of units.   

To be clear: quantity and magnitude are about taking the absolute values of things (e.g. magnitude of a complex vector, root-mean-square of a sine wave, lp-norm of a vector, volume of an n-dimensional space)

By contrast, units have to do with the units of measurement: seconds, dollars, kilograms.  So, most mathematical quantities have no unit: the root-mean-square of a sine wave is just a magnitude, a number without any units attached to it; the root-mean-square of a sine wave of electrical alternating current is measured in units of volts.  The magnitude of a sine wave is 1/1.41421 of the peak of the sine wave. The magnitude of lines voltage is 120 volts or 220 volts, which is 1/1.41421 of the peak voltage on the mains.

linas:
Oh, I see there is a wiki page. now. The name change would still be good, since this has almost nothing to do with quantity.

Issue #1680, opened 2018-05-15
AlphaConvertUTest::test_bindlink fails


ngeiswei:
It fails when comparing
```
hscox = (BindLink
  (VariableNode "$X") ; [6809953386495729144][1]
  (PresentLink
    (VariableNode "$X") ; [6809953386495729144][1]
  ) ; [16058869277486449580][1]
  (AndLink
    (VariableNode "$X") ; [6809953386495729144][1]
  ) ; [16055386024649729858][1]
) ; [12487348734513047317][1]
```
and
```
hscoy = (BindLink
  (PresentLink
    (VariableNode "$X") ; [6809953386495729144][2]
  ) ; [16058869277486449580][2]
  (AndLink
    (VariableNode "$X") ; [6809953386495729144][2]
  ) ; [16055386024649729858][2]
) ; [16295387175712849125][2]
```
which to me points to a problem with `ScopeLink::compute_hash`. Probably surfaced only now due to some alteration in the hash.

ngeiswei:
The weird shit reason is that `_vardecl` is undefined for `hscox` while clearly there is a variable declaration.

ngeiswei:
The bug has been introduced either in baf7991ee or 47add8fa8 (I can't tell since 47add8fa8 doesn't compile).

ngeiswei:
For some reason, after these commits `BindLink::extract_variables` is no longer called, that's weird, it should utterly break everything.

linas:
Closing as a duplicate of #1673

Issue #1673, opened 2018-05-11
Circular dependencies in shared libraries


mjsduncan:
what could be causing this?
```
In FreeLinkUTest::test_getlink:
/home/mozi/oc/atomspace/build/tests/atoms/../../../tests/atoms/FreeLinkUTest.cxxtest:214: Error: Expected (vars.size() == 2), found (6 != 2)
/home/mozi/oc/atomspace/build/tests/atoms/../../../tests/atoms/FreeLinkUTest.cxxtest:216: Error: Expected (vars[1] == vz), found ({ 50 6D F2 2E 71 55 00 00 ... } != { B0 75 F2 2E 71 55 00 00 ... })
[2018-05-11 19:14:48:360] [DEBUG] END TEST: test_getlink
Failed 1 and Skipped 0 of 4 tests
```

linas:
Its one of those tests that never fails; thus, the most likely cause is some version skew: the libraries it is linke to are too old, are different, were build on a different system.  before running tests, do a `make install`.  to avoid having root ownership of files in your build directory, be sure you say `make` before saying `make install`.  If you have files owned by root, then `make` will fail to replace them by the newer (and correct) versions, thus leading to mystery unit test fails. If unsure, do an `rm -r *; cmake ..; make -j; sudo make install; make -j test`

mjsduncan:
already did that, it's happening on ubuntu 17.10 and ubuntu 16.04 after your test overhaul.

linas:
Hmm. None of the code that I touched would affect this unit test, or any unit tests aside from the atomspace, in any way. There were some recent updates to cogutils; that might(??) explain incompatible shared libs.

mjsduncan:
so to be clear, on what versions of ubuntu is this passing for you?

linas:
I'm running Debian stable. I have a hard time believing that this is OS related.perhaps @ngeiswei can provide a data point.

linas:
Debugging info: what does `ldd -r ./tests/atoms/FreeLinkUTest` display?

mjsduncan:
$ ldd -r ./tests/atoms/FreeLinkUTest
	linux-vdso.so.1 =>  (0x00007fff831f8000)
	libsmob.so => /home/cog/opencog/atomspace/build/opencog/guile/libsmob.so (0x00007f2ebc9a7000)
[...]

PR #1681, opened 2018-05-15
Replace lambda by execution library

ngeiswei:
This fixes issue #1680

### DIFF
--- a/tests/atoms/CMakeLists.txt
+++ b/tests/atoms/CMakeLists.txt
@@ -21,7 +21,7 @@ IF(HAVE_GUILE)
 ENDIF(HAVE_GUILE)
 
 ADD_CXXTEST(AlphaConvertUTest)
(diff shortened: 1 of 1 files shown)


ngeiswei:
I'm not extremely confident about the fix, but it does seem right. I suppose the chain of calls was broken somewhere, as `execution` depends on `query` which depends on `lambda`, the BindLink constructor is ultimately called.

linas:
I'm merging for now, but this is really a screwball hack that plasters over a deeper problem -- described in #1673 -- the deeper problem is that there are multiple circular dependencies between some of the shared libs, which means that sometimes they are not properly linked. This in turn means that the library constructors don't run, and the library constructors are used to define the various atom types that the library contains. If the atom types aren't defined ... well .things will "mostly" work, but various C++ methods might be missing (because the atom is not defined...)

PR #1682, opened 2018-05-15
#1673 AlphaConvertUTest::test_bindlink failed

konstantin-t:
This is another approach to #1673. It looks more reliable than existing, but still ugly.


linas:
Should use `${NO_AS_NEEDED}` instead. Also, the right place for this is probably in the various shared-lib Cmakefiles, instead of the individual unit test makefiles.  This is a bug lurking below the depths, potentially biting lots of different binaries at unexpected times and places.

Issue #1687, opened 2018-05-16
Can't construct (Or (Exists ...) ...)


ngeiswei:
The following is considered ill-formed
```
(OrLink
  (ExistsLink
    (VariableNode "$X")
    (EvaluationLink
      (PredicateNode "P")
      (VariableNode "$X")
    )
  )
  (ExistsLink
    (VariableNode "$X")
    (EvaluationLink
      (PredicateNode "Q")
      (VariableNode "$X")
    )
  )
)
```
and raises the exception
```
Invalid Atom syntax: (OrLink ...
```

linas:
Closing. I think this was fixed in #1688 and #1689 and it certainly works for me, now.

PR #1688, opened 2018-05-16
Add OrLink unit test

ngeiswei:
To test `(OrLink (Exists ...` as described in issue #1687


konstantin-t:
This test fails on my computer. Is it expected behavior?

ngeiswei:
Yes, it is expected. This PR #1689 is supposed to fix it but I'm not confident enough to merge it, waiting for Linas' feedback.

linas:
Seems that `ExistsLinkUTest` would be a better name; OrLink itself is fairly trite, while ExistsLink inherits from complicated stuff.

PR #1689, opened 2018-05-16
Make ForAll and Exists links evaluatable

ngeiswei:
Possible fix for #1687 

Not sure it is right but it allows to have then as outgoings of And,
Or, Not links.

### DIFF
--- a/opencog/atoms/base/atom_types.script
+++ b/opencog/atoms/base/atom_types.script
@@ -338,8 +338,8 @@ DUAL_LINK <- SATISFYING_LINK
 // Should these inherit from EvaluatableLink ??
 // Actually, these probably should inherit from PrenexLink, because
 // we typically want to maintain these in prenex normal form.
-FORALL_LINK <- SCOPE_LINK "ForAllLink"
-EXISTS_LINK <- SCOPE_LINK
+FORALL_LINK <- SCOPE_LINK,EVALUATABLE_LINK "ForAllLink"
+EXISTS_LINK <- SCOPE_LINK,EVALUATABLE_LINK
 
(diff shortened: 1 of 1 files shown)


ngeiswei:
@linas you should probably have a look at it. According to this comment
```
// Evaluatable link. Any link type which might be evaluated, resulting
// in a truth-value, should inherit from this type.  This is a
```
I think it's appropriate because a quantifier ultimately results in a truth value, but I'm less sure about the evaluation thing, if you consider the result of proof as an evaluation, then it works obviously.


 I frankly don't fully understand the type hierarchy well enough (EvaluatableLink, BooleanLink, etc)

opencog/atoms/base/atom_types.script
```diff
@@ -338,8 +338,8 @@ DUAL_LINK <- SATISFYING_LINK
 // Should these inherit from EvaluatableLink ??
```
linas: This comment should be removed, as it's no longer appropriate...

opencog/atoms/base/atom_types.script
```diff
@@ -338,8 +338,8 @@ DUAL_LINK <- SATISFYING_LINK
...
 // Actually, these probably should inherit from PrenexLink, because
 // we typically want to maintain these in prenex normal form.
```
linas: This comment should be given some thought. But I'm not sure. PrenexLink provides assorted methods that are handy for rewriting links to make sure that they stay in prenex form, but its not at all clear if or how the ExistsLink could make use of that.   Its not urgent, we can defer this decision until later (and preserve this comment, for now)

PR #1690, opened 2018-05-16
#1673 An approach to more reliable linking

konstantin-t:
The idea is to link libraries via proxies. These proxies enforce linkage with the use of ${NO_AS_NEEDED}.
This pull requests concerns ```lambda``` library only.

If everything is fine with such approach, I can apply it to other libraries which rely on constructors.

N.B. I could not check, that it works on Macs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,6 +80,7 @@ IF (CMAKE_COMPILER_IS_GNUCXX)
 
 		# The Apple linker does not support the --no-as-needed flag.
 		SET(NO_AS_NEEDED "")
+		SET(DUMMY_CC "${PROJECT_SOURCE_DIR}/lib/dummy.cc")
 
 	ELSE (APPLE)
 		SET(CMAKE_C_FLAGS "-Wall -fPIC")
@@ -96,6 +97,7 @@ IF (CMAKE_COMPILER_IS_GNUCXX)
 		# SET(CMAKE_C_FLAGS_RELEASE "-O3 -g -fstack-protector -flto=8")
 
 		SET(NO_AS_NEEDED "-Wl,--no-as-needed")
+		SET(DUMMY_CC "${PROJECT_SOURCE_DIR}/lib/dummy.cc")
 		LINK_LIBRARIES(pthread)
 
 		# Workaround for circular dependencies problem. This causes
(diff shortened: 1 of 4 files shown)


konstantin-t:
These libraries are really different on my machine. And the sizes of the libraries are different too. The idea is that once a target links with lambda, it is implicitly being linked with lambda_, and the later linkage is made using ```${NO_AS_NEEDED}```. I checked that this approach solves the issue. Though, there is still no guarantee that the behavior of Make will not change again.

Here are the files generated by ``nm``
[x.txt](https://github.com/opencog/atomspace/files/2010559/x.txt)
[y.txt](https://github.com/opencog/atomspace/files/2010560/y.txt)

As for your last comment, agree, this is redundant.

konstantin-t:
Linas, it looks like you've missed the underscore during typing. Please try again.

linas:
So #1691 takes your core idea, and expands it to all of the libraries that have circular dependencies. It almost works, but not quite; there seem to be multiple issues. These are:
* Despite explicitly linking libraries with ctors in them to the libatomspace.so, it does not seem like the ctors will be called; but I can't quite tell.
* `nm libsmob.so` shows that it is bizarrely empty; this is unexpected.
* The opencog repo does not build.

Can you treat #1691 as a sketch for an idea, that somehow is not quite right, and maybe figure that out, and fix it?

linas:
Yeah, looking at your x.txt file, etc. something is ... not working as expected. The shared libs should contain symbols like
```
0000000000014010 T _ZN7opencog8DualLink4initEv
00000000000145b0 T _ZN7opencog8DualLink7factoryERKNS_6HandleE
0000000000014ae0 T _ZN7opencog8DualLinkC1ERKNS_4LinkE
0000000000014560 T _ZN7opencog8DualLinkC1ERKSt6vectorINS_6HandleESaIS2_EEt
0000000000014ae0 T _ZN7opencog8DualLinkC2ERKNS_4LinkE
0000000000014560 T _ZN7opencog8DualLinkC2ERKSt6vectorINS_6HandleESaIS2_EEt
```
which is what the library needs to be providing to its users.  I'm trying to guess what --no-as-needed did. I oringinally thought that it copied all symbols over to the new library; in fact, what it seems to be doing is to retain a load-time dependency to the underbar library which is passed to everyone who links to the non-underbar library. Or something like that.

konstantin-t:
I will take a look.

linas:
Clearly, efforts along the lines of #1698 are a superior approach for fixing #1673. I'm closing this.

PR #1691, opened 2018-05-16
Circular shared library dependency resolution

linas:
Yet another shot at resolving #1673 

The trick here is to have two levels of libraries: a base level, with unresolved symbols, and an upper level, linking to all the various dependent bases in such a way that all symbols are resolved in the end.

The other trick is that the atomspace is linked with `--no-as-needed`, with the hope that this really should be sufficient to force the running of all needed shared-library constructors, even in the face of newer, "smart" shared-library loaders (such as those in Ubuntu 17.10)


linas:
this cannot be merged as is, the opencog repo does not build with this in it's current state.

konstantin-t:
I made a preliminary research. This elephant needs to be eaten by pieces. Mangling with link order alone would not help.
I am going to rearrange classes in the libraries. That will also help me better understand the structure of OpenCog.

konstantin-t:
So here we are. The table below shows explicit and implicit dependencies. The implicit dependencies (marked in red) should be eliminated, as they are the source of circular references.

dependencies

A quick win target is the implicit dependence of ```truthvalue``` on ```atombase```.

So the plan for the step 001:

- Move certain files to a newly created ```proto``` library. This library will not depend on ```truthvalue```. It will contain:
--atom_types.cc
--atom_types_init.cc
--FloatValue.cc
--LinkValue.cc
--StringValue.cc
- Split ```ClassServer``` into two classes. The first will register names, the second - factories.
- Other issues:
-- Resolve the latent issue with ```AttentionValue::operator==``` and ```TruthValue::operator==```. Presently these operators inherit a method which returns false, if the target atom is not ```FLOAT_VALUE```.
-- To synchronize implementations of ```TruthValule::nearly_equal``` with ```FloatValue::operator==```
-- Remove unused ```TruthValue::isDefinedTV```

@linas, a big question. Is it a normal practice to move included to new folders? E.g. normally I would move ```atom_types.h```, ```proto_atom.h``` etc. to the ```proto``` folder. But that will break the client applications. Should I leave the includes as is?

linas:
The table looks correct.  The move to a proto library looks reasonable.  The split of classserver is reasonable too, I guess. Where would the pieces go?  Please be aware that, conceptually, the split seems reasonable, but, as you go to do it, you will typically discover that it is annoyingly hard, and you'll discover tiny little issues that seem to torpedo the whole boat.

Cleaning up attention-value, truth-value is fine with me; that code is essentially unmaintained, and I have been removing cruft for a while, now. There's still cruft left.

Moving `atom_types.h`, `proto_atom.h` etc. to the `proto` folder is a good idea. There are very few, almost no "applications" at this time, other than what is in the `openccog` repo. So you might need a simultaneous pull for the opencog repo. I think that currently, this is the only repo that would be affected. But who knows ...

linas:
You may want to copy that table to issue #1673 which is a better place for a long-term conversation, than here.

linas:
Clearly, this pull req cannot fly.

PR #1697, opened 2018-05-18
Add URE logger helpers

ngeiswei:


ngeiswei:
PR replacement for #1619

PR #1698, opened 2018-05-18
Resolve circular dependency, truthvalue and atombase

konstantin-t:
Restructure library compilation order such that protoatoms are moved to their 
own directory, independent of the atom definitions. Per discussion in issue #1691

Possible problems:
1) I did not make changes to https://github.com/opencog/opencog.git
2) I did not compile: Gearman, Haskell bindings, Hypertable, Python tests, SQL ODBC bindings, SQL Postgres bindings


opencog/atoms/base/ClassServer.cc
```diff
@@ -189,21 +61,26 @@ static Handle validating_factory(const Handle& atom_to_check)
 
 void ClassServer::spliceFactory(Type t, AtomFactory* fact)
 {
+	// N.B. it is too late to synchronize calls with NameServer using a shared mutex.
+	// If registrations of class names interleave with registration of factories,
+	// then the code will not work properly anyway. Before a factory is registered,
+	// the complete class hierarchy must be known.
+
```
linas: If this is really true, and can't be fixed, then this is a fatal design flaw. There are modules outside of the atomspace, and outside of opencog, that declare new atom types, and classes for them.

opencog/atoms/proto/README.md
```diff
@@ -0,0 +1,70 @@
+
+ProtoAtoms
+==========
+ProtoAtoms are a partially-implemented, incompletely thought-out
+concept for a way of associating arbitrary generic values with some
+given atom.  ProtoAtoms are like atoms, except that they lack a TV,
+an AV, and cannot be stored in the atomspace.  This makes them smaller,
+lighter and more efficient.
+
```
linas: Almost everything that the rest of this README says is wrong; I will try to update it.

opencog/atoms/base/Atom.h
```diff
@@ -224,6 +225,14 @@ class Atom
     virtual ~Atom();
     virtual bool is_atom() const { return true; }
 
+    /** Basic predicate */
+    bool is_type(Type t, bool subclass = true) const
+    {
```
linas: why did you move this from protoAtom to Atom? I think it needs to stay in protoatom, as one needs to be able to query the type of values, and not just atoms.

linas:
OK, I just reviewed this, I think it looks reasonable, modulo comments above. I tested python, SQL & gearman, these compile and tests pass. Hypertable can be ignored (it never worked right). Work items:

* The comment about locking needs to be clarified, as it is confusing. 
* A pull req for the opencog repo needs to be created.
* There are five failing unit tests; investigating.
```
The following tests FAILED:
	 64 - ConstantClausesUTest (Failed)
	 82 - DeepTypeUTest (Failed)
	100 - SingleUTest (Failed)
	103 - FirstNUTest (Failed)
	110 - RuleUTest (Failed)
Errors while running CTest
```

linas:
All five unit tests can be made to pass, if the following change is made:
```
--- a/tests/query/CMakeLists.txt
+++ b/tests/query/CMakeLists.txt
@@ -12,6 +12,7 @@ LINK_DIRECTORIES (
 
 LINK_LIBRARIES (
        query
+       lambda
        clearbox
        execution
        atomspace
```

opencog/atoms/base/ClassServer.cc
konstantin-t: The problem could occur if registration of class factories using ```addFactory``` precedes registration of _all_ type names. Because in such case factory methods might not be set properly for nested types, which are created after.

That might happen as the relative order of constructors execution is not guaranteed. A remedy is to define the priority in such a way, that names are always registered first. For instance, the following syntax might be used ```__attribute__((constructor (300)))```. Constructors with smaller numbers are executed first. This is weird and I am not sure, that there is an analogue for Windows.

A more general approach is to update ```_atomFactory``` each time a new type name is added. But this will look ugly too.

linas:
I just created a pull request against your code, to fix the unit tests, and update the README. If you merge that first, and then push again here, that will avoid potential merge conflicts later on.

opencog/atoms/base/ClassServer.cc
linas: I can guarantee that a type-name will be added before the factory for that type.  I can also guaranteee that some factories will be added before other type names are added.  For example, the core type-names will be added, and then their factories; then the NLP type names, and then their factories.  Its impossible to load the NLP type names before loading the core type factories.

konstantin-t:
Hm... The test did not fail on my machine. Once I finish with circular dependencies, I will apply ```no-as-needed``` where necessary.

konstantin-t:
You might have problems if your pull request will be based on mine, as I make changes in "past" commits and then force push the changes to the repo in order to keep the history neat.

linas:
Hmm. making changes in past commits, and then force-pushing is usually not a very good idea. Maybe I don't really care, if you do it just in your private branch, but generally, its a dangerous game.

linas:
The tests fail, for me, with these errors:
```
ERROR: Unbound variable: cog-bind-first-n
ABORT: unbound-variable
```
and that is defined in `liblambda.so`, which was not being loaded. Which is how we got started on this :-)

konstantin-t:
As for editing commits, I do that only because otherwise there is a risk to pollute history of the upstream.

I have a suggestion. Presently you merge all pull requests yourself. The better approach is to set up approval mechanism, and leave merges to pull request authors. Where can we discuss that?

konstantin-t:
> The tests fail, for me, with these errors:

That's strange. I will try to build the Release configuration. Maybe then I would reproduce the failure.

linas:
A bit off-topic, and I don't want to derail things, but .. there is a future need to have factories for Values, as well. For example, values could be streaming video or sound data; its not something we'd want to store as static lists or strings, but rather, have a C++ class that holds a URL or connector to that data source.  It might also be useful to have Values be ROS nodes.

linas:
Let me know when you think this is done and ready to be merged.

konstantin-t:
I have finished with this one. I was unable to reproduce the test failure, so I've just added the ```lambda```, as  you suggested.
Presently I am working on pull request for OpenCog. It compiles, but CogServerUTest fauls with SegFault.

opencog/atoms/base/ClassServer.cc
konstantin-t: Well, even this guarantee might not give absolute reliability.

Here is an example. Imagine we have two atom types: Parent and Child. The parent has explicit factory, which should be implicitly inherited by the Child. But that will not happen in the following case:

```
Load library Parent:
- Register type-name Parent
- Register factory Parent. The following code executes

        // OOPS!, the child type does not exist⤵
	for (Type chi=t; chi < _nameServer.getNumberOfClasses(); chi++) 
	{
		if (_nameServer.isAncestor(t, chi) and
		    (nullptr == _atomFactory[chi] or
		    ok_to_clobber.end() != ok_to_clobber.find(_atomFactory[chi])))
		{
			_atomFactory[chi] = fact;
		}
	}

Load library Child:
- Register type-name Child
- The factory for child is left initialized

konstantin-t:
Here is the pull request for opencog: https://github.com/opencog/opencog/pull/3132

I do not plan to make any changes there.

opencog/atoms/base/ClassServer.cc
konstantin-t: As for NameServer-ClassSever concurrent execution, I will think on it, and likely will create one more pull request.

konstantin-t: I investigated this subject. Seems like there are a lot of implicit requirements on how to register types/factories:

1. No new types can be added to an existing module once its initialization is concluded using endTypeDecls. Otherwise declType will throw.
2. Initialization of modules must be completely serialized _by_ _clients_. I.e. it is not allowed to start initializing a new module before endTypeDecls is called for the previous module. Otherwise declType will throw. So no multi-threading can be used.
3. Class factory for an atom type can be registered only after all its children types are registered. Otherwise the children will not inherit the parent's factory.
4. If a type inherits two other, and both parents have non-trivial factories, then the child must have a factory explicitly declared. Otherwise the child will inherit one of parents' factories; which one - undefined.

These requirements are met everywhere in the repo, because constructors are called in just one thread, and execution order is specified by dependencies, defined in CMakeLists.txt (though, that is unspecified behavior of gcc). 
In such case NO serialization using mutexes is needed at all - the read access is safe.

>  its not uncommon that there might be 2 or 3 modules, all trying to load new atom types, and load factories for them, and this might be happening in different threads

I did not find such scenario. And I cannot see how that could work reliably with the present implementation (see above). @linas, please give a reference to a live multi-threading example.

P.S. maybe I dug an issue which cares nobody. If so, I could switch to another task.

linas:
Thanks!

Issue #474, opened 2015-11-04
GetLink over a ChoiceLink in body isn't returning result.


AmeBel:
### Atomspace content and Scheme function

``` scheme
(StateLink
    (Node "action-on-demand")
    (ListLink
        (ConceptNode "Increase")
        (ConceptNode "Novelty")))

(define (get-current-goal)
    (cog-execute!
       (GetLink
          (TypedVariableLink
              (VariableNode "demand")
              (TypeNode "ConceptNode"))
          (StateLink
              (Node "action-on-demand")
              (ListLink
                  (ChoiceLink
                      (ConceptNode "Decrease")
                      (ConceptNode "Increase"))
                  (VariableNode "demand"))))
    )
)
```
### Present Output

``` scheme
guile> (get-current-goal)            
(SetLink                             
)                                    
```
### Expected Output

``` scheme
guile> (get-current-goal)            
(SetLink  
       (ConceptNode "Novelty")                        
)                                    
```

linas:
Has nothing to do with StateLink, fails the same way with any other link type.  I believe that the code expects ChoiceLink was to wrap things with variables inside of it ...

linas:
If the ChoiceLink wraps other links, then it works.  The fail case is only when ChoiceLink wraps nodes. ChoiceLink does NOT need to  wrap variables.

When ChoiceLink wraps nodes, the start search-trem structure gets confused,...

AmeBel:
ok thanks,

linas:
It seems that fixing this will be difficult; however, I think it will also simplify the code -- the correct fix seems to be to remove the "start_term" everywhere, ad use the normal do_term_up code from everywhere.

linas:
A fix for this would be nice, as lots of relex-related rules need this.

linas:
still a bug

vsbogd:
Would you mind if I will get it? 

I experimented with solution which uses PatternTerm tree to find thinnest term and  use thinnest term as start point for do_term_up. Such approach solves the issue.

linas:
If you have a fix, please create a pull request. However, I don't quite like that you mention PatternTerm; that is a device for managing the back-tracking state of unordered links, and should not be part of a fix for this particular bug. 

My comments from Nov 2015 above suggests that code could be simplified by removing `start_term` ...

vsbogd:
Hi @linas,

[...]

PR #1701, opened 2018-05-22
Fix choice link with nodes

vsbogd:
To review fix approach for https://github.com/opencog/atomspace/issues/474
Known issues:
- conflicts are not resolved
- failed unit tests
- old code is not removed

### DIFF
--- a/opencog/query/InitiateSearchCB.cc
+++ b/opencog/query/InitiateSearchCB.cc
@@ -312,6 +312,10 @@ PatternTermPtr InitiateSearchCB::find_thinnest_term_recursive(
 		return root;
 	}
 
+	// Ignore all dynamically-evaluatable links up front.
+	if (_dynamic->find(root->getHandle()) != _dynamic->end())
+		return PatternTerm::UNDEFINED;
+
 	size_t max_depth = 0;
 	size_t min_width = SIZE_MAX;
 	PatternTermPtr thinnest_term = PatternTerm::UNDEFINED;
@@ -455,6 +459,7 @@ bool InitiateSearchCB::neighbor_search(PatternMatchEngine *pme)
 	// no constants in them at all.  In this case, the search is
 	// performed by looping over all links of the given types.
 #if USE_PATTERN_TERM
+	_choices.clear();
 	PatternTermPtr start_term = find_thinnest_term(clauses);
 
 	// Cannot find a starting point! This can happen if:
@@ -462,7 +467,7 @@ bool InitiateSearchCB::neighbor_search(PatternMatchEngine *pme)
 	// 2) all of the clauses are evaluatable(!),
 	// Somewhat unusual, but it can happen.  For this, we need
 	// some other, alternative search strategy.
-	if (start_term == PatternTerm::UNDEFINED) {
+	if (start_term == PatternTerm::UNDEFINED  and 0 == _choices.size()) {
 		_search_fail = true;
 		return false;
 	}
--- a/opencog/query/InitiateSearchCB.cc
+++ b/opencog/query/InitiateSearchCB.cc
@@ -149,94 +149,97 @@ void InitiateSearchCB::set_pattern(const Variables& vars,
 //
 
 Handle
-InitiateSearchCB::find_starter(const Handle& h, size_t& depth,
-                                     Handle& startrm, size_t& width)
+InitiateSearchCB::find_starter(const Handle& root, size_t& depth,
+                                     Handle& parent, size_t& width)
 {
 	// If its a node, then we are done.
-	Type t = h->get_type();
-	if (_nameserver.isNode(t))
+	Type root_node_type = root->get_type();
+	if (_nameserver.isNode(root_node_type))
 	{
-		if (VARIABLE_NODE != t and GLOB_NODE != t)
+		if (VARIABLE_NODE != root_node_type and GLOB_NODE != root_node_type)
 		{
-			width = h->getIncomingSetSize();
-			startrm = h; // XXX wtf ???
-			return h;
+			width = root->getIncomingSetSize();
+			parent = root; // XXX wtf ???
+			return root;
 		}
 		return Handle::UNDEFINED;
 	}
 
 	// If its a link, then find recursively
-	return find_starter_recursive(h, depth, startrm, width);
+	return find_starter_recursive(root, depth, parent, width);
 }
 
 Handle
-InitiateSearchCB::find_starter_recursive(const Handle& h, size_t& depth,
-                                         Handle& startrm, size_t& width)
+InitiateSearchCB::find_starter_recursive(const Handle& root, size_t& depth,
+                                         Handle& starter_link, size_t& width)
 {
 	// If its a node, then we are done. Don't modify either depth or
 	// start.
-	Type t = h->get_type();
-	if (_nameserver.isNode(t))
+	Type root_node_type = root->get_type();
+	if (_nameserver.isNode(root_node_type))
 	{
-		if (VARIABLE_NODE != t and GLOB_NODE != t)
+		if (VARIABLE_NODE != root_node_type and GLOB_NODE != root_node_type)
 		{
-			width = h->getIncomingSetSize();
-			return h;
+			width = root->getIncomingSetSize();
+			return root;
 		}
 		return Handle::UNDEFINED;
 	}
 
 	// Ignore all dynamically-evaluatable links up front.
-	if (_dynamic->find(h) != _dynamic->end())
+	if (_dynamic->find(root) != _dynamic->end())
 		return Handle::UNDEFINED;
 
 	// Iterate over all the handles in the outgoing set.
 	// Find the deepest one that contains a constant, and start
 	// the search there.  If there are two at the same depth,
 	// then start with the skinnier one.
-	size_t deepest = depth;
-	startrm = Handle::UNDEFINED;
-	Handle hdeepest(Handle::UNDEFINED);
-	size_t thinnest = SIZE_MAX;
+	size_t starter_depth = depth;
+	starter_link = Handle::UNDEFINED;
+	Handle starter_node(Handle::UNDEFINED);
+	size_t starter_width = SIZE_MAX;
 
-	for (Handle hunt : h->getOutgoingSet())
+	for (Handle branch : root->getOutgoingSet())
 	{
-		size_t brdepth = depth + 1;
-		size_t brwid = SIZE_MAX;
-		Handle sbr(h);
+		size_t branch_depth = depth + 1;
+		size_t branch_width = SIZE_MAX;
+		Handle branch_link(root);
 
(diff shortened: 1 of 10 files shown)


opencog/query/InitiateSearchCB.cc
```diff
@@ -345,10 +454,39 @@ bool InitiateSearchCB::neighbor_search(PatternMatchEngine *pme)
...
+#endif
+#if USE_PATTERN_TERM
 
+	for (const Choice& choice : _choices)
+	{
+		bool found = pme->explore_term(choice.thinnest_term);
+		if (found)
+			return true;
+	}
+#else
```
ngeiswei: That's a hell of a simplification. I suppose it also means you're on your way to enable parallelization for backtracking, isn't it?

ngeiswei:
@vsbogd The PR is nicely decomposed into small consistent commits, that makes it easy to review in spite of the intricacy of that part of the code. Thanks for that!

The change itself it looks good to me, although as you seem to say, it doesn't pass all unit tests.

vsbogd:
I would like to agree on approach with Linas as he mentioned that my understanding of his suggestion probably is not correct.

It probably makes sense to break it on few PR's to simplify review. For instance variables renaming can be done in separate PR. I didn't do it because I thought these methods will be removed finally but I can split it if you think it makes sense.

Meanwhile I am going to look at failed unit tests.

opencog/query/InitiateSearchCB.cc
vsbogd: Not sure I understand you point correctly. Do you mean that this code is oversimplified comparing to previous version? I am going to revisit it when approach is confirmed and unit tests are green as I would like make sure this approach works first.

May be you mean that it was a place where matching was parallelized before? Now only place to explore incoming set is PatternMatchEngine::explore_up_branches. I didn't think what does it mean for parallelizing but PatternMatchEngine::explore_up_branches can be parallelized as well.

ngeiswei: Oh, I definitely meant "hell" in a good way.

Regarding parallelization, yes that is what I meant, I think it's a good thing and it forces us to address the problem at its core. It makes it clear that the initial suggestion we made to you about parallelization is just a subcase of backtracking parallelization.

opencog/query/InitiateSearchCB.cc
```diff
@@ -140,7 +140,7 @@ void InitiateSearchCB::set_pattern(const Variables& vars,
 // start.
 //
 // size_t& depth will be set to the depth of the thinnest constant found.
-// Handle& start will be set to the link containing that constant.
+// Handle& link will be set to the link containing that constant.
```
linas: why change this name? This finds and returns the start. The start should have a nice, clear, meaningful name, such as "start". By contrast, almost everything is a "link", we don't need more "links" in the code-base.

opencog/query/InitiateSearchCB.cc
```diff
@@ -149,94 +149,208 @@ void InitiateSearchCB::set_pattern(const Variables& vars,
...
 
 	// If its a link, then find recursively
-	return find_starter_recursive(h, depth, startrm, width);
+	return find_starter_recursive(root, depth, link, width);
 }
 
-Handle
-InitiateSearchCB::find_starter_recursive(const Handle& h, size_t& depth,
-                                         Handle& startrm, size_t& width)
+Handle InitiateSearchCB::find_starter_recursive(const Handle& root, size_t& depth,
```
linas: Similar comment here: its clearly not the "root" any more, its just some arbitrary atom somewhere far under the root, at this point. Calling it the "root" when its not the root is .. confusing.

opencog/query/InitiateSearchCB.cc
```diff
@@ -149,94 +149,208 @@ void InitiateSearchCB::set_pattern(const Variables& vars,
...
 	// the search there.  If there are two at the same depth,
 	// then start with the skinnier one.
-	size_t deepest = depth;
-	startrm = Handle::UNDEFINED;
-	Handle hdeepest(Handle::UNDEFINED);
-	size_t thinnest = SIZE_MAX;
+	depth = 0;
+	link = Handle::UNDEFINED;
+	Handle starter;
+	width = SIZE_MAX;
```
linas: I  also object to this name change.  We are explicitly looking for the least-wide atom, and not just some "wide" atom.   If you renamed it from "thinnest" to "least_wide", or "skinniest" that would be OK. But to rename it to a completely opaque, totally meaning-less "width" is just code-obfuscation.

opencog/query/InitiateSearchCB.cc
linas: one more remark about naming conventions: ideally, names should be uniquely grepable.  Thus, for example, in the current code base, I can `cd opencog; grep -r startrm` and I see exactly 6 hits, all in the same file, all for exactly this one and the same variable, and nothing else. If I instead say `cd opencog; grep -r link | wc` I see that the text string `link` occurs in 1991 distinct locations.  This makes debugging pointlessly difficult.  it breaks a basic code-debugging and automation tool.

opencog/query/PatternMatchEngine.cc
```diff
@@ -1077,6 +1077,13 @@ bool PatternMatchEngine::tree_compare(const PatternTermPtr& ptm,
 	if (hp->is_node() and hg->is_node())
 		return node_compare(hp, hg);
 
+	// CHOICE_LINK's are multiple-choice links. As long as we can
+	// can match one of the sub-expressions of the ChoiceLink, then
+	// the ChoiceLink as a whole can be considered to be grounded.
+	//
+	if (CHOICE_LINK == tp)
+		return choice_compare(ptm, hg);
+
```
linas: This code-movement means that the `link_match()` callback is no longer called for choice links, and so that callback never gets a chance to accept or reject the potential matching. 

Now, admittedly, the default callback always accepts choice links, and so perhaps this almost does not matter. The other callback, for the recognizer, does not special-case the choice link.  It is possible that the recognizer is buggy in the way that it treats the choicelink.  The recognizer is not as widely used as the pattern matcher, and has fewer test cases.

linas: Ahh, OK, I see -- the point is to move this prior to the fuzzy_match(), which throws everything off. That's good.

I might have an easy fix for all this, then ...

linas:
I'm closing this because pull req #1705 fixes this bug in a simpler, more direct fashion.

I suppose that this might be irritating to you, because you put in a lot of work ...

There were actually two bugs in the code, and you solved one of them: moving the location of the choice-link compare.  You got close to, but did not quite find the second bug -- the find_start code was incorrectly suggesting that the ChoiceLink was an OK place to start looking -- but its not. So #1705 fixes this.

opencog/query/InitiateSearchCB.cc
ngeiswei: Hmm, OK, but that's really more than ideal, I mean I don't think we need to end with weirder and weirder names just to avoid clashes, nowadays they are code indexing tools that can deal with that (CEDET, rtag, etc, on the Emacs side of things at least).

linas: aww com'on -- grep is simple, easy, available everywhere, with no special installation required, no learning curve, no unusual syntax, no special keystrokes. Its just plain-old grep.  Why use some complex tool, when there is an industry-standard simple tool?

vsbogd: Original idea was to rename fields to something more readable but I sure "link" is a bad idea because it doesn't mirror sense of the value. Probably I was just confused by the fact that method returns two values and start term is not the actual result while method's name is "find_starter". I would replace it by "start_term".

Not sure grep argument is very relevant here as this is the name of the method parameter and it has a local context you can simply search it within method. Anyway I agree link is misleading name.

opencog/query/InitiateSearchCB.cc
vsbogd: It is not only renaming - I have replaced separate variable by using method parameter. I can revert it to make code cleaner.

vsbogd:
I am going to raise separate PR with parameters renaming with all comments above incorporated if you don't mind.

Regarding other commits under this PR, I still think that using PatternTerm tree to find a start point makes code simpler and cleaner. This solution doesn't require separate "if" in find_starter_recursive() to handle ChoiceLink case. It doesn't require exploring incoming set of starter constant node in InitiateSearchCB::neighbor_search() and we have only place to do it in PatternMatchEngine::explore_up_branches().

linas:
> renaming 

I really rather that you didn't do this. 

> using PatternTerm tree 

I see two problems. 

Problem one: you want to change several hundred lines of code to "save" approximately three to six lines of code. A hundred-to-one cost ratio will almost surely be swamped by other side effects and new bugs. Nor is it clear that there will be much actual savings. 

Problem two: The PatternTerm tree is big, bulky and slow. Using it in even more places risks slowing things down even more. PatternTerm is a very heavy-weight device. 

If you want to do something useful, take a look at performance optimizations of PatternTerm. My general impression is that it is implemented "upside down" it uses weak pointers where it should use regular pointers, and it uses regular pointers where it should use weak pointers.  I'm guessing that we could get a decent performance boost by swapping these two around.

vsbogd:
> > renaming
> 
> I really rather that you didn't do this.

Could you please explain your position? I started renaming because I found it difficult understanding code with single symbol names. To known what does it mean you need to remember context. "t" is a type but type of what? "startrm" - is difficult to interpret. After some time I will remember all well known acronyms in code but each new developer who looks at the code will learn them again. If you don't like names suggested it always can be improved.

vsbogd:
> using PatternTerm tree
>
>> Problem one

Essentially the change is replacing:
https://github.com/opencog/atomspace/blob/7206242085890ec87769b263084f9fb2b3a34941/opencog/query/InitiateSearchCB.cc#L395-L407
by:

	for (const Choice& choice : _choices)
	{
                PatternTermPtr term = choice.thinnest_term;
		bool found = pme->do_term_up(term, term->getHandle(), term->getClause());
		if (found)
			return true;
	}

As I mentioned in comment to the bug https://github.com/opencog/atomspace/issues/474#issuecomment-390969067 PatternTerm tree is already used almost everywhere in PatternMatcher. The only place where it is not used is start of the clause matching where the first step is exploring all possible PatternTerms and use them further:
https://github.com/opencog/atomspace/blob/7206242085890ec87769b263084f9fb2b3a34941/opencog/query/PatternMatchEngine.cc#L2068-L2076)
If PatternTerm tree will be used from the very beginning this search will not be needed and overall algorithm will be simpler to understand. And it solves the problems like https://github.com/opencog/atomspace/issues/474 where overall pattern matching algorithm handles query correctly but only start part of the code needs additional logic to do it right. Actually it will remove almost all logic from start part of the code.

>> Problem two
I will look at PatternTerm closer if it is slow and inefficient it is already a problem as it is used in pattern matching intensively.

PR #1702, opened 2018-05-22
Added Postgres development package as dependency

marjinal1st:
**libpq-dev** package is required for SQL Postgres bindings of Atomspace. Just added it to README.


linas:
Thanks!

Issue #1630, opened 2018-04-11
Potential variable name collision in Rule::unify_source


ngeiswei:
The following
https://github.com/opencog/atomspace/blob/master/opencog/rule-engine/Rule.cc#L375
https://github.com/opencog/atomspace/blob/master/opencog/rule-engine/Rule.cc#L408
should be fixed, that is make sure that the variable names generated by random alpha conversion differ from the variables in `vardecl`.

PR #1712, opened 2018-05-28
has name capture method added

SomeUserName1:
to check wether a rule still has captured variables after alpha conversion
refering to issue/quick task #1630 

Is this done correctly? Also is this performant enough? (complexity O(n²)))

### DIFF
--- a/opencog/rule-engine/Rule.cc
+++ b/opencog/rule-engine/Rule.cc
@@ -484,6 +484,30 @@ std::string Rule::to_string(const std::string& indent) const
 	return ss.str();
 }
 
+bool Rule::has_name_capture() const
+{
+    HandleSeq vardecls = VariableListCast((this->get_vardecl()))
+            ->get_variables().varseq;
+    HandleSeq boundvars = BindLinkCast(this->get_rule())->get_variables()
+            .varseq;
+    size_t sb = boundvars.size();
+    size_t sd = boundvars.size();
+
+    for (size_t i = 0; i < sb; i++)
+    {
+        for (size_t j = i+1; j < sb; j++)
+        {
+            if(content_eq(boundvars[i], boundvars[j])) return true;
+        }
+        for (size_t k = 0; k < sd; k++)
+        {
+            if(content_eq(boundvars[i], vardecls[k])) return true;
+        }
+
+    }
+    return false;
+}
+
 Rule Rule::rand_alpha_converted() const
 {
 	// Clone the rule
--- a/opencog/rule-engine/Rule.cc
+++ b/opencog/rule-engine/Rule.cc
@@ -486,9 +486,10 @@ std::string Rule::to_string(const std::string& indent) const
 
 bool Rule::has_name_capture() const
 {
-	HandleSeq boundvars = BindLinkCast(this->get_rule())->get_variables()
+	HandleSeq boundvars = (BindLinkCast(this->get_rule())->get_variables())
 			.varseq;
-	FreeVariables fv = VariableListCast((this->get_vardecl()))->get_variables();
+	FreeVariables fv = (VariableListCast(this->get_vardecl()))
+			->get_variables();
 
 	for (const auto& v : boundvars)
 	{
--- a/opencog/rule-engine/Rule.cc
+++ b/opencog/rule-engine/Rule.cc
@@ -404,7 +404,7 @@ RuleTypedSubstitutionMap Rule::unify_source(const Handle& source,
 	// as any variable in the source. XXX This is only a stochastic
 	// guarantee, there is a small chance that the new random name
 	// will still collide.
-	Rule alpha_rule = rand_alpha_converted();
+	Rule alpha_rule = rand_alpha_converted(vardecl);
 
 	RuleTypedSubstitutionMap unified_rules;
 	Handle rule_vardecl = alpha_rule.get_vardecl();
@@ -436,9 +436,7 @@ RuleTypedSubstitutionMap Rule::unify_target(const Handle& target,
 
 	// To guarantee that the rule variable does not have the same name
 	// as any variable in the target.
-	Rule alpha_rule = rand_alpha_converted();
-	// Check for the small chance of still having name collisions
-	alpha_rule.has_name_collision(vardecl);
+	Rule alpha_rule = rand_alpha_converted(vardecl);
 
 	RuleTypedSubstitutionMap unified_rules;
 	Handle alpha_vardecl = alpha_rule.get_vardecl();
@@ -486,24 +484,30 @@ std::string Rule::to_string(const std::string& indent) const
 
 bool Rule::has_name_collision(const Handle& vardecl) const
 {
-	HandleSeq boundvars = (BindLinkCast(this->get_rule())->get_variables())
-			.varseq;
+	HandleSet boundvars = BindLinkCast(this->get_rule())->get_variables()
+			.varset;
 	Variables fv = (VariableListCast(vardecl))->get_variables();
 
 	for (const auto& v : boundvars)
 	{
-		if(fv.is_in_varset(v)) return true;
+		if (fv.is_in_varset(v)) return true;
 	}
 	return false;
 }
 
-Rule Rule::rand_alpha_converted() const
+Rule Rule::rand_alpha_converted(const Handle& vardecl) const
 {
 	// Clone the rule
 	Rule result = *this;
 
 	// Alpha convert the rule
 	result.set_rule(_rule->alpha_convert());
+
+	// Check for the small chance of still having name collisions
+	while(result.has_name_collision(vardecl))
+	{
+		result.set_rule(_rule->alpha_convert());
+	}
 
 	return result;
 }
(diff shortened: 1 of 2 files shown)


opencog/rule-engine/Rule.cc
```diff
@@ -484,6 +484,30 @@ std::string Rule::to_string(const std::string& indent) const
...
 }
 
+bool Rule::has_name_capture() const
+{
+    HandleSeq vardecls = VariableListCast((this->get_vardecl()))
+            ->get_variables().varseq;
+    HandleSeq boundvars = BindLinkCast(this->get_rule())->get_variables()
+            .varseq;
+    size_t sb = boundvars.size();
+    size_t sd = boundvars.size();
```
linas: sb and sd are identical. Why two differrent names?

linas: oh, should be vardels.

opencog/rule-engine/Rule.h
```diff
@@ -295,6 +295,8 @@ class Rule : public boost::totally_ordered<Rule>
 	// URE will always choose the one with the highest confidence.
 	TruthValuePtr _tv;
 
+	bool has_name_capture() const;
```
linas: since this is not used anywhere, why is it being added?

opencog/rule-engine/Rule.cc
```diff
@@ -484,6 +484,30 @@ std::string Rule::to_string(const std::string& indent) const
 	return ss.str();
 }
 
+bool Rule::has_name_capture() const
+{
+    HandleSeq vardecls = VariableListCast((this->get_vardecl()))
+            ->get_variables().varseq;
```
linas: The indendataion beig used here seems incosistent with the rest of the file (the rest of the file uses tabs.)

linas:
Other than the comments I made, this looks "reasonable-ish", I guess its up to @ngeiswei to figure out the next level.  I guess he asked you to do this.

Note that there already exists a fair amount of sophisticated machinery for dealing with variables, for example, `struct FreeVariables` The  `is_in_varset` looks for non-captured variables, and runs on O(nlog n) after setup. But it doesn't do quite what it is your are doing.

linas:
Also: if I recall correctly, varseq will have duplicated variable entires (as they appear in depth-first order) but varset will only have each variable appearing once and only once. 

Your boundvars loop seems to be trying to see if a vriable occurs more than once .. is that intended?  why is a variable that appears twice or more, more useful than one that appear just once?

I mean I kind-of know nothing about the rule engine, so maybe this makes sense.

linas:
p.s. I will stop commenting now. Please don't let me scare you. I was just emptying my mail queue, saw this, and was bored. So I looked and made some random superficial comments, thereby increasing the workload for @ngeiswei  who is probably cursing me now, and muttering something about how I keep interfering with things :-)  Its all good.

SomeUserName1:
Oh yes, i forgot to integrate it again after moving the lines to a clean fork (with non replaced spaces for tabs), further fixed the sd = vardecls.size() stupidity. 

> The indendataion beig used here seems incosistent with the rest of the file (the rest of the file uses tabs.)

As specified in HACKING i used spaces but i think that is deprecated/to be ignored then...

> Other than the comments I made, this looks "reasonable-ish", I guess its up to @ngeiswei to figure out the next level. I guess he asked you to do this.

actually it was on the quick tasks "list"/tag and Amen pointed me to this to get started with the atomspace code base #1630 

> Also: if I recall correctly, varseq will have duplicated variable entires (as they appear in depth-first order) but varset will only have each variable appearing once and only once.

I use the varseq s to mach for duplicates indeed as varset would remove eventual name captures.
the opposite is the case, we do not want douplicate variables as this might cause errors while beta reducing the rule/lambda calculus expression ([reference](http://www.cmi.ac.in/~madhavan/courses/pl2009/lecturenotes/lecture-notes/node85.html))

linas:
OK. One more comment: there should not be any "accidental name captures", as long as you use the provided API's correctly.  All of the code in Variables.cc and elsewhere already handles all variable substitutions, renamings, etc. completely correctly, without any name-capturing at all.  It should be perfectly safe to use the same variable name for two different things. The code can tell which is which, and handles this in an automatic fashion.

If something is not working right, then that's a bug. -- which would need a bug fix, but should not need any new code to avoid accidental "name captures".  ...   well, I don't quite have the big picture here, cause I don't understand  the rule engine, but everything else has been specifically coded to deal with this problem, there are even unit tests, somewhere, for that.

SomeUserName1:
That's a very helpful hint; so the first inner for loop is completely unnecessary and the second one may be replaced by n log n complexity
BTW i don't claim that the code is correct in any sense & am very grateful for pointers as above regarding the variable handling (as it's quite impossible to read through all the code in one pass) 
BTW2: may we have a call or sth. this week? i have some questions regarding the usage of caches, the TLB & the AtomTable

linas:
> have a call 

for me emails are simpler, faster.  TLB is used only in the postgres backend. (and also in the rest api) it should be completely invisible to users of the atomspace. It just issues temporary ID#'s that correspond to atoms, which are needed when atoms are stored on disk or sent over network. Don't know what you mean by "cache", other than that the atomspace/atomtable itself is a  "cache" of a portion of what is stored in the posgres database.  The atomspace/atomtable just hangs on to a reference to atoms that are in RAM, but those atoms can be elsewhere, too.  What are you trying to do? And we should have remainder of conversation somewhere other than here...

opencog/rule-engine/Rule.cc
```diff
@@ -486,23 +486,13 @@ std::string Rule::to_string(const std::string& indent) const
 
 bool Rule::has_name_capture() const
 {
-	HandleSeq vardecls = VariableListCast((this->get_vardecl()))
-			->get_variables().varseq;
 	HandleSeq boundvars = BindLinkCast(this->get_rule())->get_variables()
 			.varseq;
-	size_t sb = boundvars.size();
-	size_t sd = vardecls.size();
+	FreeVariables fv = VariableListCast((this->get_vardecl()))->get_variables();
```
misgeatgit: Could you change the brace the following way:
 ```FreeVariables fv = (VariableListCast(this->get_vardecl()))->get_variables();``` just to make it less ambiguous.

opencog/rule-engine/Rule.cc
```diff
@@ -484,13 +484,28 @@ std::string Rule::to_string(const std::string& indent) const
...
+			.varseq;
+	FreeVariables fv = (VariableListCast(this->get_vardecl()))
+			->get_variables();
+
+	for (const auto& v : boundvars)
+	{
+		if(fv.is_in_varset(v)) return true;
+	}
+	return false;
+}
```
ngeiswei: @SomeUserName1 I think this method always returns true. If `v` is in `varseq` it is gonna be in `varset.

Anyway, that does not solve the problem, what you want is just to make sure that the variables in bind link of `alpha_rule` https://github.com/opencog/atomspace/blob/master/opencog/rule-engine/Rule.cc#L407
are do not overlap with the variables in `vardecl`, the argument of `Rule::unify_source` https://github.com/opencog/atomspace/blob/master/opencog/rule-engine/Rule.cc#L396 that happens to be the variable declaration of `source`.

This is necessary so that subsequent unification properly works, otherwise it will assume that the variables with the same names are the same variables, while they are not.

If they are different then `rand_alpha_converted()` should rerun (or equivalently https://github.com/opencog/atomspace/blob/master/opencog/rule-engine/Rule.cc#L493 should be rerun) till all variables in the alpha converted rule are different from `vardecl` (the argument of `Rule::unify_source`).

linas:
Nil, I don't fully understand what is going on here, and it smells "fishy".  If you use the FreeVariables class, it should only report the free variables, and the bound variables should be invisible.  Are you trying to unify two different things, that accidentally have the same free variables in them?  

The reason I ask this is that all of the variables in a BindLink are bound, and are thus invisible, and impossible to unify against -- they "cannot be seen" by the external world. So why would you need to even look at them, and then try to do this calculation, if they are invisible?  As a matter of principle, its impossible to "capture" or have a collision with something that is invisible...!?

ngeiswei:
Linas, yes, but the URE produces new BindLinks (inference trees) out of existing BindLinks (inference tree + rule). In order to do that it needs to opens them up and try to stick their ends together (like pieces of a jizsaw puzzle). To stick their ends it must see if they match, i.e. if they unify. Unifying 2 terms amounts to finding a substitution (ideally the most abstract one) so that if applied to the 2 terms they end up being equal. For instance, unifying
```
F(X) =? Y
```
amounts to finding the substitution
```
sigma(Y)=F(X)
sigma(X)=X
```
Now what if you try to unify 2 terms with some shared variables, like
```
X, Y =? F(Z), X
```
where , is some non commutative operator. This still admits a solution
```
sigma(X)=F(Z)
sigma(Y)=F(Z)
sigma(Z)=Z
```
For technical reasons that I won't get into, the unifier needs to support that, I can't just forbid it, or automatically rename the variables so that the terms have different variables. However when trying to stick 2 BindLinks together, because the variables are HIDDEN and have nothing to do with each other I can afford to (and should) rename them so that the variables on the left and right terms are disjoint. Thus the alpha conversion before unification.

linas:
Hmm. OK.   I'm starting to understand why you were unhappy with the FreeLink, in other conversations.  Yeah, unification is hard.  I wuld need to set aside a few days of uniterrupted thought about how to treat unification as a connector-like problem, and a few days fee are hard to get now.

PR #1714, opened 2018-05-28
resolve #1630

### LINKED ISSUE #1630: Potential variable name collision in Rule::unify_source

SomeUserName1:


opencog/rule-engine/Rule.cc
```diff
@@ -484,14 +483,20 @@ std::string Rule::to_string(const std::string& indent) const
...
 	Rule result = *this;
 
 	// Alpha convert the rule
 	result.set_rule(_rule->alpha_convert());
 
+	// Check for the small chance of still having name collisions
+	while(is_disjoint(_rule->get_variables().varset, vardecl.varset))
+	{
+		result.set_rule(_rule->alpha_convert());
+	}
```
ngeiswei: You can save the previous conversion by using a do...while loop.

Also, the condition is wrong, it should be the varset of result, not _rule.

opencog/rule-engine/Rule.cc
```diff
@@ -511,7 +516,7 @@ Handle Rule::standardize_helper(AtomSpace* as, const Handle& h,
 		HandleSeq old_outgoing = h->getOutgoingSet();
 		HandleSeq new_outgoing;
 
-		for (auto ho : old_outgoing)
+		for (const auto ho : old_outgoing)
```
ngeiswei: This belongs to another commit. A const reference would be better BTW.

opencog/rule-engine/Rule.cc
```diff
@@ -404,7 +405,7 @@ RuleTypedSubstitutionMap Rule::unify_source(const Handle& source,
 	// as any variable in the source. XXX This is only a stochastic
 	// guarantee, there is a small chance that the new random name
 	// will still collide.
-	Rule alpha_rule = rand_alpha_converted();
+	Rule alpha_rule = rand_alpha_converted((VariableListCast(vardecl))->get_variables());
```
ngeiswei: That's better, but you still need to fetch the variables of source in case vardecl is undefined.

I'm thinking maybe you could have rand_alpha_converted take both source (or target) and vardecl as arguments, and do the possible variable extraction there. That would be cleaner I think.

SomeUserName1:
I squashed another commit into it, better now?
Is target casted correctly? Differently asked: is target (or source respectively) a Rule, or Variables?

opencog/rule-engine/Rule.cc
```diff
@@ -484,13 +481,23 @@ std::string Rule::to_string(const std::string& indent) const
...
 {
+	const Handle& m_vardecl = gen_vardecl(term, vardecl);
 	// Clone the rule
 	Rule result = *this;
 
-	// Alpha convert the rule
-	result.set_rule(_rule->alpha_convert());
+	HandleSet target_vars = VariableListCast(m_vardecl)->get_variables().varset;
+	HandleSet result_vars = BindLinkCast(result.get_rule())->get_variables()
+			.varset;
```
ngeiswei: You may directly use `result._rule` instead of casting `result.get_rule()`, that is because a class is automatically a friend of itself, see https://stackoverflow.com/questions/39404109/is-a-class-automatically-a-friend-of-itself

Also, better have a const reference of HandleSet to not waste CPU copying varset.

opencog/rule-engine/Rule.cc
```diff
@@ -484,13 +481,23 @@ std::string Rule::to_string(const std::string& indent) const
...
+
+	do {
+		// Alpha convert the rule
+		result.set_rule(_rule->alpha_convert());
+	}
+	while(!is_disjoint(target_vars, result_vars));
+
```
ngeiswei: Remove that newline.

opencog/rule-engine/Rule.cc
```diff
@@ -649,3 +656,4 @@ std::string oc_to_string(const RuleTypedSubstitutionMap& rules)
 }
 
 } // ~namespace opencog
+
```
ngeiswei: Remove that newline.

opencog/rule-engine/Rule.h
```diff
@@ -295,9 +295,15 @@ class Rule : public boost::totally_ordered<Rule>
...
-	// into random variable names.
-	Rule rand_alpha_converted() const;
+	/**
+ *
+ * @param vardecl: declarations of variables
+ * @param target: source term used by the foward chainer or target term used
+ * be backwards chainer; only used to get the variables out of the term in case
+ * there are no variable declarations
+ * @return alpha converted rule that has no collisions with the given term
+ */
```
ngeiswei: Fix indentation.

opencog/rule-engine/Rule.h
```diff
@@ -334,3 +340,4 @@ std::string oc_to_string(const RuleTypedSubstitutionMap& rule_ts_map);
 } // ~namespace opencog
 
 #endif /* RULE_H_ */
+
```
ngeiswei: Hmm, is that your editor adding these? So if I may ask, what is this unholy editor of yours?

opencog/rule-engine/Rule.cc
```diff
@@ -484,13 +481,20 @@ std::string Rule::to_string(const std::string& indent) const
...
 
-Rule Rule::rand_alpha_converted() const
+Rule Rule::rand_alpha_converted(const Handle& term, const Handle& vardecl)
+const
 {
 	// Clone the rule
 	Rule result = *this;
+	HandleSet target_vars = gen_variables(term, vardecl).varset;
+	HandleSet result_vars = BindLinkCast(result._rule)->get_variables()
+			.varset;
```
ngeiswei: You don't need to cast ;-)

ngeiswei: Also, please use const reference instead of copy.

opencog/rule-engine/Rule.cc
```diff
@@ -648,4 +652,4 @@ std::string oc_to_string(const RuleTypedSubstitutionMap& rules)
 	return oc_to_string(rules, "");
 }
 
-} // ~namespace opencog
+} // ~namespace opencog
```
ngeiswei: Now that's not enough new line, you file needs to end by one new line (not two, just one).

opencog/rule-engine/Rule.h
```diff
@@ -333,4 +339,4 @@ std::string oc_to_string(const RuleTypedSubstitutionMap& rule_ts_map);
 
 } // ~namespace opencog
 
-#endif /* RULE_H_ */
+#endif /* RULE_H_ */
```
ngeiswei: You need to add back that new line (not two, only one).

ngeiswei:
Also, you need to put that in a branch other than the master. I don't know if it's a requirement for github but it's at least a good habit to take. So please close that PR and make a new one with a branch specially for that PR.

PR #1724, opened 2018-05-30
resolve #1707

### LINKED ISSUE #1707: build python2 and/or python3 bindings
Currently, only one or the other is build depending on what is installed.  Instead, it should build bindings for both version, if both devel environments are installed.

SomeUserName1:
#1707


ngeiswei:
@SomeUserName1 I think you pushed the wrong branch.

PR #1729, opened 2018-06-11
Mac OS build fix

issacnitin:


opencog/atoms/core/ScopeLink.cc
```diff
@@ -270,7 +270,7 @@ constexpr size_t get_fvna_offset<8>(){
 template<typename T>
 ContentHash fnv1a_hash (ContentHash & hval, T buf_t)
 {
-	uint size = sizeof(buf_t);
+	uint64_t size = sizeof(buf_t);
```
linas: actually, `size_t` would have been the correct type: is always the right size.  (Its being used immediately below)

PR #1733, opened 2018-06-12
Make the matrix (atom-pair) API more flexible

linas:
The atom-pair API was not sufficiently general to allow certain types of pairs to be explore certain types of relationships.  Make the API more general.   This requires a corresponding pull for the opencog repo; it affects the language-learning code.


linas:
The corresponding pull req for language-learning is #3202

Issue #1726, opened 2018-06-01
Check hash collision between atoms at run-time


ngeiswei:
This issue is related to #1625.

To measure the quality the atom hashing function it would be convenient to be able to enable some systematic check at run-time.

So more specifically, it could be done by adding some code somewhere under the AtomSpace API that checks, whenever an atom is added, whether its hash differs from all other existing atoms in the atomspace. If it does, since the atomspace refuses to add duplicates, it implies that they differ in content and thus that there is a hash collision. 

Obviously this code should be wrapped in a macro so that it doesn't slow down the atomspace in operation when disabled.

linas:
If you just want to measure the size of a typical hash bucket, you can place a counter into the loop at AtomTable.cc line 238-245 and 284-291 -- besides finding just 0 or 1, I think you'll frequently find 2 or more.   I believe its a Poisson distribution -- there is a finite but exponentially small probability of arbitrarily many hash collisions.

ngeiswei:
@linas I'm not interested in measuring whether 2 hashes are in the same bucket which is up to `std::unordered_multimap`'s implementation, as of now anyway. What I am interested in is how often `Atom::compute_hash()` computes the same hash of atoms that differ in content. A good implementation should produce such collision fairly rarely.

I don't think placing a counter into the loop at AtomTable.cc line 238-245 and 284-291 measures the size of a hash bucket. Based on `_Hashtable::equal_range`, http://en.cppreference.com/w/cpp/container/unordered_map/equal_range and http://www.cplusplus.com/reference/unordered_map/unordered_multimap/equal_range I believe it counts the number of identical hashes, thus is actually a good suggestion for what I'm interested in. So, if I'm correct, I do not expect it would be frequently 2 or more, it will only be rare. If not, `Atom::compute_hash()` needs to be fixed.

stellarspot:
Method AtomTable::add(...) put an atom using its hash as key and the atom as value.
Because of hash collisions the unordered_multimap is used.
Is it possible to use unordered_set instead of unordered_multimap and directly put in and get atoms from it?

ngeiswei:
@stellarspot I think replacing unordered_multimap by unordered_set would work, I suspect unordered_multimap is only here for historical reason, predating atom hash (when the atom ID used to be determined by a run-time counter).

linas:
Um no. Two different atoms can still hash to the same hash. There needs to be some way of handling this. 

[...]

PR #1735, opened 2018-06-14
#1726 Check hash collision between atoms at run-time

stellarspot:
It is a proposal to print different atoms from the  _atom_store with the same hash.

What is not clear for me is it good to print it to the logger with the fine level or may be it is better to do in different way?

I also saw the code '#define DEBUG 1' in the PatternMatch.cc. Was it done by purpose and should I use it in the hash collisions logging?

### DIFF
--- a/opencog/atomspace/AtomTable.cc
+++ b/opencog/atomspace/AtomTable.cc
@@ -48,6 +48,7 @@
 
 //#define DPRINTF printf
 #define DPRINTF(...)
+#define DEBUG 1
 
 using namespace opencog;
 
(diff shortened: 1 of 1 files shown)


ngeiswei:
It works like a charm, thanks!

It has detected a few hash collisions in the unit tests that I suspect exhibit some problem residue with the hash algorithm. I'll create a separate issue about that.

Meanwhile, could you replace the DEBUG macro by CHECK_ATOM_HASH_COLLISION as it is a rather specific type of debugging? And log at the warn level instead of fine?

ngeiswei:
The hash collisions discovered from this code are listed in this issue #1739 

To do that I added an `abort()` right after the logging message, run the unit tests and see what crashed. I'm not sure I'd recommend to let the `abort()` though, it feels a bit too dangerous.

linas:
@ngeiswei The `warn` level prints a stack trace. you probably don't want that. 

There is absolutely nothing wrong, at all, with hash collisions. They are normal and expected.  I'm deeply confused about why we are obsessing about this.

linas:
Never mind, I just looked at #1739 yes, the collisions there seem to pathological. Something about the atom type is not being handled properly.

ngeiswei:
@linas that code will not be enabled by default, it's just for this specific debugging purpose so I think it's fine to have warn level, it almost guaranties that the message will be printed regardless of the log level.

Regarding collision, yes you are right, I'm only interested in exhibiting pathological ones.

ngeiswei:
I meant disabled by default, not enabled.

ngeiswei:
I needed it so I went ahead to add the little things I suggested and merged. Thanks @stellarspot !

PR #1736, opened 2018-06-14
CMake Guile required for compilation

issacnitin:
Found that libattentionbank.dylib won't be built without guile installed in Mac OS(fails at linking). Not sure if it happens in other Unix based OS, raising a PR for the concern.


linas:
Thanks!

PR #1744, opened 2018-06-18
Various fixes

roman-khimov:
Assorted fixes for different problems mostly found with static analysis tools (cppcheck and clang). Please, be very careful with this changeset as I have zero knowledge of this codebase and can easily break things.

### DIFF
--- a/opencog/cython/PythonEval.cc
+++ b/opencog/cython/PythonEval.cc
@@ -299,8 +299,6 @@ void opencog::global_python_initialize()
     // Throw an exception if this is called more than once.
     if (already_initialized) {
         return;
-        throw opencog::RuntimeException(TRACE_INFO,
-            "Python initializer global_python_init() called twice.");
     }
 
     // Remember this initialization.
--- a/opencog/cython/PythonEval.cc
+++ b/opencog/cython/PythonEval.cc
@@ -296,7 +296,7 @@ void opencog::global_python_initialize()
 
     logger().info("[global_python_initialize] Start");
 
-    // Throw an exception if this is called more than once.
+    // Don't initialize twice
     if (already_initialized) {
         return;
     }
--- a/opencog/persist/gearman/DistSCM.cc
+++ b/opencog/persist/gearman/DistSCM.cc
@@ -62,7 +62,9 @@ using namespace opencog;
 
 bool DistSCM::keep_working = true;
 
-DistSCM::DistSCM(void) : ModuleWrap("opencog dist-gearman")
+DistSCM::DistSCM(void) :
+    ModuleWrap("opencog dist-gearman"),
+    worker(nullptr)
 {}
 
 void DistSCM::init(void)
--- a/opencog/persist/gearman/DistSCM.cc
+++ b/opencog/persist/gearman/DistSCM.cc
@@ -230,12 +230,6 @@ std::string DistSCM::dist_eval(const std::string& scm_string,
 			"Gearman: Memory allocation failure on client creation");
 	}
 
-	int timeout=-1;
-	if (timeout >= 0)
-	{
-		gearman_client_set_timeout(&client, timeout);
-	}
-
 	gearman_return_t rc = gearman_client_add_server(&client,
                               "localhost", GEARMAN_DEFAULT_TCP_PORT);
 	if (rc != GEARMAN_SUCCESS)
--- a/opencog/persist/memcache/AtomCache.cc
+++ b/opencog/persist/memcache/AtomCache.cc
@@ -226,7 +226,7 @@ void AtomCache::load_list(AtomTable &table, int depth)
 	size_t klen = snprintf(keybuff, KBSIZE, "depth-list-%d", depth);
 	val = memcached_get(mc, keybuff, klen, &vlen, &flags, &rc);
 
-printf("duude depth=%d nodelist len=%d\n", depth, vlen);
+printf("duude depth=%d nodelist len=%zu\n", depth, vlen);
 	if ((vlen == 0) || (NULL == val)) return;
 
 	unsigned long ilc = load_count;
--- a/opencog/persist/memcache/AtomCache.cc
+++ b/opencog/persist/memcache/AtomCache.cc
@@ -30,7 +30,9 @@
 
 using namespace opencog;
 
-AtomCache::AtomCache(const std::string server, int portno)
+AtomCache::AtomCache(const std::string server, int portno) :
+    maxdepth(0),
+    load_count(0)
 {
 	memcached_return rc;
 	mc = memcached_create(NULL);
--- a/opencog/persist/sql/multi-driver/SQLAtomStorage.cc
+++ b/opencog/persist/sql/multi-driver/SQLAtomStorage.cc
@@ -93,14 +93,28 @@ class SQLAtomStorage::Response
 		LLConnection* _conn;
 
 	public:
-		Response(concurrent_stack<LLConnection*>& pool) : _pool(pool)
-		{
-			tname = "";
-			itype = 0;
-			intval = 0;
-			_conn = nullptr;
-			rs = nullptr;
-		}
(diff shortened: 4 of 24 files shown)


ngeiswei:
The rest looks good.

opencog/persist/gearman/DistSCM.cc
```diff
@@ -230,12 +230,6 @@ std::string DistSCM::dist_eval(const std::string& scm_string,
 			"Gearman: Memory allocation failure on client creation");
 	}
 
-	int timeout=-1;
-	if (timeout >= 0)
-	{
-		gearman_client_set_timeout(&client, timeout);
-	}
```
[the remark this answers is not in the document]
linas: Beats me. Mandeep wrote the gearman code; he writes crufty code. I cleaned up some of it but not all.

I like gearman in principle -- its small, its simple, it's direct; much easier to understand, control, debug that some of the super-whizzy big, distributed computing harnesses.  But given that no one is using it, I don't see that its worth much effort right now.

opencog/persist/memcache/AtomCache.cc
```diff
@@ -30,7 +30,9 @@
 
 using namespace opencog;
 
-AtomCache::AtomCache(const std::string server, int portno)
+AtomCache::AtomCache(const std::string server, int portno) :
+    maxdepth(0),
+    load_count(0)
 {
```
linas: All of the code in the memcache directory is based on an ancient design and will not actually work any longer.  Its probably best to discard the entire directory.

opencog/persist/sql/multi-driver/SQLAtomStorage.cc
```diff
@@ -93,14 +93,28 @@ class SQLAtomStorage::Response
...
+		    store(nullptr),
+		    pvec(nullptr),
+		    uvec(nullptr),
+		    tname(""),
+		    fltval(0),
+		    strval(nullptr),
+		    lnkval(nullptr),
+		    get_all_values(false),
+		    intval(0)
+		{}
```
linas: Is this a wise idea? Most of these values were not previously initialized, because I wanted to avoid the CPU overhead of actually initializing them. Most of these values are neither set, nor looked at; when they are looked at, they are also set. 

Perhaps the CPU overhead of initializing this is a drop in the bucket, and I shouldn't complain.  Perhaps it would be more efficient to just `bzero` the entire struct?

linas:
Yeah, mostly looks good to me. How did you find all of this? You must have some tool, as reading this much code would be .. oof. I couldn't do it.

linas:
Oh, right you explained it -- cppcheck and clang. I like almost all of the changes, except for the initialization of the `Response` object in the SQL code.  That object is always created on stack, and I'd like to avoid the overhead of clearing the stack frame.

opencog/persist/sql/multi-driver/SQLAtomStorage.cc
roman-khimov: I think with modern compiler you'll probably get something very efficient anyway, wiping a set of bytes is not that hard to optimize. Also, IIUC this code wraps around responses from the SQL DB so its whole execution time should be like nothing compared to SQL query/response time with networking or even interprocess communication.

But this improves reliability and correctness of the code, especially wrt to changes from random people from the Internet (like me) who prefer to think of constructed objects as something ready for the game in every aspect of it, so I think it's worth it.

opencog/persist/memcache/AtomCache.cc
roman-khimov: OK, less code --- less problems.

opencog/query/PatternMatchEngine.cc
```diff
@@ -2143,7 +2143,8 @@ PatternMatchEngine::PatternMatchEngine(PatternMatchCallback& pmcb)
 	: _pmc(pmcb),
 	_nameserver(nameserver()),
 	_varlist(NULL),
-	_pat(NULL)
+	_pat(NULL),
```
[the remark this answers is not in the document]
roman-khimov: Sure.

opencog/cython/PythonEval.cc
```diff
@@ -299,8 +299,6 @@ void opencog::global_python_initialize()
     // Throw an exception if this is called more than once.
```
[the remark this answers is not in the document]
roman-khimov: ACK

roman-khimov:
Yeah, that's the way static analysis works and that's one of the things I usually do when I try to get an idea of what's going on in an unknown codebase. There are some more that I didn't fix like a ton of cppchecks "Class XYZ has a constructor with 1 argument that is not explicit." (not sure all of them worth fixing) and a very hairy case of "Returning null reference" in `opencog::attentionbank()` found by Clang (but as I saw in some other code you do pass `NULL` there as a feature, so even though it scares me it's not that trivial to fix and certainly requires way better understanding of the codebase).

I'll probably close this PR as I intend to rebase the branch updating some commits and I'm not sure how GitHub will react to `push -f`.

opencog/persist/sql/multi-driver/SQLAtomStorage.cc
linas: OK,, that's fine. lets keep this change then.

linas:
You can keep this open, just add some more commits on top; you don't need to push "better" commits.  The history as is, is OK.

roman-khimov:
Well, if you're sure about that then OK, I'm just more used to cleaning up commit history before the merge, makes `git log` or `git blame` a bit less noisy.

linas:
attentionbank needs someone to, ummm, "do some gardening" . Its very old code that was never well-designed, and accumulated cruft over time. 

My biggest concern with attentionbank is that it has negative performance impacts, reaching into every nook and cranny, even when the nooks and crannies don't need or want the attentionbank.  (viz, the use of atom signals to monitor state).  The concept of "maintianing a single bank balance" is too strict, I think.  Bank balances require serialized, globalally-visibile locks (e.g. so that two crooks standing at two ATM cash machines at the same time cannot overdraw a single bank account by more money that is in there).   We don't need this metaphor -- it requires too much heavy-weight baggage.  A better metaphor would be like "keeping your blood sugar in balance" -- the exact amount of sugar in your blood is not important, as long as the approximate amount stays within bounds. 

However, re-writing the attentionbank to switch from a "economics/banking" model to a "blood-sugar" model would be a big change.  I'll open a feature for this...

linas:
attentionbank -- #1745

linas:
> just more used to cleaning up commit history before the merge, makes git log or git blame a bit less noisy.

Hmm. Yeah. Up until now, there's been so few of us, that this as not been a concern. 

By keeping this pull req going, we preserve the sequence of review comments, and that's useful too; that way, I don't have to start a review, all over again.  I can be sure that nothing strange is in this pull request, but it you open a new one, then I have to start the review all over again... and that's either more work, or I skip the review and maybe let bugs through....

ngeiswei:
I think github handles well `push -f`, it'll just delete the associated comments (viz. deleting the last commit about dropping memcache shouldn't be too problematic).

ngeiswei:
Oh, of course if you rebase the entire PR then all comments will go.

roman-khimov:
I've dropped the memcached commit, seems to be OK wrt this PR and comments.

ngeiswei:
@roman-khimov did you do a force push? Cause I'm still seeing d721617 (obviously push it to another branch before deleting it otherwise you risk to loose it, although git reflog might be able to recover it).

roman-khimov:
Yep, I did a forced push to the branch so it should be OK to merge. The memcached commit I have only locally now as that will need to be rebased after this changeset merge.

ngeiswei:
Hmm, but I see d721617. Anyway, I'm merging, it's not a big deal that commit ends up to be included.

ngeiswei:
d721617 isn't included so I guess github has a bug. Anyway, thanks man!

PR #1749, opened 2018-06-19
persist/memcache: drop it

roman-khimov:
As discussed in the #1744 review this code is old, broken and not worth
fixing.


linas:
Thanks.

Issue #1751, opened 2018-06-20
Guile needs min version increase due to ice-9 atomic


ngeiswei:
Due to https://github.com/opencog/atomspace/blob/07adea5d2e431182f9bd9fdb8b5dfdfaeec99727/opencog/matrix/object-api.scm#L200 guile requires `ice-9 atomic`, which has been introduced in version 2.2.0 of guile (according to its NEWS file).

Current minimum version requirement for guile is 2.0.9, so unless ice-9 atomic can be substituted by something else, this must be upgraded to 2.2.0.

ngeiswei:
That explains #1748

ngeiswei:
I'm having a randomly intermittent `corrupted double-linked list` error on `BackwardChainerUTest` since I've done the update...

linas:
Since an update to guile-2.2.0?  How did you update?

The nlp readme states that guile-2.2.2.1 is needed due to some kind of bug in guile-2.2.0  The instructions there also say:
```
   git clone git://git.sv.gnu.org/guile.git
   git checkout stable-2.2
```

I've been running the version
```
commit 9846178c69445142ef0b9432417453d2d4de6635
Author: Andy Wingo <wingo@pobox.com>
Date:   Wed May 31 21:52:15 2017 +0200

    Update guile.m4 to check for 2.2 by default
    
    * meta/guile.m4 (GUILE_PKG): Update default Guile versions to 2.2.
```

and have not tried anything newer yet.

ngeiswei:
Thanks. I've had updated to 2.2.2 (pre-built package) on my Fedora box, and to 2.2.0 (built from source) on my Debian Jessie box, same `corrupted double-linked list` error.

I've trying to compile the last version on git but it seems to block at
```
  BOOTSTRAP GUILEC ice-9/psyntax-pp.go
```

linas:
Some of those `BOOTSTRAP GUILEC` have taken over a hour to to compile, on my machine. 

How often does the `corrupted double-linked list` hit?

I just now updated to today's version of guile stable-2.2 (after running the previous one for a year) and have not seen any problems yet.  all unit tests, including BackwardChainerUTest passed twice so far.

linas:
This is currently running, with no problems so far:
```
#! /bin/bash

for i in `seq 1 500`;
do
    # ./tests/scm/MultiThreadUTest      >> /dev/null
    ./tests/rule-engine/backwardchainer/BackwardChainerUTest  >> /dev/null
[...]

PR #1752, opened 2018-06-20
Uplift guile min version requirement to 2.2.2

ngeiswei:
Because opencog/matrix/object-api.scm requires ice-9 atomic that has
only been introduced in guile 2.2.0.

To address #1751


ngeiswei:
@linas I let it on you to decide (guile upgrade or ice-9 atomic alternative).

ngeiswei:
Per https://github.com/opencog/atomspace/issues/1751#issuecomment-398789406 I've increased the patch number to 2, thus min requirement is 2.2.2.

Any problem merging that?

linas:
I'm concerned about raising the bar for complete beginners. But. Oh well.

Issue #1748, opened 2018-06-19
VectorAPIUtest fails


ngeiswei:
EDIT: see https://github.com/opencog/atomspace/issues/1748#issuecomment-403256028 for a description of the solution to that issue.

With the following error
```
Backtrace:
In ice-9/boot-9.scm:
 160: 13 [catch #t #<catch-closure 1667060> ...]
In unknown file:
   ?: 12 [apply-smob/1 #<catch-closure 1667060>]
In ice-9/boot-9.scm:
 160: 11 [catch #t #<catch-closure 1669c80> ...]
In unknown file:
   ?: 10 [apply-smob/1 #<catch-closure 1669c80>]
   ?: 9 [call-with-input-string "(eq? 'WordNode (bapi 'left-type))" ...]
In ice-9/boot-9.scm:
2412: 8 [save-module-excursion #<procedure 165d780 at ice-9/eval-string.scm:65:9 ()>]
In ice-9/eval-string.scm:
  44: 7 [read-and-eval #<input: string 16599c0> #:lang ...]
  37: 6 [lp (eq? (quote WordNode) (bapi (quote left-type)))]
In ice-9/eval.scm:
 387: 5 [eval # ()]
 386: 4 [eval #<memoized (bapi (quote left-type))> ()]
 393: 3 [eval #<memoized bapi> ()]
In unknown file:
   ?: 2 [memoize-variable-access! #<memoized bapi> #<directory (guile-user) c7c960>]
In ice-9/boot-9.scm:
 105: 1 [#<procedure 1608500 at ice-9/boot-9.scm:100:6 (thrown-k . args)> unbound-variable ...]
In unknown file:
   ?: 0 [apply-smob/1 #<catch-closure 1669c40> unbound-variable ...]

ERROR: In procedure apply-smob/1:
ERROR: Unbound variable: bapi
ABORT: unbound-variable
```
I can't find `bapi` anywhere in the repo. Has some file been dismissed?

linas:
Its in the tests directory, in `tests/matrix/basic-api.scm`

My guess is that the tests directory is not part of the file search path.

linas:
Hmm It does currently say 
```
   eval->eval("(add-to-load-path \"" PROJECT_SOURCE_DIR "\")");
   eval->eval("(add-to-load-path \"../../..\")");

   eval->eval("(load-from-path \"tests/matrix/basic-api.scm\")");
   eval->eval("(load-from-path \"tests/query/test_types.scm\")");
[...]

PR #1755, opened 2018-06-20
Parts of the test require postgres

linas:
Fix for issue #1748


ngeiswei:
It doesn't work for me because I have the libraries installed for enabling `HAVE_SQL_STORAGE`, I just don't have it configured.

PR #1759, opened 2018-06-21
Zermelo and Von Neumann set-theory number constructions

ngeiswei:
I actually added these for some quick memory benchmark as they kinda represent 2 extreme constructions.

Zermelo's numbers are just a linear chain where each link has an arity of 1 (or initially zero), while Von Neumann numbers have ever increasing link arities.

The results are rather interesting
```
(von-neumann 10000) ;; 10K
```
takes 4GB of RAM. While
```
(zermelo 10000000) ;; 10M
```
takes 5.5GB of RAM.

I think they could be added to the benchmark repo, (which I'm probably gonna do soon), but I added them here because I think they can be of interest beyond benchmarking.

### DIFF
--- a/examples/atomspace/set-numbers.scm
+++ b/examples/atomspace/set-numbers.scm
@@ -1,0 +1,66 @@
+;; Construct Zermelo and Von Neumann numbers
+
+(define (zermelo n)
+"
+  Construct the first n+1 Zermelo numbers
+
+  (zermelo 0) -> (Set)
+  (zermelo n) -> (Set (zermelo n-1))
+"
+  (if (= n 0)
+      (Set)
+      (Set (zermelo (+ n -1)))))
+
+(define (von-neumann n)
+"
+  Construct the first n+1 Von Neumann numbers
+
+  (von-neumann 0) -> (Set)
+  (von-neumann n) -> (Set (von-neumann n-1)) union (von-neumann n-1)
+"
+  (if (= n 0)
+      (Set)
+      (let* ((dec-n (+ n -1))
+             (dec-vn (von-neumann dec-n))
+             (dec-vn-mbrs (cog-outgoing-set dec-vn)))
+        (Set dec-vn dec-vn-mbrs))))
+
+;; Examples
+
+;; The following
+;;
+;; (zermelo 3)
+;;
+;; is expected to return
+;;
+;; (SetLink
+;;    (SetLink
+;;       (SetLink
+;;          (SetLink
+;;          )
+;;       )
+;;    )
+;; )
+
+;; The following
+;;
+;; (von-neumann 3)
+;;
+;; is expected to return
+;;
+;; (SetLink
+;;    (SetLink
+;;    )
+;;    (SetLink
+;;       (SetLink
+;;       )
+;;    )
+;;    (SetLink
+;;       (SetLink
+;;       )
+;;       (SetLink
+;;          (SetLink
+;;          )
+;;       )
+;;    )
+;; )


ngeiswei:
Kinda make sense I guess since 10M is not terribly far from 10K^2/2, the total number of outgoings of Von Neuman number 10K.

examples/atomspace/set-numbers.scm
```diff
@@ -0,0 +1,66 @@
...
+(define (zermelo n)
+"
+  Construct the first n+1 Zermelo numbers
+
+  (zermelo 0) -> (Set)
+  (zermelo n) -> (Set (zermelo n-1))
+"
+  (if (= n 0)
+      (Set)
+      (Set (zermelo (+ n -1)))))
```
linas: something wrong with saying `(- n 1)` ?

linas: You also realize that the function that you wrote is not tail-recusrsive, and thus it blows out the scheme stack?  So the RAM that you are measuring here includes guile RAM just as much as the atomspace RAM. 

Since tail-recursion tends to be example #2 in "how to program in scheme", we really should not violate that in examples.  Ditto for von neumann, below.

linas:
Several more remarks: 

1) I'm not sure what you are measuring; I consistently get 1.5KBytes/Atom over many many datasets.

2) Use of SetLink is sharply discouraged, precisely because it allows exactly things like 'von-neuman' to happen. Adding examples using a link type that we should be discouraging the use of seems like a poor idea.

ngeiswei:
I'm using List which is exactly the same as Set, except it doesn't reordered, not sure you find that much better. I guess what you dislike about the Von Neumann construct is that it allows links to have an arbitrary large outgoing set. Maybe I could reformulate the construct using `MemberLink`. But then the problem is that I need to create a `ConceptNode` for each set, which is kinda ugly. Or use a `ConsLink` that would be more elegant and wouldn't require arbitrary large outgoing set.

examples/atomspace/set-numbers.scm
ngeiswei: I hope so, I mean that's the RAM measured after algorithm has run, thus the memory stack no longer needs to be hold in memory, however I'm not sure Guile disallocates it.

BTW, I tried to save that in a file and load it to be sure it's not the stack but I failed (I think because I tried to write the string at once and Linux didn't let me allocate the space for it). I guess I should write an iterative or tail-recursive version of it, but maybe it's too atypical no bother anyway.

linas: Its just that writing recursive functions that are not tail recursive is just bad form, esp. here where it't not any harder.

As to RAM: The operating system reports the location of the `sbrk` -- the top of the heap -- as the as the size of the running process, even if the heap is almost entirely unallocated. By allocating lots of things in scheme, you force the heap to grow.  Even if they are freed later on, the increase of the size of the heap is permanent: the heap never shrinks.

linas: The one-way growth of the heap is not a guile thing, its a glibc thing.  I think I once saw some code in guile that tried to return excess guile-heap to glibc, but its not clear that glibc can actually shrink sbrk as a result (due to memory fragmentation -- just like disk fragmentation in MSDOS).  I think glibc can in some cases unmap virtual empty pages so that there is no physical page behind it; the operating system only counts physical pages in the RSS.  All of these various optimization tricks make it hard to predict what is actually happening. It's best to not allocate the memory in the first place, if you don't actually need it.

ngeiswei: I thought that sort of recursive calls would be allocated on the stack mostly, not the heap. However after reading a bunch on the subject, it seems the stack doesn't get dynamical disallocated by the OS either, so same problem.

Issue #1161, opened 2017-03-11
Provide Python API for values/protoAtoms


linas:
See issue #513 for the general description.

linas:
The correct API for this contains an abstraction for "Values".   A Value can be:

* a list of floats
* a list of strings
* a list of other Values
* (todo) an abstract object that wraps video and audio feeds, that tracks tensorflow nodes, and other rapidly changing data.

Given an atom and a key, one can fetch the Value on that atom, at that key.  One can store a new Value, at that atom, for that key.

TruthValue and AttentionValue (which are exposed in cython) are special cases of values. For historical reasons, they have a distinct API, which I guess we can keep for a while, but maybe (???) move away from eventually.

Wherever you can put a Value, you can also put an Atom.  (but not vice-versa)  So, given an Atom, and a key,  you can store another Atom at that key.  When you ask "what is the Value at this key?"  you can get back a Value, which might be a vector of floats, but it might be an atom. It might be a list of Values, and some of those Values might be Atoms.  

That's what the API should be. ProtoAtoms should be invisible, a behind-the-scenes implementation detail.  Note that there is no API for ProtoAtoms in scheme; only the Value API, described above.

I also hate the name "ProtoAtom"  That was a working title that never got fixed.  I'd like to change the name to something that starts with the letter "a" so that the git directories would be "athing", "base" and "core" in alphabetical order.

The problem with Value (upper-case V) is that its the same word as value (lower-case v)  and this can lead to a lot of confusion.  I cannot think of a good alternative.

The (atom, key,  Value) triple is a "valuation". Some people (e.g. wikipedia) call these "EAV"s, short for "Entity Attribute Value". Note that a collect of attribute-value pairs for a single entity looks like a "KVP database" or a "noSQL database" at that entity.  KVP==key-value-pair.  The name "valuation" comes from logic/model-theory.

vsbogd:
I would like to stop after https://github.com/opencog/atomspace/pull/1819 and switch to another task.
Things which are not done:

(1) LinkValue which can contain Atom instances

When LinkValue.to_list() method is called we should return the atom constructed from C++ Atom handle. This change requires additional constructor for ```atom.pyx``` Cython class to construct it from C++ Atom class instance without passing atomspace.

(2) (todo) an abstract object that wraps video and audio feeds, that tracks tensorflow nodes, and other rapidly changing data

[...]

PR #1761, opened 2018-06-22
Add ProtoAtom into Python API.

vsbogd:
Import C++ ProtoAtom class to cython (cProtoAtom). Add ProtoAtom python
wrapper. This is required to work with values using Python API.


opencog/cython/opencog/atomspace.pyx
```diff
@@ -9,3 +9,4 @@ include "nameserver.pyx"
 include "truth_value.pyx"
 include "atomspace_details.pyx"
 include "atom.pyx"
+include "protoatom.pyx"
```
ngeiswei: newline

ngeiswei:
I'm not experienced with that code so I'd rather have someone else merge, maybe @linas

opencog/cython/opencog/atomspace.pyx
vsbogd: Fixed, thanks

linas:
Um, I think this is the wrong abstraction.  The correct abstraction is "Values".   A Value can be:

* a list of floats
* a list of strings
* a list of other values
* (todo) an abstract object that video and audio feeds, that tracks tensorflow nodes, and other rapidly changing data.

Given an atom and a key, one can fetch the value on that atom, at that key.  One can store a new value, at that atom, for that key.

TruthValue and AttentionValue (which are exposed in cython) are special cases of values. For historical reasons, they have a distinct API, which I guess we can keep for a while, but maybe (???) move away from eventually.

Wherever you can put a value, you can also put an atom.  (but not vice-versa)  So, given an atom, and a key,  you can store another atom at that key.  when you ask "what is the value at this key?"  you can get back a value, which might be a vector of floats, but it might be an atom. It might be a list of values, and some of those values might be atoms.  

That's what the API should be. ProtoAtoms should be invisible, a behind-the-scenes implementation detail.  Note that there is no API for protoatoms in scheme; one the value API, described above.

I also hate the name "ProtoAtom"  That was a working title that never got fixed.  I'd like to change the name to something that starts with the letter "a" so that the git directories would be "athing", "base" and "core" in alphabetical order.

linas:
Values are almost, but not quite, like itty-bitty snippets of json that can be attached to an atom at location "key".

linas:
See issue #1161

linas:
I copied the above text into #1161 and it there are additional questions/clarifications, we should have the discussion there, not here.

vsbogd:
Hi Linas, the idea is not adding ProtoAtom as part of public API, I gave misleading name to this PR probably. The idea is add ProtoAtom as C++ ProtoAtom wrapper to allow using C++ API for adding values from Python. ProtoAtom instances cannot be constructed from Python directly. User should use Value type constructors instead, FloatValue is implemented to demonstrate the idea.

vsbogd:
Raised https://github.com/opencog/atomspace/pull/1768 to implement FloatValue API

linas:
Ahh, OK. So does this pull req need to be re-opened, or does the next one subsume everything here?

It will be about 36-48 hours before I can look at this in detail.  I looked at the example usage in the other pull req, that seemed exactly as it should be, so that's excellent.  A super-duper fast skim of the rest of the other pull req looks quite reasonable, I saw nothing to object to.

vsbogd:
PR https://github.com/opencog/atomspace/pull/1768 includes changes in this PR so it doesn't need to be reopened.
There is no rush with merging it so if you are able to look in few days it will be ok, thanks.

PR #1765, opened 2018-06-22
Replace create...Value macroses by functions

vsbogd:
Cython cannot import C macroses as functions so I replace create...Value
macroses by functions to import them in cython.
Delete createStreamValue as StreamValue doesn't have public
constructors.


linas:
Sigh. OK, I guess. But it just adds complexity without adding functionality. Which is almost never a good thing.  In the one or two places where cython needs to call these you could have just said "std::make_shared<foo>..." i.e. just hand-expanded the macro.

PR #1768, opened 2018-06-23
#1161 Add FloatValue into Python API 

vsbogd:
FloatValue is added into Python API, this PR addresses https://github.com/opencog/atomspace/issues/1161#issuecomment-399575981 and it is based on changes from https://github.com/opencog/atomspace/pull/1761

- ProtoAtom as C++ ProtoAtom wrapper is added to allow using C++ Atom::set_value()/get_value() methods from Python;
- To construct Value user should use Value type constructors; only FloatValue is implemented in this PR
- ProtoAtom instances cannot be constructed from Python directly;
- Example of float value usage is added in examples/python/values.py


linas:
This looks absolutely perfect, thanks!

linas:
Hmm. This doesn't compile on Ubuntu 14.04:
```
Error compiling Cython file:
...
    
    ctypedef shared_ptr[cProtoAtom] cProtoAtomPtr "opencog::ProtoAtomPtr"

cdef class ProtoAtom:
    cdef cProtoAtomPtr shared_ptr
    @staticmethod
   ^

/home/linas/src/atomspace/opencog/cython/opencog/atomspace.pxd:105:4: Cdef functions/classes cannot take arbitrary decorators.
```

OK, so Ubuntu 14.04 is pretty old, and I guess it's OK to stop supporting it. ... but still ... is that `staticmethod` thing really needed?

vsbogd:
Fixed in PR https://github.com/opencog/atomspace/pull/1777

PR #1772, opened 2018-06-25
Make Python loading from Scheme compatible with Python3

vsbogd:
Python3 doesn't have execfile anymore so scripts should be loaded using
exec(open('<script>')) call.


linas:
Thanks for the attention to detail!

Issue #1739, opened 2018-06-15
More hash collision between atoms


ngeiswei:
This is a continuation of issue #1625 with a list of new atom collisions. I suspect most, if not all, exhibit pathologies in the hash algorithm as opposed to being part of the random set of unavoidable collisions.

So here's the list

- [x] No longer colides
```
(SequentialAndLink
  (AbsentLink
    (EvaluationLink
      (PredicateNode "visible") ; [7535041300660910155][1]
      (ListLink
        (VariableNode "$x") ; [788143063270552312][1]
      ) ; [10028350822169473443][1]
    ) ; [11397036793678782603][1]
  ) ; [11424322274233087092][1]
  (EvaluationLink
    (GroundedPredicateNode "scm: incr-trig") ; [409184162435433586][1]
    (ListLink
    ) ; [9223882210250052939][1]
  ) ; [15246955578542806330][1]
) ; [14915126413073086491][1]
```
with
```
(SequentialOrLink
  (PresentLink
    (EvaluationLink
      (PredicateNode "visible") ; [7535041300660910155][1]
      (ListLink
        (VariableNode "$x") ; [788143063270552312][1]
      ) ; [10028350822169473443][1]
    ) ; [11397036793678782603][1]
  ) ; [11423741732093633805][1]
  (EvaluationLink
    (GroundedPredicateNode "scm: incr-trig") ; [409184162435433586][1]
    (ListLink
    ) ; [9223882210250052939][1]
  ) ; [15246955578542806330][1]
) ; [14915126413073086491][1]
```
- [x] No longer collides
```
(BindLink
  (TypedVariableLink
    (GlobNode "$star") ; [7221702073207368645][1]
    (IntervalLink
      (NumberNode "0.000000") ; [4895729370997130312][1]
      (NumberNode "1.000000") ; [2263783472620955547][1]
    ) ; [18413741763871150138][1]
  ) ; [9806233150850800417][1]
  (ListLink
    (ConceptNode "I") ; [8053501913945070544][1]
    (ConceptNode "love") ; [5772868835305624798][1]
    (GlobNode "$star") ; [7221702073207368645][1]
  ) ; [11983067370852204958][1]
  (ListLink
    (ConceptNode "Hey!") ; [518141375632718247][1]
    (ConceptNode "I") ; [8053501913945070544][1]
    (ConceptNode "like") ; [7248936160044154347][1]
    (GlobNode "$star") ; [7221702073207368645][1]
[...]

PR #1774, opened 2018-06-25
use intervalmap in hash computation(issue #1739)

noskill:


ngeiswei:
Unfortunately it visibly introduces a bug, caught by `PutLinkUTest` and `AlphaConvertUTest`.

I believe the issue is that 2 semantically equivalent scopes like
```
(Lambda
  (Evaluation
    (Predicate "P")
    (Variable "$X"))) 
```
and
```
(Lambda
  (Variable "$X")
  (Evaluation
    (Predicate "P")
    (Variable "$X")))
```
would have different hashes.

The unit test `HashUTest::test_scope_compute_hash_4()` is particular because the 2 scopes are not semantically equivalent. In `bl-3` the glob variable `"$star"` has type `(Interval (Number 0) (Number 1))`, meaning is a segment of 0 or 1 atom, while in `bl-4` the same glob variable is an unrestricted segment.

So it seems to fix it you would need to factor in the arity combined with whether or not the variable declaration is restricting the variable types or not. I suppose `ScopeLink::get_variables()` could help you there.

ngeiswei:
@vsbogd there isn't exactly a normalize method (or sort of, see below). What is happening is that if you add for instance
```
(Scope (Variable "$X") (Variable "$X"))
```
then add/get the equivalent scope
```
(Scope (Variable "$X"))
```
you get in return
```
(Scope (Variable "$X") (Variable "$X"))
```

If the first time you've had added
```
(Scope (Variable "$X"))
```
then you'd get that one in return of an equivalent one.

Applying automatic normalization (thus returning the declaration free version every time) seems sensible to me.

There is the existing normalization, there is some in `FoldLink` and its subtypes, see `FoldLink::delta_reduced` but it is only used for arithmetic, at execution time (ultimately called by `ArithmeticLink::execute()`), not insertion time.

I believe some normalization could take place at insertion time indeed, when we know the non-normalized form is never desired.

I would think it is however preferable to have a hash that works well on normalized and non-normalized terms alike (based on Richard Stallman wisdom about writing robust programs).

vsbogd:
I agree that hash should return equal value on normalized and non-normalized forms. The idea is to to implement the normalization rules once and then reuse them for normalization and for hash function calculation. Implementing separate normalization logic in hash function can be complex and error-prone.

linas:
Well, you would not need to normalize, if you just reject this pull request.  So what's the problem, really? 

This pull request does not describe what problem it is, that it is trying to solve. Without a description of the problem, it becomes hard to tell if this pull request solves it.

linas:
I don't really quite like normalization.  Maybe it's OK-ish, but it adds a lot of extra complexity to a process that is already complex. It also makes things slower. It also requires a lot of extra unit tests, to make sure that things are correctly normalized.  To me, that seems like a huge cost, a pointless cost -- and worse -- there seems to be no other reason to perform this normalization, other than for this pull req. 

So, why make extra work for yourself, and make extra work for future maintainers, to implement a function that does not seem to be needed?

ngeiswei:
@linas that pull request attempt to complete address issue #1739 as indicated in its title. The hash collision it attempts to fix is one taken from our unit tests (I forgot which, but it doesn't matter), see that comment https://github.com/opencog/atomspace/issues/1739#issuecomment-399911589 more specifically.

That is said I agree that a description in the comment section of the request would be welcome just as a reminder.

Regarding normalization, it is not needed by that pull request! It's perhaps not needed at all but it feels sensible, returning the first equivalent scope link seems rather arbitrary. But that's not urgent or anything like that.

@vsbogd I agree with you in principle, we want to avoid duplicating code of equivalent functional behaviors, however as you can see normalization isn't necessarily in the plan, while avoiding hash collision kinda is (well, I'm the one pushing for that cause I rely a lot on hashes while managing my experiments, cause a hash is easier to parse than a giant atomese program).

noskill:
HashUTest, PutLinkUTest and AlphaConvertUTest are passing now.

linas:
OK Maybe I'm too strident about normalization. It's just somewhat tricky, due to the typing that variables can have.  For example `(TypedVariable (Glob "$X") (Interval (Number 1) (Number 1)))`  is identical to `(Variable "$X")` -- there is a low-priority work item that points this out, and proposes that GlobNodes should maybe be the same thing as VariableNodes.  Or something along those lines. (we need to keep it efficient, since globs have heavy processing overhead, and also backwards compat, since there are lots of glob users) Currently Globs and variables are the same, except that globs match one or more times, variables match once and only once.

Re: the current variant of this pull-req -- it hashes in the globs ... which is kind-of-ish OK, except that:

1) Nil doesn't use Globs in his code, so this doesn't help him much or at all,

2) The normalization issue above ... which is probably not handled correctly in the current hashing code, anyway.

vsbogd:
I don't insist on normalization, separate issue is ok.

opencog/atoms/core/ScopeLink.cc
```diff
@@ -255,9 +255,14 @@ ContentHash ScopeLink::compute_hash() const
 	}
 	fnv1a_hash(hsh, vth);
 
+	const Variables & vars = this->get_variables();
+	for(const auto& pr: vars._glob_intervalmap){
```
ngeiswei: @noskill you can actually use `_varlist` instead of `get_variables()` (sorry I misguided you.)

ngeiswei:
Thanks!

PR #1777, opened 2018-06-26
Move ProtoAtom factory method out of class.

vsbogd:
Addressing https://github.com/opencog/atomspace/pull/1768#issuecomment-400109533
To make code compatible with older versions of Cython/Python, move
static methon into separate function and rename it accordingly.

### DIFF
--- a/opencog/cython/opencog/atom.pyx
+++ b/opencog/cython/opencog/atom.pyx
@@ -172,7 +172,7 @@ cdef class Atom(object):
         cdef cProtoAtomPtr value = self.get_ptr().getValue(
             deref((<Atom>key).handle))
         if (value != NULL):
-            return ProtoAtom.from_cProtoAtomPtr(value)
+            return createProtoAtom(value)
         else:
             return None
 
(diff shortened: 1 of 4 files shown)


linas:
Thanks!  You could also have said something like `Ubuntu 14.04 is too old to support`, and I probably would not have complained too much. It's what I'm running on an old laptop, which I should probably upgrade someday. Maybe.

linas:
FYI, on Ubuntu 14.04, after applying above pull req, I now get:
```
[ 22%] Built target atomutils
[ 37%] Built target atomcore
[ 37%] Cythonizing atomspace.pyx

Error compiling Cython file:
...
from libcpp.vector cimport vector
from libcpp.list cimport list as cpplist
from libcpp.memory cimport shared_ptr
^

/home/linas/src/atomspace/opencog/cython/opencog/atomspace.pxd:3:0: 'libcpp.memory.pxd' not found
```
So it seems like 14.04 simply has an old version of cython...

linas:
Yes, please update cmake to look for 0.23 or newer.

vsbogd:
Raised PR https://github.com/opencog/atomspace/pull/1786

PR #1788, opened 2018-06-27
#1161 Add ProtoAtom method to convert FloatValue to Python list

vsbogd:
Add method to convert FloatValue to Python list. Issue https://github.com/opencog/atomspace/issues/1161


linas:
thnaks!

PR #1789, opened 2018-06-27
#1161 Add scheme_eval_v which returns value to python

vsbogd:
Issue https://github.com/opencog/atomspace/issues/1161. Add scheme_eval_v method which is analog of SchemeEval::eval_v() to evaluate Scheme code which returns values from Python.


linas:
Thanks!

Issue #1754, opened 2018-06-20
MemberLink swap arguments


ngeiswei:
Currently member link is used as follows
```
MemberLink
  <element>
  <concept>
```
Linas has suggested a while ago to swap the arguments, as follows
```
MemberLink
  <concept>
  <element>
```

I agree with that for the following reasons

1. That way it would be consistent with `EvaluationLink` that follows
```
EvaluationLink
  <predicate>
  <element>
```
2. It agrees with other similar links like `AtTimeLink`
```
AtTimeLink
  <node>
  <time>
```
3. There's a more conceptual reason, coming from functional programming, that arguments should be ordered from least frequently to most frequently changing argument, because you easily partially evaluate them on the first arguments and obtain, by currying, a function on the remaining arguments. It make more sense to specialize (partially evaluate) a function by fixing its least frequently changing arguments than than the opposite. For instance if you have
```
power: Number -> Number -> Number
```
by setting the first argument as the base and the second as exponent, you can easily define specializations such as
```
power-of-2: Number -> Number
power-of-2 = power 2
```

ngeiswei:
Any opinion? In the absence of feedback with the next 48, I'll probably do the change.

vsbogd:
Arguments make sense. What about InheritanceLink(s) ? Should they be transformed by the same way?

ngeiswei:
I think `InheritanceLink`(s) should remain unchanged, first there isn't an obvious order based on frequency, and it is consistent with `ImplicationLink`.

linas:
I apologize for the bad reaction 8 days ago.  I have several large datasets that  have MemberLinks in them, so the pull request would scramble them.  However, on closer inspection, these datasets are low quality and need to be recomputed, anyway.  I'm gonna try to carefully think this through before launching the next major run.  Perhaps making the swap at that time, after data analysis on the current datasets is complete, would be a good time to make the change.

ngeiswei:
@linas just a reminder to have you ponder about it before relaunching your experiments ;-)

ngeiswei:
BTW, it's not sure that your experiments would be affected by this change. As far as I know there are only 2 core components that assume a particular order for `MemberLink`, which are ECAN and the URE (and of course components that use these like `OpenPsi`).

linas:
[...]

PR #1798, opened 2018-06-28
Swap MemberLink arguments

ngeiswei:
For issue #1754 

In the absence of feedback I'll merge in two days or so.

### DIFF
--- a/examples/c++-api/scm/bc-config.scm
+++ b/examples/c++-api/scm/bc-config.scm
@@ -6,12 +6,12 @@
 ; is currently undefined, we might settled with probabilities but it's
 ; not sure)
 (MemberLink
-   pln-rule-modus-ponens-name
    (ConceptNode "BC_DEMO_RB")
+   pln-rule-modus-ponens-name
 )
 (MemberLink (stv 1 1)
-   pln-rule-deduction-name
    (ConceptNode "BC_DEMO_RB")
+   pln-rule-deduction-name
 )
 
 ; termination criteria parameters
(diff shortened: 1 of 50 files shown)


linas:
NO!!!!!!!!!!!!!!!!!!!!!

You cannot do this!  You can invent a new link type, with the swapped arguments, but you cannot swap the arguments on the existing member links!   I have maybe 50 databases that contain MemberLinks in them, with the old order. These databases took me over a year, two years to compute, and represent maybe a dozen cpu-years of calculations.  There is no practical way for me to go into those databases, load them into RAM, find every usage of MemberLink, swap the order, and save them to disk again.   

One reason some of these can't be changed is that some are just archives/backups, to prevent data loss.  It's fundamentally wrong to try to alter an archive ... it obviates the whole point of having an archive.  The other problem is that I only have a 1 TB disk, and the databases don't all fit on there at the same time, so I have to juggle them.  I can't drop everything, spend a few weeks attempting to convert MemberLink order on umpteen databases.   This is a non-starter.

Invent a new kind of MemberLink that does the right thing. The old MemberLink is written in stone, it cannot be changed.

ngeiswei:
That's a bummer, I can't tell I don't understand though (you're actually the one who first suggested that, not so long ago, maybe a year or so).

Well, if we can't touch anything that breaks backward compatibility we might just as well adopt https://semver.org/

linas:
I apologize for the bad reaction. I have several large datasets that have MemberLinks in them, so the pull request would scramble them. However, on closer inspection, these datasets are low quality and need to be recomputed, anyway. I'm gonna try to carefully think this through before launching the next major run. Perhaps making the swap at that time, after data analysis on the current datasets is complete, would be a good time to make the change.

I'm re-opening this request, but please don't merge yet; I need to finish doing data analysis on the current datasets (which contain MemberLinks) before I can discard/archive them.

Issue #1773, opened 2018-06-25
cog-mean vs cog-tv-mean


ngeiswei:
We have the following scheme bindings
```
cog-mean
```
that returns the mean of the TV of the given atom, https://github.com/opencog/atomspace/blob/master/opencog/guile/SchemeSmobAtom.cc#L112

And
```
cog-tv-mean
```
that returns the mean of the given TV, https://github.com/opencog/atomspace/blob/master/opencog/guile/SchemeSmobTV.cc#L471

But I wonder if it shouldn't be the other way around, `cog-mean` that takes a TV and `cog-tv-mean` that takes an atom. In that case the `tv` indicates the output object type over which the property is being accessed, while current it indicates the input type.

Anyway, it's not very important but  I wonder what you think.

vsbogd:
I agree, that swapping these names seems to be reasonable. 

BTW what it the reason of introducing of cog-tv-mean (in current notation), why it is not defined purely in Scheme code as superposition of cog-mean and cog-tv? Is it because of performance reasons? To not call Sheme/C++ bindings code twice?

ngeiswei:
The scheme version `tv-mean` https://github.com/opencog/atomspace/blob/master/opencog/scm/opencog/base/utilities.scm#L80 used to be defined as
```
(assoc-ref (cog-tv->alist tv) 'mean)
```
Which was replaced by `cog-tv-mean` apparently for efficiency reasons.

linas:
Swapping the names would be reasonable, except that there is a huge amount of code all over the place, and this destablizes things. I don't know about you guys, but I have a lot of code that is not in git, it's scattered across multiple machines, its used either to run experiments, or to crawl over the output of the same, and prepare charts and graphs.  Some of this, I revisit only infrequently, and it would be a pain to track it all down to make the changes.

linas:
The original (ten-year-old) implementation used the `assoc-ref` but somewhere along the way, I discovered that direct references are faster.  For things like counting, it is a significant bottleneck.

Hmm. Actually, cog-tv-count should probably have an "increment" function on it, to avoid fetching, adding one, and then storing ....

ngeiswei:
I understand, @linas, swapping stuff is horribly error prone. On the other hand I'd rather have that sort of clean-ups occur now than later when opencog is more widely used.

[...]

PR #1799, opened 2018-06-29
Swap cog-tv-mean with cog-mean

ngeiswei:
as well as cog-tv-confidence with cog-confidence, and cog-tv-count
with cog-count.

PR for #1773

### DIFF
--- a/opencog/guile/SchemeSmob.cc
+++ b/opencog/guile/SchemeSmob.cc
@@ -324,9 +324,9 @@ void SchemeSmob::register_procs()
 	register_proc("cog-value",             2, 0, 0, C(ss_value));
 	register_proc("cog-tv",                1, 0, 0, C(ss_tv));
 	register_proc("cog-as",                1, 0, 0, C(ss_as));
-	register_proc("cog-mean",              1, 0, 0, C(ss_get_mean));
-	register_proc("cog-confidence",        1, 0, 0, C(ss_get_confidence));
-	register_proc("cog-count",             1, 0, 0, C(ss_get_count));
+	register_proc("cog-tv-mean",           1, 0, 0, C(ss_get_tv_mean));
+	register_proc("cog-tv-confidence",     1, 0, 0, C(ss_get_tv_confidence));
+	register_proc("cog-tv-count",          1, 0, 0, C(ss_get_tv_count));
 
 	// Truth-values
 	register_proc("cog-new-stv",           2, 0, 0, C(ss_new_stv));
@@ -343,9 +343,9 @@ void SchemeSmob::register_procs()
 	register_proc("cog-ftv?",              1, 0, 0, C(ss_ftv_p));
 	register_proc("cog-etv?",              1, 0, 0, C(ss_etv_p));
 	register_proc("cog-tv->alist",         1, 0, 0, C(ss_tv_get_value));
-	register_proc("cog-tv-mean",           1, 0, 0, C(ss_tv_get_mean));
-	register_proc("cog-tv-confidence",     1, 0, 0, C(ss_tv_get_confidence));
-	register_proc("cog-tv-count",          1, 0, 0, C(ss_tv_get_count));
+	register_proc("cog-mean",              1, 0, 0, C(ss_tv_get_mean));
+	register_proc("cog-confidence",        1, 0, 0, C(ss_tv_get_confidence));
+	register_proc("cog-count",             1, 0, 0, C(ss_tv_get_count));
 	register_proc("cog-tv-merge",          2, 0, 0, C(ss_tv_merge));
 	register_proc("cog-tv-merge-hi-conf",  2, 0, 0, C(ss_tv_merge_hi_conf));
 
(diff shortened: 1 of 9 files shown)


ngeiswei:
Come on, click on the merge button, you know you want to!

linas:
OK this broke absolutely everything and is now ciorrupting databases.

I told you to NOT DO THIS and YOU DID IT ANYWAY.  Even though I made it abundantly clear what would happen. WTF. This wipes out weeks worth of work or more -- keeping the various computations going is like a juggling act, and watching the computations is like watching paint dry.  Its stunningly boring.

linas:
Look, you cannot just merge stuff like this, with zero review, zero discussion, and laying the groundwork for the changes.   This is an API change that totally breaks backwards compatibility, which is always a bad thing to do in programming, precisely because it breaks code that you do not have access too.   Usually, when one breaks backwards compat, one does it over the course of 12-24 months, so that people have an opportunity to update their scripts to use the new API.

I have something like six different containers running, performing calculations.  The change here requires me to manually update maybe 3-5 scripts per container -- so maybe 20-30 scripts, which I have to manually check, update, and double-check so that I don't make an accidental mistake in teh updates.

The updates can only be done while the system is not actively running.  The typical jobs take from a few hours to a few weeks, which means that the earliest that I can cut over to the new API wold be a few weeks.   I just launched a job last night, that, at current rate of progress, might take a month to run.

The NUMBER ONE RULE OF PROGRAMMING is to not break the work that other people are doing.  This pull req breaks my work, in a bad, ugly, way.  It already corrupted one database, because I was not paying enough attention after doing a `git pull`.   This was a pure-evil change to make!

ngeiswei:
@linas, I didn't merge, my "motivational" message was addressed to you, @bgoertzel clicked on the merge button, he probably didn't know you had mixed feelings about it (or he did and that's a back clash of your incontrollable cursing).

Anyway, I think it's still worth keeping around for the day it's won't annoy you (I don't fully understand why it does, BTW, that's a rather trivial change if you use the right tool).

Since OpenCog has still a rather confined user base I'm a bit in a hurry to apply changes that breaks backward compatibility before it becomes too late.

ngeiswei:
> if you use the right tool
Vim: http://www.drchip.org/astronaut/vim/#SWAPSTRINGS
Emacs: https://github.com/xuchunyang/plur

linas:
Any change that breaks backwards compatibility to an existing API has to be dealt with in multiple steps, spread out over time. 

Here's what I deal with, on a daily basis. I am baby-sitting multiple language learning processes. One is digesting French, on a borrowed, older RedHat machine, which is sufficiently old, that I have to carry extensive patches to atomspace+opencog to get it to actually work. But it works.  Meanwhile, I have four different LXC containers running different kinds of processing on English. Two are computing clusters in two different ways, on one dataset, two more are preparing two other datasets, computing marginals for a new data structure. The first few have been running for weeks, the  other two look they'll run for days or longer.  I have to check on them all about 2x a day, to see if they're hung, crashed, whatever, make sure that there's forward progress. Its boring, and its tedious.   Anyway, all of them run custom, modified scripts, specially tailored to the specific datasets that are being handled. 

While I wait, I do development on a third machine, and test on a fourth.

The vector-API subsystem uses the tv-count in a variety of different ways.  One trick that I use is to overload the tv-count functions so that other sources of count-data can be used. (e.g. frequency, entropy, instead of count) A related trick is to filter out all atoms with a count below a certain value.  The upshot is that counts are flowing all over the place, in all kinds of fairly complicated ways. 

So I've got maybe half-a-dozen different machines with different OS'es and different versions of opencog, doing different things, which I'm trying to baby-sit and keep going, all of which are heavily dependent on getting counts correct, and then this comes in and destabilizes all of that.  I can upgrade one machine at a time, as a job finishes up.  The upgrade process is also time-consuming, boring and error prone, but it gets done eventually, as time permits.

Oh and then there's this: the above just generates data, which gets analyzed.  The analysis scripts, of which there are maybe hundreds, just litter my file-system. On several computers. Some scripts are copied into the *.lyx files which describe the data, but most are not. These kinds of scripts bit-rot over time -- this is a given -- but, as I re-do various graphs and charts, I get an opportunity to clean up those scripts. Again, its a bit tedious and boring: something old, mostly-forgotten, that doesn't work any more, and needs yet-another half-a-day of debugging.   So a change like this, which just increases the rate of bit-rot, make it just that much harder to keep things going.  There just aren't that many half-days in life, and to be forever-debugging is not really enjoyable.

This is not a tool or editor issue. This is an issue of managing hundreds of scripts for dozens of experiments littered all over multiple file systems over multiple computers.  There is no magic VIM or EMACS plugin that will track down, find all of these files, make the correct changes, perform the tests, and re-analyze the data to make sure nothing got borken.  It's got nothing to do with editors; its got everything to do with data management.

So -- its one thing to add new API's. Another to remove old, deprecated API's. But to change existing ones in an incompatible kind of way ... that just breaks lots of unknown things in unknowable ways.

ngeiswei:
Got it, thanks for the detailed explanation.

I suppose it's possible to achieve some middle ground, maybe using semver, for instance having a branch for each major version, where backward compatibility breakage is never supposed to happen, and a master or dev branch that allows (concerted) breakages, tagged as the next major version. Of course maintaining all that also takes extra half-days (i.e. merge backward compatible commits to the version branches)... but it seems we'll have to because the imperativeness to remain backward compatible and the pressure to introduce new changes that breaks it will simultaneously grow over time.

PR #1808, opened 2018-07-02
#1161 Add AtringValue python api

vsbogd:
Workin on https://github.com/opencog/atomspace/issues/1161
Add StringValue tests and implementation

### DIFF
--- a/lib/OpenCogAtomTypes.cmake
+++ b/lib/OpenCogAtomTypes.cmake
@@ -182,14 +182,14 @@ FOREACH (LINE ${TYPE_SCRIPT_CONTENTS})
         # identifier as the Python Atom object.
         IF (NOT TYPE_NAME STREQUAL "Atom")
             IF (ISVALUE STREQUAL "VALUE" OR ISSTREAM STREQUAL "STREAM")
-                # Implementing only FloatValue for now, other types of values
-                # will be added after design is agreed
-                IF (TYPE_NAME STREQUAL "FloatValue")
+                # Implementing only FloatValue and StringValue for now, 
+                # other types of values will be added later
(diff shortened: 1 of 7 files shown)


linas:
great, thanks!

linas:
Hmm. On Debian-stable, which I think is a pretty modern version, I get this error during compilation:
```

Error compiling Cython file:
...

cdef list vector_of_strings_to_list(const vector[string]* cpp_vector):
    list = []
    it = cpp_vector.const_begin()
    while it != cpp_vector.const_end():
        list.append(deref(it).decode('UTF-8'))
                            ^

/src/atomspace-git/opencog/cython/opencog/protoatom.pyx:23:29: Object of type 'string &' has no attribute 'decode'
```

vsbogd:
@noskill also reported it under https://github.com/opencog/atomspace/issues/1812
Probably this code works differently under python2 and python3, will look at it

noskill:
https://stackoverflow.com/questions/3826458/cython-conditional-compile-based-on-external-value
We can support both python versions by checking at compile time

vsbogd:
I cannot reproduce it: have tried clean build with python2 and python3.
According cython documentation it should work since cython 0.17
http://cython.readthedocs.io/en/latest/src/tutorial/strings.html#c-strings

@noskill, could you please try to compile, changing it on

    cdef list vector_of_strings_to_list(const vector[string]* cpp_vector):
        list = []
        it = cpp_vector.const_begin()
        while it != cpp_vector.const_end():
            list.append((<bytes>deref(it).c_str()).decode('UTF-8')) # convert std::string to bytes explicitly

noskill:
Code compliles without error with your fix

vsbogd:
Raised https://github.com/opencog/atomspace/pull/1813 to fix

Issue #1812, opened 2018-07-03
Error compiling Cython file


noskill:
I can't build atomspace anymore:

> [ 29%] Linking CXX shared library libatomcore.so
> [ 29%] Built target atomcore
> [ 29%] Cythonizing atomspace.pyx
> 
> Error compiling Cython file:
> ------------------------------------------------------------
> ...
> 
```
 cdef list vector_of_strings_to_list(const vector[string]* cpp_vector):
     list = []
     it = cpp_vector.const_begin()
     while it != cpp_vector.const_end():
         list.append(deref(it).decode('UTF-8'))
                             ^
```

> home/alex/cog/atomspace/opencog/cython/opencog/protoatom.pyx:23:29: Object of type 'string &' has no attribute 'decode'

atomspace version f7d3c2ed8f3d89b001e1310c5751892f34d96d8d
Cython version 0.25.2

vsbogd:
Hi Anatoly, have you tried making clean build? Cmake sometimes fails calculate cython dependencies correctly.

noskill:
Yes, i deleted the build directory. Didn't touch ccache though - i build with ccache.

vsbogd:
@linas, @noskill, if you don't have the issue I think we can close it.

noskill:
I don't have the issue anymore.

PR #1813, opened 2018-07-03
#1812 Fix cython compilation issue.

vsbogd:
Convert std::string to bytes explicitly before calling decode().
Fix for https://github.com/opencog/atomspace/pull/1808#issuecomment-401922293 and https://github.com/opencog/atomspace/issues/1812


ngeiswei:
I'm merging (based on trust rather than understanding given my lack of Cython familiarity).

PR #1814, opened 2018-07-03
Rebuild core_types.pyx when OpenCogAtomTypes.cmake updated

vsbogd:
Add OpenCogAtomTypes.cmake as dependency for OPENCOG_ADD_ATOM_TYPES
macros to trigger the build when OpenCogAtomTypes.cmake is touched.


opencog/atoms/proto/CMakeLists.txt
```diff
@@ -5,7 +5,10 @@ OPENCOG_ADD_ATOM_TYPES(atom_types.script
   atom_types.inheritance
   core_types.scm
   core_types.pyx)
-ADD_CUSTOM_TARGET(opencog_atom_types DEPENDS atom_types.h)
+
+ADD_CUSTOM_TARGET(opencog_atom_types DEPENDS atom_types.h
```
linas: this is overkill, since all five files will change, when the types.script file changes.

vsbogd: Yes, I was going to write comment on it but then thought that it is simpler add all files as dependencies.

PR #1815, opened 2018-07-03
#1161 Add RandomStream into Python api

vsbogd:
Work on https://github.com/opencog/atomspace/issues/1161. Add RandomStream value unit tests and implementation.


linas:
I'm kind-of not entirely thrilled with this.  There's nothing technically wrong with the actual code in the pull request; rather, the problem is that RandomStream is meant to be a proof-of-concept or example demo for a generic API that could allow time-varying values.  Adding formal API's and unit-tests for a provisional prototype is .. just not a good idea.

During email conversations with Alexey, a question came up: how should one map systems like tensorflow into the atomspace? The provisional answer was that "values can be time-varying", and a time-varying value does not have to actually store the actual value (which might be changing a trillion times a second e.g. pixels from a video camera), but instead should just provide methods to access that value. So the next question is "how would that work?"  The simplest thing I could think of for a "time-varying value" was a random number generator, so that got turned into a prototype demo.

Now, I think that I like the "stream-value" API. I guess. Maybe. Probably.  But we have very little experience using it, and it might need to change, or be discarded and replaced by some other idea. So writing a formal python API, and a formal unit test for this thing seems terribly premature.  Its not really ready for general use.

linas:
What would be interesting would be a generic python API for any kind of generic stream value.  That is, an AI that says "Create a value of type FooValue", and then "get the current value of FooValue".  So, the python API should be something like
```
    import Values from atomspace
    vvv = createValue ("FooValue")
    print ("The current Foo is=" + vvv)

   # FooValue is just like any other value
   atom.setValue (key, vvv)
...etc.
```

and then you could use RandomValue for Foo in the above.

linas:
viz something like this:
```
   vvv = createValue("FloatValue", 1, 2, 3)
  # or maybe this:
   vvv = createValue ("FloatValue", [1,2,3])

  # and thus, the api for RandomStream would be:
   vvv = createValue("RandomStream", 7)
  # Should print a list of random numbers:
  print ("its this:", vvv)
```

linas:
I don't care if the actual function name is `createValue` or something else.  I don't care if the first argument is a string, or something else .. e.g. a symbol.   It might be more efficient to use symbols instead of strings. 

However, it is important to assume that there might be all kinds of user-defined  Value types.  Not all of these will be defined in the atomspace -- e.g. some might be defined in agi-bio, e.g. for some polymerase chain-reaction PCR robot, or in some other repo, e.g. for interfacing with ROS, or some tensorflow repo, etc.   So we will never have a definitive list of all possible value types.  I think it is reasonable to assume that all values will be written in C++; I don't think it is currently worth the effort to invent non-C++ Value sources.

Hmm. Although perhaps it might be cute to write a C++ wrapper, say a "PythonListValue", that holds on to the name of a python function that returns a list, and, whenever you ask for that value, the python function gets called, and the resulting list gets returned.   Right now, this would be a kind-of stupid-computer-trick kind of thing to do, as no one needs this, and its not obviously useful for anything (yet).  But it does kind of give a hint of how varied and different and wide-spread values could be.

linas:
I hope I did not un-nerve you with all the comments above, and, if you feel like I gave you an impossible task or an impossible API, let me know.

vsbogd:
No problem. I actually didn't notice that RandomStream is fast and dirty experiment so design API correctly looks like the right way of doing things. I agree with Python Values idea but I think use cases should be more clear to design and implement it accurately. I will move it further if I have clear usecases.

PR #1818, opened 2018-07-04
Allow matching of quoted arithmetic link with unquoted args

vsbogd:
To match for example Times link without executing it one should write something like (GetLink (Variable "$A") (Quoted (Times (Unquoted (Variable "$A")) (Number 3)))). It doesn't work because arithmetic link arguments are checked to be numeric. Fixing it using check_evaluatable() as example.

### DIFF
--- a/opencog/atoms/core/Checkers.cc
+++ b/opencog/atoms/core/Checkers.cc
@@ -90,6 +90,9 @@ bool check_numeric(const Handle& bool_atom)
 		// Oddly enough, sets of numbers are allowed.
 		if (SET_LINK == t and check_numeric(h)) continue;
 
+		if (QUOTE_LINK == t) continue;
+		if (UNQUOTE_LINK == t) continue;
+
 		if (not h->is_type(NUMERIC_OUTPUT_LINK)) return false;
 	}
 	return true;
--- a/tests/query/QuoteUTest.cxxtest
+++ b/tests/query/QuoteUTest.cxxtest
@@ -203,7 +203,7 @@ void QuoteUTest::test_gpn_bindy(void)
 }
 
 /*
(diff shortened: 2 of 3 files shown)


linas:
Grrr.  I'm merging this, but I cannot begin to describe how much I hate this.

The entire concept of QuoteLink and UnquoteLink is deeply and fundamentally broken, and it is extremely vital and urgent to find some other alternative, better solution that will allow QuoteLink and UnquoteLink to be removed from Atomese!

As far as I can tell, neither ProLog, nor SQL has the concept of quote and unquote, and, insofar as atomese is a lot like SQL or ProLog, my intuition says that it should not have, and not need the concept of quotation, either.   Quotation was invented as a temporary, hacky solution to an icky problem, and now, like a virus, a cancer, kudzu, it is growing and infesting every corner of the C++ code. 

Fundamentally: any concept that causes cancerous infestations is a symptom that something is wrong.  This pull request is an example of that infestation.  I'll merge this, for now, but we have to figure out how to fight the cancer itself.

vsbogd:
As I see this problem: if we have some atomspace state and this state contains executable or evaluatable atoms (like TimesLink) and we would like to have ability doing both (1) matching such atoms themselves or (2) matching result of the execution or evaluation of the atoms  then we need somehow denote it. 

At the moment default way of matching is (2) and so one need some way to point that way (1) should be used. It can be done vice versa probably: have (1) as default and special link type for (2). But if we need both ways of matching I don't see another way as somehow denote it.

ngeiswei:
My understanding agrees with @vsbogd's . I don't especially like `QuoteLink` and such (well, implementing it was a pain and maintaining it still somewhat is) but I don't see another way that would be dramatically better, do you (@linas)?

PR #1819, opened 2018-07-05
#1161 Add LinkValue Python API

vsbogd:
Issue https://github.com/opencog/atomspace/issues/1161. LinkValue is implemented for case when it holds other Values not Atoms. Atoms holding requires (1) making Python Atom child of Python ProtoAtom and (2) Atom constructor to create Atom from cHandle only without AtomSpace.


opencog/cython/opencog/atomspace.pxd
```diff
@@ -85,6 +85,10 @@ cdef extern from "opencog/atoms/proto/atom_types.h" namespace "opencog":
 cdef extern from "opencog/atoms/proto/ProtoAtom.h" namespace "opencog":
     cdef cppclass cProtoAtom "opencog::ProtoAtom":
         Type get_type()
+        int is_atom()
```
noskill: May it be bool or bint?

vsbogd: Thanks, fixed.

linas:
Looks good. You understand I am not reviewing this as closely as Anatoly just did; (thanks Anatoly!) rather, that the general form looks good.

PR #1821, opened 2018-07-08
Fix Atomspace Python README.md

vsbogd:
Add required packages description, remove old commend about moses.cpp which is copy of https://github.com/opencog/moses/tree/master/tests/cython.

### DIFF
--- a/tests/cython/README.md
+++ b/tests/cython/README.md
@@ -1,8 +1,20 @@
 
-HOWTO run these tests by hand:
-------------------------------
+## Preconditions
 
-You need to set up the PYTHON path:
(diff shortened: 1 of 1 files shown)


tests/cython/README.md
```diff
@@ -1,27 +1,33 @@
 
-HOWTO run these tests by hand:
-------------------------------
+## Preconditions
 
-You need to set up the PYTHON path:
+```nose``` Python testing framework is required.
```
linas: fwiw -- I've been installing with `apt-get install nose` -- I also have both nose and nose3 isntalled; I'm not sure that is even possible with pip and conda.

Issue #1825, opened 2018-07-09
GlobNode sometimes misses trailing matches


linas:
Consider the following search:
```
(define b

(BindLink
   (VariableList
     (TypedVariableLink
        (GlobNode "$begin")
        (IntervalLink (NumberNode 0) (NumberNode -1)))
     (TypedVariableLink
        (GlobNode "$end")
        (IntervalLink (NumberNode 0) (NumberNode -1))))
   (List
       (GlobNode "$begin")
       (Concept "foo")
       (GlobNode "$end"))
   (OrderedLink
       (Concept "begin")
       (GlobNode "$begin")
       (Concept "end")
       (GlobNode "$end")))
)
```
where the atomspace contains `(List (Concept "foo")(Concept "foo"))`

Run  `(cog-execute! b)`

The expected output is:
```
(SetLink
   (OrderedLink
      (ConceptNode "begin")
      (ConceptNode "foo")
      (ConceptNode "end")
   )
   (OrderedLink
      (ConceptNode "begin")
      (ConceptNode "end")
      (ConceptNode "foo")
   )
)
```
The current actual output is
```
(SetLink
   (OrderedLink
      (ConceptNode "begin")
      (ConceptNode "foo")
      (ConceptNode "end")
   )
)
```

linas:
This is super-urgent for me, its a blocker bug, so I'll try to fix it asap.

linas:
If instead, the atomspace contains `(List (Concept "foo")(Concept "foo")(Concept "foo")(Concept "foo"))` then, again, only one solution is found, even though four solutions are expected. It appears that there is no retry for a different number of matching possibilities.

linas:
If the body is altered to 
```
(List (List
       (GlobNode "$begin")
       (Concept "foo")
       (GlobNode "$end")))
```
and the data to `(List (List (Concept "foo")(Concept "foo")))`  then everything works, because `explore_up_branches()` does contain the logic to explore all the possibilities.  The fix is to move that logic to a different place ...

ngeiswei:
In the example you use -1 to represent +infinity I suppose, if so this should be added to http://wiki.opencog.org/w/IntervalLink (I would do it but for some reason I cannot log into wiki.opencog/org, crap, I haven't been able for 2 weeks now).

linas:
Ugh, the fix is a simple one-liner fix, where the intermediate state was being incorrectly cleared.  However, that then triggers a different bug...

linas:
[...]

PR #1826, opened 2018-07-09
Minor cleanup, simplification, minor bug fix.

linas:
This simplifies the glob-matching code, and fixes a bug in it.  
The bug does not seem to affect any unit tests, so clearly 
the glob-matching unit tests are incomplete.


linas:
Minor cleanup, before attempting #1825

PR #1828, opened 2018-07-09
fix for findcython version check

Dagiopia:
the cython version check wasn't working because cmake version check wasn't working. 
compiling cythonized files would have errors with older cython.


linas:
thanks.

README.md
```diff
@@ -320,7 +320,7 @@ during the build, will be more precise as to which parts will not be built.
 * C bindings for Python.
 * Strongly recommended, as many examples and important subsystems
   assume python bindings.
-* https://cython.org | `apt-get install cython`
+* https://cython.org | `pip install cython`
```
linas: well, as a general rule, we encourage people to use `apt-get` and not `pip`, because `apt-get` gives us a stable version that is reproducible. This is useful, when dealing with bug reports. By contrast, `pip` gives some unstable version that may or may not be buggy, and is, at any rate, changing all the time.

Dagiopia: noted! didn't see it that way.
I was only thinking about update frequency. Specially for ubuntu 14 guys.

vsbogd: I agree with ```apt-get install``` default. People which use old software versions usually able managing dependencies without simple clues, using backporting for instance.

PR #1830, opened 2018-07-10
A DontExecLink fix

linas:
The simple case worked fine. The variant that was encountered inside a SetLink was not.

### DIFF
--- a/opencog/atoms/execution/Instantiator.cc
+++ b/opencog/atoms/execution/Instantiator.cc
@@ -248,17 +248,32 @@ Handle Instantiator::walk_tree(const Handle& expr, bool silent)
 		// awkward.  I'm confused about how to handle this best.
 		// The behavior tree uses this!
 		// Anyway, do_evaluate() will throw if rex is not evaluatable.
+		//
+		// The DontExecLink is a weird hack to halt evaluation.
+		// We unwrap it and throw it away when encountered.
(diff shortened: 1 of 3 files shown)


linas:
This highlights some frequent comments: two deep redesigns are needed here:

1) Stop using SetLink to wrap up intermediate results
2) Find some alternative to DontExec, Quote and Unquote.

ngeiswei:
I should note that I (think that I) rely on the fact the `DontExec` doesn't get consumed during rewriting (in particular during inference control learning since it needs to reason on URE rule without executing them). However the inference control learning examples works fine after merging that PR so, I'm a bit confused (maybe it's because it uses `Bind` instead of `Put`, I'd have to look deeper into it, but so far it works so...)

linas:
FYI, you might be interested to know this:  When bindlink matches multiple different patterns, but they produce the same output, that output is reported only once.  When Get  matches multiple different patterns, it reports each one. Then Put converts them into a common form but reports them separately (i.e. "with multiplicity")  I needed one copy for each distinct case so used Get+Put. (I almost hacked Bind to get the desired effect)   So I thought that was a pretty cool difference between the two -- so BindLink is almost equivalent to Get+Put, except for multiplicity.

But now you say it also differs in how DontExec gets handled.  :-/

PR #1831, opened 2018-07-10
Replace pip and conda instructions by apt-get

vsbogd:
To fix https://github.com/opencog/atomspace/pull/1821#discussion_r200854882

### DIFF
--- a/tests/cython/README.md
+++ b/tests/cython/README.md
@@ -3,13 +3,9 @@
 
 ```nose``` Python testing framework is required.
 
-Installation using ```pip``` package manager :
+Installation using ```apt-get``` package manager :
 ```
-pip install nose
-```
-or install it using ```conda``` package manager:
-```
-conda install nose
+apt-get install python-nose
 ```
 
 ## Running tests
--- a/README.md
+++ b/README.md
@@ -320,7 +320,7 @@ during the build, will be more precise as to which parts will not be built.
 * C bindings for Python.
 * Strongly recommended, as many examples and important subsystems
   assume python bindings.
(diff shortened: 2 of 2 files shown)


Dagiopia:
best change the [one](https://github.com/opencog/atomspace/blob/b5c4426db65b007f7343788dd39f152758b4bc2b/README.md#cython) in the main readme too on this pr

vsbogd:
Ok, added

tests/cython/README.md
```diff
@@ -3,13 +3,9 @@
...
-Installation using ```pip``` package manager :
+Installation using ```apt-get``` package manager :
 ```
-pip install nose
-```
-or install it using ```conda``` package manager:
-```
-conda install nose
+apt-get install python-nose
 ```
```
linas: Not to drive you crazy, but the above is for python 2.  There is also a python3-nose or a python-nose3 which can be installed at the same time, without conflicts ... however, I don't think our current cmakefiles are smart enough to use it.  Or something like that.  It might be why cmake is currently not launching the nose tests.

linas:
anyway thanks

tests/cython/README.md
vsbogd: Yes, you're right, the same thing with ```cython```. May be using ```apt-get``` is not a good idea because all of python packages in Ubuntu have two versions ```python``` and ```python3``` and you should always choose version which corresponds to your environments.

At the moment I use ```conda``` package management on my machine because it allows opaque switching between python2/python3 environments and have richer package base.

PR #1832, opened 2018-07-11
reorganize cmake files

noskill:
It's my attempt at issue https://github.com/opencog/cogutil/issues/113
1) Moving Find*.cmake files to cogutil for installation
2) Use Find*.cmake files from cogutil installation in atomspace


linas:
Some fixes are needed:

* Valgrind has nothing to do with cmake, and so does not belong in a dirctory called `cmake`.

* Vagrant has nothing to do with cmake, or with valgrind ... 

* The file `COPYING-CMAKE-SCRIPTS` needs to travel with the `Find.*.cmake` files because it is a copyright notice that applies to them.

* The file `OpenCogAtomTypes.cmake` should move as well; it is currently shared between four repos: the atomspace, opencog, agi-bio and as-moses.

noskill:
@linas i updated as you suggested, moving OpenCogMacros.cmake to cogutil as well

linas:
It be great if Nil and Amen took a look. Nil has been maintaining cogutils; Amen has done some serious CMake hacking.

CMakeLists.txt
```diff
@@ -55,7 +55,20 @@ ADD_DEFINITIONS(-DPROJECT_SOURCE_DIR=\\"${CMAKE_SOURCE_DIR}\\"
                 -DPROJECT_BINARY_DIR=\\"${CMAKE_BINARY_DIR}\\")
 
 # add the 'lib' dir to cmake's module search path
-SET(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/lib/")
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/lib/")
+
+# Cogutil
+FIND_PACKAGE(CogUtil)
```
vsbogd: According to [find-package documentation](https://cmake.org/cmake/help/latest/command/find_package.html#command:find_package) should be:

    FIND_PACKAGE(CogUtil CONFIG)

to skip searching FindCogUtil.cmake and look for CogUtilConfig.cmake

noskill: updated

ngeiswei:
Testing this is proving more difficult. I could persevere but it seems it has been sufficiently tested by you guy (thanks a lot for your effort!).

I'll wait 24h in case you have something else to say and I'll merge.

PR #1833, opened 2018-07-13
fix minor cython c string compile error

SomeUserName1:
according to [the cython doc](http://docs.cython.org/en/latest/src/tutorial/strings.html) one needs to either specify the encoding or use a python string.


linas:
Maybe this fixes bugs mentioned elsewhere? If so, should mention that in the commit message.

PR #1836, opened 2018-07-18
Remove redundant tv-mean, tv-conf and tv-count

soulbliss:
Changes done in the utilities.scm file only.

### DIFF
--- a/opencog/scm/opencog/base/utilities.scm
+++ b/opencog/scm/opencog/base/utilities.scm
@@ -77,7 +77,7 @@
 (define-public (etv pos-count total-count) (cog-new-etv pos-count total-count))
 
 ; Fetch the mean, confidence and count of a TV.
-(define-public (tv-mean TV)
+(define-public (cog-tv-mean TV)
 "
   Warning: this function is obsolete, use cog-tv-mean instead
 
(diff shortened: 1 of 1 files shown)


ngeiswei:
`cog-tv-mean` and such are already defined in https://github.com/opencog/atomspace/blob/master/opencog/guile/SchemeSmobTV.cc#L471 and such. So what we want is to remove `tv-mean` and such from `utilities.scm` and replace their calls (`tv-mean` -> `cog-tv-mean`, etc) all over the code.

In order to find them (`tv-mean`, `tv-conf` and `tv-count`) you may use the github search bar, at the top left of https://github.com/opencog

Thanks!

ngeiswei:
I'm closing as it's best if you start from a fresh one.

soulbliss:
A neophyte here, so please bear with me. Let me know if i got this right.

1. What I did with the `utilities.scm` has to be done with all the other redundancies that occur throughout atomspace repository.

2. Is the first point applicable only to the atomspace repo of opencog or other repos as well like opencog,  moses, link grammer.

ngeiswei:
What we want is to replace all instance of `tv-mean` listed here https://github.com/search?q=org%3Aopencog+tv-mean&type=Code by `cog-tv-mean`, so yes it touches multiple repos. And do the same for `tv-conf` and `tv-count`.

Issue #1837, opened 2018-07-18
MapLink probably wrongly execute inputs


ngeiswei:
The following
```
scheme@(guile-user)> (define ml (MapLink
  (LambdaLink
    (VariableList
      (VariableNode "$PM-9d63faa") ; [2876229365943642109][3]
      (VariableNode "$PM-70483054") ; [395373704462619396][3]
    ) ; [16529408717632465190][3]
    (TypedVariableLink
      (VariableNode "$PM-9d63faa") ; [2876229365943642109][3]
      (VariableNode "$PM-70483054") ; [395373704462619396][3]
    ) ; [15593697934161382659][3]
  ) ; [10365331990018569528][3]
  (QuoteLink
    (MemberLink
      (DefinedSchemaNode "specialization-rule") ; [8392826406491317314][3]
      (ConceptNode "pm-rbs") ; [6196355960147000852][3]
    ) ; [10981008991023846440][3]
  ) ; [11675264712320902971][3]
) ; [15987864180456164925][3]
)
... ... ... ... ... ... ... ... ... ... ... ... ... ... ... ... ... ... scheme@(guile-user)> (cog-execute! ml)
```
Produces the following error
```
ERROR: In procedure opencog-extension:
Throw to key `C++-EXCEPTION' with args `("cog-execute!" "Cannot find defined hypergraph for atom (DefinedSchemaNode \"specialization-rule\") ; [8392826406491317314][1]
 (/home/nilg/OpenCog/atomspace/opencog/atoms/core/UniqueLink.cc:125)
Function args:
((MapLink
   (LambdaLink
      (VariableList
         (VariableNode \"$PM-9d63faa\")
         (VariableNode \"$PM-70483054\")
      )
      (TypedVariableLink
         (VariableNode \"$PM-9d63faa\")
         (VariableNode \"$PM-70483054\")
      )
   )
   (QuoteLink
      (MemberLink
         (DefinedSchemaNode \"specialization-rule\")
         (ConceptNode \"pm-rbs\")
      )
   )
)
)")'.
```
Which indicates that the input
```
  (QuoteLink
    (MemberLink
      (DefinedSchemaNode "specialization-rule") ; [8392826406491317314][3]
      (ConceptNode "pm-rbs") ; [6196355960147000852][3]
    ) ; [10981008991023846440][3]
  ) ; [11675264712320902971][3]
```
gets executed in spite of being a `QuoteLink`.

In contrast the following
```
(define gl (Get     (VariableList
[...]

PR #1838, opened 2018-07-19
Fix MapLink with quoted argument

ngeiswei:
Addresses issue #1837

### DIFF
--- a/opencog/atoms/execution/MapLink.cc
+++ b/opencog/atoms/execution/MapLink.cc
@@ -322,8 +322,10 @@ bool MapLink::extract(const Handle& termpat,
 
 Handle MapLink::rewrite_one(const Handle& cterm, AtomSpace* scratch) const
 {
+	// Execute the ground, including consuming its quotation as part of
+	// the MapLink semantics
 	Instantiator inst(scratch);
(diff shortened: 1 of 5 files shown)


ngeiswei:
@linas I'm fairly confident that is the right way to fix that (consuming of quotation is part of the semantics of unification/matching and substitution). You may have a look at it and merge if you wish, or not if you're too busy, I'll take care of it.

opencog/atoms/execution/MapLink.cc
```diff
@@ -322,8 +322,10 @@ bool MapLink::extract(const Handle& termpat,
 
 Handle MapLink::rewrite_one(const Handle& cterm, AtomSpace* scratch) const
 {
+	// Execute the ground, including consuming its quotation as part of
+	// the MapLink semantics
 	Instantiator inst(scratch);
-	Handle term(HandleCast(inst.execute(cterm)));
+	Handle term(HandleCast(inst.instantiate(cterm, HandleMap())));
```
linas: The new comment seems to describe the old code...

linas:
The fix looks entirely reasonable to me.  Merge when ready.

PR #1849, opened 2018-07-30
Fix build for unusual protobuf environments

vsbogd:
Use ```protoc``` executable and Protobuf include path found by Protobuf
cmake module instead of relying that they are located in standard places.


linas:
I'm merging this, without testing. My general impression is that this subsystem needs redesign, or needs to be marked obsolete, or something.   I might be confused, but if I recall correctly, this was designed to move data to and from the Java Neo4J graph database.  However, that backend never worked well... and probably does not work at all, currently.

I'm also uncertain as to whether or not the agi-bio guys are using this interface to talk to the atomspace, so it cannot be marked obsolete without asking around.

PR #1850, opened 2018-07-31
Update GearmanUTest cmake configuration

AmeBel:
Depends on https://github.com/opencog/cogutil/pull/122


ferrouswheel:
I mention this on another commit, but it seems to have been eaten...

This test is leaving `gearmand` running after the test... tests should clean up after themselves and not leave processes running the background.

pennachin:
To be precise, since the test checks whether gearmand is runnig -- it should clean up only if it had to start it.

AmeBel:
ok will update it

AmeBel:
Done by https://github.com/opencog/atomspace/pull/1852

PR #1867, opened 2018-09-16
add cog-

soulbliss:
addressing issue #1836

tv-mean 	 --->  cog-tv-mean

tv-conf		--->   cog-tv-conf	

tv-count	--->	cog-tv-count

made these changes in the atomspace repo.
will be doing the same in other repos of opencog.

### DIFF
--- a/opencog/scm/opencog/base/utilities.scm
+++ b/opencog/scm/opencog/base/utilities.scm
@@ -77,7 +77,7 @@
 (define-public (etv pos-count total-count) (cog-new-etv pos-count total-count))
 
 ; Fetch the mean, confidence and count of a TV.
-(define-public (tv-mean TV)
+(define-public (cog-tv-mean TV)
 "
   Warning: this function is obsolete, use cog-tv-mean instead
 
@@ -87,7 +87,7 @@
 	(cog-tv-mean TV)
 )
 
-(define-public (tv-conf TV)
+(define-public (cog-tv-conf TV)
 "
   Warning: this function is obsolete, use cog-tv-confidence instead
 
@@ -108,7 +108,7 @@
 ;
 ; Simple truth values won't have a count. Its faster to just check
 ; for #f than to call (cog-ctv? tv)
-(define-public (tv-count TV)
+(define-public (cog-tv-count TV)
 "
   Warning: this function is obsolete, use cog-tv-count instead
 
@@ -1278,10 +1278,10 @@
 'itv
 'ctv
 'etv
-'tv-mean
-'tv-conf
+'cog-tv-mean
+'cog-tv-conf
 'tv-non-null-conf?
-'tv-count
+'cog-tv-count
 'gar
 'gdr
 'gadr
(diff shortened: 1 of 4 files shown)


opencog/scm/opencog/base/utilities.scm
```diff
@@ -77,7 +77,7 @@
 (define-public (etv pos-count total-count) (cog-new-etv pos-count total-count))
 
 ; Fetch the mean, confidence and count of a TV.
-(define-public (tv-mean TV)
+(define-public (cog-tv-mean TV)
 "
```
linas: ahem, cough cough cough.   Did you actually look at what you are doing? 

This defines a backwards-compatibility function. You just changed it to call itself, in an infinite-recursive way.

linas:
The changes to the rule-engine test files make sense, The other changes do no. Can you create a pull req for the rule-engine tests, only?

tests/scm/SCMExecutionOutputUTest.cxxtest
```diff
@@ -221,8 +221,8 @@ void SCMExecutionOutputUTest::test_recursive(void)
 		"(define (make-queen x)"
 		"   (InheritanceLink x (ConceptNode \"queen\")))"
 		"(define (stv mean conf) (cog-new-stv mean conf))"
-		"(define (tv-mean tv) (assoc-ref (cog-tv->alist tv) 'mean))"
-		"(define (tv-conf tv) (assoc-ref (cog-tv->alist tv) 'confidence))"
+		"(define (cog-tv-mean tv) (assoc-ref (cog-tv->alist tv) 'mean))"
+		"(define (cog-tv-conf tv) (assoc-ref (cog-tv->alist tv) 'confidence))"
```
linas: Again, this is just plain wrong.

linas:
I'm going to close this pull req, for now; please create a new one, for the rule-engine unit tests, only.

tests/scm/SCMExecutionOutputUTest.cxxtest
```diff
@@ -231,18 +231,18 @@ void SCMExecutionOutputUTest::test_recursive(void)
 	eval->eval(
 		"(define (royal-incr atom cnt)"
 		"   (cog-set-tv! atom"
-		"       (stv (+ (tv-mean (cog-tv atom)) cnt)"
-		"           (tv-conf (cog-tv atom)))))"
+		"       (stv (+ (cog-tv-mean (cog-tv atom)) cnt)"
```
soulbliss: are the changes made from here on till the last one, also incorrect?

linas: Hmm. Its ... well, its a coin-flip.  The unit test defined a custom version of `tv-mean`, which just happened to do exactly the same thing as the "official" version.  The reason for doing this is because the unit test was old, and it was trying to avoid having to load `utilities.scm`.  What should we do now? There are two choices:

1) continue to define a custom version of `tv-mean` and use that customer version in the body of the test.

2) Remove the definition of `tv-mean` and use `cog-tv-mean` in the body of the text.

Either 1) or 2) is OK, I guess. I have a slight preference for 1) because it tests a more obscure code path, which is probably not tested anywhere else. On the other hand, 2) makes the unit test easier to understand.  So its a coin-flip.

linas: BTW, you have run the unit tests, and they pass, right?

Issue #1868, opened 2018-09-21
Pass truth values from GroundedPredicateNodes to rule engine from pattern matcher.


noskill:
We need to use rule engine along with virtual evaluation links. 
Current pattern matcher implementation does not store truth value returned by grounded predicate.
Instead matched evaluation link keeps its default value of stv(1 0). And here matched means having mean tv more than hardcoded threshold of 0.5.

Example:

```scheme
(use-modules (opencog logger))

(load-from-path "conjunction-rule-base-config.scm")


(Inheritance (stv 0.99 0.99) (Concept "BB1") (Concept "BoundingBox"))
; default truth value 1.0 0.0 give 1.0 0 for any inference 
(Inheritance (stv 0.99 0.99) (Concept "BB2") (Concept "BoundingBox"))

(define Red (Predicate "Red"))


(define (redness box ) 
    (stv 0.55 (random 1.0)) 
)

(define RedBox1
    (EvaluationLink
         (GroundedPredicateNode  "scm: redness")
         (Variable "$X")
    )
)

(define RedBox2
   (EvaluationLink (stv 0.56 0.7)
         Red
         (ConceptNode "BB2")
   )
)


; need to restrict the search by inheritance link
(define AndRedBox1 (AndLink RedBox1 (InheritanceLink (VariableNode "$X") (ConceptNode "BoundingBox")) ) )
; don't need to restrict the search by inheritance link here
; do it anyway
(define AndRedBox2 (AndLink RedBox2 (InheritanceLink (ConceptNode "BB2") (ConceptNode "BoundingBox"))  ))
   

(define (run-bc and-box) 
    (define result (conj-bc and-box))
    (display "\n")
    (display result)
    (display "truth value for the first element in the set\n")
    (display (cog-tv (car (cog-outgoing-set result))))
    (display "\n")
    result
)

(ure-logger-set-level! "fine")
(cog-logger-set-level! "fine")
(ure-set-num-parameter conjunction-rule-base "URE:maximum-iterations" 20)
(display "query with virtual evaluation links\n")
(run-bc AndRedBox1)
(display "\n\nquery with real evaluation links\n")
(run-bc AndRedBox2)

```
====================================================================
Current implementation gives (stv 1.000000 0.000000) for all **and** links for the first conjunction **(run-bc AndRedBox1)**:

```scheme
query with virtual evaluation links

(SetLink
   (AndLink
      (EvaluationLink
         (GroundedPredicateNode "scm: redness")
[...]

PR #1870, opened 2018-09-26
Test if pattern matcher hold grounded predicate evaluation TV

ngeiswei:
Unit test (adapted from @noskill 's https://github.com/noskill/atomspace/commit/6292a636d56e58c9add319d41860db32f003e897) for issue #1868.

It tests if the pattern matcher properly hold the TV set by an grounded predicate call to store the result of its own call.

It turns out to pass?!!


ngeiswei:
@noskill, maube it's a detail but note I have converted your unit test from whitespace indentation to tabs.

I should mention that 2/3 of the atomspace repo code is indented with tabs, and actually I have set tabs as the preferred choice in opencog's dev standard (for C++ at least) https://wiki.opencog.org/w/Development_standards

The great holy war of tab vs whitespace IMO has been won by tabs ever since humanity realized that you can indent with tabs while aligning with spaces...

noskill:
@ngeiswei It was my mistake, sorry for that. I copy-pasted SingleUTest.cxxtest without checking..

ngeiswei:
The fact that the code doesn't uniformly use the same indentation style is annoying, but hopefully this annoyance will decrease overtime as we stick to the same style.

PR #1875, opened 2018-10-02

ngeiswei:
I'm thinking of attempting to merge singnet/atomspace into opencog/atomspace every month or two just to be sure that they do not diverge too much.

@linas you may want to have a look at it, although frankly nothing is really touching the atomspace core code. Here are the following changes 

1. Improve URE
2. Improve Haskell bindings
3. Improve calling C function from ExecutionOutput or Evaluation links

You're probably only interested in the last item, maybe.

I tested it on opencog and it doesn't seem to make any unit tests fail (`AttentionParamQueryUTest` fails but it already did before).


linas:
This particular merge request did not contain any changes worth arguing about.

However, I'm concerned that we're using a new methodology for code development, that we never talked about.  I did talk to Cassio about this; he promised me that there would NOT be any development happening in the singnet forks, and yet clearly, here we are, and there is. 

There are both good and bad things that can come out of this style of software development.   I am not reviewing the bug reports and pull reqs in singnet, I'm effectively cut out of the loop, removed from the day-to-day operations in the code base.  I really really don't like it that this was done without anyone having the grace to tell me; it is a de facto a political coup, a maneuver, and frankly it sucks. I fundamentally believe that conducting affairs in such a low-down, under-handed kind of way is bad; it's not constructive, its not positive, its not the kind of behavior patterns I want to be around.

Anyway, there's some good things that come out of the reorganization.  By no longer being involved in the day-to-day development of the code, I have more time for other things. And that's cool.  It does require me to trust that what is going on in the singnet side is good: That means I have to trust you and Vitaly, and I think I mostly do trust you two.

There is a risk, however, that there will be some future pull request that will contain 9 good new features, and one very objectionable one. It is not entirely clear how that can be resolved.  For now, we can play it by ear, and see what happens.

ngeiswei:
Yeah, I also feel ambivalent about it, but I'm taking it as an experiment. It has good aspects like we're freer to experiment with more stringent forms of continuous integration on singnet. And bad aspects like we run into the risk of building up conflicting changes (which should be mitigated if we stay in sync relatively often).

I'm not worried too much about the atomspace repo because we're making an effort (@vsbogd et al and I) at only touching portions that do not affect the atomspace core, and working on `opencog/atomspace` otherwise. I'm more worried about the opencog repo which seems to have already somewhat diverged.

linas:
OK, we'll play it by ear.

PR #1877, opened 2018-10-03
Add opencog test runner that works with ctest

AmeBel:
This makes it easy to write scheme tests without having to write c++ codes. See included unit tests for examples.

Depends on https://github.com/opencog/cogutil/pull/131

### DIFF
--- a/opencog/scm/opencog/base/utilities.scm
+++ b/opencog/scm/opencog/base/utilities.scm
@@ -62,7 +62,6 @@
 ;
 
 (use-modules (srfi srfi-1))
-(use-modules (srfi srfi-64))  ; For opencog-test-runner
 (use-modules (ice-9 threads))  ; needed for par-map par-for-each
 
 ; See below; the compiler step is not needed for guile-2.1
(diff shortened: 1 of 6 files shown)


opencog/scm/opencog/base/utilities.scm
```diff
@@ -1277,6 +1278,49 @@
...
+    (test-begin \"Name of the test suite\")
+    (test-assert ...)
+    ...
+    ...
+    (test-end \"Name of the test suite\")
+    ; Any code written after 'test-end' is called will not be run as
+    ; 'test-end calls 'exit'.
+
+  For details on the apis for writing srfi-64 test-suites see
+  https://srfi.schemers.org/srfi-64/srfi-64.html
```
noskill: It looks like the site is down

linas: It works for me, right now.

AmeBel:
Added the (opencog test-runner) module.

opencog/scm/opencog/base/utilities.scm
noskill: Works for me too

PR #1878, opened 2018-10-03
Unit test for EvaluationLink retaining default TV in pattern matcher

noskill:
Run cog-execute! with BindLink returning virtual evaluation link, check that returned evaluation link retains default truth value. See the discussion in issue #1868

### DIFF
--- a/tests/query/CMakeLists.txt
+++ b/tests/query/CMakeLists.txt
@@ -42,6 +42,7 @@ IF (HAVE_GUILE)
 	ADD_CXXTEST(ImplicationUTest)
 	ADD_CXXTEST(ExecutionOutputUTest)
 	ADD_CXXTEST(BindTVUTest)
+	ADD_CXXTEST(EvalLinkDefaultTVUTest)
 	ADD_CXXTEST(BuggyStackUTest)
 	ADD_CXXTEST(VarTypeNotUTest)
 	ADD_CXXTEST(BuggyNotUTest)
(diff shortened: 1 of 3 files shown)


linas:
It's not obvious that adding this unit test is a good idea.  Why?  The unit test suggests that there is a promise, from now until forever, that the EvaluationLink will work in this certain way.  That future users of the EvaluationLink can depend on getting exactly this behavior.  That some important application will break if this unit test fails. 

But is there some important application that will fail, if this unit test breaks? Not that I can see; nothing in #1868 depended or required this behavior.

Do we really want to commit to EvaluationLink working in exactly this way, forever more?  I made a strong argument that we should, because that is the way that the pattern matcher works.   I want the pattern matcher to use crisp-logic. I also want "cached results" to be multi-thread (multi-user)-consistent, which means the TV should not be set (because then different users would see different values in a racey way).  So, yes, right now I feel pretty strongly that the TV should not be set.

But do we actually need to nail this down, specify an exact behavior, right now? Can't we leave the behavior "undefined"? Maybe there will be a good reason to change it, in the future.  Leaving it "undefined" gives more freedom in the future.

(Side remark; hitchhikers-guide-to-the-galaxy: the larger the number of possible alternative futures in front of you, the greater your chances of survival. Having few, or having no choices is a dangerous place to be.  This is a deep observation about nature.)

noskill:
@linas We would like to use workaround suggested by Nil in #1868, likely till better solution - getting rid of truth values in pattern matcher is ready. This unit test should allow us to notice if someone is going to break current behavior, it shouldn't prevent changes, one just would modify the unit test as well.

ngeiswei:
@noskill maybe just add a comment explaining that in `EvalLinkDefaultTVUTest`, pointing to https://github.com/opencog/atomspace/issues/1868 , etc, to help our future selves ;-).

noskill:
@ngeiswei Sure!

PR #1883, opened 2018-10-16
Create a Value factory 

linas:
Having explicit factories for values allows other other modules, outside of the atomspace (such as the octomap) to define values, and have them be treated in an automatic way.

Per private conversations with Misgana, this needs about 5-6 minor fixes/tweaks before being merged.


opencog/atoms/proto/FactoryLoader.cc
```diff
@@ -0,0 +1,31 @@
...
+#include <opencog/atoms/proto/StringValue.h>
+#include <opencog/atoms/proto/ValueFactory.h>
+
+using namespace opencog;
+
+static __attribute__ ((constructor)) void init(void)
+{
+    valuefactory().register_factory(LINK_VALUE, (CreateProto) & (createLinkValue<std::vector<ProtoAtomPtr>>),
+                                    std::vector<std::type_index> {std::type_index(typeid(std::vector<ProtoAtomPtr>))});
+
```
linas: The above 4 lines should be copied into `LinkValue.cc`  .. similarly, the lines below should be copied to the corresponding `*.cc` files.

opencog/atoms/proto/ValueFactory.cc
```diff
@@ -0,0 +1,33 @@
+#include "ValueFactory.h"
+
+using namespace opencog;
+
+void ValueFactory::register_factory(Type vtype, CreateProto func, std::vector<std::type_index> args)
+{
+    FuncRegister fr = {func, args};
+
+    if(func_register.find(vtype) != func_register.end()) {
```
linas: whitespace, here and elsewhere: write `if (` instead of `if(`

opencog/atoms/proto/ValueFactory.h
```diff
@@ -0,0 +1,98 @@
...
+
+namespace opencog
+{
+using CreateProto = ProtoAtomPtr (*) (...);
+
+struct FuncRegister
+{
+    CreateProto func;
+    std::vector<std::type_index> args;
+};
```
linas: `struct FuncRegister` should be cleared frivate, inside of `class ValueFactory` below.

opencog/atoms/proto/ValueFactory.h
```diff
@@ -0,0 +1,98 @@
...
+
+class ValueFactory
+{
+private:
+
+#define THROW_ERROR \
+    (throw std::invalid_argument("There was no function found for the given type."))
+
```
linas: For historical reasons, we should continue to use the exceptions defined in cogutils.  They have a standard framework for printing, error logging, and the scheme code knows how to catch and rethrow them to the user.

opencog/atoms/proto/ValueFactory.h
```diff
@@ -0,0 +1,98 @@
...
+public:
+
+    friend ValueFactory& valuefactory();
+
+    /** Registers the creator functions.
+     *  @param vtype The value type.
+     *  @param func a pointer to the creator function of the value.
+     *  @param args  an ordered vector the types of arguments the creator takes.
+     */
+     void register_factory(Type vtype, CreateProto func, std::vector<std::type_index> args);
```
linas: Change name to be consistent with `ClassServer.h`, where this is called `addFactory()`

opencog/atoms/proto/ValueFactory.h
```diff
@@ -0,0 +1,98 @@
...
+
+     void register_castor(Type vtype, CreateProto func);
+
+     ProtoAtomPtr cast(Type vtype, ProtoAtomPtr ptr);
```
linas: The `Type vtype` argument is not needed.

linas: ... also, change the name from `cast` to something else .. maybe 'recreate()` ... the problem is that "cast" already means something else in C++.

opencog/atoms/proto/ValueFactory.cc
```diff
@@ -0,0 +1,33 @@
...
+    }
+}
+
+void ValueFactory::register_castor(Type vtype, CreateProto func)
+{
+    cast_register[vtype] = func;
+}
+
+ProtoAtomPtr ValueFactory::cast(Type vtype, ProtoAtomPtr ptr) 
+{
```
linas: The `vtype` argument is not needed. Write instead:
```
ProtoAtomPtr ValueFactory::cast(ProtoAtomPtr ptr) 
{
    Type vtype = ptr->get_type();
   ...
}
```

opencog/guile/SchemeSmobValue.cc
```diff
@@ -144,9 +145,8 @@ SCM SchemeSmob::ss_new_value (SCM stype, SCM svalue_list)
 	{
```
linas: This is currently OK as is, but in the future, you will probably need to change this to
```
  if (nameserver().isA(t, FLOAT_VALUE)
  { ...
```

misgeatgit:
@linas I have made the changes you requested.

linas:
looks good, thanks.

linas:
Next steps: add `((constructor)) init () {...}` to the truth values and attention values, and also kill the file `opencog/atoms/truthvalue/TruthFactory.cc`

PR #1885, opened 2018-10-16
Overload oc_to_string for TruthValueSeq

ngeiswei:


ngeiswei:
Just added a `TruthValueSeq` typedef for `std::vector<TruthValuePtr>` and overload `oc_to_string` for it.

Issue #1801, opened 2018-06-29
tests/cython cxxtests are broken


vsbogd:
Few issues here:
1. tests/CMakeLists.txt checks PYTHONINTERP_FOUND before including cython tests folder:
```
	IF (HAVE_CYTHON AND PYTHONINTERP_FOUND)
		ADD_SUBDIRECTORY (cython)
	ENDIF (HAVE_CYTHON AND PYTHONINTERP_FOUND)
```
But PYTHONINTERP_FOUND is unset in lib/FindCxxtest.make and lib/FindPython3Interp.cmake so cython folder is never included into list of unit tests.

In particular ```make test_python``` command fails:
```
make[3]: *** No rule to make target 'atomspace/PythonEvalUTest', needed by 'CMakeFiles/test_python'.  Stop.
CMakeFiles/Makefile2:279: recipe for target 'CMakeFiles/test_python.dir/all' failed
make[2]: *** [CMakeFiles/test_python.dir/all] Error 2
CMakeFiles/Makefile2:286: recipe for target 'CMakeFiles/test_python.dir/rule' failed
make[1]: *** [CMakeFiles/test_python.dir/rule] Error 2
Makefile:225: recipe for target 'test_python' failed
```

2. Tests in tests/cython folder fails:
```
57% tests passed, 3 tests failed out of 7

Total Test time (real) =   4.25 sec

The following tests FAILED:
	  1 - PythonEvalUTest (Child aborted)
	  4 - CythonAtomSpace (Failed)
	  7 - CythonGuile (Failed)
```

linas:
FYI, for me, running Debian stable, and having installed whatever it is I have installed, all tests pass 100% of the time, more or less always, and its been like that for maybe a year now.

vsbogd:
Yes, ```make test``` passes but C++ Python related unit tests are not started. This is a subject of the issue.

linas:
Sorry, please ignore my earlier comment. yes, `make test_python` fails for me with same error as above.  In addition, for me, the `tests/cython` directory is never even visited, the tests are never even built. 

I got very used to seeing these tests run and pass, that I never noticed when they stopped getting run.  I'm thinking that the test report should include, at the bottom, a note "the python tests were not run", and for the sql tests, "the sql tests were not run", (as those are the other optional set of unit tests)

noskill:
I made a new issue for remaining test https://github.com/opencog/atomspace/issues/1888

linas:
[...]

PR #1887, opened 2018-10-25
enable python tests in cmake

noskill:
enable python tests in cmake

fix python errors decoding for python3

Should partially fix python tests https://github.com/opencog/atomspace/issues/1801 , i have only one test failing PythonEvalUTest


noskill:
I made new issue for remaining test https://github.com/opencog/atomspace/issues/1888

CMakeLists.txt
```diff
@@ -544,10 +544,7 @@ IF (CXXTEST_FOUND)
 		)
 	ENDIF (CMAKE_BUILD_TYPE STREQUAL "Coverage")
 	ADD_CUSTOM_TARGET(test_python
-		DEPENDS PythonEvalUTest
-				PythonUtilitiesUTest
-				bindlink_cython atomspace_cython
-				scheme_wrapper logger_cython utilities_cython
+		DEPENDS tests
```
vsbogd: Why test_python depends on tests? 
Should it be just removed instead?

tests/CMakeLists.txt
```diff
@@ -39,9 +39,10 @@ IF (CXXTEST_FOUND)
 	# Reasoning depends on query, above.
 	ADD_SUBDIRECTORY (rule-engine)
 
-	IF (HAVE_CYTHON AND PYTHONINTERP_FOUND)
+	IF (HAVE_CYTHON AND HAVE_NOSETESTS)
+		MESSAGE(STATUS "CYTHON")
```
vsbogd: I would add more verbose message like "Cython found, enabling Cython unit tests"

vsbogd:
Please ignore the tests which are failed (PythonEvalUTest as I understand) https://github.com/opencog/atomspace/issues/1888

CMakeLists.txt
noskill: @inflector Do you think we can remove test_python and run python tests together with all other tests?

CMakeLists.txt
```diff
@@ -544,10 +544,7 @@ IF (CXXTEST_FOUND)
 		)
 	ENDIF (CMAKE_BUILD_TYPE STREQUAL "Coverage")
 	ADD_CUSTOM_TARGET(test_python
-		DEPENDS PythonEvalUTest
-				PythonUtilitiesUTest
-				bindlink_cython atomspace_cython
-				scheme_wrapper logger_cython utilities_cython
+		DEPENDS tests PythonEval logger_cython scheme_wrapper bindlink_cython
```
linas: This seems to be a misunderstanding. The original intent was to be able to say `make test_python` and as a result, ONLY the python-related tests would run. This can save a lot of time during development, by avoiding the overhead of re-running all of the other tests.   This change seems to completely wreck the intended behavior.

linas: Just to be clear: if one says `make test`, then all of the python tests should *also* run -- The `make test_python` is just a short-cut, to run a subset of all of the unit tests.

noskill: Thanks Linas, there were some misunderstanding on my side, now make test should successfully run python tests, i updated dependencies

tests/cython/CMakeLists.txt
```diff
@@ -26,20 +26,20 @@ LINK_LIBRARIES(
...
+#	clearbox
+#	PythonEval
+#	atomspace_cython
+#	${COGUTIL_LIBRARY}
+#)
+
+#SET_TESTS_PROPERTIES(PythonEvalUTest
+#	PROPERTIES
+#	ENVIRONMENT "PYTHONDONTWRITEBYTECODE=1"
+#	ENVIRONMENT "PYTHONPATH=${PROJECT_BINARY_DIR}/opencog/cython")
```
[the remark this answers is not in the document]
noskill: Reverted

vsbogd:
Looks good to me

CMakeLists.txt
```diff
@@ -544,14 +544,12 @@ IF (CXXTEST_FOUND)
 		)
 	ENDIF (CMAKE_BUILD_TYPE STREQUAL "Coverage")
 	ADD_CUSTOM_TARGET(test_python
-		DEPENDS PythonEvalUTest
-				PythonUtilitiesUTest
-				bindlink_cython atomspace_cython
-				scheme_wrapper logger_cython utilities_cython
+		DEPENDS PythonEval logger_cython scheme_wrapper bindlink_cython atomspace_cython
```
linas: PythonEvalUTest and PythonUtilitiesUTest are still missing.  However, PythonEval is not a unit test, so it cannot actually run. 

Did you actually try to run `make test_python` ?  when you say that, it should run all (seven) python tests.

CMakeLists.txt
```diff
@@ -544,14 +544,12 @@ IF (CXXTEST_FOUND)
...
 		WORKING_DIRECTORY tests/cython
 		COMMAND ${CMAKE_CTEST_COMMAND} $(ARGS)
 		COMMENT "Running Python and Cython tests..."
 	)
+	add_dependencies(tests PythonEval logger_cython scheme_wrapper bindlink_cython atomspace_cython)
```
linas: Why is this needed? The tests themselves should already declare the correct dependencies; you should not need to also declare additional dependencies here.

CMakeLists.txt
linas: Oh, I see ... OK, dependencies. Yes, I guess these are sqonky.  Never mind; I'll just merge this.

CMakeLists.txt
noskill: I didn't figure out how, `add_dependencies` won't work on tests, so I used this workaround.

PR #1889, opened 2018-11-01
python wrapper for backward chainer

noskill:
This PR provides wrapper for backward chainer
port of examples/rule-engine/simple/ to python for testing
addressing issue https://github.com/singnet/atomspace/issues/39


linas:
I skimmed this very quickly and it looks plausible (I have no complaints or comments). I added @ngeiswei as reviewer because he will be better at spotting any issues.

ngeiswei:
@noskill the wrapper itself looks good (except some trailing spaces in `backwardchainer.pxd` and `test_bc.py` but that's a minor detail).

However, I don't think the config files and rules should be in `tests/scm`, it seems it should contain only files pertaining to the unit tests of that folder (`BasicSCMUTest.cxxtest`, etc), (well, visibly this folder seems to contain some junk like `relex-questionXX.scm`, these file don't seem to be used anywhere at all, but that's another problem). I would instead, maybe, create an `scm` folder under `tests/cython/backwardchainer`.

ngeiswei:
Thanks.

PR #1900, opened 2018-11-07
Fix EqualLink quote consumption

ngeiswei:
Fixing a bug in quotation consumption of EqualLink and other virtual links, for instance without that fix the following
```
(EqualLink
  (QuoteLink
    (LambdaLink
      (UnquoteLink
        (Variable "$X"))
      (UnquoteLink
        (Variable "$X"))))
  (Lambda
    (Variable "$X")
    (Variable "$X")))
```
instead of
```
(EqualLink
  (Lambda
    (Variable "$X")
    (Variable "$X"))
  (Lambda
    (Variable "$X")
    (Variable "$X")))
```
thus falsely discarding the match. See `GetLinkUTest::test_quote_equal` for more info.

### DIFF
--- a/opencog/atoms/core/Variables.cc
+++ b/opencog/atoms/core/Variables.cc
@@ -109,6 +109,23 @@ FreeVariables::FreeVariables(const std::initializer_list<Handle>& variables)
 		index.insert({var, i++});
 }
 
+bool FreeVariables::is_identical(const FreeVariables& other) const
+{
+	Arity ary = varseq.size();
+	if (ary != other.varseq.size()) return false;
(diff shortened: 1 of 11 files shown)


ngeiswei:
@linas, although I'm reasonably confident the fix is correct (not complete though, but nearly) you may want to have a look at it as it's a rather subtle issue.

ngeiswei:
@linas the main commit of interest is https://github.com/opencog/atomspace/pull/1900/commits/1aa31b3bf5ee815c62229d499158dec374feb745 all the rest is just cleaning/preparing stuff.

linas:
Looks good, to me!

PR #1915, opened 2018-11-22

ngeiswei:
Ports various goodies from the backward chainer to the forward chainer. More specifically

* Thompson Sampling rule selection, to account for the confidence of a rule, not just its strength.
* Complexity Penalty, to tweak degrees of breadth vs depth first search.
* Keep track of the inference evolution tree to not re-apply the same rule over the same source (optional).
* Various optimizations.

Do not merge now as it will break the pattern miner on `opencog/opencog`.

Other work include
1. @rTreutlein improving circleci config file for compiling Haskell binding.
2. Some comments in `Handle.h` regarding content vs pointer comparison + an extra content_eq overload.
3. Move `cog-new-flattened-link` from `utilities.scm` to `rule-engine-utilities.scm`.


ngeiswei:
@linas, sorry to throw 49 commits at you, but truly there are only 2 commits you might (and frankly it's big might) care about

https://github.com/opencog/atomspace/pull/1915/commits/70c822b5c07b96f6a4461461e8407c79cd8ae4c7
https://github.com/opencog/atomspace/pull/1915/commits/14a203dc56aa3459adac26e32d07c2526d0a6501

For the rest it's just forward chainer improvements (+ a bit of circleci config but I don't think you care at all).

Again these should be not be merged right now because it would break the pattern miner on the opencog repo.

ngeiswei:
The associated work on the pattern miner on the opencog has been merged, see https://github.com/opencog/opencog/pull/3346 thus this branch is ready to be merged.

ngeiswei:
@linas, I'm going ahead and merging it as I do not anticipate any resistance on your part.

PR #1926, opened 2018-11-30

ngeiswei:
Regular singnet to opencog merging...


ngeiswei:
@vsbogd @Necr0x0Der there's something slightly suspicious about https://github.com/opencog/atomspace/pull/1926/commits/10853d7e22afd8fd9dd6882be672802a1a95f993 and https://github.com/opencog/atomspace/pull/1926/commits/4b5373a011e5ed6496dd61959500d4c0e56c6e61 they seems almost identical.

Besides, I would recommend that this sort of work be done on `opencog/atomspace` rather than `singnet/atomspace` as it's somewhat "core" (well, as far as python bindings is concerned at least).

ngeiswei:
@linas, you should have a look, I'm largely unfamiliar with the Python bindinds, so I wouldn't dare to commit this on my own.

opencog/cython/PythonEval.cc
```diff
@@ -4,6 +4,7 @@
  *         Ramin Barati <rekino@gmail.com>
  *         Keyvan Mir Mohammad Sadeghi <keyvan@opencog.org>
  *         Curtis Faith <curtis.m.faith@gmail.com>
+ *         Alexey Potapov <alexey@singularitynet.io>
```
linas: Umm, why isn't Alexey creating this pull req? If this code really is from him, he should be the one doing the legwork on it?

linas:
I merged, despite several problems, that really should be fixed.  First the good news: the changes to the python bindings look to be quite nice and professionally done (although there are a few minor whitespace issues). So overall, I like them.

1) The merge history and sequence of commits is completely confused and confusing. As Nil points out, it seems that 10853d7 and 4b5373a  are almost identical, but I think this is due to all of the interwoven git merges.  It is really kind-of important to avoid tangled and confusing commits and tangled and confusing pull reqs. Again -- this makes debugging hard, this makes it hard for developers to understand what is going on.  In the long run, a development process that causes confusing and tangled pull requests is not sustainable.  We have to stop using a tangled and confusing process. Please.

In particular, the first pull req should go to the opencog repo, and not to the singnet repo. Only after it has been proven to compile, be stable, not wreck unit tests, and not disturb the robot, ghost, etc. and other dependent projects and systems, only then should it be merged into singnet.   The singnet repos should be stable. Development should happen in opencog.

2) Who is @Necr0x0Der ? Can i assume this is Alexey Potapov? If so, can I get Alexy to update his github profile to make this clear?  If it's not Alexey, then who is it?

3) The examples are missing all explanation and comments. They need comments and explanation.

4) The examples read more like unit tests. Perhaps they should be converted to unit tests.

I

Necr0x0Der:
> In particular, the first pull req should go to the opencog repo, and not to the singnet repo. Only after it has been proven to compile, be stable, not wreck unit tests, and not disturb the robot, ghost, etc. and other dependent projects and systems, only then should it be merged into singnet. The singnet repos should be stable. Development should happen in opencog.

Well, singnet repo has CircleCI for atomspace and opencog, and doesn't allow merging if any unit test fails. Maybe, there are other dependent projects that can be disturbed, but we don't change the core functionality, but extend it while commiting to singnet, so this should be relatively safe. We were asked to commit to singnet more, so that's why I prefer to commit to singnet. But maybe this policy is already outdated or doens't apply to commits touching the core functionality. I should consult with @pennachin about this...

> 
> If so, can I get Alexy to update his github profile to make this clear? 

Done

> 2. The examples are missing all explanation and comments. They need comments and explanation.
> 3. The examples read more like unit tests. Perhaps they should be converted to unit tests.
Well, they were initially designed as unit tests. But python unit tests are currently broken in atomspace (there is an issue about this afaik). Also, it has sense as an example of possible calls to methods (the example is quite transparent, but for a stand-alone example some general description might be desirable). So, do you think it's better as a (not runnable now?) unit test, or as an example?

linas:
>  But python unit tests are currently broken in atomspace

? They've always worked, so far as I know. They've been working for the last year, at least:
```
        Start 129: PythonEvalUTest
129/136 Test #129: PythonEvalUTest ..................   Passed    1.11 sec
        Start 130: PythonUtilitiesUTest
130/136 Test #130: PythonUtilitiesUTest .............   Passed    0.13 sec
        Start 131: Cython
131/136 Test #131: Cython ...........................   Passed    3.09 sec
        Start 132: CythonAtomSpace
132/136 Test #132: CythonAtomSpace ..................   Passed    2.05 sec
        Start 133: CythonUtilities
133/136 Test #133: CythonUtilities ..................   Passed    0.96 sec
        Start 134: CythonBindlink
134/136 Test #134: CythonBindlink ...................   Passed    0.96 sec
        Start 135: CythonBackwardchainer
135/136 Test #135: CythonBackwardchainer ............   Passed    1.63 sec
        Start 136: CythonGuile
136/136 Test #136: CythonGuile ......................   Passed    1.53 sec

100% tests passed, 0 tests failed out of 136
```

noskill:
@linas @ngeiswei  What do you think about turning on CircleCI for https://github.com/opencog/atomspace?

linas:
Circleci should be turned on (it should never have been turned off). It should be configured so that it does not block merges.

opencog/cython/PythonEval.cc
ngeiswei: Because I have an incentive to keep `opencog/atomspace` and `opencog/singnet` in sync.

However, that sort of work should take place on `opencog/atomspace`.

ngeiswei:
I'm for turning on CircleCI on `opencog/atomspace` and other repos on the opencog account, and keeping blocking merge off (though as I recently understood, it's not too hard to temporarily disable it when it becomes annoying).

I actually tried to enable that but it is seems I don't have the credentials (there's some password required in the github settings that I don't have). I think @AmeBel can do it, but he seems unresponsive these days.

vsbogd:
Python tests will not be run in CI see https://github.com/singnet/atomspace/issues/48

PR #1929, opened 2018-12-03
print in cmake if nosetest not found

noskill:
Reporting to user if nosetest not found


noskill:
python tests are not running - https://github.com/singnet/atomspace/issues/48, I think due to missing nosetest

linas:
thanks

Issue #1880, opened 2018-10-08
Rename FloatValue to FloatSeq?


ngeiswei:
ProtoAtoms such as `FloatValue`, `LinkValue`, etc actually are sequences. For instance `FloatValue` is
```
std::vector<double>
```
Why not rename them `FloatSeq`, etc?

It would be consistent with their meanings, and leave some rooms for other names the day someone wants to introduce single values as opposed to sequences.

Also, maybe `LinkValue` should be renamed `AtomSeq`.

vsbogd:
As far as I understand it is named so because it is a parent class of the TruthValue. And TruthValue in turn is 2 floats actually. I agree that name is confusing, but I think we should better change implementation instead of renaming. 

Thus FloatValue should contain single float. TruthValue should contain two float fields instead of vector. And they should not inherit each other. And so on. I am not sure what is impact on compatibility though.

ngeiswei:
@vsbogd SimpleTruthValue has 2 values but other truth value types such as CountTruthValue have different numbers of values, this is thus why TruthValue inherits from `FloatValue` (note saying it is the best design, I simply don't know, but just to explain why it inherits from `FloatValue` as opposed to say `FloatPair`).

vsbogd:
@ngeiswei, yes, sure, there are other values in this hierarchy which inherit FloatValue and for some of them using vector as internal representation makes sense. 

What I am trying to say in other words:
- my opinion TruthValue is not FloatSequence, so simple renaming will introduce another sort of confusion
- we still need FloatValue to represent single float instead of vector

ngeiswei:
@vsbogd agreed, for instance generalized distributional truth values are not FloatSeq (more like float counter, or something).

Also agreed about having FloatValue eventually representing single floating number rather than vector, just need to define it strictly after this first round of renaming to not create lots of confusion.

rTreutlein:
Is anybody working on this? If not I might do it as I could use a FloatValue that's just a single float.

ngeiswei:
Let's wait for @linas' feedback.

And let me stress again that turning FloatValue into a single value will have to be in a separate PR, doing otherwise would be way to0O DANGEROUS!

Also beware that it might affect the opencog repo...

linas:
Hi, Sorry for late reply. 

1) Regarding the renaming: Now that you know its actually a sequence, what's the problem? Are you afraid that you'll wake up tomorrow morning, having forgotten that its a sequence?

[...]

PR #1930, opened 2018-12-03
Rename protoatom

kasimebrahim:
Renaming the value directory and ProtoAtom.h as per this issue [1880](https://github.com/opencog/atomspace/issues/1880#issuecomment-430348458). This will require fixing in all repos that use protoatom [opencog, agi-bio, language-learning, TinyCog and benchmark]. I have already fixed opencog if this PR is to be accepted I will fix the rest.


linas:
This is obviously a big change. I'm OK with it; however, it cannot be merged until all dependent repos are fixed at the same time.  So please o make changes to agi-bio, language-learning, TinyCog and benchmark.

@ngeiswei @vsbogd now is the time to express any reservations or doubts about the wisdom of doing this.

(A related change: the `/opencog/truthvalue` dir needs to be moved to `/opencog/atoms/truthvalue`, and that is also an invasive change.)

ngeiswei:
I've carefully reviewed it. It looks good to me.

ngeiswei:
@linas said
> (A related change: the `/opencog/truthvalue` dir needs to be moved to `/opencog/atoms/truthvalue`, and that is also an invasive change.)

I suppose you mean: `/opencog/truthvalue` dir needs to be moved to `/opencog/value/truthvalue`, right?

Which also makes me wonder do we want the directory to be called `value` or `values`?

ngeiswei:
Which kinda makes wonder, maybe we want the directory tree
```
opencog/values
opencog/values/atoms
opencog/values/truthvalues
```

(with or without `s` at the end of `value` and `truthvalue`, I'm completely neutral on that).

linas:
The directory structure does not have to match the class hierarchy. I kind of prefer flatter directory structures, it makes it easier to find things, to navigate around. So just having `/opencog/atoms/value` and `opencog/atoms/truthvalue` seems fine to me.

ngeiswei:
OK, I can see some sense in that, given that `atoms` is the main class of interest.

vsbogd:
I don't see any technical issues with the change. This change highlighted some basic questions to me. I have found that I don't have clear understanding what is the difference between value and atom and why does Atom inherit Value. In other words what was the role of the ProtoAtom?

linas:
Good question. 
-- Atoms can be searched by name, contents. Which means they are held in index in atomspace. Which means they are slow to insert.  (Values are not and of these, they are fast to create)
-- Atoms are unique - only one of a kind, in the atomspace (values are not)
-- Atoms cannot be "changed"  after creation (immutable) values can be (and quickly)
-- Atoms form a graph, can be traced upwards and downwards. (The incoming set is stored, and can be traversed. Keeping track of the incoming set adds some performance overhead) Values can only be trace downwards. (no incoming set)
-- Each atom plays host to a full KVP database (key-value-pair dictionary) values are just .. values
-- Atoms can be used as keys or values, so can be stored in the KVP (thus atoms derive from values)
-- Values are meant to be fast, fluid, changable; atoms hold permanent, fixed graphical structure.

linas:
Performance: the above means e.g. truth values can be changes maybe 50x faster than a single atom can be inserted into the atomspace.  Maybe 100x difference if you also consider insert-remove performance.  As per current benchmark.

linas:
Also -- Atoms use up about 1KByte each, when counting the RAM in the atomspace needed to store an atom, plus RAM for incoming set, misc flags, mutex lock, etc. So atoms are "fat".  The history of protoatom was that it would be "just like an atom, but thin and fast", so it discarded as much as possible, while still being "usable, useful and interesting".

ngeiswei:
It looks like the only affected repos are `opencog` and `as-moses`. There is a PR for `opencog`. And `as-moses` will be fixed shortly (which is not ultra urgent since nobody relies on it yet). Thus, I'm merging.

vsbogd:
@linas, @misgeatgit, I am afraid I cannot help as I am 100% involved into the singnet platform work at the moment.

linas:
@kasimebrahim after this merge, we now have the somewhat strange/confusing situation that atom definitions are in the `value` directory. Could I ask you to move `atom_types.script`, `atom_types.cc`, `atom_types_init.cc`, `types.h` , `NameServer.*` and `README-Adding-New-Atom-Types.md` to a new directory? Say, `opencog/atoms/atom_types/`  ?  The library should be `libatom_types`.

This should be easy to do, there should be very few dependencies ... or that is my guess...

kasimebrahim:
@linas sure I can do that.

linas:
@kasimebrahim Thank you! And now, one last massive-change-everything move: could you move the directory `opencog/truthvalue` to `opencog/atoms/truthvalue` ? This is yet another change-all-header-files type change.  It should be smaller than the last one.

kasimebrahim:
Yeah sure I am on it.

PR #1931, opened 2018-12-04
Minor fix in copy on write example

ngeiswei:
Fix
+ typo in README.md
+ indentation in copy-on-write.scm (no idea why there is such indentation, was it intended?)


linas:
Indentation was due to a propagated cut-n-paste from some email.

PR #1932, opened 2018-12-05
Attempted fix for non-cython builds

linas:
Cleanup/partial fix for opencog/opencog#3298


linas:
Passes all unit tests, with and without cython installed.

PR #1937, opened 2018-12-06
Add OctoValue type support

misgeatgit:


opencog/guile/SchemeSmobValue.cc
```diff
@@ -141,7 +141,16 @@ SCM SchemeSmob::ss_new_value (SCM stype, SCM svalue_list)
...
+    if (OCTO_VALUE == t) {
+        SCM sl = svalue_list;
+        SCM satom = SCM_CAR(sl);
+        SCM svalue = SCM_CDR(sl);
+        Handle hlist = verify_handle(satom, "cog-new-value",2);
+        HandleSeq hseq = hlist->getOutgoingSet();
+        std::vector<double> valist = verify_float_list(svalue, "cog-new-value",2);
+        pa = valueserver().create(t,hseq,valist);
+    }
+    else if (FLOAT_VALUE == t)
```
linas: bad whitespace indentation.

opencog/atoms/value/ValueFactory.h
```diff
@@ -103,6 +103,46 @@ class ValueServer
...
+        {
+            try
+            {
+                // First, find the list of factories for this type.
+                std::vector<ProtoFactory> func_vec = _factories.at(vtype);
+
+                // Second, find the matching arglist.
+                for (const ProtoFactory& fr : func_vec)
+                {
+                    // At this time, only one arg is supported. FIXME
```
linas: This comment is clearly incorrect!

opencog/guile/SchemeSmobValue.cc
```diff
@@ -326,7 +335,7 @@ SCM SchemeSmob::ss_value_to_list (SCM svalue)
 	ValuePtr pa(verify_protom(svalue, "cog-value->list"));
 	Type t = pa->get_type();
 
-	if (nameserver().isA(t, FLOAT_VALUE))
+	if (nameserver().isA(t, FLOAT_VALUE) or nameserver().isA(t, OCTO_VALUE))
 	{
 		const std::vector<double>& v = FloatValueCast(pa)->value();
```
linas: If this works, then it only works by accident... because, currently, `OCTO_VALUE` does not actually inherit from `FLOAT_VALUE`.  See line 49 of `opencog/atoms/value/atom_types.script` It says:
```
OCTO_VALUE <- VALUE
```
You can change this if you want ... but the OctoValue C++ class should inherit the same way that the `atom_types.script` file says it does; otherwise, there's a problem.

misgeatgit: The C++ class inherits from FloatValue. I will modify the atom_types.script to reflect that.

Issue #1916, opened 2018-11-23
Support of gcc-4.9


kasimebrahim:
I am having error building with g++-4.9. I might be able to solve this problem by editing some code but I was thinking of may be we should drop 4.9 support.
```
proto/ValueFactory.cc:25:43: error: cannot pass objects of non-trivially-copyable type ‘opencog::ProtoAtomPtr {aka class std::shared_ptr<opencog::ProtoAtom>}’ through ‘...’
         return (*cast_register[vtype])(ptr);
```

ngeiswei:
I'm personally OK to drop gcc-4.9 support. Has anybody a problem with that?

linas:
I guess we can drop gcc-4.9 support, but that specific bug showed up due to sloppy coding. I think I fixed it in #1918

linas:
I mean to say, I think it will now compile with gcc-4.9, after #1918

linas:
After the fix #1918, was the bug fixed?

kasimebrahim:
No @linas it is still there 
<pre>
ValueFactory.h:101:31: error: cannot pass objects of non-trivially-copyable type ‘class std::vector<double>’ through ‘...’
             return (*fptr)(arg);

</pre>

linas:
Hmm I don't ming dropping support for 4.9, but I'm just wondering why you are even testing it. Is this needed for some icog labs project?

Maybe the following will work, to fix this?
```
--- a/opencog/atoms/proto/ValueFactory.h
+++ b/opencog/atoms/proto/ValueFactory.h
@@ -67,7 +67,7 @@ public:
       * @throws invalid_argument exception.
       */
     template <typename TYP, typename ARG>
-    ProtoAtomPtr create(TYP vtype, ARG arg) const
+    ProtoAtomPtr create(TYP vtype, const ARG& arg) const
     {
         // Look up the factory only once; cache the result.
         static ValueFactory fptr = nullptr;
```

Dagiopia:
I think the problem have to do with sending a vector object to the variadic ValueFactory function. It expects an array (c data) rather than objects. This should be an issue for value types float, string and link (excluding randomstream)

Dagiopia:
abandoning the 4.9 seems like a better solution than rewriting the value factory though

kasimebrahim:
@linas 

> but I'm just wondering why you are even testing it.

I thought I had to.
@Dagiopia 

> I think the problem have to do with sending a vector object to the variadic ValueFactory function

[...]

PR #1941, opened 2018-12-10
Change minimum version to 5.4.0

yitbrek:
as per #1916

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,7 +79,7 @@ include(${COGUTIL_DATA_DIR}/cmake/Summary.cmake)
 # Detect different compilers and OS'es, tweak flags as necessary.
 
 IF (CMAKE_COMPILER_IS_GNUCXX)
-	# gcc before version 4.8 does not support the c++11 definition
+	# gcc before version 5.4  does not support the c++11 definition
 	# of thread_local.
 	IF (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 5.4.0)
 		MESSAGE(FATAL_ERROR "GCC version must be at least 5.4.0!")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -81,9 +81,9 @@ include(${COGUTIL_DATA_DIR}/cmake/Summary.cmake)
 IF (CMAKE_COMPILER_IS_GNUCXX)
 	# gcc before version 4.8 does not support the c++11 definition
 	# of thread_local.
```
ngeiswei: Changing that message would be welcome as well. Beside the desire to support Ubuntu xenial https://github.com/opencog/atomspace/issues/1916#issuecomment-444601566, I think @Dagiopia pinpointed the problem https://github.com/opencog/atomspace/issues/1916#issuecomment-443631710

Anybody knows what would an adequate message to justify 5.4 (I haven't followed this very closely)?

yitbrek: As discused earlier the gcc with version 4.9 supports only primitive data types while we should use  objects. So, that is better to upgrade the library instead of configuring functions which accept objects. The best gcc version is 5.4 because it is used on ubuntu xenial(16.4) and by most containers.

Dagiopia: the change in g++ that resulted in our upgrade is this "G++ now allows passing a non-trivially-copyable class via C varargs, which is conditionally-supported with implementation-defined semantics in the standard. This uses the same calling convention as a normal value parameter." from https://www.gnu.org/software/gcc/gcc-5/changes.html

@ngeiswei maybe we should have a quote from the above next to "GCC version must be at least 5.4.0!"  to make it clear to the user.

Dagiopia: Actually 5.0

ngeiswei: Thanks @Dagiopia , so yes I suppose we should set the minimum version to 5.0 with a comment explaining about gcc support for passing non-trivially copyable class via C varargs.

PR #1942, opened 2018-12-10
Initial Version of Distributional Values

rTreutlein:
So after a bunch of experimenting, I have a basic version that works and can be merged. The plan is to extend this using an object-api similar to what matrix is using so that the dv's will be representable any kind of way.

This branch also contains the addition of an < operator to Values as this was needed during some of my experiments. The final version no longer requires it but I think that it's still a useful addition as it allows us to use Values in Maps/Counters or other ordered Containers. The ordering uses first the Type and if that's the same it's based on the content of the Value. This seems more appropriate for Values than using a hash-based comparison as with Atoms.
I was thinking about cherry-picking those changes into their own branch but then it could lead to duplicated commits or would require some rebasing/rewriting of history so I decided against it for now.

depends on https://github.com/opencog/cogutil/pull/143


opencog/atoms/base/Node.cc
```diff
@@ -102,6 +102,14 @@ bool Node::operator<(const Atom& other) const
         return get_type() < other.get_type();
 }
 
+bool Node::operator<(const Value& other) const
+{
+	if (get_type() != other.get_type())
+		return get_type() < other.get_type();
+ 	const Node* nov = (const Node*) &other;
+ 	return get_name() < nov->get_name();
```
ngeiswei: This would crash is `other` is not a node, right?

This would really need some unit tests, especially since it is apparently used nowhere ATM, so there is no way to evaluate it, even indirectly.

opencog/atoms/base/Link.cc
```diff
@@ -159,6 +159,16 @@ bool Link::operator<(const Atom& other) const
...
 }
 
+bool Link::operator<(const Value& other) const
+{
+	if (get_type() != other.get_type())
+		return get_type() < other.get_type();
+
+	const Link* lov = (const Link*) &other;
+
+	return this < lov;
```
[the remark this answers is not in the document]
rTreutlein: It was my intention that this would call operator<(const Atom& other) but now that you bring it up there should probably be a test for it or something since it could validly call operator<(const Value& other) which would be an infinite loop.

The operator<(const Atom& other) uses a hashbased comparison so I thought that comment makes sense but I guess I can just say that it uses operator<(const Atom& other).

opencog/atoms/base/Node.cc
rTreutlein: Why would it crash if other is not a node? That would mean that the types are different which goes to this linke:  
`return get_type() < other.get_type();` 
Which should work whatever the type of other is.
But it's still a good idea to add some tests for this.

opencog/atoms/base/Link.cc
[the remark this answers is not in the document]
ngeiswei: I understand but `Link::operator<(const Value&)` as implemented here does not call `Link::operator<(const Atom&)` instead it just compares their addresses.

opencog/atoms/base/Node.cc
ngeiswei: You're right, forget about my dumb comment.

opencog/atoms/base/Link.cc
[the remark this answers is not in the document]
linas: Yeah, I really don't want this. We used to do address compares exactly like this, some umpteen decades ago, and it was found wanting, troublesome, prone to bugs and mis-use. (I spent nearly a year trying to brow-beat Nil int thinking that address-compare was a good idea, until I threw in the towel.)  The atom-hash was invented in order to solve this problem. Please use that.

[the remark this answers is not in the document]
linas: oh, hang on, my last comment isn't right. Deleting it.

opencog/atoms/base/Node.cc
linas: instead of doing `get_name() < nov->get_name();` you should be comparing hashes. This should be identical to `Node::operator<(const Atom& other)`

opencog/atoms/value/LinkValue.cc
```diff
@@ -40,6 +40,23 @@ bool LinkValue::operator==(const Value& other) const
...
 }
 
+bool LinkValue::operator<(const Value& other) const
+{
+	if (LINK_VALUE != other.get_type())
+		return LINK_VALUE < other.get_type();
+
+	const LinkValue* lov = (const LinkValue*) &other;
+
+	return _value < lov->_value;
```
linas: huh??? is there some existing `operator<(std::vector<Value>&)` ??? How can this possibly work?

opencog/atoms/value/StreamValue.cc
```diff
@@ -32,4 +32,13 @@ bool StreamValue::operator==(const Value& other) const
 	return &other == this;
 }
 
+bool StreamValue::operator<(const Value& other) const
+{
+	if (STREAM_VALUE != other.get_type())
+		return STREAM_VALUE < other.get_type();
+
```
linas: If instead, you just add `if (this->get_type() != other.get_type())` to `Value.h` and `Value.cc` then you won't have to nickel-n-dime to death all the derived classes.

opencog/guile/SchemeSmob.cc
```diff
@@ -231,7 +231,7 @@ void SchemeSmob::module_init(void*)
 
 	// Set the library load path, so that other modules can find
 	// thier libraries. Copied from `scm/opencog.scm` and should stay
-	// in sync with that file.  This is NOT needed for ordinary usage
+	// in sync with that file.	This is NOT needed for ordinary usage
 	// from the guile REPL, but is needed by the unit tests.  The problem
```
linas: ???

opencog/guile/SchemeSmobDV.cc
```diff
@@ -0,0 +1,372 @@
+/*
+ * SchemeSmobDV.cc
+ *
```
linas: Although putting everything n the guile directory is in keeping with the historical style, its ... well, the goal was to not do this any more, and so `SchemePrimitive.h` and `SchemeModule.h` were invented to allow new guile modules to be easily created, and new API's to be defined. I would kind-of be a lot more comfortable if DV's were done that way, instead of adding them here.  (and maybe/probably TV and AV should be refactored to use `SchemePrimitive` etc. as well, but that's rather low priority).

See, for example: `persist/guile/PersistSCM.cc` which defines `fetch-atom` and etc. up at the top.  other, maybe more opaque examples:  `opencog/query/PatternSCM.cc` which defines `cog-bind-first-n` and `cog-satisfy` etc. also, `persist/gearman/DistSCM.cc`  In principle, files like `guile/RandGenSCM.cc` and `guile/LoggerSCM.cc` should live in thier own distinct directory, with their own CMakefile, but historical accident means they don't.

opencog/guile/SchemeSmobNew.cc
```diff
@@ -148,6 +148,95 @@ SCM SchemeSmob::handle_to_scm (const Handle& h)
...
+            res = scm_list_1(sh);
+        }
+        else {
+            res = scm_cons(sh,res);
+        }
+    }
+    return res;
+}
+
+SCM SchemeSmob::protomseq_to_scm (const ProtomSeq& ps)
```
linas: why is this, and all the above needed?  Its adding a bunch of code that .. It seems to be useless, un-needed code. 

Maintaining code that no one uses, no one needs creates a burden for future maintainers: raises the bar for learning what the hell it does,  and general mental overload for remembering what is where, why, how.  Increased mental burden non developers means increased bugs, glitches, security crashes, etc.

linas:
It would be best to have the `operator<(const Value&) const` code as a distinct pull request! I think its uncontroversial, other than formating, and teh fact that it needs to work just like `operator<(const Atom&) const` whenever Value is actually an Atom.

opencog/guile/SchemeSmobDV.cc
ngeiswei: I think having DV scheme bindings in it's own module makes sense. In fact, I think the whole code should be in its own library, especially since it's rather big and experimental at this point.

Could be under `opencog/distributional_value`, `opencog/atoms/distributional_value` or such, what you guys think?

rTreutlein: Yea I agree that moving it into its own directory makes sense. If we do choose `opencog/atoms/distributional_value` i think we should also move truth values to `opencog/atoms/truthvalue` which i think is the correct thing to do.

ngeiswei:
@rTreutlein the way I would deal with that is just split the 2 branches by cherry picking the commits associated to `Value::operator<` and apply them to another branch, and cherry picking the remaining commits and apply them to another branch.

Or I would soft-reset the whole thing and new commits from scratch, since you're gonna have to move your DV code to its own library that might actually be easier/cleaner.

Cherry picking or crafting commits can be easier with a front-end. I'm using magit for Emacs. If I want to cherry pick commits I select them and press `A A`. If I want to stage some code I just select it and press `s`, etc. There's hopefully something similar for neovim.

linas:
I just asked @kasimebrahim to move truthvalue to `atoms/truthvalue`.  Since I like medium-size names, I propose:  `atoms/distribution` or `atoms/truthdist` or `atoms/distvalue`  I think I like  `atoms/distribution` the most!?

medium-sized names: long enough to be memorable; short enough to not make typing a chore.

ngeiswei:
I like `atoms/distribution` too (in spite of being perhaps confusable with distributed "something", maybe).

rTreutlein:
I would prefer `atoms/distvalue` as it makes clear that this is about values but I also don't mind `atoms/distribution` so i will use that.

opencog/atoms/value/StreamValue.cc
rTreutlein: I am probably missing something obvious but I can't see how that could work. I need the if to decide between 2 options one which is the same in every class and one which is different.

I could put the 1st option in the parent somehow but the decision if we should use that or the specialized one still needs to happen in the child classes. Because If I put the branch into Value it can't know what to do in case the types are equal.

opencog/atoms/value/LinkValue.cc
rTreutlein: Well std::vector has templates for the operator< so as long as you can compare the elements you can compare vectors. https://en.cppreference.com/w/cpp/container/vector/operator_cmp

rTreutlein:
I am going to close this and open 2 new PRs for the respective changes.

opencog/atoms/value/StreamValue.cc
linas: It will always be true, all of the time, that `STREAM_VALUE == this->get_type()` whenever `this` is a `StreamValue` class. Likewise, it will always be true that  `FLOAT_VALUE == this->get_type()` whenever `this is a `FloatValue`.  Thus, it is enough to provide a default implementation  in the base class, instead of cut-n-pasting nearly identical code in each derived class.

opencog/atoms/value/LinkValue.cc
linas: Ah, well, now that you say this, I suppose that it's true; of course they'd provide something like this.

For readability, it would be nicer to write
```
    return _value.operator<(lov->_value);
```
because that naked less-than seems confusing, a least to me, as a veteran c++ programmer...

opencog/atoms/value/StreamValue.cc
rTreutlein: I still don't get it. The code kinda looks like
```
If (this->get_type() != other.get_type())
    return this->get_type() < other.get_type();
else
    Do something custom for this type
```
I haven't explicitly used an else but maybe I should. Anyway it's true that the condition and the first option are the same for all Values but I can't just move out a part of the Code and if I try to move the whole block it fails because I can't know what the custom part is. Unless you want me to do a switch statement that checks the Type we have and then does something custom according to that. But that seems like bad design if the code in Value is responsible for other Types.

opencog/atoms/value/LinkValue.cc
rTreutlein: I guess having worked so much with Haskell I am just more used to things like that. I find it annoying when I can't print a vector automatically when I know how to print an element.

I mean I can change that no problem just don't really see the difference between the 2 version other than yours having more characters.

opencog/atoms/value/StreamValue.cc
linas: Never use `else` if you can avoid it. Indentation is evil.

custom: but all of your code is just cut-n-paste, all of the "custom" parts are identical.  If/when some value actually does need to do anything actually custom, it can always implement this method. But right now, they are all cut-n-paste of one-another.

However, I am starting to question the need for an `opeator<()` for Values, anyway. What does this do? What's the point? What's it good for?  Consider an audio stream -- should    `operator<()` compare the time-stamps on the audio streams? should it compare the loudness of the audio? Should it compare the URL of the audio source, ordered lexicographically? What, exactly, are we trying to accomplish, here?

rTreutlein: Well in this case the custom part looks like:
```
const StreamValue* sov = (const StreamValue*) &other;
return _value < sov->_value;
```
But since _value is not define in Value.h but instead in each of it's sub-classes I can't move that part into Value.h

Well, the reason I originally added this was so that I could use Values in an ordered container. For that, the ordering doesn't really need a semantic meaning just like the ordering of Atoms has no meaning. It's just an added benefit that for some Values like StringValue or FloatValue it does have a meaning.
But I mean we don't have to add this now while nobody is using it. It just seemed like a useful addition overall.

linas: Hmm. philosophically speaking "It just seemed like a useful" is considered to be a dangerous, bad coding habit. It is something that junior programmers love to do, and senior programmers (and project managers) detest.  Here's why (and let me pretend this is a commercial setting, but its also true for open source): It's not only code-bloat, its pernicious: three years from now, someone realizes that they need some kind of ordering that is different from what the current `operator<()` does. They add that. They might spend a lot of time trying to retrofit the existing `operator<()` be backwards-compat.  If no one is actually using `operator<()`, then this is time-wasted on the part of the developer.  Worse, some customer might discover this, and start using it (probably for some spurious, totally bogus reason) but "the customer is always right" so the project manager has to assign someone to write better unit tests, better integration tests, and check system test coverage. And if the customer gets borken, deal with that.   Meanwhile the system architect gets pulled in and proclaims "this is stupid, we need to rip this out". So, now we have to figure out how to deprecate the function, while maintaining backwards-compat for N years.

What has happened is that an idle afternoon of a junior developer has snowballed into ten-person-days, or more, of headache-inducing fixing, tracking, refactoring and redesign work down the road -- all to support something that really wasn't really needed, to begin with. The general moral of the story:  don't just add spurious code to a code-base just because it seems cute.  There really should be a strong reason for it: an immediate need or an obviously required feature.

rTreutlein: Well then I'll just get rid of these changes but it would have been nice if you could have told me that earlier. Because exactly for that reason I wrote in my initial message that these changes are currently not used by anything. I thought it was useful but I did want confirmation that adding code that's not used is okay. Guess next time I'll ask more directly.

PR #1943, opened 2018-12-10
Change minimum version to 5.4.0

yitbrek:
as per #1916

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,11 +79,11 @@ include(${COGUTIL_DATA_DIR}/cmake/Summary.cmake)
 # Detect different compilers and OS'es, tweak flags as necessary.
 
 IF (CMAKE_COMPILER_IS_GNUCXX)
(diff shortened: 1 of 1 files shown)


linas:
OK w/ me. @ngeiswei last chance? Anyone else care about this?

ngeiswei:
I'm OK but the comment isn't accurate. On the other hand it's not very important.

ngeiswei:
Also as @Dagiopia pointed out it seems 5.0 would work, so maybe it's better to set the minimum version to 5.0 as who knows what distro might use it.

yitbrek:
So, may I change it to 5.0?

ngeiswei:
Thanks @yitbrek. If you can change the comment as well with something like "Version 5.0 of gcc is required for passing non-trivially copyable class via C varargs" or anything that you think is more adequate, that would be great.

linas:
closing, this seems to be obsoleted by #1946

PR #1945, opened 2018-12-11
Update add guile test

amebel:


linas:
Uh, this reverts an earlier fix, from just a few days ago, and now causes the unit tests to fail again ... So um, what's up with that?

linas:
Specifically, this reverts #1936  -- why are you reverting?

linas:
Oh, I see, is the claim that the new install to the site directory means this is un-needed? Let me check ...

linas:
OK, so .. never mind.  The unit tests do pass, the new install location fixes the need for the `-L` flag.

PR #1947, opened 2018-12-11
Factorize atom types

kasimebrahim:
This will remove the confusion mentioned in issue #1930.
opencog, agi-bio,benchmark and as-moses will require fixing.


linas:
Thanks. This is just what I wanted. However, some unit tests are failing; I'm debugging them now.

linas:
found it; fixing shortly.

linas:
will merge this first, then push a fix on top of it.

PR #1950, opened 2018-12-12
Move truthvalue2atoms

kasimebrahim:
per #1930.
opencog,benchmark and as-moses will require fixing.


kasimebrahim:
Oh man I am so sorry, I will fix it as soon as I get office.

examples/c++-api/CRUDExample.cc
```diff
@@ -2,9 +2,9 @@
 
 #include <opencog/atomspace/AtomSpace.h>
 #include <opencog/attentionbank/AttentionBank.h>
-#include <opencog/truthvalue/AttentionValue.h>
-#include <opencog/truthvalue/SimpleTruthValue.h>
-#include <opencog/truthvalue/TruthValue.h>
+#include <opencog/atoms/truthvalue/AttentionValue.h>
```
ngeiswei: I just realized that `AttentionValue.h` was under `truthvalue` but it has nothing to do here, right? @linas? Shouldn't it moved in its own directory? Anyway, I don't want to add confusion to that PR, we can really discuss that in another PR.

linas: Don't care. We could leave it there, or move it to the `attentionbank` directory (thereby indicating that one should never use it, unless one is also using the attentionbank.)

PR #1952, opened 2018-12-14
Use SchemeModule ModuleWrap in PersistSCM

rTreutlein:
PersistSCM had duplicated code from ModuleWrap. So now PersistSCM
inherits from ModuleWrap.


linas:
Yep. The persist code was older than the module wrap code. Did things that way, for a while, and then realized there was a recurring pattern, and created  ModuleWrap to capture that recurring pattern.

Issue #833, opened 2016-07-19
Generalized Distributional TV proposal


ngeiswei:
# Generalized Distributional TV

## Overview

This is a proposal for a new TV type that encompasses probabilistic,
fuzzy, distributional TV types and more. In short it is a
distributional TV that may work over any domain, not just
boolean/fuzzy, equipped with counts to quantify uncertainty.

Disclaimer: things are presented under the angle of Kolmogorov
probability theory, mostly cause I feel more comfortable with it.

## Omega

Omega is a probability space representing the subjective universe. We
learn about Omega via random variables. A random variable X: Omega->E
is a measurable function from domain Omega to co-domain E.

Any atom in the AtomSpace associated with a TV represents a random
variable. The TV (or GDTV as defined below) describes the distribution
over the co-domain of that random variable, and the atom defines the
name (or the structure) of that random variable and its relationships
to other random variables, i.e. other atoms.

When talking about the co-domain of an atom A we mean the co-domain of
the associated random variable of the TV of that atom.

## Definition

We introduce a new TV type called Generalized Distributional Truth
Value, or GDTV for short. It could also be called merely
Distributional Value, or DV for short.

### Unconditional GDTV

An unconditional GDTV of atom A is a pair <D,N> where

1. D is a probability distribution over E, the co-domain of A.
2. N is a count, a non negative real number.

In practice, discrete case, the probability measure will be described
as a mass function over a partition of E summing up to 1, following
the format

```
<{k1:p1, ..., kn:pn}, N>
```

where k1 to kn are the partition blocks (often singletons or
intervals), and p1 to pn are probabilities summing up to 1.

### Conditional GDTV

More generally a GDTV is a conditional probability distribution. In
the discrete case, a GDTV over 2 atoms A and B is defined as a map
from a partition of the co-domain of A to a unconditional GDTV of B, as
defined above. The whole thing forms a conditional probability matrix,
equipped with a count on each row. The format may look like

```
{ka1: <{kb1:p11, ..., kbm:p1m}, N1>,
 ...
 kan: <{kb1:pn1, ..., kbm:pnm}, Nn>}
```

where ka1 to kan are partition blocks over the co-domain of A, and kb1
to kbm are partition blocks over the co-domain of B.

#### Perfect boolean sensor

Let's assume our window to the universe is a perfect and timeless
boolean sensor A. Since the sensor is boolean the co-domain E of the
associated random variable is {0,1}. One may represent an active
sensor as follows

```
[...]

PR #1957, opened 2018-12-17
Initial Version of Distributional Values

rTreutlein:
This should now contain all the suggestion I received on the last PR and it no longer contains the addition of a comparison operator for values.


opencog/atoms/distvalue/DistributionalValueSCM.cc
```diff
@@ -0,0 +1,463 @@
...
+			DVKey v = verify_DVKey(svalue,"scm_to_DVKeySeq",0);
+			valist.emplace_back(v);
+		}
+		sl = SCM_CDR(sl);
+	}
+	return valist;
+}
+
+SCM DistributionalValueSCM::float_list_to_scm(const std::vector<double>& v)
+{
```
linas: FYI, there is a scheme function `cog-value->list` implemented in `ss_value_to_list` that accepts an SCM value, and returns a list of whatever is inside of it. It is currently implemented as one big case statement; it should probably be converted to some kind of per-value callback (so that each value knows how to convert itself into a list).  We've never done this kind of per-object scheme callback, so not clear how that might work.

Its plausible to think that someone someday might expect `cog-value->list` to do something reasonable for distributional values; however, implementing this, today, is low priority.

opencog/atoms/value/FloatValue.cc
```diff
@@ -35,16 +36,10 @@ bool FloatValue::operator==(const Value& other) const
...
 	size_t len = _value.size();
 	for (size_t i=0; i<len; i++)
-		// Compare floats with ULPS, because they are lexicographically
-		// ordered. For technical explanation, see
-		// http://www.cygnus-software.com/papers/comparingfloats/Comparing%20floating%20point%20numbers.htm
-		// if (1.0e-15 < fabs(1.0 - fov->_value[i]/_value[i])) return false;
-#define MAX_ULPS 24
-		if (MAX_ULPS < llabs(*(int64_t*) &(_value[i]) - *(int64_t*)&(fov->_value[i])))
+		if (not is_approx_eq_ulp(_value[i],fov->_value[i],24))
 			return false;
```
linas: this change should not be a part of this pull req; its completely unrelated.

linas: anyway, you should not remove the first half of the comment, and you should not hard-code the magic number 24.  The `MAX_ULPS` define was specifying that literal; its bad C/C++ style to have literals that don't have an associated `#define`  (because if some future programmer wants to change the 24, they have to find all other uses of 24 in the code base, and wonder if they have to change them too.  If they forget to change one of the other 24's, bugs will typically happen.)

opencog/scm/opencog/rule-engine/rule-engine-utils.scm
```diff
@@ -43,6 +43,7 @@
 ;; -- tv->bool -- Convert TRUE_TV to #t, anything else to #f
 ;; -- atom->number -- Convert NumberNode into its corresponding number
 ;; -- gt-zero-confidence -- Return TrueTV iff A's confidence is greater than 0
+;; -- has-dv -- Return TrueTV iff A has a Distributional Value
 ;; -- absolutely-true -- Return TrueTV iff A's TV is TrueTV
```
linas: This too, should probably be a distinct pull request.  For small pull reqs, ti would not matter.  For something this big, the pull req defining the thing should be distinct from the pull req that uses it.

linas:
OK, so far, I've only examined the "style", not the "substance". There's one thing that catches my eye: a proliferation of different C++ types, which then requires a proliferation of scheme types, which then requires 4 or 5 different conversion routines to convert between one and the other.  So, for example ` scm_to_DVKey` and `scm_to_DVKeySeq` which are almost the same, but different, and also `float_list_to_scm` and `dvkey_to_scm` and `dvkeyseq_to_scm` which are almost the same, but different.

So -- please recall one of the meta-goals of atomese, which is meant to govern the API design: Atomese is not meant to be a programming language for humans; it is meant to be a programming language for other algorithms.  In particular, it should be easy for other algos to grab Atomese expressions, and do interesting things with them.  This has usually meant that the number of new types are kept minimal: there are only Atoms and Values. In C++ this means only Handles and ValuePtrs.  And then we accidentally allowed someone to typedef things like HandleSeq and ProtomSeq, which is almost surely a design mistake, and then we have allowed functions that that HandleSeq and ProtomSeq as function arguments -- another design mistake. But this is a subtle point: there's nothing wrong with HandleSeq per se, as long as it's kept private, hidden away, part of the internal implementation, and not visible to the public. Kind of like how you hide the internals of a coffee maker or a car behind a nice shiny surface.

Ideally, the DV API should be so simple, that other things, like MOSES or Neural Nets or (well, the matrix-like things) or ghost or whatever, can latch on to it, and manipulate the contents. And so the 18 brand new `cog-cdv` methods seem like maybe 15 too many.  OK, so, yes, agreed, in the "short term" we need humans as well as algos to write in atomese.  So maybe these are needed.  But methods like `cog-cdv-get-conditions` have names that suggest that they should be generically valid, instead of being limited to just DV's.

Perhaps its too early in the implementation to make these things generic.  Also, I have not yet studied what the code actually does -- I only looked at its form. 

In the meanwhile, some long-running discussions about "how to do things":  https://en.wikipedia.org/wiki/Form_follows_function  and an interesting side-track to East Germany, where artists and designers were politically forbidden from doing their work, and so untrained factory-floor workers took over the job of designing: https://www.taschen.com/pages/en/catalogue/architecture/all/40201/facts.sed_design.htm  It wasn't just East Germany; you could see the same issues arising in all Soviet consumer products. The book appears to be out of print; a very miniscule taste from google image search: https://www.google.com/search?tbm=isch&q=soviet+design So those pictures are about consumer goods. Here, the intended "consumer" of atomese is not a human, but another algorithm; thus, atomese needs to resemble GIL or gimple, or java byte-code or Gnu lightning: https://www.gnu.org/software/lightning/manual/lightning.html except targeting a prolog-like datastore, instead of targeting a hardware chip.

rTreutlein:
The extra types are there for humans to be able to understand the code. As the representation has multiple nested vectors writing out the types would be very verbose and I think confusing. I know multiple nested vectors doesn't sound good but each level has a clear semantic meaning so it makes working with it easier.
The different conversion functions are there for different levels in that structure. Most aren't actually used in something other than the function that converts the 1 higher level. If we had an overloaded scm_to/from then with the addition of some template functions that could be simplified to get rid of the redundant code.
The reason why there are so many functions in the API is that they implement functionality required for PLN which up till now were written in Scheme (for TVs). Maybe these shouldn't be part of the AtomSpace I don't really know. But the point is that all that exists for TVs already I just moved it to C++. 
But this is a good point I have primarily been thinking about how the API would be used for PLN and not how it could be used from/for Atomese.

linas:
Hard for me to say more until I read through it and get the general idea. Re: "nested vectors" -- this was (partly) the intent of the LinkValue: its a vector of values.  So, for example, an NxM matrix could be represented as a LinkValue of N FloatValues, each FloatValue holding M floats. 

Re: "getting the general idea" -- there should be accompanying documentation. Traditionally, this has been in the wiki.

linas:
It's possible that we need a `KeyValue` or `KVPValue` or a `MapValue` which is a `std::map` under the covers. The key would always have to be an atom. This seems top be what you are defacto using, after a very very fast skim of your code.

linas:
Re; TV's and AV's: perhaps in some distant future, we could make TV and AV be special cases of DV.  Some of the defined TV's are unused, and are essentially junk: FuzzyTV, EvidenceCountTV, etc. and need to be discarded.

ngeiswei:
> Re; TV's and AV's: perhaps in some distant future, we could make TV and AV be special cases of DV. Some of the defined TV's are unused, and are essentially junk: FuzzyTV, EvidenceCountTV, etc. and need to be discarded.

Yes, that is the plan, once DV has matured enough.

ngeiswei:
@rTreutlein, just a detail for future PR, it's better if your commits are on top rather interleaved with a master merge, that way we don't run into the risk of missing some.

Also only merge the master into your branch if you need to. Alternatively you can rebase (before the PR of course), which will puts your commits on top.

rTreutlein:
@ngeiswei I was considering a rebase but there were a bunch of conflicts in the codebase and I would have probably missed something leading to a lot of broken commits. Which if you ever wanted to do a git bisect would be problematic.

And I had to do a merge before the last PR and before this one to fix the conflicts so that it could be merged. But I guess I could have done the PR as it was to more easily discuss it and then afterwards merge in Master to fix conflicts.

ngeiswei:
@rTreutlein, actually it would be helpful if you could rebase at that point. Your commits are scattered at 3 different places over 2 months of history. You could rebase onto the master and force push, I think it would work well. Thanks.

ngeiswei:
OK, I get it, @rTreutlein. The other option would be to reconstruct the git history (after rebase and soft reset), but it's a rather laborious task and it's not clear it's worth it. I'll start reviewing in that state, maybe I'll ask you to reconstruct if there's too much mess but otherwise I think it's OK as it is.

opencog/atoms/distvalue/DistributionalValue.cc
```diff
@@ -0,0 +1,455 @@
...
+		{
+			return false;
+		}
+	}
+	return true;
+}
+
+//Add Evidence for the provided key i.e increment the count of this Interval
+//Returns a new DVPtr with the update count
+DistributionalValuePtr DistributionalValue::add_evidence(DVKey h) const
```
ngeiswei: Better pass by const ref.

opencog/atoms/distvalue/DistributionalValue.cc
```diff
@@ -0,0 +1,455 @@
...
+			return 0;
+
+		//Two Intervals
+		//Calculate the overlapp
+		Interval res;
+		//Find the start of the Overlapp
+		if (k1[0] > k2[0])
+			res.push_back(k1[0]);
+		else
+			res.push_back(k2[0]);
```
ngeiswei: Question of taste but I would go with
```c++
		res.push_back(k1[0] > k2[0] ? k1[0] : k2[0]);
```

opencog/atoms/distvalue/DistributionalValue.cc
```diff
@@ -0,0 +1,455 @@
...
+
+		//Find the end of the Overlapp
+		if (k1[1] < k2[1])
+			res.push_back(k1[1]);
+		else
+			res.push_back(k2[1]);
```
ngeiswei: Same thing, taste, do what you prefer.

opencog/atoms/distvalue/DistributionalValue.cc
```diff
@@ -0,0 +1,455 @@
...
+			return false;
+	}
+	return true;
+}
+
+std::string oc_to_string(const DistributionalValuePtr dvp)
+{
+	std::string part = dvp->to_string("");
+	return part;
+}
```
ngeiswei: 1. `oc_to_string` now always takes an optional indent argument. I had done 2 versions, one without indent and one with indent, to work around some gcc/gdb issue, but it's been fixed in gcc 5.
2. Why not just `return dvp->to_string(indent);` instead of creating the intermediary variable `part`?

opencog/atoms/distvalue/DistributionalValue.h
```diff
@@ -0,0 +1,157 @@
...
+	// return std::const_pointer_cast<Value>(tv);
+
+	// This, too, should have worked!?
+	// return std::shared_ptr<Value>(tv, const_cast<Value*>(tv.get()));
+
+	// This works...
+	return std::shared_ptr<Value>(dv, (Value*) dv.get());
+}
+
+std::string oc_to_string(const DistributionalValue& hc);
```
ngeiswei: Require optional indent argument.

opencog/atoms/distvalue/DistributionalValueSCM.cc
```diff
@@ -0,0 +1,463 @@
...
+	                        &DistributionalValueSCM::ss_new_dv_simple,
+	                        this,"distvalue");
+	define_scheme_primitive("cog-dv?",
+	                        &DistributionalValueSCM::ss_dv_p,
+	                        this,"distvalue");
+	define_scheme_primitive("cog-dv-divide",
+	                        &DistributionalValueSCM::ss_dv_divide,
+	                        this,"distvalue");
+	define_scheme_primitive("cog-dv-sum-joint",
+							&DistributionalValueSCM::ss_dv_sum_joint,
```
ngeiswei: Should be aligned with whitespaces.

opencog/atoms/distvalue/DistributionalValueSCM.cc
```diff
@@ -0,0 +1,463 @@
...
+        if (first) {
+            first = false;
+            res = scm_list_1(sh);
+        }
+        else {
+            res = scm_cons(sh,res);
+        }
+    }
+    return res;
+}
```
ngeiswei: The previous 50 lines should be smart/mixed tab indented.

opencog/atoms/distvalue/DistributionalValueSCM.cc
```diff
@@ -0,0 +1,463 @@
...
+   return cdv_to_scm(res);
+}
+
+SCM DistributionalValueSCM::ss_cdv_cde(SCM scdv1,SCM scdv2)
+{
+   ConditionalDVPtr cdv1 = verify_cdv(scdv1,"cog-cdv-cde",1);
+   ConditionalDVPtr cdv2 = verify_cdv(scdv2,"cog-cdv-cde",2);
+   ConditionalDVPtr res = DVFormulas::consequent_disjunction_elemination(cdv1,cdv2);
+   return cdv_to_scm(res);
+}
```
ngeiswei: The previous 20 lines should be tab indented.

opencog/atoms/distvalue/DistributionalValue.h
```diff
@@ -0,0 +1,157 @@
...
+		throw RuntimeException(TRACE_INFO, "Cannot modify truth values!");
+	}
+
+
+public:
+	static count_t DEFAULT_K;
+
+	DistributionalValue();
+	DistributionalValue(DVCounter);
+	DistributionalValue(double,double);
```
ngeiswei:
```suggestion
	DistributionalValue(double, double);
```

opencog/atoms/distvalue/DistributionalValue.h
```diff
@@ -0,0 +1,157 @@
...
+	DistributionalValue& operator=(const DistributionalValue& rhs) {
+		throw RuntimeException(TRACE_INFO, "Cannot modify truth values!");
+	}
+
+
+public:
+	static count_t DEFAULT_K;
+
+	DistributionalValue();
+	DistributionalValue(DVCounter);
```
ngeiswei:
```suggestion
	DistributionalValue(const DVCounter&);
```
This will avoid to copy twice.

opencog/atoms/distvalue/DistributionalValue.h
```diff
@@ -0,0 +1,157 @@
...
+	DistributionalValue(double,double);
+
+	const DVCounter& value() const { return _value; }
+
+	static DistributionalValuePtr UniformDistributionalValue(DVKey,int);
```
ngeiswei:
```suggestion
	static DistributionalValuePtr UniformDistributionalValue(const DVKey&, int);
```

opencog/atoms/distvalue/DistributionalValue.h
```diff
@@ -0,0 +1,157 @@
...
+	static DistributionalValuePtr UniformDistributionalValue(DVKeySeq,int);
```
ngeiswei:
```suggestion
	static DistributionalValuePtr UniformDistributionalValue(const DVKeySeq&, int);
```

opencog/atoms/distvalue/DistributionalValue.h
```diff
@@ -0,0 +1,157 @@
...
+	static DistributionalValuePtr TRUE_TV();
+	static DistributionalValuePtr FALSE_TV();
+	static DistributionalValuePtr DEFAULT_TV();
+	static DistributionalValuePtr createDV(double,double);
```
ngeiswei:
```suggestion
	static DistributionalValuePtr createDV(double, double);
```

opencog/atoms/distvalue/DistributionalValue.h
```diff
@@ -0,0 +1,157 @@
...
+	static DistributionalValuePtr createDV(DVCounter);
```
ngeiswei:
```suggestion
	static DistributionalValuePtr createDV(const DVCounter&);
```

opencog/atoms/distvalue/DistributionalValue.h
```diff
@@ -0,0 +1,157 @@
...
+	static int to_count(double);
+
+	bool is_uniform() const;
+
+	std::vector<double> get_mode() const;
+	std::vector<double> get_mean() const;
+	std::vector<double> get_var() const;
+
+	double get_fstord_mean() const;
+	DVec middle_of_interval(DVKey) const;
```
ngeiswei:
```suggestion
	DVec middle_of_interval(const DVKey&) const;
```

opencog/atoms/distvalue/DistributionalValue.h
```diff
@@ -0,0 +1,157 @@
...
+
+	double get_mode_for(double) const;
+	double get_mean_for(double) const;
+	double get_var_for(double) const;
+
+	DistributionalValuePtr add_evidence(DVKey) const;
```
ngeiswei:
```suggestion
	DistributionalValuePtr add_evidence(const DVKey&) const;
```
etc...

opencog/atoms/distvalue/DVFormulas.cc
```diff
@@ -116,7 +118,7 @@ DVFormulas::conjunction(DistributionalValuePtr dv1,
 	double m1 = 1;
 	double m2 = 1;
 
-	while ((abs(m1) > 0.0000001) && (abs(m2) > 0.0000001))
+	while (not is_within(m1,0.0,EPSILON) && not is_within(m2,0.0,EPSILON))
```
ngeiswei: You should use space after comma (I've upgraded the dev standard accordingly).

opencog/atoms/distvalue/DistributionalValue.h
```diff
@@ -0,0 +1,157 @@
...
+	DistributionalValuePtr merge(DistributionalValuePtr) const;
+	DistributionalValuePtr negate() const;
+
+	double min_count() const;
+	double max_count() const;
+
+	double total_count() const;
+	double get_confidence() const;
+
+	static double key_contained(const DVKey&, const DVKey&);
```
ngeiswei: @rTreutlein, I wonder about https://www.boost.org/doc/libs/1_69_0/libs/numeric/interval/doc/interval.htm ? Would that be overkill in your opinion? Or maybe at least using their terminology.

opencog/atoms/distvalue/DistributionalValue.cc
```diff
@@ -0,0 +1,455 @@
...
+
+//Find out how much of Key1 is contained in Key2
+double DistributionalValue::key_contained(const DVKey &ks1,const DVKey &ks2)
+{
+	Interval k1,k2;
+	//Start with the assumption that 100% of Key1 is in Key2
+	double sum = 1;
+	for (auto zipped : boost::combine(ks1,ks2))
+	{
+		boost::tie(k1,k2) = zipped;
```
ngeiswei: Wow, that `tie` function is really bad ass! Actually it seems to be in the standard https://en.cppreference.com/w/cpp/utility/tuple/tie and I would be bet it is supported by gcc 5.0 so if you can replace by `std::tie` it would be better.

ngeiswei: What's more, it is optimized away https://stackoverflow.com/questions/30212307/does-gcc-optimize-stdtie-used-only-for-readability and works for pairs https://stackoverflow.com/questions/31272966/does-there-exist-something-like-stdtie-for-stdpair#31273053 this thing :-)

opencog/atoms/distvalue/DistributionalValue.cc
```diff
@@ -0,0 +1,455 @@
...
+		//Singleton Intervals are equal => assumption is correct and we can continue
+		if (k1.size() == 1 && 1 == k2.size() && k1[0] == k2[0])
+			continue;
```
ngeiswei: Would that return 1?

opencog/atoms/distvalue/DistributionalValue.cc
```diff
@@ -0,0 +1,455 @@
...
+			res.push_back(k1[1]);
+		else
+			res.push_back(k2[1]);
+
+		//Check that the Overlapp starts before it ends
+		if (res[0] <= res[1])
+			//Figure out how much smaler the Overlapp is then K1
+			sum *= (res[1] - res[0]) / (k1[1] - k1[0]);
+		else
+			return 0;
```
ngeiswei: That looks like a Jaccard index to me, maybe that's how it should be called. There's BTW a `jaccard_index` function in cogutil but it's for sets, not intervals.

ngeiswei: Sorry, it's not the Jaccard index. It must have a standard name though...

ngeiswei: It's basically a conditional probability, maybe that's how it should be called...

opencog/atoms/distvalue/DistributionalValue.h
rTreutlein: I looked at that in the beginning and decided against it but maybe it's worth considering again. The main issue is that they don't support multidimensional Intervals or actually they might if I have an Interval over a vector of doubles that might work. I'll have to try that and see how it works out. But in any case using their terminology makes sense.

opencog/atoms/distvalue/DistributionalValue.cc
rTreutlein: Oh, you are right the intervals are basically continuous uniform distributions this calculates the conditional probability. Going to rename it and add some comments.
Would be nice if we could use other distributions but I don't know of a library that implements that nicely. std/boost have distributions but they are mainly for generating random numbers not to calculate conditional probabilities between them.

opencog/atoms/distvalue/DistributionalValue.h
ngeiswei: Looking the git history of the project it seems rather mature. It's been in dev for 17 years and hasn't got a code change for at most 2 years.

opencog/atoms/distvalue/DistributionalValue.cc
ngeiswei: Maybe have a look at https://www.boost.org/doc/libs/1_69_0/libs/math/doc/html/dist.html

The URE code uses https://www.boost.org/doc/libs/1_69_0/libs/math/doc/html/math_toolkit/dist_ref/dists/beta_dist.html to calculate beta distribution pdf and other things.

Though I suppose uniform distribution is OK for starter.

rTreutlein: hm, I guess using the pdf and some integration I could calculate the conditional probability on my own.

opencog/atoms/distvalue/DistributionalValue.cc
rTreutlein: Hm, not as straightforward as I hoped. boost::combine uses boost::tuples so std::tie doesn't work for that. So I also need to replace boost::combine and there is no combine in std yet.
So I could leave as is or maybe switch to using 2 independent iterators or write my own combine function that uses std::tuple. I am leaning towards option 2 as that is also what I have done in some other places. Option 3 seems a bit overkill for this one case but if it were used in more places it might be worth it.

ngeiswei: Do what you feel is right, keeping `boost::tie` isn't a big deal if necessary.

opencog/atoms/distvalue/DistributionalValue.cc
ngeiswei: Boost can handle the cdf as well.

linas:
OK, I am starting to review what this code actually does. And, between my earlier comments, and Nil's comments, and the comments I want to make, but have not yet made, I feel like we are "piling on" and that Roman will crack with frustration.  And I'd rather not have that. So, I'm not quite sure how to proceed. So, here we go:

Now that I'm reading what the code actually does:

* Technically, what is implemented here is not a distribution, its a histogram. Calling it a distribution is a common mis-understanding or just casual conversational sloppiness.  https://en.wikipedia.org/wiki/Histogram  Rather than enshrining casual sloppiness in permanent code, it should probably be renamed. 

* DVKey also seems to be an unusual name choice. Normally, a "key" is the first element in a pair. And, OK, in `Counter<DVKey, double>` it is indeed the first element, but .. semantically, it is used as a list of bins. So I propose changing the name to BinList.
 
* Interval is an OK name, cause that's what it is, syntactically. But semantically, what it really is is a Bin. and so it probably should be called that.  Even something goofy like `typedef Interval Bin` would be better.

I have more comments, to be made line-by-line.  Again, apologies, as a review this close and nit-picky is frustrating and irritating. However, in terms of developer-time and effort, its much easier cheaper and quicker to get it fixed up now, rather than later.

opencog/atoms/distvalue/DistributionalValue.cc
```diff
@@ -0,0 +1,455 @@
...
+}
+
+DistributionalValuePtr DistributionalValue::createDV(double mode,
+                                                     double conf)
+{
+	return std::make_shared<const DistributionalValue>(mode,conf);
+}
+
+DistributionalValuePtr
+DistributionalValue::UniformDistributionalValue(const DVKey &k,int c)
```
linas: why `int`, why not `double`?

opencog/atoms/distvalue/DistributionalValue.cc
```diff
@@ -0,0 +1,455 @@
...
+{
+	DVCounter dvctr;
+	dvctr[k] = c;
```
linas: This seems incorrect. This will cause a vector to be created, consisting of exactly k+1 bins.  The width of each bin will be exactly zero.  The location of each bin will be at 0.0.  Surely this is neither intended nor desirable.

opencog/atoms/distvalue/DistributionalValue.cc
```diff
@@ -0,0 +1,455 @@
...
+}
+
+
+DistributionalValuePtr
+DistributionalValue::UniformDistributionalValue(const DVKeySeq &ks,int c)
+{
+	DVCounter dvctr;
+	for (auto k : ks)
+	{
+		dvctr[k] = c;
```
linas: as above: the bins have un-known widths. There's no reason that I can see for `c` to be an int.

opencog/atoms/distvalue/DistributionalValue.cc
```diff
@@ -0,0 +1,455 @@
...
+	}
+	return createDV(dvctr);
+}
+
+DistributionalValuePtr DistributionalValue::TRUE_TV()
+{
+	DVKey v1{Interval{1.0}};
+	DVKey v2{Interval{0.0}};
```
linas: How can an interval consist of only one point? Intervals must neccesarily consist of two and exactly two points ...

opencog/atoms/distvalue/DistributionalValue.cc
```diff
@@ -0,0 +1,455 @@
...
+	DVCounter dvc;
+	dvc[v1] = 1;
+	dvc[v2] = 0;
+	return std::make_shared<const DistributionalValue>(dvc);
+}
```
linas: This should return a pointer to a singleton. Do NOT create a new instance with every call!  Why?
1) This splurges memory  2) it makes compares harder.  If two pointers point at the same object, you can trivially compare pointers. If they don't then you have to compare contents, which will always be slower.

opencog/atoms/distvalue/DistributionalValue.cc
```diff
@@ -0,0 +1,455 @@
...
+
+bool DistributionalValue::is_uniform() const
+{
+	double val = _value.begin()->second;
+	for (auto p : _value)
+	{
+		if (val != p.second)
+		{
```
linas: val is a `double`. This won't work. You have to `is_approx_equal_ulp` or something like that.

Furthermore, a distribution will NEVER be flat to within an ulp or two. A distribution is considered "flat" when the variation between bins is less than sqrt(N)/N where N is the average count per bin.   Also -- I'm missing a factor of sqrt(pi/2)  Also: when computing the "average count per bin" one must NOT include the counts from the overflow bins in the average!!

This is subtle stuff, and easy to get wrong, if you just wing it.  Should be coding this with a book or good reference by your side.

opencog/atoms/distvalue/DistributionalValue.cc
```diff
@@ -0,0 +1,455 @@
...
+	}
+	return true;
+}
+
+//Add Evidence for the provided key i.e increment the count of this Interval
+//Returns a new DVPtr with the update count
+DistributionalValuePtr DistributionalValue::add_evidence(const DVKey &h) const
+{
+	DVCounter newdvc = _value;
+	newdvc[h] += 1;
```
linas: why just plus-one?  It could be a positive or a negative double, in general.

Also, this cannot possibly work as written. Recall that `DVKey` is not a single bin, but is a list of bins.  So this doesn't make sense, as written.

There are also other issues:  after you fix `DVKey` to be `Bin` you still have the issue where `DVCounter` might be holding a bin [0.4, 0.5] bu you accidentally supplied a bin of width [0.40001, 0.500001] and so now you increment the count in a bin that completely overlaps an existing bin.  So this API is borken.

opencog/atoms/distvalue/DistributionalValue.cc
```diff
@@ -0,0 +1,455 @@
...
+}
+
+//Flip all the counts
+DistributionalValuePtr DistributionalValue::negate() const
+{
+	double total = max_count() + min_count();
+	DVCounter res;
+	for (auto elem : _value)
+	{
+		res[elem.first] = total - elem.second;
```
linas: Whit is implemented here is not negation. Its mirroring about the L-infinity mean.

opencog/atoms/distvalue/DistributionalValue.cc
```diff
@@ -0,0 +1,455 @@
...
+	}
+	return createDV(res);
+}
+
+//Get the lowest count of all Interals
+double DistributionalValue::min_count() const
```
linas: `min_count` and `max_count` need to be compbined into one, and return an `Interval` (not a `Bin`) 

Everyone who asks for min will always ask for max immediately afterwards, causing you to run the loop twice, paying all cache-line misses twice.  you can get both min and max "for free", because the additional branch will be almost completely seamped by cache-line effects.

opencog/atoms/distvalue/DistributionalValue.cc
```diff
@@ -0,0 +1,455 @@
...
+//Get the variance for a Count
+double DistributionalValue::get_var_for(double ai) const
+{
+		double a0 = total_count();
+		return ai*(a0-ai) / a0*a0*(a0+1);
+}
+
+double DistributionalValue::total_count() const
+{
+	return _value.total_count();
```
linas: Without guard-bins, this will return a total that is 100% garbage.  Sorry.

linas:
I'm not sure what to say. I'm giving up on reviewing. There are a very large number of fundamental design flaws shot throughout the code. It would probably be a good idea to get a book on this subject, and read through it.  This is not difficult math, but its clearly the first-time ever that you are seeing it. 

The wikipedia article https://en.wikipedia.org/wiki/Histogram does list a few books in it's references; no clue how good they are. In ye olden days, these books had titles like "lab technique for scientists" and "data analysis for scientists", and "good experimental design" or "designing experiments and analyzing data".  Track down one of these books and go through it. It will clear up a lot of the design issues that I'm seeing.  This will take some effort on your part -- sorry.

linas:
I'm closing for now.  This needs a rethink; just a few patches here and there won't rescue this.

linas:
p.s. If I'm being too harsh about this or being an asshole about it, let me know.  I've been rubbing people the wrong way; am trying to stop.

opencog/atoms/distvalue/DistributionalValue.cc
ngeiswei: A point would be a bin with a width of epsilon > 0. I think having pointy bins, like one at 0, the other at 1 allows crisp values for crisp logic, etc.

ngeiswei:
@linas, I'd like to review it in depth first, especially as it addresses an issue that I have created #833. I'm reopening because otherwise @rTreutlein can't push commits, which doesn't help progress, regardless of whether or not the PR gets merged in the end.

opencog/atoms/distvalue/DistributionalValue.h
```diff
@@ -0,0 +1,157 @@
...
+	DistributionalValue& operator=(const DistributionalValue& rhs) {
+		throw RuntimeException(TRACE_INFO, "Cannot modify truth values!");
+	}
+
+
+public:
+	static count_t DEFAULT_K;
+
+	DistributionalValue();
+	DistributionalValue(const DVCounter&);
```
ngeiswei: What do the values of DVCounter represent? According to get_mean and get_mode they represent parameters of a Dirichlet distribution. But in the absence of comments the user might think that they represent evidence counts.

@rTreutlein, you need to better comment your code. My personal policy regarding comments is as follows
1. Put the "what" in header file. Everything the user needs to know to use the code. What each function does, using math formulas if you need to, pointing to wikipedia pages, articles, github issues if you need to, adding examples if you need to. The user should be able to read your comments and know what it is doing without having to read the code.
2. Put the "how" in the source file. Everything the user may not care but the developer/maintainer may need.

See my last piece of code, the pattern miner, as examples of comments https://github.com/opencog/opencog/blob/master/opencog/learning/miner/Miner.h https://github.com/opencog/opencog/blob/master/opencog/learning/miner/MinerUtils.h (it's not a coincidence that my example comments are in my last code, that's because I've come to learn over the years how important good comments are).

You should also add a README.md, see for instance https://github.com/opencog/opencog/blob/master/opencog/learning/miner/README.md

Writing good comments and good documentation takes time, but it's just as important as good code.

ngeiswei:
@rTreutlein, maybe I'm influenced by @linas, but I'm bending toward closing that PR, and I would recommend to

1. comment-proof the whole thing
2. simplify the git history, there are about 35 commits scattered at 3 places, over a couple of months, while the code is mostly localized in the `distvalue` directory, which should make re-crafting the git history fairly painless.
3. and obviously take into considerations all other comments of that PR so far.

Other than that I'm think it's heading in the right direction.

opencog/atoms/distvalue/DistributionalValue.cc
```diff
@@ -0,0 +1,455 @@
...
+	}
+	return probs;
+}
+
+double DistributionalValue::get_mode_for(double ai) const
+{
+	return (ai - 1) / (total_count() - _value.size());
+}
+
+std::vector<double> DistributionalValue::get_mean() const
```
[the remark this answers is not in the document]
ngeiswei: Maybe `get_bin_means` or `get_interval_means` or `get_key_means` or something like that would be the correct name. Possible also without the `get_`.

linas:
I'm also thinking that a minimal pull, without the conditional stuff in it, would make it easier. When that looks good, then the conditional variant can be added.

PR #1959, opened 2018-12-19
Replace minimal PostgreSQL version by 9.5

vsbogd:
I have tried to build it with Postgre 9.3 and found that it doesn't work because this version doesn't support `NOT EXISTS` conditions.
So updating minimal version to 9.5.
Postgre 10 doesn't work as well as it doesn't support `increment_by`

### DIFF
--- a/opencog/persist/sql/README.md
+++ b/opencog/persist/sql/README.md
@@ -319,7 +319,7 @@ Each step discussed below.
 
 Database Install
 ----------------
-Download and install Postgres version 9.3 or newer.  The current design
+Download and install Postgres version 9.5.  The current design
 simply won't work with MySQL, because of a lack of array support.
 Same holds true for SQLite.  Sorry. There is some work in the
 code-base to support these other databases, but the work-arounds for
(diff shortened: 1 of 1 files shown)


opencog/persist/sql/README.md
```diff
@@ -319,7 +319,7 @@ Each step discussed below.
 
 Database Install
 ----------------
-Download and install Postgres version 9.3 or newer.  The current design
+Download and install Postgres version 9.5.  The current design
```
linas: Hmm. Can you make it 9.5 or newer? It does work with 9.6, and the change in version 10 seems easy to fix.

Any clue about circleci? Its failing with `Permission denied (publickey).`

vsbogd: Ok, added 'or newer'.

Regarding `circleci` I think it fails because I did create branch in the main repo not in the fork, but it is just hypothesis.

linas: Oh. Please don't create branches in the main repo...

linas: I mean, not for ordinary pull requests; create branches only for, aumm "major" things.

linas: Ah, OK. Not big a deal. I was afraid that this might be the start of a pull-req trend.

Issue #1888, opened 2018-10-25
PythonEvalUTest fails


noskill:
The test fails on imports from atomspace with error:
```

[2018-10-25 08:44:28:729] [INFO] PythonEval::initialize_python_objects_and_imports Finished initialising python evaluator.
[2018-10-25 08:44:28:729] [DEBUG] [PythonEval] get line:
from opencog.atomspace import *

[2018-10-25 08:44:28:729] [DEBUG] [PythonEval] eval_expr length=32:
from opencog.atomspace import *

[2018-10-25 08:44:28:729] [ERROR] Python error : <built-in method fileno of _io.FileIO object at 0x7f39c7c1e870> returned a result with an error set. (/home/noskill/projects/atom2/opencog/cython/PythonEval.cc:1141)
        Stack Trace:
        2: /usr/local/lib/libcogutil.so (opencog::Logger::logva(opencog::Logger::Level, char const*, __va_list_tag*) +0xd4) [0x7f39c7921d34]

```

Imports like "import opencog.atomspace" are fine

linas:
It has been passing for me, ever since #1887 turned testing back on. I think (I'm not sure) that I only build python3 by default; perhaps this is a python2 issue?

noskill:
PythonEvalUTest links to libPythonEval.so and  libPythonEval.so links to /usr/lib/x86_64-linux-gnu/libpython3.6m.so.1.0 on my machine.

linas:
Does it work from the command line? i.e.
```
$  python3
>>> from opencog.atomspace import *
>>> a = AtomSpace()
>>> TV = TruthValue(1, 0.8)
>>> A = a.add_node(types.ConceptNode, 'Apple', TV)
```

linas:
And does the stack trace really just end with `2: /usr/local/lib/libcogutil.so (opencog::Logger::logva` ? Normally, there should be a full stack trace, locating where the problem is. Is there a chance that libcogutil is not linked correctly, or that some old or stale version of cogutil is found and linked?

linas:
Ohh .. that's weird ... libcogutil is mis-installed; it should be in `/usr/local/lib/opencog/libcogutil.so` but is currently in `/usr/local/lib/libcogutil.so` ... wow. That could explain a LOT of bizarro bugs I've been hunting for .. years.

vsbogd:
What I did see on my machine is:

[...]

Issue #1835, opened 2018-07-17
Some C++ unit tests fails without ```make install```


vsbogd:
All tests below except VectorAPIUTest which suffers from separate issue https://github.com/opencog/atomspace/issues/1834

```
The following tests FAILED:
	 49 - MultiAtomSpaceUTest (Failed)
	 53 - VectorAPIUTest (Failed)
	 99 - NestedPutUTest (Failed)
	110 - BackwardChainerUTest (Failed)
```
Steps to reproduce:
```
TESTDIR="${HOME}"
mkdir -p "${TESTDIR}/prefix"
cd "${TESTDIR}"
git clone https://github.com/opencog/cogutil.git
mkdir cogutil-build
cd cogutil-build
cmake -DCMAKE_INSTALL_PREFIX="${TESTDIR}/prefix" ../cogutil
make install
cd "${TESTDIR}"
git clone https://github.com/opencog/atomspace.git
mkdir atomspace-build
cd atomspace-build
cmake -DCMAKE_INSTALL_PREFIX="${TESTDIR}/prefix" ../atomspace
make test
```

linas:
closing seems to be fixed.

PR #1960, opened 2018-12-19
Fix unit tests opencog

vsbogd:
Fix for #1835 

What is done:
- necessary libraries are added as dependencies
- opencog guile code is added as dependency for guile relating tests
- PythonEvalUTest is skipped because of  #1888

### DIFF
--- a/tests/scm/CMakeLists.txt
+++ b/tests/scm/CMakeLists.txt
@@ -31,4 +31,9 @@ ADD_GUILE_TEST(SCMOpenogTestRunnerFail scm-opencog-test-runner-fail.scm)
 # The SCMOpenogTestRunnerFail fails purposely, as it is testing that the
 # scheme opencog-test-runner exits properly; thus requiring the flipping
 # of the test result by setting its properties.
-SET_TESTS_PROPERTIES(SCMOpenogTestRunnerFail PROPERTIES WILL_FAIL TRUE)
+SET_PROPERTY(TEST SCMOpenogTestRunnerFail PROPERTY WILL_FAIL TRUE)
+
+# Add guile code to the load path to make it possible run tests before
+# executing "make install"
+SET_PROPERTY(TEST SCMOpenogTestRunnerPass SCMOpenogTestRunnerFail
+	PROPERTY ENVIRONMENT "GUILE_LOAD_PATH=${PROJECT_SOURCE_DIR}/opencog/scm")
--- a/tests/scm/CMakeLists.txt
+++ b/tests/scm/CMakeLists.txt
@@ -15,6 +15,7 @@ LINK_LIBRARIES(
 	smob
 	clearbox
 	execution
+	logger
 	${GUILE_LIBRARIES}
 )
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,17 +46,17 @@ jobs:
           name: Build
           command: cd build && make -j2
       - run:
-          name: Install AtomSpace
-          command: cd build && make -j2 install && ldconfig
-      - run:
-          name: Build examples
-          command: cd build && make -j2 examples
-      - run:
           name: Build tests
           command: cd build && make -j2 tests
       - run:
           name: Run tests
           command: cd build && make -j2 test ARGS=-j2
+      - run:
+          name: Install AtomSpace
+          command: cd build && make -j2 install && ldconfig
+      - run:
+          name: Build examples
+          command: cd build && make -j2 examples
       - run:
           name: Print test log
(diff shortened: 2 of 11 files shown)


tests/scm/CMakeLists.txt
```diff
@@ -30,4 +31,9 @@ ADD_GUILE_TEST(SCMOpenogTestRunnerFail scm-opencog-test-runner-fail.scm)
 # The SCMOpenogTestRunnerFail fails purposely, as it is testing that the
 # scheme opencog-test-runner exits properly; thus requiring the flipping
 # of the test result by setting its properties.
-SET_TESTS_PROPERTIES(SCMOpenogTestRunnerFail PROPERTIES WILL_FAIL TRUE)
+SET_PROPERTY(TEST SCMOpenogTestRunnerFail PROPERTY WILL_FAIL TRUE)
```
linas: Well, if `SCMOpenogTestRunnerPass` below needs `PROPERTY ENVIRONMENT "GUILE_LOAD_PATH=${PROJECT_SOURCE_DIR}/opencog/scm")` then so does `SCMOpenogTestRunnerFail` -- both should look the same, from this point of view

vsbogd: Yes, I have fixed it in later commit.

vsbogd:
> it also passes in circleci, (whatever env that is using)
In CircleCI Python tests are not running, I have already prepared change to turn them on, but didn't applied it because of this test.

This test fails for me and @noskill and @Necr0x0Der, and there is #1888 about it so I afraid of turning on tests in CircleCI before test is skipped. I can check its behavior in CircleCI docker and will revert the change if it is passed.

linas:
Stubbing out unit tests because they fail for some people, and because those people are too lazy to debug -- this is fundamentally wrong behavior. The unit tests pass just fine for at least three different kinds of configs; if you want to have #1888 debugged .. you have to make the effort. 

The problem is that when unit tests get stubbed out -- people forget. Then, half a year later when something does break, then the time and effort to fix the bug skyrockets. Its as if bugs have a very high interest rate: debug early, and they are cheap; debug later, they are very expensive.

linas:
Why TF are the python tests turned off in CircleCI? Who turned them off? Why do we keep shooting ourselves in the foot?

vsbogd:
Actually I have spent some time in debugging of #1888, it is not simple for me, it segfaults at some  point in `PyImport_Import` call to import `opencog.atomspace` cython module. For me reason is still not clear.

I have checked that this test works in CI, so I will turn it on back. 

They are not explicitly disabled but as current `opencog-deps` doesn't have `nosetests` tool they are not started. I have raised PR https://github.com/opencog/ocpkg/pull/128 to fix it.

linas:
I asked several questions on #1888 that got no reply... I'm hoping gdb should be enough to figure it out ... 

The autogen'ed code that calls PyImport_Import is in the build directory .. you can hand-edit it, put a `printf` in it, and make sure that the correct arguments are being passed in.

vsbogd:
Added my findings to https://github.com/opencog/atomspace/issues/1888#issuecomment-448735896

tests/scm/CMakeLists.txt
linas: oh duh kant read

linas:
Hmm. Well, OK to merge now, I guess, except `GearmanUTest` fails in circleci with `Unbound variable: start-work-handler`  I've seen stuff like this in the past, with other unit tests .. it always had something crazy to do with `liblogger` .. is that why youd added `logger` to some of the CMakes?

tests/scm/CMakeLists.txt
```diff
@@ -30,4 +31,9 @@ ADD_GUILE_TEST(SCMOpenogTestRunnerFail scm-opencog-test-runner-fail.scm)
...
+
+# Add guile code to the load path to make it possible run tests before
+# executing "make install"
+SET_PROPERTY(TEST SCMOpenogTestRunnerPass SCMOpenogTestRunnerFail
+	PROPERTY ENVIRONMENT "GUILE_LOAD_PATH=${PROJECT_SOURCE_DIR}/opencog/scm")
```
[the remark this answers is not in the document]
linas: Hey, @AmeBel you (or someone else?) recently (less than a week ago?) changed the install paths, so that they no longer install into `${PROJECT_BINARY_DIR}/opencog/scm`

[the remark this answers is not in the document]
AmeBel: The issues mentioned are fixed by https://github.com/opencog/atomspace/pull/1964

linas:
Thank you!

PR #1966, opened 2018-12-22
Attempt to avoid minor inefficiencies

linas:
There's a functional change in here, too. It doesn't affect the unit tests, but .. it still smells fishy.

### DIFF
--- a/opencog/rule-engine/Rule.cc
+++ b/opencog/rule-engine/Rule.cc
@@ -552,7 +552,7 @@ Handle Rule::standardize_helper(AtomSpace* as, const Handle& h,
 		for (auto ho : old_outgoing)
 			new_outgoing.push_back(standardize_helper(as, ho, dict));
 
-		Handle hcpy(as->add_atom(createLink(new_outgoing, h->get_type())));
+		Handle hcpy(as->add_link(h->get_type(), new_outgoing));
 		hcpy->copyValues(h);
 		return hcpy;
 	}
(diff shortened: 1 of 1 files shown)


opencog/rule-engine/Rule.cc
```diff
@@ -237,10 +237,13 @@ void Rule::add(AtomSpace& as)
...
-	HandleSeq outgoings;
-	for (const Handle& h : _rule->getOutgoingSet())
-		outgoings.push_back(as.add_atom(h));
-	_rule = createBindLink(outgoings);
+	// XXX FIXME ... this seems wrong ... why would we EVER create
+	// a BindLink, and NOT put it in the atomspace?  The correct
+	// fix would seem to be to say
+	// _rule = BindLinkCast(as.add_link(BIND_LINK, _rule->getOutgoingSet()));
+	// but doing this causes BackwardChainerUTest to consistently crash.
+	// This sure smells like a bug, somewhere, to me.
```
ngeiswei: That's because the rule is then used to expand an inference tree. That particular rule has been alpha-converted to guaranty that variable names do not collide with the variable names of the inference tree to expand. Adding the rule in an atomspace may undo the alpha-conversion. See the comment of Rule::add in the header file https://github.com/opencog/atomspace/blob/master/opencog/rule-engine/Rule.h#L152-L172

ngeiswei: I think the correct fix would be to alpha-convert during expanding, on the other hand it works at it is so...

ngeiswei:
I'm merging. I'll rework/complete the FIXME comment when I get a chance.

opencog/rule-engine/Rule.cc
ngeiswei: Now I remember why I don't do the alpha-conversion during expansion. Unification also requires alpha-conversion. I have to carry out unification to check rule validity (if 2 jigsaw puzzle pieces match, one jigsaw piece = inference tree premise or conclusion, another jigsaw piece = rule premise or conclusion, a rule conclusion must match an inference tree premise and such). So after unification I already have an alpha-converted rule that I can pass to the inference tree expansion procedure, by not inserting it to the atomspace I avoid the cost of redoing the alpha-conversion.

linas: Ohhhhhh  . I think I get it.  I see why. I will have to think about alternative approaches, and if they would be better, the same or worse.

PR #1969, opened 2018-12-24
make function for error checking in PythonEval

noskill:
Adding new error check with some refactoring. Adding this new check since it gives usefull error message - see https://github.com/opencog/atomspace/issues/1888#issuecomment-449730655

### DIFF
--- a/opencog/cython/PythonEval.cc
+++ b/opencog/cython/PythonEval.cc
@@ -26,6 +26,7 @@
 
 #include <dlfcn.h>
 #include <unistd.h>
+#include <memory>
 
 #include <boost/filesystem/operations.hpp>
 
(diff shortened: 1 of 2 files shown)


opencog/cython/PythonEval.cc
```diff
@@ -1175,38 +1176,36 @@ void PythonEval::apply_as(const std::string& moduleFunction,
...
+
 std::string PythonEval::apply_script(const std::string& script)
 {
     std::lock_guard<std::recursive_mutex> lck(_mtx);
 
     // Grab the GIL
     PyGILState_STATE gstate = PyGILState_Ensure();
-
+    std::shared_ptr<void> _(nullptr, [&gstate](void * ptr){
+                                   // Release the GIL. No Python API allowed beyond this point.
```
linas: I don't recall, but maybe `std::unique_ptr` is faster/slimmer than `std::shared_ptr` ? Although it really pretty much doesn't matter here. 

Also, using a single underscore as a variable name is really, um, well, I get what you are doing, but this obfuscates the code. Call it `raii` because that is what it is, and future developers will read this and sai "Oh I know what RAII is" instead of puzzling over "what the heck is he trying to do here"?

I'm going to merge this anyway, but just thought I'd mention this because I'm nit-picky.

linas:
Also there are some whitespace issues. These should be fixed...

linas:
Thanks, BTW.

opencog/cython/PythonEval.cc
vsbogd: You can also use `BOOST_SCOPE_EXIT`, see https://theboostcpplibraries.com/boost.scopeexit

noskill: @linas Yea, i tried unique_ptr first, but it doesn't work. It turns out that according to standard shared_ptr should call destructor callback on null pointer and unique_ptr shouldn't. 
Updated with typedef for better clarity. Btw what is the issue with whitespaces??

noskill: Ok, replaced shared_ptr with boost' scope guard. It seems it doesn't require adding new dependencies.

linas:
I just merged this. I'm now going to add a comment about what that boost thing does, since its relatively magic .. and any "magic" in the code should always be explained...

PR #1982, opened 2019-01-09
Include given atom headers file name instead of hardcoded one

vsbogd:
Allow custom header file name to be used in `OPENCOG_ADD_ATOM_TYPES` macro call.


linas:
I'm trying to merge this, but github is not working despite my logging out and back in on two different browsers. Merge whenever.

Issue #1965, opened 2018-12-22
Question about atom copying procedure


vsbogd:
`AtomTable` uses `createNode()` function to copy node. 
https://github.com/opencog/atomspace/blob/e47dc46daa278c22cf3ba0c503e12b0e1e9b025d/opencog/atomspace/AtomTable.cc#L394

`createNode` uses constructor of `Node` class to create `Node` instance and then calls the `classserver().factory(tmp->get_handle())` to create the new instance of proper class. 
https://github.com/opencog/atomspace/blob/e47dc46daa278c22cf3ba0c503e12b0e1e9b025d/opencog/atoms/base/Node.h#L128-L134

Such way of copying has two disadvantages:
- temporary node is created though it will not be used further;
- as Node copies only type and name of the copied node all other information which node can potentially contain will be lost.
The same is true for link copying procedure.

I would like to discuss and implement another way of copying, which allows using copying constructor of method which is specific for original atom class if it is different from Node or Link.
One of the simplest and safe ways is to add another method for copying existing atom and return it from `ClassServer` (the way as `factory` method works).

linas:
Oh, huh. Sure. Wouldn't the easiest fix just be to call the factory directly, in `AtomTable.cc` line 394?  "Other information" is not lost; it's copied at line 422.  I'm tempted to try that right now, to see what kind of impact it makes on performance.

Please note the following:

* In "principle", one is not supposed to copy atoms: the original idea is that they are unique: there's only one of a given atom in the atomspace.  This assumption does get violated in a few places, but I don't want to abandon it.

* AtomTable.cc line 389-390 says "We MUST maintain our own private copy", but perhaps this is no longer needed.  Maybe we can skip copying, entirely. In the past, there were unit tests that crashed ... !?? Why?  Why is a private copy needed?   Maybe the unit tests are badly designed?

* In order to help enforce the idea of a single, globally unique atom, the AtomTable API originally did not have methods that took `const Handle&`.  The original methods were `getHandle(Type, const std::string&)` and `add(Type, const std::string&)`.  Should we return to this API? Or maybe the opposite: should we get rid of it? 

* Why do we need both `AtomPtr` and `Handle` anyway? Every time I think of getting rid of one or the other, I get tired and confused and lazy and give up.  Maybe there is a valid reason for having both, but I can't remember what it is, and its not obvious.

[...]

PR #1986, opened 2019-01-11
Export get value ptr cython function

vsbogd:
Adding `get_value_ptr` to the `atomspace.pxd` in order to export it from Cython `atomspace.so` library and reuse it to implement custom `Value` wrappers.

### DIFF
--- a/opencog/cython/opencog/atomspace.pxd
+++ b/opencog/cython/opencog/atomspace.pxd
@@ -88,12 +88,12 @@ cdef extern from "opencog/atoms/value/Value.h" namespace "opencog":
         bint is_atom()
         bint is_node()
         bint is_link()
-        
+
         string to_string()
         string to_short_string()
         bint operator==(const cValue&)
         bint operator!=(const cValue&)
-    
+
     ctypedef shared_ptr[cValue] cValuePtr "opencog::ValuePtr"
 
 cdef class Value:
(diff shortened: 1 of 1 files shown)


vsbogd:
Current Python API implementation uses `get_value_ptr` in any `Value` implementation to get original C++ `Value`. The reason of exporting is I would like to start forming API which allows creating Python API for values which are introduced outside of `opencog/atomspace` repo.

I am not going to sneak `PtrValue` into the `atomspace` repo but as I said [here](https://github.com/opencog/atomspace/issues/1965#issuecomment-453472987) I would like to be able experimenting with any strange idea just to understand its downsides.

linas:
Due to a design flaw/limitation, you cannot define new Value types outside of the atomspace.  I think there is only one trouble-spot, not sure. 

Note in in `atom_types.script` explain it:
```
// ====================================================================
// Octomap related. The Octomap code is (currently) in the opencog repo.
// We need to define these here, until SchemeSmob::ss_new_value() gets
// converted into some kind of factory. Once we get a factory, these
// should be moved to live with the octomap code.
OCTO_VALUE <- FLOAT_VALUE
``` 
You can either try to fix this, or hack around it.

linas:
Maybe if you don't use scheme, this is not a problem.

vsbogd:
Right now I am using only Python API, but yes I think if I need Scheme API I will fix it.

PR #1989, opened 2019-01-17
Rename FloatValue to FloatSeqValue and such

ngeiswei:
I went ahead and renamed

1. `FloatValue` -> `FloatSeqValue`
2. `StringValue` -> `StringSeqValue`
3. `LinkValue` -> `SeqValue`

I wanted to see if it would open a can of worms or not, but the process was fairly painless.


ngeiswei:
This is too possible address issue #1880

PR #1990, opened 2019-01-17
 Overload operator<< for Incoming set

yitbrek:
Overloading operator<< for Incoming Set. As per https://github.com/singnet/as-moses/issues/39#issuecomment-451862474

### DIFF
--- a/opencog/atoms/base/Atom.cc
+++ b/opencog/atoms/base/Atom.cc
@@ -78,6 +78,12 @@ equal_to<opencog::WinkPtr>::operator()(const opencog::WinkPtr& lw,
     return hl == hr;
 }
 
+// Overloading operator<< for Incoming Set
+ostream& operator<<(ostream& out, const opencog::IncomingSet& iset)
+{
+ return out << opencog::oc_to_string(iset);
+}
+
 } // namespace std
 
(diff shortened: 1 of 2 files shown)


opencog/atoms/base/Atom.cc
```diff
@@ -78,6 +78,12 @@ equal_to<opencog::WinkPtr>::operator()(const opencog::WinkPtr& lw,
     return hl == hr;
 }
 
+// Overloading operator<< for Incoming Set
+ostream& operator<<(ostream& out, const opencog::IncomingSet& iset)
+{
+ return out << opencog::oc_to_string(iset);
```
ngeiswei: The indentation requires 4 whitespaces.

ngeiswei:
@yitbrek, as an alternative to closing the PR and create a new one, note that you also can push commits on top of the existing ones, github will be able to take them into account. You are not happy of some existing commits and would rather replace them, you may even force-push commits and github will still be able take them into account.

ngeiswei:
Ah, I saw that you actually updated, I don't understand why you closed then?

ngeiswei:
OK, I just saw you created a new PR. ;-)

PR #1996, opened 2019-01-19
Return ValuePtr from EvaluationLink::do_evaluate()

vsbogd:
Return `ValuePtr` from `EvaluationLink` methods.

What is done:
- Return `ValuePtr` from `EvaluationLink::do_evaluate(AtomSpace*, const Handle&, bool silent=false)` and `EvaluationLink::do_eval_scratch` methods
- Remove not used `EvaluationLink::do_evaluate(AtomSpace*, const HandleSeq& schema_and_args, bool silent=false)` method
- Rename `EvaluationLink::do_evaluate(AtomSpace*, const Handle& schema, const Handle& args, bool silent=false);` to `EvaluationLink::do_evaluate_evaluation_link`
- Fix Scheme, Python and Haskell bindings to return `Value`
- Fix Python bindings to compare `TruthValue` and `Value` properly
- Fix PattenMatcher code to cast `ValuePtr` to `TruthValuePtr` with checking type


vsbogd:
@linas, I would like to know your opinion about this change before looking into broken opencog unit test.

linas:
I'll look at this "real soon now". I'm not convinced that this change is this easy -- it requires rethinking a lot of the demos and examples, and changing explanations of how cog-evaluate works.  Maybe this is a good start, but the bigger, harder part of this is getting all of the documentation and examples revise and refined.

I'm working on an overhaul of examples right now, many of them are celebrating 10-year anniversaries, and it shows. 

Also, why aren't unit tests passing?

linas:
Note also: something else to think about: Traditionally, the `cog-execute` function always returned atoms, the `cog-evaluate` always returned truth values.  Perhaps the change should be the other way around: have `cog-execute` return values in general, and leave cog-evaluate alone.   Or maybe merge `cog-execute!` and `cog-evaluate!` so that they are the same function. 

The `ValueOfLink` was one early attempt to try to untangle this. It should continue to work "as expected" after this change.

vsbogd:
Thing I don't like in this change is casting. Two kinds of casting:

(1) in `EvaluationLink.cc` it is required to cast `TruthValuePtr` to `ValuePtr` because `TruthValuePtr` is `shared_ptr<const ...>` and `ValuePtr` is not const.

(2) in PatternMatcher and unit tests one need to cast `ValuePtr` to expected value, can be probably avoided using type parameterized `do_evaluate` but doing it doesn't seems right without having type parameterized atoms or something, probably casting is a best way.

vsbogd:
`atomspace` unit tests are passed without any change of logic. Broken one is in `opencog`. I will look at it.

linas:
Hmm. Well, I skimmed it. You did the "obvious" thing of just casting everything to ValuePtr... which we could have done many years ago. That's not hard to do. That's easy. (Well there is one technical issue: calling ValueCast, or calling any FooBarCast, takes hundreds of CPU cycles, and so this change will create some kind of slowdown. I wish there was some way of working with smart-pointers without this huge penalty, but I haven't found it yet. It's not your fault; it's just smart pointers. That's why other languages garbage-collect; smart-pointers kind-of suck.) 

The real issue with converting from TruthValue to generic Value is: how will the users use it? Answering this question is the hard part. Say that I've got some FloatValue here or there, and I want to pass it to somewhere else, and there's this StreamValue here, containing a video/audio feed, and I want to connect it to this and that. How does it work? Do I call cog-evaluate on it? Or cog-execute?  I think that  cog-execute! already handles values "correctly" (whatever that means), and so maybe the question is: do we really even need `cog-evaluate!` at all? Or can we rip it out?  (Seriously .. can we? Do we really need it? Except for examples and demos .. who uses it?)

A related question is: there are lots and lots of places where the pattern matcher is using TruthValues, when all it needs are just crisp boolean yes/no answers. And so maybe the correct fix is to keep `cog-evaluate`, but change it to use crisp boolean c++ yes/no values. Let `cog-execute!` handle TruthValues.  

The last suggestion might be the best: it replaces the slow and awkward TruthValue with fast std::bool  in all the places that matter, and then uses `cog-execute!` to handle everything else.  Would that work, and how hard would that be?

linas:
I just wrote up #2004 as a proposal that I think is fairly elegant, solves at least three different technical issues, and improves performance to boot. Of course, I wrote it on an empty stomach, so maybe tomorrow it will look less keen. But right now, I'm pretty excited about it. It would be easy to code, it would take 3 or 4 small, modular pull reqs to make it work.

So the question is: does #2004 fail in some way? Did it overlook something?  Forget about something?

vsbogd:
Closing it for now till end of the issue #2004 discussion

PR #1998, opened 2019-01-20
Add missing type constructors

linas:
Oddly, no one noticed these were missing. But the demos and examples need them.


linas:
MinerUTest is failing. wtf.

linas:
Merging despite failure of MinerUTest.  I find it hard to believe that this pull req is causing MinerUTest to fail.

Issue #1993, opened 2019-01-18
ExecutionOutputLink child class is not called


vsbogd:
I would like to make an extension of `ExecutionOutputLink` by defining new child class `VsbogdExecutionOutputLink` and overriding `ExecutionOutputLink::execute(AtomSpace*,bool)` method.

It doesn't work as `Instantiator::walk_tree(Handle, bool)` method checks that atom type is equal to `EXECUTION_OUTPUT_LINK`: https://github.com/opencog/atomspace/blob/062469366854bc856dac9c914f030d6ad0461337/opencog/atoms/execution/Instantiator.cc#L346

Is it made by intention or `==` can be replaced by `isA` here?

linas:
Yes, this should be an `classserver().isA`

linas:
I mean `nameserver().isA`

PR #2003, opened 2019-01-21
Allow adding children of execution output link

vsbogd:
Fix for issue #1993 to allow adding children of `ExecutionOutputLink` which extends its functionality.


linas:
This fixes #1993

PR #2011, opened 2019-01-23
Cython api refactoring

vsbogd:
Atomspace Python API refactoring, no changes in functionality.
- handles are passed to Atom constructor without conversion to `long` and back
- Value constructor is modified to allow inheritance
- separate value classes are added to make code more modular


opencog/cython/opencog/value.pyx
```diff
@@ -1,60 +1,47 @@
...
+        return Value(ValuePtr.create(shared_ptr))
+
+    def __init__(self, value_ptr):
+        self.value_ptr = value_ptr
+
+    cdef cValuePtr get_c_value_ptr(self):
+        """Return C++ ValuePtr instance"""
+        return self.value_ptr.shared_ptr
+
+    cdef cValue* get_c_value(self):
```
noskill: Names are confusing: get_c_value_ptr returns shared pointer and get_c_value returns raw pointer, which is pointer type too. Clear names are get_shared_ptr and get_raw_ptr. But it seems shared pointer should be enough for most cases.

noskill: or just get_ptr and get_raw_ptr

vsbogd: I have removed `get_c_value` method and replaced it by `get_c_value_ptr` calls

ngeiswei:
@vsbogd , I don't feel confident enough to merge it, so either you have to wait for @linas, or if you feel confident enough, you can merge it yourself.

vsbogd:
@linas, I am not sure if you are going to go through it as it touches only Python API and doesn't contain any functionality changes?

linas:
I skimmed this quickly, it seems quite reasonable. The python bindings have long-needed someone willing to devote the time to clean them up, so, yes, thank you for doing this.

There is only one potential land-mine lurking in around in some uncertain location: the cogserver atomspace. In the past, the atomspace that the cogserver used was not always the same one that python used (which was not always the same one as the one scheme used.) Tracing down the root cause of this was hard. The void* pointer cast that this pull req removed was a part of the attempt to get python and the cogserver to agree to use just one atomspace.  I am not sure if there is a unit test for this, or not. There might be; it might not work well.

Since that time, there has been another change that might make things confusing: one can now intentionally create multiple atomspaces. In the scheme bindings, you can use different atomspaces in different threads. (recall, the scheme bindings use the concept of "the implicit current atomspace"; every scheme thread can have it's own "current" atomspace). This potentially multiplies confusion: if the scheme code in one thread calls GroundedPredicate "py:foobar", then the py function should use whatever atomspace is current in that thread (and not current in py or the cogserver).  I doubt there's an explicit unit test for this, although some unit tests might implicitly assume this.

vsbogd:
Thank you for explaining it. I have tried to find unit tests which demonstrate `CogServer` and `PythonEval` integration. Found three candidates: [ShellUTest.cxxtest](https://github.com/opencog/opencog/blob/master/tests/server/ShellUTest.cxxtest), [PyEvalUTest.cxxtest](https://github.com/opencog/opencog/blob/master/tests/cython/PyEvalUTest.cxxtest), [PythonModuleUTest.cxxtest](https://github.com/opencog/opencog/blob/master/tests/cython/PythonModuleUTest.cxxtest), but they doesn't include real python code execution examples. It makes sense to add such tests if we touch this integration. In this PR I didn't touch this functionality as I just replaced passing pointer to `Handle` as `long` by passing reference to `Handle` directly.

Also I see that `PythonEval` uses its own global `ATOMSPACE` definition to work which is different from global `atomspace` variable which is used by `Cython` type constructors.

linas:
Wandering off-topic a bit: the contents of the opencog git repo is an ad hoc collection of loosely-related things. and should probably be split up into distinct projects/repos. (you once did a clean split of the benchmark code; I'm vaguely interested in the script-fu you did to do that; I've never been able to get as clean as that.)

The cogserver is mashup of three unrelated things; it too should be split up.  It is
* A network shell server. Like telnet, but different. 
* A command shell. Like bash, but different. With "pluggable modules" (e.g. PythonModule)
* A cooperative-multi-tasking framework, called "agents".

These really need to be split into distinct parts. Two of the three are hokey; the multi-tasking framework is outright lame. The command-shell is kind of goofy, like somebody's school project. The network server is the most debugged and mostly quite stable. I would have recommended discarding it, except I once searched the internet for a drop-in replacement, and found none.  I can kind-of see why all three are interesting, but they really need to be reimagined from the ground-up.

I should open a distinct issue to discuss all this.

linas:
OK so https://github.com/opencog/opencog/issues/3391 exists. @vsbogd could you post the scripts that you used to split out the benchmark code? (at that issue?)

linas:
And this one https://github.com/opencog/opencog/issues/3414 talks about splitting up the cogserver.

For adding new cross-subsystem atomspace unit tests .. just create new tests, as needed; don't try to modify existing unit tests.

linas:
> Also I see that `PythonEval` uses its own global `ATOMSPACE` definition

Yeah. Cause why not? its presumably harmless, until some unsuspecting developer trips over it.  Right now, leaving things alone, as they are, is not a bad policy. Once you start to try to fix things, it will be like pulling on a thread from a knit sweater.

linas:
I think this implements #1161

Issue #1994, opened 2019-01-18
remove check for arity of python callbacks


noskill:
There is method that checks that arity of callback in GroundedSchemaNode is the same as number of arguments defined in ExecutionOutputLink:
https://github.com/opencog/atomspace/blob/062469366854bc856dac9c914f030d6ad0461337/opencog/cython/PythonEval.cc#L702 It returns incorrect argument count for python functions with variable number of arguments, like

```
def test(a, b, *args):
      pass
```

It is possible correctly determine that argument count is variable with **inspect** module. On the other hand why not wait for error from python interpreter in case of argument count mismatch. That would cause RuntimeException to be thrown just like in case of argument count mismatch.
So should we remove this check?

vsbogd:
I agree with @noskill suggestion, it is not clear why we are checking arity manually instead of delegating it to the Python interpreter.

linas:
Yes, removing this seems like a reasonable suggestion, if you are certain that python is also performing this check.  Are you certain? If python is not performing this check, then removing this turns an error into a silent error. 

Silent errors are notoriously hard to debug: someone young and inexperienced, or maybe tired and not paying attention, accidentally passes two arguments to a function that expects only one. If no exception is thrown, everything seems fine, at first ... it can take hours to figure out why its not working as hoped. Silent errors are deadly.

It would be great if there was a unit test added, which pases one argument too many, and one too few, and checks for the exception.

linas:
This was #2012

PR #2012, opened 2019-01-23
remove arity check of python callbacks

noskill:
see issue https://github.com/opencog/atomspace/issues/1994

### DIFF
--- a/opencog/cython/PythonEval.cc
+++ b/opencog/cython/PythonEval.cc
@@ -699,33 +699,6 @@ void PythonEval::execute_string(const char* command)
     if (f) fsync(PyObject_AsFileDescriptor(f));  // Force a flush
 }
 
-int PythonEval::argument_count(PyObject* pyFunction)
-{
-    PyObject* pyFunctionCode;
-    PyObject* pyArgumentCount;
-    int argumentCount;
-
-    // Get the 'function.func_code.co_argcount' Python internal attribute.
-    pyFunctionCode = PyObject_GetAttrString(pyFunction, "__code__");
-    if (pyFunctionCode) {
-        pyArgumentCount = PyObject_GetAttrString(pyFunctionCode, "co_argcount");
-        if (pyArgumentCount) {
-            argumentCount = PyLong_AsLong(pyArgumentCount);
-        }  else {
-            Py_DECREF(pyFunctionCode);
-            return MISSING_FUNC_CODE;
-        }
-    } else {
-        return MISSING_FUNC_CODE;
-    }
-
-    // Cleanup the reference counts.
-    Py_DECREF(pyFunctionCode);
-    Py_DECREF(pyArgumentCount);
-
-    return argumentCount;
-}
-
 /**
  * Find the Python object by its name in the given module'.
  */
@@ -850,9 +823,10 @@ PyObject* PythonEval::call_user_function(const std::string& moduleFunction,
     if (!pyUserFunc) {
         if(pyObject) Py_DECREF(pyObject);
         PyGILState_Release(gstate);
+        const char * moduleName = PyModule_GetName(pyModule);
         throw RuntimeException(TRACE_INFO,
-            "Python function '%s' not found!",
-            moduleFunction.c_str());
+            "Python function '%s' not found! in module '%s'",
+            moduleFunction.c_str(), moduleName);
     }
 
     // Promote the borrowed reference for pyUserFunc since it will
@@ -869,16 +843,6 @@ PyObject* PythonEval::call_user_function(const std::string& moduleFunction,
             "Python function '%s' not callable!", moduleFunction.c_str());
     }
 
-    // Get the expected argument count.
-    int expectedArgumentCount = this->argument_count(pyUserFunc);
-    if (expectedArgumentCount == MISSING_FUNC_CODE) {
-        Py_DECREF(pyUserFunc);
-        PyGILState_Release(gstate);
-        throw RuntimeException(TRACE_INFO,
-            "Python function '%s' error missing 'func_code'!",
-            moduleFunction.c_str());
-    }
-
     // Get the actual argument count, passed in the ListLink.
     if (arguments->get_type() != LIST_LINK) {
         Py_DECREF(pyUserFunc);
@@ -886,20 +850,7 @@ PyObject* PythonEval::call_user_function(const std::string& moduleFunction,
         throw RuntimeException(TRACE_INFO,
             "Expecting arguments to be a ListLink!");
     }
-
-    // Now make sure the expected count matches the actual argument count.
     int actualArgumentCount = arguments->get_arity();
-    // If this is really a method, it has 'self' argument, which is passed automatically
-    if (PyMethod_Check(pyUserFunc)) expectedArgumentCount--;
-    if (expectedArgumentCount != actualArgumentCount) {
-        Py_DECREF(pyUserFunc);
-        PyGILState_Release(gstate);
-        throw RuntimeException(TRACE_INFO,
-            "Python function '%s' which expects '%d arguments,"
-            " called with %d arguments!", moduleFunction.c_str(),
-            expectedArgumentCount, actualArgumentCount);
-    }
-
     // Create the Python tuple for the function call with python
     // atoms for each of the atoms in the link arguments.
     PyObject* pyArguments = PyTuple_New(actualArgumentCount);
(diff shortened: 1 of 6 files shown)


opencog/cython/PythonEval.cc
```diff
@@ -850,9 +823,10 @@ PyObject* PythonEval::call_user_function(const std::string& moduleFunction,
     if (!pyUserFunc) {
         if(pyObject) Py_DECREF(pyObject);
         PyGILState_Release(gstate);
+        const char * moduleName = PyModule_GetName(pyModule);
         throw RuntimeException(TRACE_INFO,
-            "Python function '%s' not found!",
-            moduleFunction.c_str());
+            "Python function '%s' not found! in module '%s'",
```
vsbogd:
```suggestion
            "Python function '%s' not found in module '%s'!",
```

tests/cython/atomspace/test_callback_arity.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def tearDown(self):
+        finalize_opencog()
+        del self.space
+
+    def test_correct_argcount(self):
+        atom1 = ConceptNode("atom1")
+        exec_link = ExecutionOutputLink(GroundedSchemaNode("py:return_concept"),
+                                        ListLink(atom1))
+        result = execute_atom(self.space, exec_link)
```
vsbogd: I would check that returned value is equal to expected

tests/cython/atomspace/test_callback_arity.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def test_too_many_args(self):
+        atom1 = ConceptNode("atom1")
+        exec_link = ExecutionOutputLink(GroundedSchemaNode("py:return_concept"),
+                                        ListLink(atom1, atom1))
+        try:
+           result = execute_atom(self.space, exec_link)
+           self.assertFalse("call should fail")
+        except RuntimeError:
+           pass
```
vsbogd: I would check that exception message here and in next test is equal to expected.

tests/cython/atomspace/test_callback_arity.py
```diff
@@ -0,0 +1,58 @@
...
+        atom1 = ConceptNode("atom1")
+        exec_link = ExecutionOutputLink(GroundedSchemaNode("py:return_concept"),
+                                        ListLink())
+        try:
+           result = execute_atom(self.space, exec_link)
+           self.assertFalse("call should fail")
+        except RuntimeError:
+           pass
+
+    def test_incorrect_atomspace_init(self):
```
vsbogd: Looks like this test doesn't call `PythonEval::apply_as`, looks like Scheme call `python-call-with-as` should trigger the method.

opencog/cython/opencog/bindlink.pxd
```diff
@@ -7,7 +7,7 @@ cdef extern from "opencog/cython/opencog/BindlinkStub.h" namespace "opencog":
     #   Handle stub_bindlink(AtomSpace*, Handle);
     #
     cdef cHandle c_stub_bindlink "stub_bindlink" (cAtomSpace*, cHandle)
-    cdef cHandle c_execute_atom "do_execute"(cAtomSpace*, cHandle)
+    cdef cHandle c_execute_atom "do_execute"(cAtomSpace*, cHandle) except +
```
linas: Does python really need this? In the old days, C++ used to allow you to say "this function throws an exception", and eventually realized this was a bad idea; they mostly ripped it all out. So I'm a little surprised to see the above.

tests/scm/scm-test-arity.scm
```diff
@@ -0,0 +1,43 @@
...
+(python-eval "
+from opencog.atomspace import AtomSpace, TruthValue
+from opencog.atomspace import types
+def foo(atspace):
+    TV = TruthValue(0.2, 0.69)
+    atspace.add_node(types.ConceptNode, 'Apple', TV)
+")
+
+(python-call-with-as "foo" (cog-atomspace))
+(cog-node 'ConceptNode "Apple")
```
linas: this line is a no-op. `(cog-node 'ConceptNode "Apple")` returns `(ConceptNode "Apple")` if it exists, else it returns `#f`.  One uses `cog-node` only on those rare occasions where you want to test to see if an atom exists already, without creating it.

Note also that an autogenerated header file defines `(ConceptNode "Apple")`  to be `(cog-new-node 'ConceptNode "Apple")`   (more or less, its slightly more complicated than that, but effectively only `cog-new-node` actually creates new nodes.)

tests/scm/scm-test-arity.scm
```diff
@@ -0,0 +1,43 @@
...
+
+; Return the strength of a simple truth value
+(define (get-tv-strength tv) (cdr (assoc 'mean (cog-tv->alist tv))))
```
linas: this is harmless, but not needed. The built-in `cog-tv-mean` does the same thing. I do not know why it is sometimes called "mean" and sometimes called "strength".

in case you did not know typing `,a` and `,d` at the guiel prompt brings up documentation `,a` is the same as `,apropos` and `,d` is the same as `describe`. So:
```
 $ guile
Enter `,help' for help.
scheme@(guile-user)> (use-modules (opencog))
scheme@(guile-user)> ,a cog-tv
(opencog): cog-tv-mean	#<procedure cog-tv-mean (_)>
(opencog): cog-tv-merge-hi-conf	#<procedure cog-tv-merge-hi-conf (_ _)>
(opencog): cog-tv->alist	#<procedure cog-tv->alist (_)>
(opencog): cog-tv-count	#<procedure cog-tv-count (_)>
(opencog): cog-tv	#<procedure cog-tv (_)>
(opencog): cog-tv?	#<procedure cog-tv? (_)>
(opencog): cog-tv-confidence	#<procedure cog-tv-confidence (_)>
(opencog): cog-tv-merge	#<procedure cog-tv-merge (_ _)>
scheme@(guile-user)> ,d cog-tv-mean

 cog-tv-mean TV
    Return the `mean` of the TruthValue TV. This is a single
    floating point-number.

scheme@(guile-user)> 
```

tests/scm/scm-test-arity.scm
```diff
@@ -0,0 +1,43 @@
...
+
+(define strength (get-tv-strength
+ (cog-tv (cog-node 'ConceptNode "Apple"))))
```
linas: ... and you can skip even one more step:
```
scheme@(guile-user)> ,d cog-mean

 cog-mean ATOM
    Return the `mean` of the TruthValue on ATOM. This is a single
    floating point-number.
```

tests/scm/scm-test-arity.scm
```diff
@@ -0,0 +1,43 @@
...
+  (catch #t 
+    thunk
+    (lambda (key . parameters)
+      (format (current-error-port)
+              "caught throw to '~a: ~a\n" key parameters)
+      "catch")))
+
+(define failed-result
+  (catch-wrong-args
+   (lambda () (python-call-with-as "foo" (cog-atomspace) (cog-node 'ConceptNode "Test")))))
```
linas: I don't think this does what you think it does. Because you said `cog-node` instead of `cog-new-node`, the `(Concept "test")` was never created. Instead, you are passing `#f` to "foo".  I'm not sure what happens if you pass `#f` instead of passing an actual Atom. Maybe some other exception gets thrown, but not the wrong-arguments exception??

tests/scm/scm-test-arity.scm
linas: You might want ti say `(test-assert "atom was created" (not (eq? #f (cog-node 'ConceptNode "Apple"))))` to verify that the atom got created by the python code.

opencog/cython/opencog/bindlink.pxd
linas: "eventually realized" -- it was the C++ standards committee realized it was a bad idea. (and not me)

noskill: Checked exceptions were poorly implemented in c++, but the idea is not bad by itself. *except +* causes cython to wrap c++ exception with PyErr_SetString call, so it is not c++ throws keyword. Without it throwing an exception from c++ causes SIGABRT

linas: OK

tests/scm/scm-test-arity.scm
```diff
@@ -0,0 +1,40 @@
...
+    TV = TruthValue(0.2, 0.69)
+    atspace.add_node(types.ConceptNode, 'Apple', TV)
+")
+
+(python-call-with-as "foo" (cog-atomspace))
+(test-assert "atom was created" (not (eq? #f (cog-node 'ConceptNode "Apple"))))
+
+(define strength (cog-mean (cog-node 'ConceptNode "Apple")))
+
+(test-assert "strength value is wrong" (< (- 0.2 strength) 0.00001))
```
linas: This will fail to actually catch the error: if something screws up, the default TV on Apple is TRUE_TV, which has strength 1 and so `(< (- 0.2 strength) 0.00001))` will be true, even though the the strength is wrong.  You really do want the "abs" for absolute value, in there.

tests/scm/scm-test-arity.scm
linas: This remains unfixed.

linas:
I'll just merge this now, and fix up the remaining issues by hand. Hopefully, after fixing these, the unit test will still pass :-)

linas:
... Argh. And as I fix the unit test, I am now tripping over at least four different python bugs. I just fixed one. Like I say, it seems that no one actually uses python for anything, and so its as buggy as shit.

linas:
pull req #2019 fixes the unit test, and fixes the six bugs that were probably making you wonder why nothing worked as expected. :-/

noskill:
Thanks

PR #2014, opened 2019-01-25
Allow extending of ExecutionOutputLink::check_schema() in child classes

vsbogd:
Adding protected ExecutionOutputLink constructor to extend `check_schema()` method in child classes.
Related to issue https://github.com/opencog/atomspace/issues/1993

### DIFF
--- a/opencog/atoms/execution/ExecutionOutputLink.cc
+++ b/opencog/atoms/execution/ExecutionOutputLink.cc
@@ -34,7 +34,7 @@
 
 using namespace opencog;
 
-void ExecutionOutputLink::check_schema(const Handle& schema) const
+void ExecutionOutputLink::check_schema(const Handle& schema)
 {
 	if (not nameserver().isA(schema->get_type(), SCHEMA_NODE) and
 	    LAMBDA_LINK != schema->get_type() and
@@ -47,7 +47,7 @@ void ExecutionOutputLink::check_schema(const Handle& schema) const
 	}
 }
 
-void ExecutionOutputLink::init()
+void ExecutionOutputLink::init(check_schema_function check_schema)
 {
 	if (!nameserver().isA(get_type(), EXECUTION_OUTPUT_LINK))
 		throw SyntaxException(TRACE_INFO,
@@ -62,23 +62,29 @@ void ExecutionOutputLink::init()
 	check_schema(oset[0]);
 }
 
+ExecutionOutputLink::ExecutionOutputLink(const HandleSeq& oset, Type t,
+		check_schema_function check_schema_external) : FunctionLink(oset, t)
+{
+	init(check_schema_external);
+}
+
 ExecutionOutputLink::ExecutionOutputLink(const HandleSeq& oset, Type t)
 	: FunctionLink(oset, t)
 {
-	init();
+	init(check_schema);
 }
 
 ExecutionOutputLink::ExecutionOutputLink(const Handle& schema,
                                          const Handle& args)
 	: FunctionLink({schema, args}, EXECUTION_OUTPUT_LINK)
 {
-	init();
+	init(check_schema);
 }
 
 ExecutionOutputLink::ExecutionOutputLink(const Link& l)
 	: FunctionLink(l)
 {
-	init();
+	init(check_schema);
 }
 
 /// execute -- execute the function defined in an ExecutionOutputLink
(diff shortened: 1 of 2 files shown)


linas:
I will try to think about this real soon now, but you should know that there already exist two different mechanisms for checking schema.  One system is in the atom factory. Depending on the atom type, the factory will verify that the expected inputs are all there, and have the correct types, and also that the correct output type is generated. This allows type-checking when composing atoms.  Of course, the factory assumes that the schemasare determined by the atom types; i.e. have been specified in the `atom_types.script` file.  I presume that is not what you are trying to do here.

The second schema-checking subsystem is called "deep types"; its got an entire run-time type-checking system built around it. Read the deep-types wiki pages and also the SignatureLink wiki page. https://wiki.opencog.org/w/SignatureLink and also https://wiki.opencog.org/w/ArrowLink

Please pay particular attention to the ArrowLink. I think that wiki page might already explain how to do everything that you are trying to do in this pull request, so I am very tempted to reject this pull request as "not needed" and "already solved".  To the best of my knowledge, the ArrowLink works fine; it works with ExecutionOutputLinks just fine, and I think its unit tested. I think the agi-bio folks actually use it. 

If what you need is somehow more than what is defined on that wiki page, please explain it...

linas:
Actually, I think it is wisest if I close this pull req now.  I know that this feels rude and insulting, and I don't want it to be. However, there has been a whole lot of confusion about how the atomspace works and what it does, and I am very concerned that you are busy inventing a completely different way of doing things that the atomspace can already do. 

Again: I very very very strongy suggest: try to avoid using ExecutionOutputLinks and GroundedSchemaNodes. I do not know who told you to use these, but .. its a really fundamentally bad idea. I'm quite surprised; it should not be.

vsbogd:
As far as I see from your explanation in https://github.com/opencog/atomspace/pull/2014#issuecomment-457890016 `ExecutionOutputLink::check_schema` here doesn't belong to neither first nor second type checking system.

The reason of raising this PR is that class which inherits `ExecutionOutputLink` cannot link anything which type is not derived from `SCHEMA_NODE` or is not equal to `LAMBDA_LINK` or `UNQUOTE_LINK` (surprisingly). And there is no manner for child class to override this check even if it implements `ExecutionOutputLink` interface correctly. This PR adds protected  constructor which allows overriding the check in child classes.

What problem do you see with this PR?

vsbogd:
Possible alternative solution is representing `ExecutionOutputLink::check_scheme` as `ClassServer::Validator` instance.

linas:
Can you give an example of what requires this change? What do you mean when you say "class which inherits ExecutionOutputLink cannot link anything which type is not derived from SCHEMA_NODE" ?

> or UNQUOTE_LINK

that's a bug.

> What problem do you see with this PR?

Code  bloat. Adding yet more code, that creates more complexity, requires the programmer to understand more parts, increases the combinatorial possibilities of combining parts together.  This is acceptable, when the added function is completely new and completely different. but this change is offering a new API, when it seems like there are already existing API;s do deal with this.

vsbogd:
> Code bloat.

Ok, will it work for you if I will replace `ExecutionOutputLink::check_scheme` call by adding `ClassServer::Validator` instance for `ExecutionOutputLink`?

I also think it is better to keep validator and call `addValidator` near the class itself, not in `Checkers.cc`. For example it can be done adding macros like `DEFINE_ATOM_VALIDATOR`.

vsbogd:
> Can you give an example of what requires this change?

I mean if I will add `VsbogdExecutionOutputLink` which inherits `ExecutionOutputLink` and will try to construct the following:
```scheme
(VsbogdExecutionOutputLink
  (VsbogdDefinedSchemaLink ...)
  (ListLink ...)
)
```
Then `ExecutionOutputLink` will throw exception from constructor. While I was writing it I have realized that I can use `DefineLink` to define `DefinedSchemaNode` externally.

Anyway what do you think about making type checking in atoms unified using `ClassServer::Validator`?

linas:
I cannot quite figure out what you are trying to do.  `check_shema()` is a virtual method.  You can just write 
```
class VsbogdExecutionOutputLink {
    virtual void check_schema(const Handle& schema) const {
            do whatever your want;
   }
};
```
and that overloads `ExecutionOutputLink::check_schema()` and your function gets called instead.  In C++, that is the normal way of altering the behavior of a derived class.

linas:
Also, disregard my comments about factories; what you are doing is probably sufficiently uncommon, that you don't need the factory. And you probably don't need the other devices I mentioned.  Eventually, in the future, we will need to extend/improve how the type system works, but maybe not yet.

PR #2023, opened 2019-01-31
Add ClassServer unit tests

vsbogd:
I added unit tests for the class server to understand current behavior.
Couple of things works strange but seems they don't affect current functionality:
- segmentation fault, if `ClassServer::addValidator` is called before any factory is added - `_atomFactory` was not resized in this case - fixed it
- if B inherits A and  `addValidator(B)` is called before `addValidator(A)` then validator for B is overridden by validator for A (see `test_explicitly_set_validator_is_not_replaced` unit test);
- if B inherits A and `addFactory(A)` is called before `addValidator(B)` then factory for B will be replaced by `validating_factory` (see `test_implicitly_set_factory_is_not_replaced_by_validating_factory` unit test);


tests/atoms/base/ClassServerUTest.cxxtest
```diff
@@ -0,0 +1,236 @@
...
+		class_server->addFactory(NODE_B, factory_b);
+
+		TS_ASSERT_EQUALS(factory_b, class_server->getFactory(NODE_B));
+	}
+
+	void test_set_factory_to_descendants()
+	{
+		class_server->addFactory(NODE_A, factory_a);
+
+		TS_ASSERT_EQUALS(factory_a, class_server->getFactory(NODE_B));
```
linas: I think you are testing a bug here. Setting the factory on A should not clobber the factory for B.

I recall spending a lot of time fiddling with the factory code to make sure that this would not happen: that, no mater what, A got factory for A, and B got factory for B, no matter what order in which they were declared.  However, guess I failed to unit test; it looks to me like you found a bug, and are testing it.

linas: Sorry, I mis-read the unit test. Erasing earlier comment. Let me start all over again.

linas: Yes, this test looks good.

tests/atoms/base/ClassServerUTest.cxxtest
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	void test_explicitly_set_validator_is_not_replaced()
+	{
+		TS_SKIP("fails");
+		class_server->addValidator(NODE_C, validator_c);
+
+		class_server->addValidator(NODE_B, validator_b);
+
+		TS_ASSERT_EQUALS(validator_c, class_server->getValidator(NODE_C));
```
linas: When writing the original code, I worked quite hard to try to get this test case to pass; it is possible that I gave up before I got it working. It was surprisingly complicated.

tests/atoms/base/ClassServerUTest.cxxtest
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	void test_implicitly_set_factory_is_not_replaced_by_validating_factory()
+	{
+		TS_SKIP("fails");
+		class_server->addFactory(NODE_A, factory_a);
+
+		class_server->addValidator(NODE_B, validator_b);
+
+		TS_ASSERT_EQUALS(factory_b, class_server->getFactory(NODE_B));
```
linas: I think you want instead
```
TS_ASSERT_EQUALS(validator_b, class_server->getFactory(NODE_B));
```

linas:
Merging. I think the last unit test fails because the unit test has a bug, and the middle unit test fails because the validator code is buggy.  I'm going to merge.  Sorry for earlier confusion, I failed to pay attention to what `Setup()` did.

tests/atoms/base/ClassServerUTest.cxxtest
vsbogd: Thank you for catching it!
I was going to check: `TS_ASSERT_EQUALS(factory_a, class_server->getFactory(NODE_B));`
Raised PR https://github.com/opencog/atomspace/pull/2028 with use case description

PR #2025, opened 2019-01-31
Fixing a few typos spotted while skimming.

matt2000:


linas:
Thanks!  That document is unfinished, and I plant to finish it "any day now".

Issue #2018, opened 2019-01-28
Question about ForAllLink


stellarspot:
I have cat and dog which are mammalia and have 4 legs. And ant which is insect and has 6 legs.
Is it right to use ForAllLink for a question like: Find a class each member of it has 4 legs?

The code below shows nothing as result instead of expected cat and dog:
```
(use-modules (opencog) (opencog query) (opencog exec))

(define cat (Concept "cat"))
(define dog (Concept "dog"))
(define mammalia (Concept "mammalia"))

(define ant (Concept "ant"))
(define insect (Concept "insect"))

(define key (Predicate "legs-number"))

(define (legs-number who legs)
 (Evaluation key who (Number legs)))

(Inheritance cat mammalia)
(Inheritance dog mammalia)
(legs-number cat 4)
(legs-number dog 4)

(Inheritance ant insect)
(legs-number ant 6)


; Find $X that
; For All (Inheritance $Y $X)
; (legs-number $Y 4)

(define pattern
 (Bind
  (Variable "$X")
  (ForAll
   (Variable "$Y")
   (And
    (Inheritance
     (Variable "$Y")
     (Variable "$X"))
    (Evaluation
     key
     (Variable "$Y")
     (NumberNode 4))))
  (Variable "$X")))

(display
 (cog-execute! pattern)
)
```

linas:
This will do what you want (I think)
```
(define pattern
    (Bind
      ;; Variable declarations: Search requires two variables
      (VariableList (Variable "$X") (Variable "$Y"))
      (And
         (Inheritance (Variable "$Y") (Variable "$X"))
         (Evaluation key (Variable "$Y") (Number 4)))

      ;; Return only one of the two variables
      (Variable "$X")))
```

Comments: 
* the BindLink is always implicitly "for all".
* the variable declarations can sometimes be optional, but things work best if they are explicit  (they are optional for the above query; the pattern matcher can guess just fine that you want X and Y; try it, comment out the VariableList)
* the "return value" is always the last part of the BindLink.
* Although we have a ForAllLink and ExistsLink, they are no-ops; they do not do anything (they "have no meaning"), we haven't yet figured out what to do with them. Don't use them.

linas:
Thinking to myself... perhaps `BindLink` should be renamed to `ForAllLink` ... BindLinks has the structure
```
for all x,y,z,...
    if p(x,y,z,...)
    then return q(x,y,z,...)
```
[...]

PR #2027, opened 2019-02-01
Bugfix temporary atomspace management

linas:
Temporary atomspaces (created in guile) were  not  being garbage-collected! Fixes bug noticed in #2018

### DIFF
--- a/opencog/guile/SchemeSmobAS.cc
+++ b/opencog/guile/SchemeSmobAS.cc
@@ -19,24 +19,11 @@ using namespace opencog;
 // to update this map.  The map contains a use-count for the number of
 // threads that are currently using this atomspace as the current
 // atomspace.  When the count drops to zero, the atomspace will be
-// reaped if the number of SCM references also drops to zero (and the
-// GC runs).
+// reaped if the guile GC finds that the number of SCM references
+// has also dropped to zero.
+
 std::mutex SchemeSmob::as_mtx;
 std::map<AtomSpace*, int> SchemeSmob::deleteable_as;
-
-AtomSpace* SchemeSmob::encrypt(AtomSpace* as)
-{
-	if (as == nullptr)
-		return as;
-	return (AtomSpace*)((long int)as xor 0x5555555555555555);
-}
-
-AtomSpace* SchemeSmob::decrypt(AtomSpace* as)
-{
-	// Encryption and decryption happen to be the same
-	return encrypt(as);
-}
-
 
 /* ============================================================== */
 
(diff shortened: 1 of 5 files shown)


ngeiswei:
After this PR and in particular commit https://github.com/opencog/atomspace/pull/2027/commits/0255e0b83eecfcd22dd752f40d93f3482001099e the pattern miner example https://github.com/opencog/opencog/tree/master/examples/learning/miner/ugly-male-soda-drinker fails, although the corresponding unit test passes. The difference between the two is that the unit tests uses the C++ pattern miner API while the example uses the scheme API.

linas:
fails == crashes? or fails == gives wrong results?  The changes here don't change how atomspaces are used; they do change how they get garbage-collected. This commit removed the work-around for crashes, basically, the use-count was being counted incorrectly, and this fixes teh use-count.

ngeiswei:
fails = segfault

linas:
Examining now. There seem to be several bugs. It doesn't help that the scheme miner API is insane; it needs some love and attention. (One of the arguments to cog-mine, as currently written, can be either a scheme list, an atom, or an atomspace! Which is not normal. And if its an atomspace, it attempts to copy every atom in it .. which is going to be unhappy-land if your atomspace has several millions of atoms in it.)  I'll try to push a fix soon. This should kind-of be reported as a distinct bug report.

linas:
Pull req #2073 should fix it. I'm not convinced that all bugs are actually gone; I can still imagine some might-mare scenarios.

ngeiswei:
I don't like that the argument can be an atomspace either, I did that for the crazy folks who wish to mine their entire atomspace, but now that you bring that up I think I should rather provide a separate function for it (turn an atomspace into a texts list).

ngeiswei:
It seems to work, no more crash, thanks!!!

opencog/guile/SchemeSmobAS.cc
```diff
@@ -90,8 +77,13 @@ void SchemeSmob::release_as (AtomSpace *as)
...
 
 		// (Recursively) decrement the use count on the parent.
-		// XXX This is confusing, I'm not sure its right.
-		if (env) release_as(env);
+		// We had incremented it earlier, in `ss_new_as`, when
+		// creating it.
+		if (env and deleteable_as.end() != deleteable_as.find(env))
+		{
+			--deleteable_as[env];
+			if (0 == deleteable_as[env]) release_as(env);
```
vsbogd: Am I right, that this and above operations on `deleteable_as` should also be under lock `lck`?

vsbogd: Ok, I see you have fixed it in https://github.com/opencog/atomspace/pull/2073

linas:
Phew. Well, this precipitated another marathon bug-fixing session, in #2074 and #2073 ... the attentionbank code is unmaintained, had a could of bugs. Cython had several use-after-free bugs. Once those got cleared up, then the central bug involving atomspace-use-counts could be fixed.

I'm thinking that the attentionbank should be moved out of the atompace repo, and re-united with all the other attention tools in the opencog repo. And then, after that, all of the attention-stuff should git its own repo...

The example code now runs. It does not seem to provide the answers that it promises to provide, however.

Issue #2024, opened 2019-01-31
Split atoms copying and atom creation factory methods and generalize ClassFactory


vsbogd:
This issue is to discuss plan of removing unnecessary double atoms creation which were discussed at issue #1965, especially in comments: 
- https://github.com/opencog/atomspace/issues/1965#issue-393676171
- https://github.com/opencog/atomspace/issues/1965#issuecomment-449579254
- https://github.com/opencog/atomspace/issues/1965#issuecomment-449588049
- https://github.com/opencog/atomspace/issues/1965#issuecomment-449594328

At the moment `createNode` and `createLink` functions create instance of unspecific instance of `Node` or `Link` class and passes them into `ClassServer::factory` which creates another atoms with proper class.

Suggested plan:
- move validator call from `DEFINE_LINK_FACTORY` `DEFINE_NODE_FACTORY` to the `ClassFactory::factory` method
- stop replacing factories by `validating_factory` in `ClassServer::addValidator` method, just call proper validator instead in `ClassFactory::factory`
- reuse `ClassServer::spliceFactory` method to splice not only factories but validators as well
- as side effect it should fix unit tests from PR https://github.com/opencog/atomspace/pull/2023#issue-249267032
- add additional `ClassServer::_copier` collection to keep methods to copy atoms
- add additional method `copyAtom` to use it in `AtomTable::add` method for atoms copying
- replace `Handle (*)(const Handle& )` factory method signature by method with vararg like in `ValueFactory`
- in `createNode` and `createLink` replace new atom creation by arguments forwarding

linas:
I think I like this. But first some general commentary, just so that things are clear.

*  `methods to copy atoms` There is one and only one valid place where it is acceptable to copy an atom: this is during atom-table insertion, so that the AtomTable can have it's own private copy. It is currently "illegal" to make copies of atoms; it should never be done. (There might exist bizarro corner cases, but I cannot think of any) The goal is to always have only one copy, ever, of any atom.  So please to NOT add a `_copier()`

* The atom factory should only be called only once, ever, and that is when an atom is inserted into the atomspace. It is used for only one purpose: to create the one-and-only, single, final version of the atom. The factory should never-ever be called from anywhere else.

* Some atom constructors are very expensive. For example, the `PatternLink` does a HUGE amount of calculation. That's OK, because that calculation is only done once, ever, when the PatternLink is inserted into the atomspace.  We want to be careful, and not accidentally cause a PatternLink costrutor to run more than once.

[...]

PR #2029, opened 2019-02-01
Fix class server behavior

vsbogd:
First steps from issue https://github.com/opencog/atomspace/issues/2024:
- move validator call from DEFINE_LINK_FACTORY DEFINE_NODE_FACTORY to the ClassFactory::factory method
- stop replacing factories by validating_factory in ClassServer::addValidator method, just call proper validator instead in ClassFactory::factory
- reuse ClassServer::spliceFactory method to splice not only factories but validators as well
- fix unit tests from PR https://github.com/opencog/atomspace/pull/2023

This PR incorporates unit test fix from https://github.com/opencog/atomspace/pull/2028

### DIFF
--- a/opencog/atoms/base/ClassServer.cc
+++ b/opencog/atoms/base/ClassServer.cc
@@ -45,20 +45,6 @@ ClassServer::ClassServer(const NameServer & nameServer):
 	_nameServer(nameServer)
 {}
 
-static Handle validating_factory(const Handle& atom_to_check)
-{
-	ClassServer::Validator* checker =
-		classserver().getValidator(atom_to_check->get_type());
-
-	/* Well, is it OK, or not? */
-	if (not checker(atom_to_check))
-		throw SyntaxException(TRACE_INFO,
-		     "Invalid Atom syntax: %s",
-		     atom_to_check->to_string().c_str());
-
-	return atom_to_check;
-}
-
 void ClassServer::spliceFactory(Type t, AtomFactory* fact)
 {
 	// N.B. it is too late to synchronize calls with NameServer using a shared mutex.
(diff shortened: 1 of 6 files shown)


opencog/atoms/base/ClassServer.cc
```diff
@@ -45,34 +45,23 @@ ClassServer::ClassServer(const NameServer & nameServer):
...
-		throw SyntaxException(TRACE_INFO,
-		     "Invalid Atom syntax: %s",
-		     atom_to_check->to_string().c_str());
-
-	return atom_to_check;
-}
-
-void ClassServer::spliceFactory(Type t, AtomFactory* fact)
+template<typename T>
+void ClassServer::splice(std::vector<T>& methods, Type t, T fact)
```
noskill: I suggest adding some comments about what this method does, what are it's arguments

linas:
I just added the following explanation:
```

+/// ClassServer::splice -- install a factory for an atom type.
+///
+/// This inserts a factory `fact` for the atom type `t` into the
+/// array `methods`.  The insertion is performed for not only `t`
+/// but also for all of it's sub-types. The insertion for the
+/// subtypes is done such that any existing factory is not clobbered
+/// by the new one. This allows factories to be installed in any
+/// order, and still have them mirror the correct type hierarchy.
+///
+/// As currently designed, this will 'work correctly' only if the
+/// full atom type hierarchy has already been set up. If a new atom
+/// subtype is delcared, after the factories have been set up, then
+/// the new subtype will not automatically inherit a factory from the
+/// supertype; you will have to write new code for that. XXX FIXME.
+/// So I think that is a bug, as many atom types live outside of the
+/// atomspace. Fixing this bug does NOT require changes to the below;
+/// instead, it requires copying factories whenever the new atom type
+/// is added.
```

Notice the FIXME comments :-)

vsbogd:
Thanks, looks clear to me. Yes copying factories when new types are added is another unsolved issue :-)

PR #2035, opened 2019-02-02
Resolve circular shared-lib dependency

linas:
There was a circular dependency between atomutils, and other atom libraries.  This resolves one of the last few circular dependencies from issue #1673


linas:
unit tests for opencog won't pass till this is merged. Merging now

linas:
this is a pre-req to opencog/opencog#3422

Issue #1968, opened 2018-12-24
opencog/atomspace tries to install scheme files to  /usr/share/guile/site/2.2/opencog/base


vsbogd:
Recently opencog started to install some files into  `/usr/share/guile/site/2.2/opencog/base`. Changing prefix via `CMAKE_INSTALL_PREFIX` doesn't work.
```
Install the project...
-- Install configuration: "Debug"
-- Installing: /home/vital/projects/opencog/builds/current/lib/opencog/libatom_types.so
-- Set runtime path of "/home/vital/projects/opencog/builds/current/lib/opencog/libatom_types.so" to "/home/vital/projects/opencog/builds/current/lib/opencog:lib/python3.6/dist-packages/opencog:/home/vital/projects/opencog/builds/current/lib"
-- Installing: /home/vital/projects/opencog/builds/current/include/opencog/atoms/atom_types/atom_types.h
CMake Error at opencog/atoms/atom_types/cmake_install.cmake:78 (file):
  file cannot create directory: /usr/share/guile/site/2.2/opencog/base.
  Maybe need administrative privileges.
Call Stack (most recent call first):
  opencog/atoms/cmake_install.cmake:42 (include)
  opencog/cmake_install.cmake:42 (include)
  cmake_install.cmake:43 (include)
```

linas:
Huh. Yeah, that should probably be `CMAKE_INSTALL_PREFIX/share/guile/site/2.2`

linas:
So is this an issue or can it be closed?

vsbogd:
My point of view it is a guile issue, as if guile was installed from package is has only `/usr/share` paths in `%load-path`, so one cannot install new modules into `/usr/local`.

I think the fix is to allow set guile site-directory as `cmake` variable.

vsbogd:
raised https://github.com/opencog/atomspace/pull/2044 to fix

AmeBel:
@vsbogd why would you want to install in `/usr/local/` if guile expects the modules in `/usr/share/`? Or am I missing something?

vsbogd:
@AmeBel , it is not about `/usr/local` only. I have raised the issue because I was building `atomspace` using `CMAKE_INSTALL_PREFIX` to override prefix. My `CMAKE_INSTALL_PREFIX` points to the my folder which I can modify without using root permissions. So I execute just `make install` without `sudo`. And at the moment this fails because of Guile install path.

[...]

PR #2044, opened 2019-02-04
Allow guile site dir overriding

vsbogd:
Fix for issue https://github.com/opencog/atomspace/issues/1968

Allow using `cmake -DGUILE_SITE_DIR` to override default dir.

Have checked it with `-DGUILE_SITE_DIR` and without it.


AmeBel:
Would you please add a comment on what the usage is for as you described it in https://github.com/opencog/atomspace/issues/1968#issuecomment-460187016 ?

vsbogd:
Sure, added comment to the cmake file.

PR #2046, opened 2019-02-04
Last of the cyclic dependency unwinding

linas:
This finally allows the execution atoms to not depend on the pattern matcher. I think this now enables simplified future development, as components are now more modular.


opencog/atoms/base/Atom.h
```diff
@@ -253,6 +253,14 @@ class Atom
         throw RuntimeException(TRACE_INFO, "Not a link!");
     }
 
+    virtual TruthValuePtr evaluate(AtomSpace*, bool silent=false) {
```
vsbogd: Why it is needed to add `evaluate` and `execute` to `Atom` API:
- `SatisfuctionLink` is evaluatable and evaluate can be called with casting it to appropriate atom type
- if `DualLink` or `SatisfyingLink` need `execute` method than it can be made executable
- if execute should have `AtomSpace` as parameter then it makes sense modify FunctionLink interface as well.

linas: :-)  Good question. I spent a day trying to create a mixin class called "EvaluatableLink" and I could not get the mixin to work, because I could not get the pointer cast to work right. Maybe-possibly I might have gotten it to work, by using a virtual base-class, and accepting the "dreaded diamond" but the diamond is dreaded for a reason: big performance hit, and hard-to-debug when it fails.  The simple, high-performance direct solution was to add `virtual execute()` to the base class. This has the benefit of avoiding the (expensive) cast, as explained in #2053

Cannot actually cast to `SatisfactionLink` because that cast was the cause of the circular dependency: these casts force `libexecution.so` on every library that contains some `BlahCast` function in it.  Basically, this is why C++ has the concept of overloaded virtual methods: it allows one library to implement a function, without forcing all other libraries to know about it.

linas: > execute should have AtomSpace as parameter then it makes sense modify FunctionLink interface as well

I think I did that in a later pull req.

linas: I edited the above comments for readability; please see the web page, and not the email.

vsbogd: > > execute should have AtomSpace as parameter then it makes sense modify FunctionLink interface as well

> I think I did that in a later pull req.

I mean instead of adding it into `Atom` it can be added into `FunctionLink` only. But ok as far as I understand you added these methods to `Atom` because you got stuck with adding `EvaluatableLink`. Not clear what was the problem with `EvaluatableLink` though? Which API did you try to implement?

linas: As I slept last night, I realized that there is probably a simple way to make the multiple-inheritance work. But first, some side comments:

* Not everything that is executable is going to inherit from FunctionLink, because not everything executable is actually a function. I'm not going to try to explain this here; its off-topic, but it has to do with whether or not variables are bound, whether functions have arbitrary arity, and how many results are returned. So, a typical example of FunctionLink is `PlusLink`, which has zero bound variables, and can add an arbitrary list of "addable" atoms.  A typical example of not-a-function-link is GetLink, which always binds it's variables, and can return more than one result.

What I tried to do is here, https://github.com/linas/atomspace/tree/la and if you look at it, you will say "linas is an idiot". (I now know how to fix it, I think, but I also kind-of like the way it is now). I tried to do this:
```
// Everything executable should inherit from this (pure-virtual, "interface" class) class:
class ExecutableLink {
    virtual Handle execute() = 0;
}

// Multiple inheritance; PatternLink does the work, ExecutableLink provides the API.
class DualLink: public PatternLink, public ExectuableLink {
    virtual Handle execute() { return stuff; }
}
```
and that way, you can execute without having to know what class it is, by simply saying:
```
Handle foo = ...;
ExecutableLinkPtr elp = ExecutableLinkCast(foo);
Handle result = elp->execute();
```
I could not make `ExecutableLinkCast` work because I was being stupid. I now know of a way of making the above work.  But I also kind-of like the current solution, too. I think the current solution has higher performance than the above, because it avoids the CPU overhead of a cast and of the thunk (multiple inheritance in C++ always requires a thunk)

vsbogd: > Not everything that is executable is going to inherit from FunctionLink

Yes, it can be seen from the atomspace code and the logical step here is to introduce EXECUTABLE_LINK to designate atoms which are executable. I see that you did it in your branch.

> What I tried to do is here, https://github.com/linas/atomspace/tree/la

This is exactly the way I meant in my comment above.

> I think the current solution has higher performance than the above, because it avoids the CPU overhead of a cast and of the thunk (multiple inheritance in C++ always requires a thunk)

Part of casting overhead can be removed by using interfaces (or abstract classes) and calling virtual methods in places where type is clear as you said at https://github.com/opencog/atomspace/pull/2046#discussion_r254088874 . But anyway it will be a problem for graph traversal algorithms.

vsbogd: Actually we can use `pointer_static_cast` almost everywhere because we are checking types using atomspace typesystem.

linas: I have no objection to `pointer_static_cast`, other than that maybe just maybe it might maybe lead to subtle and hard-to-find bugs. But hopefully not. My biggest concern is that I don't want to be the one trying to track down such bugs, if/when they show up. You take responsibility for that, I'm happy. :-)  If random users make random complaints about random pointless things, forcing me to janitor, I'm unhappy. :-(

I'm guessing that `pointer_static_cast`, will probably improve performance, but I think we still don't have any benchmarks that would detect this. I run the old benchmarks twice a year, not much changes, these days. (which is good, I guess).

vsbogd: Agree, that it doesn't make sense to improve performance without having benchmarks in place.

To make it more safe it is possible to ensure the type is correct before conversion like in following snippet (not sure if it can be compiled, just to demonstrate the idea):
```c++
template<typename T>
bool call_if_is_a(Handle& handle, std::function<void(T&)> callable) {
	if (!nameserver().isA(handle->get_type(), T::type)) {
		return false;
	}
	T* p = static_cast<T*>(h.get());
	callable(*p);
	return true;
}
...
call_if_is_a<ExecutionOutputLink>(h, EXECUTION_OUTPUT_LINK, [](ExecutionOutputLink& eol) {
	eol.execute(as);
}); 
```

linas: I want to introduce type-safety gradually and conservatively. Code like the above can sometimes be like quick-set concrete. By the time you realize you made a design mistake, its too late, it's hard as a rock, and you have to use a jack-hammer to break it up.  So I am mostly trying to move as slow as practical, to make as few design assumptions as possible, and make only those that are "really needed" or "really obvious" (whatever that means), or "steps along the right path"  (the right path is uncertain and unknown ...).

In this case, there still is a tension between "execute" and "evaluate". They are more similar, now, than they used to be, a few years ago. But they are still different. Is that good? Is that bad? Should they become more similar? Less? I don't know. It's unwise to make a quick decision; its better to wait, and see what kind of example problems and example solutions show up. These will help make the various possible answers more clear. In the meantime, I want to avoid complicated apparatus that forces a certain design.  Something light, fluid, easy to change seems best, for this problem here, for just right now.

PR #2050, opened 2019-02-05
Remove obsolete python functions

linas:
This forces a wide-spread change to the function signatures.


linas:
This is co-dependent on pull req opencog/opencog#3423 and will not pass unit tests without it.

PR #2052, opened 2019-02-05
Add Python ForwardChainer wrapper

stellarspot:
This is a proposal  to add Python ForwardChainer wrapper. It is created in the same way as BackwardChainer wraper:
- forwardchainer.pxd and forwardchainer.pyx files are added into opencog/cython/opencog folder
- CMakeLists.txt is updated
- ForwardChainer unit test is added


tests/cython/forwardchainer/test_fc.py
```diff
@@ -0,0 +1,52 @@
+import os
+import unittest
+from opencog.forwardchainer import ForwardChainer
```
vsbogd: I would suggest merge `backwardchainer` and `forwardchainer` modules into single `rule-engine` module. As they are part of the same rule-engine module. May be do it in separate PR. @noskill, @stellarspot  what do you think?

vsbogd: There is some risk to break current code, but as far as I know we have only one usage at the moment.

ngeiswei: Agreed, especially since the forward and backward chainer might get unified some day.

noskill: backwardchainer is used in opencog/vqa demo jupyter notebook https://github.com/singnet/dev-portal/tree/master/workshops/docker-opencog

ngeiswei: I like short names. The guile module (which I didn't create) is called `rule-engine`. However python doesn't accept `-` and `ruleengine` is kinda sucky, so yes maybe `ure` is the best.

Maybe I'll rename the `rule-engine` guile module to `ure` (or keep both names for backward compatibility).

linas:
Oh, I notice unit tests are failing. I think they will probably "fix themselves" if you `git merge master; git push` .   If that doesn't fix it, ask me, I can explain/fix.

stellarspot:
I have fixed the Python ForwardChainer, now it passes.

vsbogd:
Looks good to me, merging it as @linas approved it as well

PR #2053, opened 2019-02-05
Add universal casting functions

vsbogd:
This PR adds universal cast functions for `Handle`, `AtomPtr` and `ValuePtr` references to not generate `Cast..` functions for each new `Atom` or `Value` instance. Cast functions are defined for `Handle`, `AtomPtr` and `ValuePtr` subclasses only.

### DIFF
--- a/opencog/atoms/base/Handle.h
+++ b/opencog/atoms/base/Handle.h
@@ -322,6 +322,30 @@ std::string oc_to_string(const TypeSet& types,
 std::string oc_to_string(const AtomPtr& aptr,
                          const std::string& indent=empty_string);
 
+/**
+ * Cast Handle to the specific Atom subclass. This function is defined only
+ * for T which are subclasses of Atom.
+ */
+template<typename T>
+static inline
+typename std::enable_if< std::is_base_of<Atom, T>::value, std::shared_ptr<T> >::type
+CastFromHandle(const Handle& handle)
+{
+	return std::dynamic_pointer_cast<T>(handle);
+}
+
+/**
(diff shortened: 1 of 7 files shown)


linas:
Thanks!

The concern I had was this: `return std::dynamic_pointer_cast<Value>(std::const_pointer_cast<T>(value));`  Last time I looked (many years ago), the various std:: cast's generated almost 100 lines of assembly code consisting almost entirely of branch statements, and so seems to be guaranteed to cause i-TLB and i-cache misses, and maybe d-cache misses too. I would be surprised if this got cleaned up in newer compiler versions. Casting of smart pointers is heavy-weight, and best avoided, if possible. Easy to say, but hard to do.

I did run the google profiler a few weeks ago, and it showed that the C++ smart-pointers in the atomspace used about 25% of all CPU time. In theory, we could gain a huge performance gain by switching to garbage collection.   However: "been there, done that": there was a version of the atomspace that was garbage-collected using BDW-GC. It sort-of worked. The problem is that C++ templates and standard classes create very deeply connected graphs, and the GC has a hard time tracing these to realize that they are disconnected, and can be reaped.  I looked -- the GC gives up after gong about 12-13 pointers deep; and the combo of C++ std::library plus atomspace went far deeper than that, even though the graph was disconnected, and thus could have been collected. Oh well. We are stuck with smart-pointers, and all their inefficiencies, it seems.

Above comments are just FYI, things you should know.  No action required. Other than that maybe you might want to get familiar with the performance profile of smart-pointers and casting, so as to minimize the performance hit.

PR #2057, opened 2019-02-07
Make python api classes hierarchy consistent

vsbogd:
- replace `ValuePtr` by general `PtrHolder`
- make `TruthValue` and `Atom` children of `Value`
- some code cleanup

### DIFF
--- a/opencog/cython/opencog/atom.pyx
+++ b/opencog/cython/opencog/atom.pyx
@@ -7,32 +7,29 @@ cdef cAtom* get_atom_ptr(Atom atom):
 
 
 # Atom wrapper object
-cdef class Atom(object):
+cdef class Atom(Value):
 
     @staticmethod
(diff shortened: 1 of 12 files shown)


opencog/cython/opencog/atom.pyx
```diff
@@ -1,38 +1,31 @@
...
+        """Return C++ shared_ptr from PtrHolder instance"""
+        return <cHandle&>(self.ptr_holder.shared_ptr)
 
     def __nonzero__(self):
         """ Allows boolean comparison, return false is handle is
         UNDEFINED or doesn't exist in AtomSpace """
         if self.handle:
             return self.atomspace.is_valid(self)
-        else: return False
+        else:
```
noskill: else not needed

vsbogd: ok, will  it in next pr

PR #2061, opened 2019-02-08
Move Python Forward and Backward chainers to opencog.ure package

stellarspot:
Move Python Forward and Backward chainers to opencog.ure package


vsbogd:
Looks good to me

PR #2063, opened 2019-02-08
Return correct value class from atom

vsbogd:
NB: This PR breaks compatibility as it moves TruthValue type constructor to the `opencog.type_constructors` package. Similar changes in `opencog` repository are required.
- move `TruthValue` type constructor into `opencog.type_constructors` package
- add values constructor to construct from c shared_ptr
- add separate type constructors to construct from Python values 
- return correct value class from `Atom.get_value`
- add unit test to check that `Atom.get_value` returns correct type

### DIFF
--- a/examples/atomspace/my_py_func.py
+++ b/examples/atomspace/my_py_func.py
@@ -5,8 +5,9 @@
 # example. The code below is called, when a GroundedSchemaNode and
 # a GroundedPredicateNode is triggered.
 #
-from opencog.atomspace import AtomSpace, TruthValue
+from opencog.atomspace import AtomSpace
 from opencog.atomspace import types
+from opencog.type_constructors import TruthValue
 
 asp = AtomSpace()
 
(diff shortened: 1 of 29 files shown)


vsbogd:
NB: This PR breaks compatibility as it moves TruthValue type constructor to the `opencog.type_constructors` package. Similar changes in `opencog` repository are required.

While preparing this change (see my branch https://github.com/vsbogd/opencog/tree/import-python-truth-value-from-type-constructors-package) I thought that may be it is simpler and more logical to move all type constructors into `opencog.atomspace` package. But it will require having some separate namespace for value classes as it is natural to have them using same names.

vsbogd:
Ok, I will close this PR for a while and think about better way of exporting TruthValue, or at least making this change

linas:
?  Like I said, I have no objections to this change; there are some existing users but it is possible that they don't really care.

vsbogd:
Yes, I think I will break this PR on two:
- first to return correct value class
- second to cleanup namespaces

PR #2065, opened 2019-02-11
Return correct value class from atom

vsbogd:
Version of PR https://github.com/opencog/atomspace/pull/2063, without moving TruthValue type constructor into `opencog.type_constructors` module. So it doesn't break compatibility.

Things done:
* add values constructor to construct from c shared_ptr
* add separate type constructors to construct from Python values
* return correct value class from `Atom.get_value`
* add unit test to check that `Atom.get_value` returns correct type


vsbogd:
Merging it as previous PR https://github.com/opencog/atomspace/pull/2063 was reviewed by @linas

Issue #2070, opened 2019-02-13
Automatic alpha-conversion during pattern matching


ngeiswei:
The following

```scheme
(define query (Get
                (VariableList
                  (Variable "$vardecl")
                  (Variable "$vardecl-0")
                  (Variable "$vardecl-1")
                  (Variable "$body-0")
                  (Variable "$body-1"))
                (Present
                  (Quote (Lambda
                           (Unquote (Variable "$vardecl"))
                           (And
                             (Unquote (Variable "$body-0"))
                             (Unquote (Variable "$body-1")))))
                  (Quote (Lambda
                           (Unquote (Variable "$vardecl-0"))
                           (Unquote (Variable "$body-0"))))
                  (Quote (Lambda
                           (Unquote (Variable "$vardecl-1"))
                           (Unquote (Variable "$body-1")))))))
```

attempts to grab a pattern, lambda link of 2 conjuncts, only if each conjunct is also a pattern (wrapped in a `LambdaLink` as well), for instance, given the KB

```scheme
(define X (Variable "$X"))
(define Y (Variable "$Y"))
(define A (Concept "A"))
(define InhXA (Inheritance X A))
(define InhYA (Inheritance Y A))
(define PX (Lambda X InhXA))
(define PY (Lambda Y InhYA))
(define VarXY (VariableList X Y))
(define PXY (Lambda VarXY (And InhXA InhYA)))
```

We would expect to get

```
$vardecl = (VariableList X Y)
$vardecl-0 = X
$vardecl-1 = Y
$body-0 = Inheritance X A
$body-1 = Inheritance Y A
```

but instead get nothing.

It is explained by the fact that

1. The atomspace alpha-convert PY to PX
2. The pattern matcher doesn't realized that PX could be alpha-converted back to PY (or any alpha-convertion such that it doesn't collide with PX) during pattern matching.

linas:
I don't see how this has anything to do with alpha conversion, since nothing in the search pattern is forcing the variable names to match.  It's just looking for two bodies -- those should be findable, and then after that (after finding them) it looks for what is attached to the bodies -- which should be the variable declarations, whatever they may be, converted or not.  What am I missing?

linas:
Ohh, I see, they bodies cannot match up. Hmmm.

linas:
OK, why are the quotes needed? Would it be OK to search, without the quotes?

linas:
So I poked around just for a bit. I notice that this simpler quoate-less search also fails:
[...]

PR #2071, opened 2019-02-14
Partial fix for bug #2070

linas:
This allows the following query to succeed:
```
(use-modules (opencog) (opencog exec))

(define X (Variable "$X"))
(define Y (Variable "$Y"))
(define A (Concept "A"))
(define InhXA (Inheritance X A))
(define InhYA (Inheritance Y A))
(define PX (Lambda X InhXA))
(define PY (Lambda Y InhYA))
(define VarXY (VariableList X Y))
(define PXY (Lambda VarXY (And InhXA InhYA)))

(define q3 (Get (VariableList
                  (Variable "$vardecl")
                  (Variable "$body-0")
                  (Variable "$body-1"))
                (Present (Lambda
                           (Variable "$vardecl")
                           (And (Variable "$body-0")
                             (Variable "$body-1"))))))

(cog-execute! q3)
```

I think that this is an over-all beneficial change.  I think this makes sense ...

### DIFF
--- a/opencog/atoms/core/Variables.cc
+++ b/opencog/atoms/core/Variables.cc
@@ -519,7 +519,7 @@ bool Variables::is_type(const Handle& var, const Handle& val) const
  */
 bool Variables::is_type(const HandleSeq& hseq) const
 {
-	// The arity must be one for there to be a match.
+	// The arities must be equal for there to be a match.
 	size_t len = hseq.size();
 	if (varset.size() != len) return false;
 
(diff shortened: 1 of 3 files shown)


linas:
@ngeiswei if this change seems wrong to you, let me know.  I think the query `g3` should work just fine, without the need for quotes .... but I don't have a strong opinion ... I don't see any good reason to disallow this match.

linas:
After this change, the following also works:
```
(define q5 (Get (VariableList
                  (Variable "$vardecl")
                  (Variable "$vardecl-0")
                  (Variable "$body-0")
                  (Variable "$body-1"))
                (Present (Lambda
                           (Variable "$vardecl")
                           (And
                             (Variable "$body-0")
                             (Variable "$body-1")))
                  (Lambda
                           (Variable "$vardecl-0")
                           (Variable "$body-0")))))

(cog-execute! q5)
```

linas:
Merging for now, because this blocks the alpha-equal experiment. To revert the changes here, only line 305 of  `opencog/query/DefaultPatternMatchCB.cc` needs to be reverted.

ngeiswei:
Cool, this is freeing us a bit more from having to use quotations. I don't fully understand its repercussions but I guess we can wait and see.

Obviously that doesn't solve #2070 but no rush, it's not a blocker or anything ATM.

PR #2072, opened 2019-02-14
Add AlphaEqualLink

linas:
Useful as a work-around for issue #2070 This allows the following query to work:
```
(use-modules (opencog) (opencog exec))

(define X (Variable "$X"))
(define Y (Variable "$Y"))
(define A (Concept "A"))
(define InhXA (Inheritance X A))
(define InhYA (Inheritance Y A))
(define PX (Lambda X InhXA))
(define PY (Lambda Y InhYA))
(define VarXY (VariableList X Y))
(define PXY (Lambda VarXY (And InhXA InhYA)))

(define b6 (Bind (VariableList
                  (Variable "$vardecl")
                  (Variable "$vardecl-0")
                  (Variable "$vardecl-1")
                  (Variable "$body-0")
                  (Variable "$body-1")
                  (Variable "$body-0-eqv")
                  (Variable "$body-1-eqv"))
                (And (Present (Lambda
                           (Variable "$vardecl")
                           (And (Variable "$body-0")
                             (Variable "$body-1")))
                  (Lambda (Variable "$vardecl-0")
                           (Variable "$body-0-eqv"))
                  (Lambda (Variable "$vardecl-1")
                           (Variable "$body-1-eqv")))
              (AlphaEqual (Variable "$body-0") (Variable "$body-0-eqv"))
              (AlphaEqual (Variable "$body-1") (Variable "$body-1-eqv"))
                )
(List (Concept "here are the vardecls")
                  (Variable "$vardecl")
                  (Variable "$vardecl-0")
                  (Variable "$vardecl-1")
 (Concept "here are the bodies")
                  (Variable "$body-0")
                  (Variable "$body-1")
 (Concept "here are the qivalents")
                  (Variable "$body-0-eqv")
                  (Variable "$body-1-eqv"))
      ))
(cog-execute! b6)
```

### DIFF
--- a/opencog/atoms/atom_types/atom_types.script
+++ b/opencog/atoms/atom_types/atom_types.script
@@ -560,6 +560,10 @@ GREATER_THAN_LINK <- VIRTUAL_LINK,NUMERIC_LINK
 IDENTICAL_LINK <- UNORDERED_LINK,VIRTUAL_LINK
 EQUAL_LINK <- UNORDERED_LINK,VIRTUAL_LINK
 
+// AlphaEqualLink tests to see if both sides are the same, up to an
+// alpha-renaming of any variables that they may have.
+ALPHA_EQUAL_LINK <- UNORDERED_LINK,VIRTUAL_LINK
+
 // --------------------------------------------------------------
 // Functions that do NOT bind variables.
 
--- a/opencog/atoms/execution/EvaluationLink.cc
+++ b/opencog/atoms/execution/EvaluationLink.cc
@@ -230,6 +230,28 @@ static TruthValuePtr equal(AtomSpace* as, const Handle& h, bool silent)
 		return TruthValue::FALSE_TV();
 }
 
+/// Check for alpha equivalence
+static TruthValuePtr alpha_equal(AtomSpace* as, const Handle& h, bool silent)
+{
+	const HandleSeq& oset = h->getOutgoingSet();
+	if (2 != oset.size())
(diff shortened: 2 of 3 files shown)


linas:
Note: lines 241-243
```
	Instantiator inst(as);
	Handle h0(HandleCast(inst.execute(oset[0], silent)));
	Handle h1(HandleCast(inst.execute(oset[1], silent)));
```

are overkill and probably not needed by most users, but required to be consistent with general expectations with how other similar things work.

linas:
The best long-term plan might be to change `EqualLink` to just do what `AlphaEqualLink` does. !?

linas:
XXX This cannot be merged as is, the implementation is buggy! Will fix later!

ngeiswei:
Thanks @linas. Maybe I should have mentioned in the issue that it's a not a blocking bug (I thought it was but then realized I actually don't need it).

Ideally I suppose it should be silent, not requiring `AlphaEqualLink` or `EqualLink` since alpha-conversion itself is silent at insertion, etc. But for the time being having an `AlphaEqualLink` seems best. I would however not recommend to merge `AlphaEqualLink` into `EqualLink`, I fear that this might generate some weird behaviors that the user would find very hard to wrap his head around. So again, better either be completely silent, or completely explicit using `AlphaEqualLink`.

linas:
OK. I just had a few hours to kill, so I thought I'd look at this.

Issue #1967, opened 2018-12-24
Replace AtomSpace with AtomSpaceLink


linas:
This is an old idea, but without it's own issue. Time to describe it more carefully. This would resolve issues #1855 and #1921 This depends on #1502 being implemented first.

Proposal: Implement a new kind of very special link, called an AtomSpaceLink, that behaves a lot like a MemberLink.  Move Values from Atoms to AtomSpaceLinks.  Design pros and cons follow.

Why?
* There is a generic need to limit some operation to some smaller set of atoms. Examples include the AttentionBank, the "focus set" during inference (issue #1984), and the MapLink inputs. (https://wiki.opencog.org/w/MapLink)
* The above is currently handled in different ad hoc ways. The "focus set" is kept in C++ code, and is passed around with C++ pointers in an adhoc way. There is only one focus set (because there is only one attention bank).  Meanwhile, the MapLink inputs are kept in a SetLink. Temporary inference results are kept in temporary atomspaces....
* Having a generic mechanism to manage multiple "focus sets", or "atomspaces" or "SetLinks" would be beneficial.
* Having a generic mechanism that is fast, easy-to-use, easy-to-understand, and avoiding crazy side-effects would be best!

Pros:
* By moving values to the AtomSpaceLink, the bungled incoming set issue of #1921 is solved. Atoms become, once-again, globally unique.
* The AtomSpaceLink becomes a kind-of ContextLink, and so contextualized TV's become "easy to handle" (viz the sky is blue has a big truth value on earth, a small one on mars)
* Confusion about nested atomspaces, and multiple atomspaces is lessened. Atomspaces no longer have to be strict nestings; they can intersect in non-trivial ways.
* Data sharing, i.e. making some atomspace be read-only, while other, smaller one read-write, are simpler. The agi-bio/genomic people want this; see issue #1855 
* MapLink/BindLink can be unified. So can GetLink/FilterLink. No need for two similar-but-different code bases  to do almost the same thing.
* @ngeiswei  claims that this enables an "efficient URE-based pattern miner", in a comment in #1502
* The set of all AtomSpaces, and their relationships to one-another, becomes a Kripke frame (a modal frame) Thus, each atomspace is a "possible world", and a `FrameLink` describes accessibility (the FrameLink replaces the current parent-child relationship in atomspaces). 
[...]

PR #2074, opened 2019-02-15
Count atomspace nesting usage.

linas:
Needed sanity check for mis-bhaving code.

### DIFF
--- a/opencog/cython/opencog/backwardchainer.pyx
+++ b/opencog/cython/opencog/backwardchainer.pyx
@@ -41,6 +41,7 @@ cdef class BackwardChainer:
 
     def __del__(self):
         del self.chainer
+        self._as = None
 
     def do_chain(self):
         return self.chainer.do_chain()
--- a/opencog/cython/opencog/backwardchainer.pyx
+++ b/opencog/cython/opencog/backwardchainer.pyx
@@ -39,9 +39,11 @@ cdef class BackwardChainer:
                                         deref(focus_set.handle))
         self._as = _as
 
-    def __del__(self):
-        del self.chainer
-        self._as = None
+    # according to PEP 442, the below simply won't work.
+    # def __del__(self):
+    # def __dealloc__(self):
+    #    del self.chainer
+    #    self._as = None
 
     def do_chain(self):
         return self.chainer.do_chain()
@@ -49,5 +51,9 @@ cdef class BackwardChainer:
     def get_results(self):
         cdef cHandle res_handle = self.chainer.get_results()
         cdef Atom result = Atom.createAtom(res_handle, self._as)
-        return result
 
+        # Delete the chainer now. There does not appear to be
+        # any other way of avoiding the mem leak. See PEP 442
+        # for details.
+        del self.chainer
+        return result
--- a/opencog/cython/opencog/backwardchainer.pyx
+++ b/opencog/cython/opencog/backwardchainer.pyx
@@ -39,6 +39,9 @@ cdef class BackwardChainer:
                                         deref(focus_set.handle))
         self._as = _as
 
+    def __del__(self):
+        del self.chainer
+
     def do_chain(self):
         return self.chainer.do_chain()
 
--- a/tests/cython/backwardchainer/test_bc.py
+++ b/tests/cython/backwardchainer/test_bc.py
@@ -69,6 +69,7 @@ class BCTest(TestCase):
         self.assertTrue(resultAC.tv == AB.tv)
         self.assertEquals("A", resultAC.get_out()[0].name)
         self.assertEquals("C", resultAC.get_out()[1].name)
+        chainer = None
 
     def test_conjunction_fuzzy_with_virtual_evaluation(self):
         """Test for correct vardecl parameter initialization in BackwardChainer
@@ -101,6 +102,7 @@ class BCTest(TestCase):
         bc.do_chain()
         results = bc.get_results()
         self.assertTrue(len(results.get_out()) == 2)
+        bc = None
 
 
 if __name__ == '__main__':
--- a/tests/cython/backwardchainer/test_bc.py
+++ b/tests/cython/backwardchainer/test_bc.py
@@ -69,7 +69,7 @@ class BCTest(TestCase):
         self.assertTrue(resultAC.tv == AB.tv)
         self.assertEquals("A", resultAC.get_out()[0].name)
         self.assertEquals("C", resultAC.get_out()[1].name)
-        chainer = None
+        del chainer
 
     def test_conjunction_fuzzy_with_virtual_evaluation(self):
         """Test for correct vardecl parameter initialization in BackwardChainer
@@ -102,7 +102,7 @@ class BCTest(TestCase):
         bc.do_chain()
         results = bc.get_results()
         self.assertTrue(len(results.get_out()) == 2)
-        bc = None
+        del bc
 
 
 if __name__ == '__main__':
--- a/opencog/atomspace/AtomTable.cc
+++ b/opencog/atomspace/AtomTable.cc
@@ -83,6 +83,8 @@ AtomTable::AtomTable(AtomTable* parent, AtomSpace* holder, bool transient) :
 {
     _as = holder;
     _environ = parent;
+    if (_environ) _environ->_num_nested++;
+    _num_nested = 0;
     _uuid = _id_pool.fetch_add(1, std::memory_order_relaxed);
     _size = 0;
     _num_nodes = 0;
@@ -121,6 +123,10 @@ AtomTable::~AtomTable()
             }
         }
     }
+    if (_environ) _environ->_num_nested--;
+    if (0 != _num_nested)
+        throw opencog::RuntimeException(TRACE_INFO,
+                "AtomTable - deleteing atomtable with subtables!");
 }
 
 void AtomTable::ready_transient(AtomTable* parent, AtomSpace* holder)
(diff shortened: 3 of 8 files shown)


linas:
BackwardChainer is failing this one.

linas:
Fixed a memory leank in the BackwardChainerUTest.

The cython bindings for the backward chainer still have a mem leak somewhere.

ngeiswei:
@linas, if you still have troubles with the backward chainer utest, rebasing (or merging) onto the last master may fix it.

tests/cython/backwardchainer/test_bc.py
```diff
@@ -69,6 +69,7 @@ def test_bc_deduction(self):
         self.assertTrue(resultAC.tv == AB.tv)
         self.assertEquals("A", resultAC.get_out()[0].name)
         self.assertEquals("C", resultAC.get_out()[1].name)
+        chainer = None
```
noskill: this line looks like no-op

linas: I was hoping that this would cause the chainer destructor to run.  In python, how does one trigger destructors or finalizers?

linas:
I think I fixed one memleak in the cython backchainer bindings, but now there is a deadlock in attentionbank, because the backchainer constructor is accessing an attentionbank that no longer exists :-(

linas:
There was an attentionbank use-after-free, because some of the cython bindings deleted the atomspace that the attentionbank depended on.

linas:
woo hoo! all tests pass! this simple trick exposed half-a-dozen hard-to-find use-after-free bugs.

opencog/cython/opencog/backwardchainer.pyx
```diff
@@ -39,11 +39,21 @@ cdef class BackwardChainer:
                                         deref(focus_set.handle))
         self._as = _as
 
+    # according to PEP 442, the below simply won't work.
```
noskill: Why it wouldn't? Is there a circular reference somewhere between python and c++?

linas: I don't know. You'd have to ask the people who made that decision.

noskill: PEP 442 doesn't say that finalizer shouldn't run.
And what is the point of clearing fields in finalizer? This what is done by default at time of object disposal.

linas: The python bindings for the backward chainer had a memleak.  I hacked around the memleak in a very hacky way. You should alter it to suit your tastes -- I don't have any particular opinion about the best way to do this.

linas: I mean -- try it -- uncoment the `__del__` it will never get called. Googling this, one is lead to a variety of stackexchange discussions. Apparently, is something that worked in python2 and was removed in python3.. The most telling remark said something similar to this: "with garbage collection, its like having an infinite amount of memory; there are no guarantees that your finalizer will ever be called. If you need to delete an object, you muse provide a `def close()` method and teach your users to use it."  I wasn't particularly interested in understanding the internals of the python implementation; I just was trying to get e atomspace unit tests to not crash.

noskill: __del__ doesn't called since BackwardChainer is cython class https://cython.readthedocs.io/en/latest/src/userguide/special_methods.html
__dealloc__  gets called as expected

noskill: tested on python 3.7.0

vsbogd: @linas, how do you see the memory leak you mention?

linas: This patch originally added some very simple use-counting to the atomspace constructors and destructors. I was trying to track down mysterious crashes, which seemed to involved a use-after-free of the atomspace.  I found maybe six such bugs, all patched in this pull req. The one with the backward chainer was this: 

* create atomspace
* create backchainer, givening atomsace to it
* backchainer ctor calls BIT ctor which makes child atomspace
* use backhainer
* delete atomspace
* null-pointer deref because attentionbank is sending signals to the deleted atomspace, via the child atomspace, (which does not know that the parent was deleted). Child is being held in the BIT class.
* solution A: delete BIT class, after query is finished. I could not figure out how to do this. Maybe @ngeiswei knows how... So I did not implement solution A.
* solution B: delete backchainer (which deletes BIT) BEFORE deleting atomspace.
* ... except that cython wasn't deleting the back-chainer, either by explicit GC (setting `self.backchainer=None` or by explicit destructor (with the `__del__` destructor)

noskill: I suggest reverting changes in backwardchainer.pyx and test_bc.py

linas: Jesus duude, why don't you actually try to do that? Have fun watching it crash and burn.  It would save me a lot of time arguing about this stuff, if, instead, you took the effort to actually try some of the things that you are arguing about.  This is not magic theoretical philosophy here. The code actually does stuff. You can find out what it does by changing it, and seeing what happens.  Stop being lazy.

noskill: I actually did run the code, as i said above with python 3.7
I added a lot of prints to see what is going on:
```

atomspace destructor 0x7ffee363eb90
atomspace destructor 0x7ffee363eb90
atomspace destructor 0x7ffee363eb90
atomspace destructor 0x7ffee363eb90
atomspace destructor 0x7ffee363eb90
atomspace destructor 0x7ffee363eb90
atomspace destructor 0x7ffee363eb90
atomspace destructor 0x7ffee363eb90
atomspace destructor 0x7ffee363eb90
atomspace destructor 0x7ffee363eb90
atomspace destructor 0x7ffee363eb90
atomspace destructor 0x7ffee363eb90
:74: DeprecationWarning: Please use assertEqual instead.
deleting
bc destructor0x560b00e80b40
atomspace destructor 0x560b01a8f5d8
atomspace destructor 0x560b00e80b40
.Item1
large
Item1
small
Item0
large
Item0
small
atomspace destructor 0x7ffee363eb90
deleting
bc destructor0x560b00e80b40
atomspace destructor 0x560b01a9f338
atomspace destructor 0x560b00e80b40
.
----------------------------------------------------------------------
Ran 2 tests in 0.357s

OK


```

I believe that there might be bug somewhere but how deleting local variable could help? I mean del chainer in test_bc.py?

noskill: https://github.com/noskill/atomspace/tree/testtest1

noskill: in backward chainer destructor i print atomspace address

linas: Many different fixes went in; it is possible that one of the other fixes is hiding the bug, or changed the way it manifests.  I already explained the bug, above, but let me repeat. The observed crash was due to this scenario:
* Cython creates atomspace
* cython gives atomspace to backhainer.
* backchainer creates child atomspace, in the c++ `class BIT`
* backchainer is used, everything is good.
* cython deletes atomspace.
* BIT is still holding child of the deleted atomspace. 
* Anything that might cause the backward chainer to run again might cause a fault, due to access of freed memory.
* Anything that subscribes to atom signals (e.g. the attentionbank) will cause a fault, due to access of freed memory.  (AttentionBank has been moved to a place where it can no longer hurt us...)

However, the general pattern, of having BIT holding on to a child atomspace whose parent has been deleted is .. bad design. It is just asking for future trouble. Two solutions are possible: either don't delete the parent atomspace, or cleanup backchainer to not hold long-term references.

The first solution is not possible, because when cython exists, the parent atomspace wa being deleted, *before* the backchainer.  This caused a crash.  The fix was to ... delee the backchainer.

You are welcome to find an alternate solution.  The "best" solution is probably to change the C++ code to delete the `BIT` immediately after the chainer is done with it. Rhis might be easy, this might be hard. @ngeiswei can tell you if that is easy or hard.

The second-best solution is to change the cython bindings for backchainer, so that all users of the backchainer must call a `close()` or a `finish()` or a `release_results()` method when they are done.

There may be other solutions. I don't know. I mostly don't care. What I do care about is bad coding/design/programming practice: one should not write code that leaks memory, that maintains dangling pointers to deleted memory. Its asking for trouble. 

Now, of course, the whole parent-child atomspace idea has ... problems. Its ... not ideal. Its imperfect. The perfect is the enemy of the good.  Fixing the parent-child atomspace problem is much much harder than simply calling the destructor on the backchainer. 

(Fixing the parent-child atomspace problem is documented in multiple issues, #1967 and teh others referenced there)

noskill: Ok, i see that BackwardChainer cython class is holding reference to the result atomspace, so it can't be deleted by cython without deletion of backward chainer instance. But references to trace and control atomspaces are not being held, so they might get deleted before the backward chainer instance. I'll make PR with storing reference to trace and control atomspaces.

linas: I hope you meant that the other way around. The backward chainer should NOT store references!

linas: The ONLY legal way to store references would be to create a `BackwardChainerLink` -- which can store all the references that it wants to.  Otherwise ... it should not store any references.

noskill: I mean by reference cython atomspace wrapper object.

noskill: https://github.com/opencog/atomspace/pull/2102

PR #2079, opened 2019-02-16
Remove code that appears to do nothing at all.

linas:
Cruft.

### DIFF
--- a/opencog/rule-engine/UREConfig.cc
+++ b/opencog/rule-engine/UREConfig.cc
@@ -32,7 +32,6 @@ using namespace opencog;
 const std::string UREConfig::top_rbs_name = "URE";
 
 // Parameters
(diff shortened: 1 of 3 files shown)


ngeiswei:
Well, it's true that code isn't doing anything today but is planned to do something in the future. No need to add it back, I'll do that when finally it's needed, and I agree that dead code is better removed, although it does feels a bit like the vandalism you like to rant about... ;-)

linas:
I'm getting set to kick out attention allocation from the atomspace. It's been a source of performance issues and bugs for a decade. It's unmaintained, no one ever uses it for anything. I can see why it could be useful, but currently, nothing in the atomspace actually depends on it. I figure that it could be a stand-alone module, and if someone comes up with some other different or better way of doing inference control, it could work along-side attention allocation.  There's no need to give it preferential treatment by tangling it into the atomspace.

PR #2084, opened 2019-02-17
Remove backwards-compat shims

linas:


linas:
should probably make sure that agi-bio and as-moses build, before merging this pull req

linas:
merging; the test case failure has nothing to do with this.

PR #2086, opened 2019-02-18
Remove unused rule-base inheritance with (Concept "URE")

ngeiswei:
This was planned to be able to inherit rule based systems but it hasn't been used so far.


ngeiswei:
I highly doubt the failure is due to that PR so I'm merging.

linas:
Yeah, cython code is crashing for unknown reasons. simple tests that just print something crash...

PR #2102, opened 2019-02-21
add destructor to cython ure classes

noskill:
Prevent deletion of parent atomspace before child atomspace in backward and forward chainers

### DIFF
--- a/opencog/cython/opencog/backwardchainer.pyx
+++ b/opencog/cython/opencog/backwardchainer.pyx
@@ -12,6 +12,8 @@ from ure cimport cBackwardChainer
 cdef class BackwardChainer:
     cdef cBackwardChainer * chainer
     cdef AtomSpace _as
+    cdef AtomSpace _trace_as
+    cdef AtomSpace _control_as
 # scheme interface
 #    (define* (cog-bc rbs target
 #                 #:key
@@ -38,12 +40,8 @@ cdef class BackwardChainer:
                                         <cAtomSpace*> (NULL if control_as is None else control_as.atomspace),
                                         deref(focus_set.handle))
         self._as = _as
-
-    # according to PEP 442, the below simply won't work.
-    # def __del__(self):
-    # def __dealloc__(self):
-    #    del self.chainer
-    #    self._as = None
+        self._trace_as = trace_as
+        self._control_as = control_as
 
     def do_chain(self):
         return self.chainer.do_chain()
@@ -51,9 +49,11 @@ cdef class BackwardChainer:
     def get_results(self):
         cdef cHandle res_handle = self.chainer.get_results()
         cdef Atom result = Atom.createAtom(res_handle, self._as)
-
-        # Delete the chainer now. There does not appear to be
-        # any other way of avoiding the mem leak. See PEP 442
-        # for details.
-        del self.chainer
         return result
+
+    def __dealloc__(self):
+        self._control_as = None
+        self._trace_as = None
+        self._as = None
+        del self.chainer
+
--- a/opencog/cython/opencog/forwardchainer.pyx
+++ b/opencog/cython/opencog/forwardchainer.pyx
@@ -49,3 +49,7 @@ cdef class ForwardChainer:
 
         result_set = self._as.add_link(types.SetLink, list)
         return result_set
+
+    def __dealloc__(self):
+        self._as = None
+        del self.chainer


noskill:
HebbianCreationModuleUTest in opencog failed

noskill:
@linas @vsbogd test is passing on local machine. What do you do to debug such failures?

linas:
> test is passing on local machine. What do you do to debug such failures?

Me? I curse, mutter, and think ugly thoughts. 

1.  Hit the re-run-tests button, see if it passes a second time.

2.  Since HebbianCreationModule has nothing to do with your change, I hit the "merge pull request" button, anyway.

3. Oh wait, maybe it does? All of the `*ModuleUTests` load cython, even if they don't use it. So maybe Cython broke? 

4. Download the Testlog file and look at actual error message.

5. If its not a broken cython, then go to step 2.  If it is, then add printf's to a commit, push that commit, look at what circle-ci printed. You can even run GDB that way. Check this, for nosetests:

```
ADD_TEST(CythonGuile gdb -batch -x /ws/atomspace/tests/cython/gdb.cmds python3)
```
and `gdb.cmds` has 
```
handle SIGPWR nostop noprint
handle SIGXCPU nostop noprint
r /usr/bin/nosetests3 -vs /ws/atomspace/tests/cython/guile/
bt
q
```

I'm going to hit "merge pull request" because I'm 99% sure this pull req is not the reason for the failure.

linas:
I hit "retest". This time, `HebbianCreationModuleUTest` passed but `ShellUTest` failed.

opencog/cython/opencog/backwardchainer.pyx
```diff
@@ -38,22 +40,20 @@ cdef class BackwardChainer:
...
+        return result
 
-        # Delete the chainer now. There does not appear to be
-        # any other way of avoiding the mem leak. See PEP 442
-        # for details.
+    def __dealloc__(self):
+        self._control_as = None
+        self._trace_as = None
+        self._as = None
         del self.chainer
```
noskill: I now thinking that chainer should be deleted before atomspaces. Could order in destructor be important?
@vsbogd What do you think?

linas: As a general programming principle, deletion should always be in the opposite order of creation. Since the chainer was created last, it should be deleted first. 

In this case, the C++ code for the chainer is holding references to the atomspace. (the BIT class...) when you delete the atomspace first, you are left with a chainer pointing at deleted memory. If you are lucky, nothing bad happens. If you are unlucky, something in the chainer accidentally touches that deleted memory, and a crash results.  That's the basic reason for the generic "last created first deleted" policy, in general. 

There are two alternatives to  "last created first deleted":

1. don't store references

2. use RAII

vsbogd: There is a case when Cython AtomSpace instance owns the pointer to the C++ AtomSpace instance and deletes C++ instance when Cython object is being collected. So I would propose change the order of statements in `__dealloc__` to:
```Python
def __dealloc__(self):
    del self.chainer
    def __dealloc__(self):
    ...
```

noskill: Ok. Also it seems necessary to keep reference to child atomspace in cython atomspace wrapper class. Otherwise parent could possibly get deleted somehow before child.

vsbogd: > Also it seems necessary to keep reference to parent atomspace in cython atomspace wrapper class. Otherwise parent could possibly get deleted somehow before child. 

There is no case, when Cython `AtomSpace` class owns C++ `AtomSpace` with pointer to parent.

noskill: It is quite possible:

```
a = opencog.atomspace.AtomSpace()
b = opencog.atomspace.create_
opencog.atomspace.create_child_atomspace(  opencog.atomspace.create_value_by_type(    
b = opencog.atomspace.create_child_atomspace(  opencog.atomspace.create_value_by_type(    
 b = opencog.atomspace.create_child_atomspace(a)
del a
terminate called after throwing an instance of 'opencog::RuntimeException'
  what():  AtomTable - deleteing atomtable with subtables! (/home/noskill/projects/atomspace/opencog/atomspace/AtomTable.cc:128)
Aborted (core dumped)
```

I'll make pr with fix

Issue #2094, opened 2019-02-19
execute_atom does not execute LambdaLink in ExecutionOutputLink using Python


stellarspot:
Here is a simple LambdaLink usage which just sums two numbers which works when written in Scheme:
```scheme
(use-modules (opencog) (opencog exec))

(Define
 (DefinedSchema "add")
 (Lambda
  (VariableList
   (Variable "$X")
   (Variable "$Y"))
  (Plus
   (Variable "$X")
   (Variable "$Y"))))

(display
 (cog-execute!
  (ExecutionOutput
   (DefinedSchema "add")
   (List
    (Number "2")
    (Number "3")))))
```
The result is (NumberNode "5.000000")

The same code written in Python which uses execute_atom prints:
```python
from opencog.type_constructors import *
from opencog.utilities import initialize_opencog
from opencog.bindlink import execute_atom

# Initialize AtomSpace
atomspace = AtomSpace()
initialize_opencog(atomspace)

(DefineLink(
    DefinedSchemaNode('add'),
    LambdaLink(
        VariableList(
            VariableNode('$X'),
            VariableNode('$Y')),
        PlusLink(
            VariableNode('$X'),
            VariableNode('$Y')))))

res = execute_atom(atomspace,
                   ExecutionOutputLink(
                       DefinedSchemaNode('add'),
                       ListLink(
                           NumberNode("3"),
                           NumberNode("4"))))
print(res)
```
Output:
```scheme
(ExecutionOutputLink
  (DefinedSchemaNode "add")
  (ListLink
    (NumberNode "3.000000")
    (NumberNode "4.000000")
  )
)
```

linas:
Yeah. Looks like a bug. Fix should be easy!? Care to learn how to fix python code?

linas:
So pull req #2100 provides a quick, cheap fix for this bug. It's deficient in multiple ways. Maybe I could get you to fix up the rest?
* It should get a unit test, so that it stays fixed.
* `execute_atom` should return a ValuePtr, not a Handle.
* It should not be called "bindlink", its got nothing to do with bindlink.

And more general python cleanup: 
* probably all of type constructors, the `execute_atom` should be moved into the main atomspace python/cython module. 
* The whole thing should be called "atomspace" not "opencog". 
* And the whole `initialize_opencog(atomspace)` is a dorky hack that should not be needed.  I mean, if you already did a "import *" you should not need to "initialize opencog" 
[...]

PR #2103, opened 2019-02-22
Feature/python execute atom return valueptr

stellarspot:
These are initial changes for the request #2094 
* a unit test is added
* `execute_atom` returns a ValuePtr

Packages renaming will be addressed in a separate fix.

### DIFF
--- a/opencog/cython/opencog/bindlink.pyx
+++ b/opencog/cython/opencog/bindlink.pyx
@@ -1,12 +1,14 @@
 from opencog.atomspace cimport Atom, AtomSpace, TruthValue
 from opencog.atomspace cimport cHandle, cAtomSpace, cTruthValue
 from opencog.atomspace cimport tv_ptr, strength_t, count_t
+from opencog.atomspace cimport handle_cast
 from cython.operator cimport dereference as deref
 
 def execute_atom(AtomSpace atomspace, Atom atom):
     if atom == None: raise ValueError("execute_atom atom is: None")
-    cdef cHandle c_result = c_execute_atom(atomspace.atomspace,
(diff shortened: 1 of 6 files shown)


opencog/cython/opencog/bindlink.pyx
```diff
@@ -1,12 +1,14 @@
...
+from opencog.atomspace cimport handle_cast
 from cython.operator cimport dereference as deref
 
 def execute_atom(AtomSpace atomspace, Atom atom):
     if atom == None: raise ValueError("execute_atom atom is: None")
-    cdef cHandle c_result = c_execute_atom(atomspace.atomspace,
+    cdef cValuePtr c_value_ptr = c_execute_atom(atomspace.atomspace,
                                            deref(atom.handle))
+    cdef cHandle c_result = handle_cast(c_value_ptr)
     return Atom.createAtom(c_result, atomspace)
```
linas: This will work only if the eturned value is actually an atom, otherwise it will crash. If the returned value is a ...value, then the `handle_cast` will return a null, and the `createAtom` will crash. For example the python version of `(cog-execute! (TruthValueOf (Concept "foo")))`   or more generally `(cog-execute! (ValueOf ("Predicate "some-key") (Concept "foo")))` assuming you set the key, previously.

linas: `examples/atomspace/stream.scm` for ValueOf

vsbogd: `create_value_by_type` can be used instead, see:
https://github.com/opencog/atomspace/blob/5b47b937b7cb1d03785042bf636ae0e2e90a5444/opencog/cython/opencog/atom.pyx#L71-L78

PR #2106, opened 2019-02-26
cython context manager to create temporary atomspaces

noskill:
cython method to create temporary atomspaces from current default one

fix \_\_contains\_\_ method for cython atomspace wrapper

### DIFF
--- a/opencog/cython/opencog/utilities.pyx
+++ b/opencog/cython/opencog/utilities.pyx
@@ -1,5 +1,7 @@
+from contextlib import contextmanager
 from opencog.atomspace cimport AtomSpace
-from opencog.type_constructors import set_type_ctor_atomspace
+from opencog.atomspace import create_child_atomspace
+from opencog.type_constructors import get_type_ctor_atomspace, set_type_ctor_atomspace
 
 # Avoid recursive intialization
 is_initialized = False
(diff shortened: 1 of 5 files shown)


opencog/cython/opencog/utilities.pyx
```diff
@@ -1,5 +1,7 @@
+from contextlib import contextmanager
 from opencog.atomspace cimport AtomSpace
-from opencog.type_constructors import set_type_ctor_atomspace
+from opencog.atomspace import create_child_atomspace
+from opencog.type_constructors import get_type_ctor_atomspace, set_type_ctor_atomspace
```
linas: Someday we should rename these to get and `set_current_atomspace` since the phrase `type_ctor` is hard to remember and confusing.  And also, probably all of `opencog.type_constructors` should be merged into `opencog.atomspace`

opencog/cython/opencog/utilities.pyx
```diff
@@ -23,3 +25,19 @@ def finalize_opencog():
...
+@contextmanager
+def tmp_atomspace():
+    """
+    Context manager, to create child atomspace from current default 
+    """
+    parent_atomspace = get_type_ctor_atomspace()
+    if parent_atomspace is None:
+        raise RuntimeError("Default atomspace is None")
+    atomspace = create_child_atomspace(parent_atomspace)
+    initialize_opencog(atomspace)
```
linas: We also need to make `initialize_opencog` go away. If you look at what it does, (`utilities.pyx`) it does two things: intialize the C++ python code, and call ` set_type_ctor_atomspace`. The first part -- initializing the C++ python code, that should happen automatically, when one does `from opencog.atomspace import foo` -- just make the `__init__()` method initialize the C++. 

I don't know if should ask you to do it, or just open another issue for this, or what; this already came up earlier in conversation with stellarspot Sorry to mention it again

noskill: Ok, it's on my todo list.

PR #2112, opened 2019-02-27
Replace result list by result set

linas:
This is a pre-requisite for other changes that are attempting
to simplify and speed up the bindlink API.


linas:
Requires opencog/opencog#3447

PR #2113, opened 2019-02-27
#2094 Handle returning Value from execute_atom method in Python

stellarspot:
* Handle returning Value from execute_atom method (#2094)
* Unit test is added

### DIFF
--- a/opencog/cython/opencog/bindlink.pyx
+++ b/opencog/cython/opencog/bindlink.pyx
@@ -11,8 +11,6 @@ def execute_atom(AtomSpace atomspace, Atom atom):
     cdef cValuePtr c_value_ptr = c_execute_atom(atomspace.atomspace,
                                            deref(atom.handle))
     type = deref(c_value_ptr).get_type()
-    if is_a(type, types.Atom):
-        return Atom.createAtom(handle_cast(c_value_ptr), atomspace)
     cdef PtrHolder ptr_holder = PtrHolder.create(<shared_ptr[void]&>c_value_ptr)
     return create_value_by_type(type, ptr_holder, atomspace)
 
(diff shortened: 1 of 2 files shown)


opencog/cython/opencog/bindlink.pyx
```diff
@@ -1,15 +1,21 @@
...
+from opencog.atomspace import is_a, types
+
 def execute_atom(AtomSpace atomspace, Atom atom):
     if atom == None: raise ValueError("execute_atom atom is: None")
     cdef cValuePtr c_value_ptr = c_execute_atom(atomspace.atomspace,
                                            deref(atom.handle))
-    cdef cHandle c_result = handle_cast(c_value_ptr)
-    return Atom.createAtom(c_result, atomspace)
+
+    if is_a(deref(c_value_ptr).get_type(), types.Atom):
```
vsbogd: Such implementation will not return correct class of Atom or Value. Please use `create_value_by_type` instead, see:
https://github.com/opencog/atomspace/blob/5b47b937b7cb1d03785042bf636ae0e2e90a5444/opencog/cython/opencog/atom.pyx#L71-L78

It handles both Atom and Value cases.

stellarspot: I updated the code to use create_value_by_type and updated the test to check that value.to_list()  works as expected.

opencog/cython/opencog/bindlink.pyx
```diff
@@ -1,15 +1,22 @@
...
 
+from opencog.atomspace import is_a, types, create_value_by_type
+
 def execute_atom(AtomSpace atomspace, Atom atom):
     if atom == None: raise ValueError("execute_atom atom is: None")
     cdef cValuePtr c_value_ptr = c_execute_atom(atomspace.atomspace,
                                            deref(atom.handle))
-    cdef cHandle c_result = handle_cast(c_value_ptr)
-    return Atom.createAtom(c_result, atomspace)
+    type = deref(c_value_ptr).get_type()
```
vsbogd: Following three lines are not needed, `create_value_by_type` will construct `Atom` instance if required

stellarspot: Updated.

PR #2117, opened 2019-03-01
Implement a parallelizable query

linas:
The `QueryLink` is (almost) exactly like the `BindLink` but it can be run in parallel threads.


examples/pattern-matcher/query.scm
```diff
@@ -0,0 +1,196 @@
...
+	(Predicate "foobar") (List (Concept "funny") (Concept "thing")))
+(Evaluation
+	(Predicate "foobar") (List (Concept "funny") (Concept "story")))
+(Evaluation
+	(Predicate "foobar") (List (Concept "funny") (Concept "joke")))
+
+; -------------
+; Define a simple query. It looks for the funny stuff, and attaches
+; the result to an AnchorNode
+(define query
```
noskill: Is this query going to run in separate thread?

linas: Yes ... the `ParallelLink` creates three threads, running that query in one of them.

linas: at bottom:
```
(define threads
   (Parallel
      (SequentialAnd
         (prtime "step one A")
         (True query)
         (True (Sleep (Number 4)))
... etc.
```

PR #2119, opened 2019-03-02
Allow ExecutionOutputLinks to return values

linas:
ExOutLinks can now return values, in general.  This brings them into line with other executable link types.


linas:
Merging. I think this is a step in the correct direction for handling generic Values in the long-term

PR #2125, opened 2019-03-07
Return correct atom class in Atom outgoing and incoming sets in Python API

vsbogd:
What is done:
- add unit test to check that `Atom.out` and `Atom.incoming` work properly
- call `create_value_by_type` to return correct `Atom` class from `Atom.out`
- rename `create_value_by_type` to `create_python_value_from_c_value` and simplify signature


linas:
I merged because it looks plausible after a quick skim. I did not review carefully. If you want careful review, (from me or others) let me know.

PR #2135, opened 2019-03-22
Bug-fix #2134

### LINKED ISSUE #2134: sql-open problems with postgres 10
sql-open command fails if using postgres 10 with the following message:
```
Backtrace:
           6 (apply-smob/1 #<catch-closure 7f466c9d41e0>)
In ice-9/boot-9.scm:
    705:2  5 (call-with-prompt _ _ #<procedure default-prompt-handle…>)
In ice-9/eval.scm:
    619:8  4 (_ #(#(#<directory (guile-user) 7f46664c3140>)))
In ice-9/boot-9.scm:
   2312:4  3 (save-module-excursion _)
  3822:12  2 (_)
In launch-cogserver.scm:
     56:0  1 (_)
In unknown file:
           0 (opencog-extension sql-open ("postgres:///test"))

ERROR: In procedure opencog-extension:
Throw to key `C++-EXCEPTION' with args `("sql-open" "Failed to execute SQL command!\nPQresult message: ERROR:  column \"increment_by\" does not exist\nLINE 1: SELECT 1            - (SELECT increment_by FROM uuid_pool)\n                                      ^\nQUERY:  SELECT 1            - (SELECT increment_by FROM uuid_pool)\nCONTEXT:  PL/pgSQL function inline_code_block line 1 at assignment\n\nPQ query was: DO $$DECLARE    under BIGINT;BEGIN   IF (SELECT count(*) FROM pg_stat_activity WHERE           datname=(SELECT current_database())) = 14 THEN      under := 1            - (SELECT increment_by FROM uuid_pool);      IF (1 < under) THEN          RAISE NOTICE 'Set uuid_pool sequence to %', under;         PERFORM (SELECT setval('uuid_pool', under));      END IF;   END IF;END $$; (/home/andres/MyOpenCogSources/atomspace/opencog/persist/sql/multi-driver/ll-pg-cxx.cc:130)\nFunction args:\n(postgres:///test)")'.
```

From this site (and others) https://help.heroku.com/WKJ027JH/rails-error-after-upgrading-to-postgres-10 it seems like there was a change in the way id sequences work in postgres now.
If not an easy fix, we should specify that a previous postgres version should be used.

linas:
This should fix bug #2134 - incompatibility between postgres version 9 and version 10.

### DIFF
--- a/opencog/persist/sql/multi-driver/SQLAtomStorage.cc
+++ b/opencog/persist/sql/multi-driver/SQLAtomStorage.cc
@@ -127,6 +127,9 @@ void SQLAtomStorage::init(const char * uri)
 
 	if (!connected()) return;
 
+	// Need the server version before init'ing the UUID pool.
+	get_server_version();
+
 	_uuid_manager.that = this;
 	_uuid_manager.reset_uuid_pool(getMaxObservedUUID());
 	_vuid_manager.that = this;
@@ -188,6 +191,15 @@ bool SQLAtomStorage::connected(void)
 	bool have_connection = db_conn->connected();
 	conn_pool.push(db_conn);
 	return have_connection;
+}
+
+/** get_server_version() -- get version of postgres server */
+void SQLAtomStorage::get_server_version(void)
+{
+	Response rp(conn_pool);
+	rp.exec("SHOW server_version_num;");
+	rp.rs->foreach_row(&Response::intval_cb, &rp);
+	_server_version = rp.intval;
 }
 
 /// Rethrow asynchronous exceptions caught during atom storage.
--- a/opencog/persist/sql/multi-driver/SQLAtomStorage.h
+++ b/opencog/persist/sql/multi-driver/SQLAtomStorage.h
@@ -71,6 +71,8 @@ class SQLAtomStorage : public BackingStore
 
 		void init(const char *);
 		std::string _uri;
+		int _server_version;
+		void get_server_version(void);
 
 		// ---------------------------------------------
 		// Handle multiple atomspaces like typecodes: we have to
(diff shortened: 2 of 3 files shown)


linas:
@glicerico I think this fixes the bug, but I can't test. Please test; if it passes, please merge. If it fails, can you fix whatever is failing about it?

glicerico:
Cool. Unfortunately, this doesn't fix the problem and I get the same error with the new name:
`ERROR:  column \"increment\" does not exist` (so I guess that website I linked was not telling the whole thing).

Digging a bit more, apparently there's a whole new catalogue that handles sequences:
https://www.postgresql.org/docs/10/catalog-pg-sequence.html

I tried modifying your changes to:
`SELECT seqincrement FROM pg_sequence WHERE seqrelid = `
but it also doesn't work and now complains about not finding column uuid_pool:
```
ERROR: In procedure opencog-extension:
Throw to key `C++-EXCEPTION' with args `("sql-open" "Failed to execute SQL command!\nPQresult message: ERROR:  column \"uuid_pool\" does not exist\nLINE 1: ...CT seqincrement FROM pg_sequence WHERE seqrelid = uuid_pool)\n                                                             ^\nQUERY:  SELECT 1            - (SELECT seqincrement FROM pg_sequence WHERE seqrelid = uuid_pool)\nCONTEXT:  PL/pgSQL function inline_code_block line 1 at assignment\n\nPQ query was: DO $$DECLARE    under BIGINT;BEGIN   IF (SELECT count(*) FROM pg_stat_activity WHERE           datname=(SELECT current_database())) = 14 THEN      under := 1            - (SELECT seqincrement FROM pg_sequence WHERE seqrelid = uuid_pool);      IF (1 < under) THEN          RAISE NOTICE 'Set uuid_pool sequence to %', under;         PERFORM (SELECT setval('uuid_pool', under));      END IF;   END IF;END $$; (/home/andres/MyOpenCogSources/atomspace/opencog/persist/sql/multi-driver/ll-pg-cxx.cc:130)\nFunction args:\n(postgres:///test)")'.
```
What would be the way to reference uuid_pool to get seqincrement from its pg_sequence?

glicerico:
BTW, I don't have permission to merge this, but I can test if you have some other idea.

linas:
OK, I'll have to read the docs, I guess.

Issue #2139, opened 2019-03-26
Protobuf causes build failure


glicerico:
Atomspace doesn't compile in Ubuntu 18.04, complaining about protobuf. but I can build properly inside an Ubuntu 16.04 docker container.
Here's the error:

```
Scanning dependencies of target dist-gearman
[ 77%] Serializing with protobuf
Scanning dependencies of target persist
Scanning dependencies of target persist-sql
[libprotobuf WARNING google/protobuf/compiler/parser.cc:562] No syntax specified for the proto file: ZMQMessages.proto. Please use 'syntax = "proto2";' or 'syntax = "proto3";' to specify a syntax version. (Defau
lted to proto2 syntax.)
Scanning dependencies of target exec
[ 77%] Building CXX object opencog/persist/guile/CMakeFiles/persist.dir/PersistSCM.cc.o
[ 77%] Building CXX object opencog/atoms/execution/CMakeFiles/exec.dir/ExecSCM.cc.o
Scanning dependencies of target zmqatoms
[ 77%] Building CXX object opencog/persist/gearman/CMakeFiles/dist-gearman.dir/DistSCM.cc.o
Scanning dependencies of target utilities_cython
[ 77%] Building CXX object opencog/persist/sql/multi-driver/CMakeFiles/persist-sql.dir/llapi.cc.o
[ 77%] Building CXX object opencog/persist/zmq/atomspace/CMakeFiles/zmqatoms.dir/ZMQMessages.pb.cc.o
[ 77%] Building CXX object opencog/cython/opencog/CMakeFiles/utilities_cython.dir/Utilities.cc.o
In file included from /home/andres/MyOpenCogSources/atomspace/build/opencog/persist/zmq/atomspace/ZMQMessages.pb.cc:4:0:
/home/andres/MyOpenCogSources/atomspace/build/opencog/persist/zmq/atomspace/ZMQMessages.pb.h:12:2: error: #error This file was generated by a newer version of protoc which is
 #error This file was generated by a newer version of protoc which is
  ^~~~~
/home/andres/MyOpenCogSources/atomspace/build/opencog/persist/zmq/atomspace/ZMQMessages.pb.h:13:2: error: #error incompatible with your Protocol Buffer headers. Please update
 #error incompatible with your Protocol Buffer headers.  Please update
  ^~~~~
/home/andres/MyOpenCogSources/atomspace/build/opencog/persist/zmq/atomspace/ZMQMessages.pb.h:14:2: error: #error your headers.
[...]

PR #2141, opened 2019-03-27
Comment out references to protobuf and zmq

glicerico:
Fixing issue #2139 , commented out references to protobuf and zmq in CMakeFiles.txt . @linas , this workaround fixes the issue, but let me know if I should do it in a different way.


linas:
This is in no sense a "fix", its a terrible hack. It's got more than a few problems-- you comment out much much more than you need to, and you fail to identify the root cause. It's OK for you to trash your own systems, but this is not acceptable for systems that others depend on.

PR #2142, opened 2019-03-28
Commented out flag for protobuf

glicerico:
Fixing issue #2139 , commented out the line setting HAVE_PROTOBUF to true in CMakeLists.txt


linas:
@ngeiswei @TScottJ @AmeBel  @mjsduncan @Selameab @Dagiopia @skoumas -- question for you -- so - I believe that the REST interfaces, despite being effectively obsolete and unmaintained, are still actively used in two or three or four projects -- the atomspace visualizer (in the opencog/external-tools repo)? the shanghai agi-bio project? the teaching tablet? one or more iCogLabs projects?  Is this the case, or are all of these abandoned projects?

Merging this pull request would cause the REST interfaces for the atomspace to not build any more. This would cause the dependent projects to break. Is this acceptable? 

Perhaps the best short-term solution would be to move the REST code over to the opencog/external-tools repo?

ngeiswei:
I've never needed REST but others might need it, I can't tell.

AmeBel:
@linas The change will not affect the REST api

AmeBel:
It will affect https://github.com/opencog/opencog-neo4j but since it is not maintained, it is a non-issue.

The REST api is found at https://github.com/opencog/opencog/tree/master/opencog/python/web/api

TScottJ:
Yes the latest visualizer still uses the REST interface, but as @AmeBel said we should be fine as that all is in /python/web now.

linas:
Oh. Huh. OK. So I'm thinking that, in terms of simplicity and maintainability, it would be best to remove protobuf and neo4j support? Any opinions on that?

linas:
closing; the original issue was resolved.

PR #2144, opened 2019-03-29
Add case when vardecl is Handle::UNDEFINED in filter_vardecl comment

ngeiswei:


linas:
As a general rule, I've been trying to remove checks for error conditions, and replace them with exceptions.  The idea is that its always faster to handle an exception once a week, instead of performing `if(h == Handle::UNDEFINED)` a million times a second.

Of course, this works only if the client knows how to deal with the exception.  And also, I guess that, for some utilities, maybe having `Handle::UNDEFINED` actually makes sense?

ngeiswei:
Yeah, maybe using exceptions would be better in this instance, we'd have to study the code to determine whether we want to consider this as an error or as just another case.

linas:
OK, thanks, yeah this is not urgent. Just a sort-of "keep it in mind" reminder.

Issue #2054, opened 2019-02-06
Rename rule-engine guile module to ure


ngeiswei:
As the URE python module is gonna be called `ure` is would be good if the guile module is called `ure` as well. I suppose one could keep `rule-engine` too for as long as needed to preserve backward compatibility.

PR #2146, opened 2019-04-02
Rename rule-engine to ure

ngeiswei:
- Rename rule-engine module to ure
- Rename rule-engine directories to ure
- Add back rule-engine for backward compatibility only

From now on the rule engine is imported with

```scheme
(use-module (opencog ure))
```

similar to that of the Python ure module.

A rule-engine module has been kept for backward compatibility though.


ngeiswei:
Fixes #2054 

I don't expect anybody to be against it so I'm merging. Besides, as I said `rule-engine` can still be used. More controversial might be the renaming of the `rule-engine` directories to `ure`, but it only causes a C++ backward compatibility break and is more cohesive that way.

PR #2148, opened 2019-04-11
Use open(...).read() for exec method

stellarspot:
Add read() call for open(...) method to fix exception:
"python-eval" "Python error: exec() arg 1 must be a string, bytes or code object.

Actually it works if I run the python.scm file from the directory where it is placed.
If I call it from different directory I get exception:
 "python-eval" "Python error: [Errno 2] No such file or directory: 'my_py_func.py'.


linas:
thanks. I think its a python2 vs python3 incompatibility.

PR #2149, opened 2019-04-16
Add cog-get-all-roots in base utilities

ngeiswei:
I found this to be useful to the pattern miner, maybe it can be useful outside of it.

### DIFF
--- a/opencog/scm/opencog/base/utilities.scm
+++ b/opencog/scm/opencog/base/utilities.scm
@@ -333,6 +333,13 @@
 
 ; -----------------------------------------------------------------------
 (define-public (cog-get-all-roots)
+"
+  cog-get-all-roots -- Return the list of all root atoms.
+
+  cog-get-all-roots
+  Return the list of all root atoms, that is atoms with
+  no incoming set.
+"
   (define roots '())
   (define (cons-roots x) (set! roots (cons x roots)))
   (traverse-roots cons-roots)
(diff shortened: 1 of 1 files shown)


ngeiswei:
@linas, I let you to decide whether you want to add it to the base utilities.

linas:
1) you should provide documentation for this

2) A better implementation might be to change `traverse-roots` slightly:

* make it `define-public` instead of `define`
* replace the final `for-each` with the srfi-1 `filter-map`

That way, the implementation for `(cog-get-all-roots)` would be a single line of code:
```
(define (cog-get-all-roots) (traverse-roots (lambda (x) #f)))
```

linas:
Other than point 1) - missing docs, I'm OK with it.

ngeiswei:
@linas, I like the idea, but do we want to replace `for-each` by `filter-map` or `map`? And if so why?

For instance `cog-prt-atomspace` would be easier to implement if `for-each` is replaced by `map` (by discarding the results with `*unspecified*`). I don't have much strong feeling about it... What do you think?

ngeiswei:
OK, I get why you suggest a `filter-map`, because that way `func` has the same output type as `apply-if-root`, which is the algo a bit simpler. Anyway, I'm fine with `filter-map`.

ngeiswei:
Actually the following doesn't work
```scheme
(define (cog-get-all-roots) (traverse-roots (lambda (x) #f)))
```
because `filter-map` ignores false values. There's certainly a way to make it work but it's not high on my priority list so, I'll just add a comment to `cog-get-all-roots` for now.

linas:
OK. yeah, `cog-map-type` is mis-named. Long run, best design might be some kind of root-only traversal in C++ but its too much work with too little payoff right now.

PR #2151, opened 2019-04-19
Scheme documentation updates

linas:
Point out that some of the API calls are nearly-trivial wrappers.


linas:
Haskell unit tests are failing -- but not cause of this pull-req, which changes documentation only. Merging anyway.

PR #2158, opened 2019-04-30
Handle the case where And, Or or Not are used as logical connectors

ngeiswei:
LocalQuote at the root of the pattern body should not be consumed when
wrapping logical connectors And, Or or Not, otherwise the semantics of
the pattern is modified.

### DIFF
--- a/opencog/atoms/core/RewriteLink.cc
+++ b/opencog/atoms/core/RewriteLink.cc
@@ -434,7 +434,7 @@ Handle RewriteLink::consume_quotations(const Variables& variables,
 	                       (t == PUT_LINK or
 	                        nameserver().isA(t, FUNCTION_LINK) or
 	                        nameserver().isA(t, EVALUATION_LINK) or
-	                        (t == AND_LINK and clause_root) or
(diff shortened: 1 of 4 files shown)


ngeiswei:
The following unit tests are failing
```
	 54 - MultiUserUTest (Failed)
	 55 - FetchUTest (Failed)
```
I'm merging, as it's pretty clear the unit tests are not related to that PR, and I think the PR itself is uncontroversial enough.

linas:
> unit tests are failing

That's due to a test-harness bug. Circleci is trying to run these tests in parallel; both of them write incompatible data to the database, they clobber one-another. They must only be run in serial order.  However, there does not seem to be any way of specifying, in a cmake makefile, that they must run serially. Googling this, many people have tripped over this in many projects; there does not seem to be any solution at all. (I tried forcing dependencies in cmake; it didn't work). Well - there is a solution - change the circleci config to not run tests in parallel. ...

ngeiswei:
Maybe I'm at fault here, as I've enabled 2 jobs to complete unit tests faster

https://github.com/opencog/atomspace/blob/master/.circleci/config.yml#L53

PR #2160, opened 2019-05-07
atom-to-string scheme function

Dagiopia:
not sure if it's needed very much from your end or a similar functionality exists but I needed such a utility function and couldn't find one.

### DIFF
--- a/opencog/guile/SchemeSmob.cc
+++ b/opencog/guile/SchemeSmob.cc
@@ -313,6 +313,7 @@ void SchemeSmob::register_procs()
 	register_proc("cog-number",            1, 0, 0, C(ss_number));
 	register_proc("cog-type",              1, 0, 0, C(ss_type));
 	register_proc("cog-arity",             1, 0, 0, C(ss_arity));
+	register_proc("cog-to-str",            1, 0, 0, C(ss_to_string));
 	register_proc("cog-incoming-set",      1, 0, 0, C(ss_incoming_set));
 	register_proc("cog-incoming-by-type",  2, 0, 0, C(ss_incoming_by_type));
 	register_proc("cog-outgoing-set",      1, 0, 0, C(ss_outgoing_set));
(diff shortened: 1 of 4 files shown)


ngeiswei:
I find strange we don't have this already, @linas?

Also, a better name would be
```
atom->string
```

Dagiopia:
yes that's a better name.

opencog/guile/SchemeSmob.cc
```diff
@@ -313,6 +313,7 @@ void SchemeSmob::register_procs()
 	register_proc("cog-number",            1, 0, 0, C(ss_number));
 	register_proc("cog-type",              1, 0, 0, C(ss_type));
 	register_proc("cog-arity",             1, 0, 0, C(ss_arity));
+	register_proc("atom->string",            1, 0, 0, C(ss_to_string));
```
linas: mis-matched whitespace indentation

linas:
I don't think you need this function ... or rather, you can implement it in one line of scheme code as
```
(define (atom->string ATOM) (format #f "~A" ATOM))
```

Recall that `format` prints things (including atoms) and that `(format #t ...)` prints to stdout, and `(format #f ...)` prints to a string, and `(format DEV ...)` prints to an open file descriptor.

I'm closing this, because we don't need C++ code to do this; it already works

linas:
FYI, the existing printing code is in `SchemeSmobGC.cc` from line 98 to the end-of-file.

Dagiopia:
> I don't think you need this function ... or rather, you can implement it in one line of scheme code as
> 
> ```
> (define (atom->string ATOM) (format #f "~A" ATOM))
> ```
> 
> Recall that `format` prints things (including atoms) and that `(format #t ...)` prints to stdout, and `(format #f ...)` prints to a string, and `(format DEV ...)` prints to an open file descriptor.
> 
> I'm closing this, because we don't need C++ code to do this; it already works

`format` is just great! plus, unlike `oc_to_string` the id isn't printed wiith `format` 
sorry for the premature edit. will use this instead.

linas:
no problem. Take a look at https://www.gnu.org/software/guile/manual/html_node/Formatted-Output.html for more.

linas:
and in case you want to print to something other than a string, or to stdout, see https://www.gnu.org/software/guile/manual/html_node/Default-Ports.html

ngeiswei:
BTW, the opencog logger supports format, so if you do
```scheme
(cog-logger-debug "my atom is ~a" atom)
```
it will log the content of `atom`.

ngeiswei:
See https://wiki.opencog.org/wikihome/index.php/Scheme#Logger for more info.

Dagiopia:
Thank you very much both! this is very helpful.

PR #2161, opened 2019-05-09
Allow clang/libc++ compilation

maparent:
I am using gcc-9 from homebrew
I have not installed haskell, so not sure if that works
There is still a path error when installing cython libraries
This allows building and most tests pass, except 27 that require libsmod. I will need help for this.

Alternative: make it work on latest clang, as per issue #2163 
The commits and comments until 2019-05-09 pertain to gcc, the ones from 2019-05-11 pertain to clang, running with 
`env CXXFLAGS="-std=c++11" LDFLAGS="-lstdc++ -undefined dynamic_lookup" cmake ..`


maparent:
```
The following tests FAILED:
	 16 - ParallelUTest (Child aborted)
	 17 - RandomUTest (Child aborted)
	 21 - PutLinkUTest (Child aborted)
	 26 - DefineLinkUTest (Child aborted)
	 27 - DeleteLinkUTest (Child aborted)
	 28 - EqualLinkUTest (Child aborted)
	 30 - StateLinkUTest (Child aborted)
	 39 - CheckersUTest (Child aborted)
	 44 - TVMergeUTest (Child aborted)
	 50 - SCMPrimitiveUTest (Child aborted)
	 54 - SCMExecutionOutputUTest (Child aborted)
	 55 - SCMOpencogTestRunnerPass (Failed)
	 58 - SCMPythonSharedAtomspace (Failed)
	 59 - SCMPythonArity (Failed)
	 60 - VectorAPIUTest (Child aborted)
	 70 - DisconnectedUTest (Child aborted)
	 77 - VarTypeNotUTest (Child aborted)
	 84 - ArcanaUTest (Child aborted)
	 85 - SubstitutionUTest (Child aborted)
	 88 - GetStateUTest (Child aborted)
	 92 - SequenceUTest (Child aborted)
	 95 - QuoteUTest (Child aborted)
	 97 - LocalQuoteUTest (Child aborted)
	 99 - BuggyQuoteUTest (Child aborted)
	102 - BuggyBindLinkUTest (Child aborted)
	106 - AbsentUTest (Child aborted)
	107 - ScopeUTest (Child aborted)
	108 - NestedPutUTest (Child aborted)
	112 - NoExceptionUTest (Child aborted)
	117 - ForwardChainerUTest (Child aborted)
	118 - BackwardChainerUTest (Child aborted)
Errors while running CTest
```

linas:
Never heard of `libsmod` google doesn't know about it either.

linas:
what's the error message?

maparent:
typo: I meant libsmob.so (sorry)
The error messages are always SIGABORT.
Example backtrace:

```
lldb ./tests/atoms/RandomUTest                                                                                                       Jeu  9 mai 15:55:25 2019
(lldb) target create "./tests/atoms/RandomUTest"
  File "<input>", line 1, in <module>
  File "/usr/local/Cellar/python@2/2.7.16/Frameworks/Python.framework/Versions/2.7/lib/python2.7/copy.py", line 52, in <module>
    import weakref
  File "/usr/local/Cellar/python@2/2.7.16/Frameworks/Python.framework/Versions/2.7/lib/python2.7/weakref.py", line 14, in <module>
    from _weakref import (
ImportError: cannot import name _remove_dead_weakref
Current executable set to './tests/atoms/RandomUTest' (x86_64).
(lldb) run
Process 41539 launched: '/Users/maparent/OpenSource/atomspace_maparent/build/tests/atoms/RandomUTest' (x86_64)
Running cxxtest tests (2 tests)[2019-05-09 19:55:40:881] [DEBUG] BEGIN TEST: test_weights
Process 41539 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x00007fff64a342c6 libsystem_kernel.dylib`__pthread_kill + 10
libsystem_kernel.dylib`__pthread_kill:
->  0x7fff64a342c6 <+10>: jae    0x7fff64a342d0            ; <+20>
    0x7fff64a342c8 <+12>: movq   %rax, %rdi
    0x7fff64a342cb <+15>: jmp    0x7fff64a2e457            ; cerror_nocancel
    0x7fff64a342d0 <+20>: retq
Target 0: (RandomUTest) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0x00007fff64a342c6 libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff64ae9bf1 libsystem_pthread.dylib`pthread_kill + 284
    frame #2: 0x00007fff6499e6a6 libsystem_c.dylib`abort + 127
    frame #3: 0x0000000100b7926c libgcc_s.1.dylib`uw_init_context_1.cold + 5
    frame #4: 0x0000000100b77187 libgcc_s.1.dylib`_Unwind_Resume + 55
    frame #5: 0x00000001002e1780 libatomcore.dylib`opencog::UniqueLink::get_unique(opencog::Handle const&, unsigned short, bool) (.cold) + 22
    frame #6: 0x00000001002cba49 libatomcore.dylib`opencog::StateLink::get_link(opencog::Handle const&) + 25
    frame #7: 0x0000000100157594 libatomspace.dylib`opencog::AtomTable::add(std::shared_ptr<opencog::Atom>, bool, bool) + 4420
    frame #8: 0x000000010014a6e6 libatomspace.dylib`opencog::AtomSpace::add_link(unsigned short, std::vector<opencog::Handle, std::allocator<opencog::Handle> > const&, bool) + 150
    frame #9: 0x000000010010d298 libsmob.so`opencog::SchemeSmob::ss_new_link(scm_unused_struct*, scm_unused_struct*) + 392
    frame #10: 0x00000001006ee862 libguile-2.2.1.dylib`vm_regular_engine + 1263
    frame #11: 0x00000001006ed5aa libguile-2.2.1.dylib`scm_call_n + 380
    frame #12: 0x00000001006a1be8 libguile-2.2.1.dylib`scm_primitive_load_path + 904
    frame #13: 0x00000001006ee862 libguile-2.2.1.dylib`vm_regular_engine + 1263
    frame #14: 0x00000001006ed5aa libguile-2.2.1.dylib`scm_call_n + 380
    frame #15: 0x000000010068504b libguile-2.2.1.dylib`scm_call_3 + 50
    frame #16: 0x00000001006ee862 libguile-2.2.1.dylib`vm_regular_engine + 1263
    frame #17: 0x00000001006ed5aa libguile-2.2.1.dylib`scm_call_n + 380
    frame #18: 0x00000001006eae92 libguile-2.2.1.dylib`catch + 490
    frame #19: 0x00000001001020e6 libsmob.so`opencog::SchemeEval::do_eval(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 358
    frame #20: 0x00000001001021ac libsmob.so`opencog::SchemeEval::c_wrap_eval(void*) + 28
    frame #21: 0x0000000100680422 libguile-2.2.1.dylib`c_body + 15
    frame #22: 0x00000001006ee862 libguile-2.2.1.dylib`vm_regular_engine + 1263
    frame #23: 0x00000001006ed5aa libguile-2.2.1.dylib`scm_call_n + 380
    frame #24: 0x00000001006eae92 libguile-2.2.1.dylib`catch + 490
    frame #25: 0x00000001006803b1 libguile-2.2.1.dylib`scm_i_with_continuation_barrier + 137
    frame #26: 0x0000000100680408 libguile-2.2.1.dylib`scm_c_with_continuation_barrier + 60
    frame #27: 0x0000000102b0544d libgc.1.dylib`GC_call_with_gc_active + 294
    frame #28: 0x00000001006eaa3d libguile-2.2.1.dylib`with_guile + 150
    frame #29: 0x0000000102b005c8 libgc.1.dylib`GC_call_with_stack_base + 26
    frame #30: 0x00000001006e8f2d libguile-2.2.1.dylib`scm_with_guile + 45
    frame #31: 0x000000010010216e libsmob.so`opencog::SchemeEval::eval_expr(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 46
    frame #32: 0x000000010000cac6 RandomUTest`RandomUTest::test_weights() + 230
    frame #33: 0x0000000100008f61 RandomUTest`CxxTest::RealTestDescription::run() + 17
    frame #34: 0x000000010000c4ac RandomUTest`int CxxTest::Main<CxxTest::ErrorPrinter>(CxxTest::ErrorPrinter&, int, char**) + 2588
    frame #35: 0x00000001000105ec RandomUTest`main + 156
    frame #36: 0x00007fff648f93d5 libdyld.dylib`start + 1
    frame #37: 0x00007fff648f93d5 libdyld.dylib`start + 1
```
Just noticed, quite strange that we have `python2.7` in the path, both `RandomUTest` and `/usr/local/lib/opencog/libPythonEval.dylib` clearly link against Python3.7.
I am currently able to run the guile and python base examples, for the record.

maparent:
P.S. You can contact me on skype with a username equal to my github handle if you want to look at this with me. Or @gmail.com if you want to set a good time to do it (I'm in EDT timezone.)

linas:
> skype

How about IRC?

> python@2  ... _remove_dead_weakref

Please fix that .. it is probably the root cause.  Because ... in your stack trace:

> frame #5: 0x00000001002e1780 libatomcore.dylib`opencog::UniqueLink::get_unique(opencog::Handle const&, unsigned short, bool) (.cold) + 22

So just two insns inside of `UniqueLink::get_unique` an exception is thrown.  Why, the very first thing that `get_unique` does is `IncomingSet defs = alias->getIncomingSetByType(type);` which is actually a loop over weak refs! (Not obvious, but I happen to know this cause I wrote the code). 

One weak-ref message is rare. Two unrelated weak-ref messages is an astronomical rarity.

linas:
And if you want a run-it-by-hand sanity check, all that `./tests/atoms/RandomUTest` is doing is to load the file `tests/atoms/random-choice.scm` and then verify return values.  So load that by hand, it should load without error.  I'm not sure, there might might be an existing example that is almost the same.

maparent:
Well. I solved the python2 issue: lldb loads python2, and got confused between my system python and the homebrew python. But even with the message eliminated, the segfault happens. So maybe a coincidence after all. I tried executing `tests/atoms/random-choice.scm` line by line, and no problem. 
I wonder if there is interference between components built with clang (like guile) and gcc9.
But I did check that the `libgcc_s.1.dylib` in the stack trace is gcc9's. 
I logged onto IRC, we can try to dig deeper tomorrow.

ngeiswei:
@maparent, here are the various ways you can get in touch

https://wiki.opencog.org/w/The_Open_Cognition_Project#Participate

@linas, I've removed the IRC channel from the list as it seems to no longer exist.

linas:
@ngeiswei three people posted on the IRC channel in the last week. (no one stick around long enough to get an answer, though)

linas:
> But I did check that the libgcc_s.1.dylib in the stack trace is gcc9's.
>  components built with clang (like guile) and gcc9.

Yes, absolutely there will be interference, especially for things like exceptions.  Now, there is a general claim that both compilers generate compatible binary objects, according to the ABI and for the most part that is true. However: 

1) the interaction between exceptions, threading and debugging is quite complex, and I can see how there might be bugs/incompatibilities there. Also, exceptions, threading and debugging also require cooperation from the C library, which provides the linker-loader. The MacOS C library differs from glibc

2) Each compiler/C-library combination specifies different header files. Things like reference counting and weak pointers will be implemented in those header files, and they will be different.  In particular, gcc requires libstdc++ which clang requires libc++ and these two are NOT compatible with one-another (and that is where the implementation of things like weak pointers will live)

net-net -- do not mix-n-match compilers; do a `make clean` in between.

maparent:
Yes, but it's beyond the issue of make clean; I have homebrew components like guile or system components like Python that are built with clang, and trying to make them use modules from atomspace built with gcc9 is not working as well as it should. Hence the ongoing effort to build everything with clang.

opencog/guile/SchemePrimitive.h
```diff
@@ -66,8 +68,39 @@ namespace std {
   /// Alias template index_sequence_for
   template<typename... _Types>
     using index_sequence_for = make_index_sequence<sizeof...(_Types)>;
-}
 
+#elif defined(_LIBCPP_VERSION)
+	// Copied/pasted  from libc++ v1 utility
+	template<class _Tp, _Tp... _Ip>
+	struct _LIBCPP_TEMPLATE_VIS integer_sequence
+	{
```
linas: as far as I can tell, `integer_sequence` is not used anywhere at all. Do you really need this?  What happens if it is removed?

linas: never mind. I just can't read the code.

opencog/guile/SchemePrimitive.h
```diff
@@ -24,11 +24,13 @@
 #include <opencog/atoms/truthvalue/TruthValue.h>
 #include <opencog/guile/SchemeSmob.h>
 
-// Copied/pasted from gcc 4.9 utility. Remove as soon as C++14 is
-// enabled and gcc 4.9 is a minimum requirement.
+// Remove as soon as C++14 is enabled and gcc 4.9 or clang is a minimum requirement.
```
linas: Sooo ... we require gcc version 5 in `cogutils-git/cmake/OpenCogGccOptions.cmake`  so in fact, all of this can be removed, right? Or do you still need this for clang?

maparent:
Status: There's still an obscure compilation issue around AndBIT::HandleBITNodeMap. But if I comment out the relevant lines of code, most tests pass! One notable exception is in `./tests/atoms/MapLinkUTest` where the order changes unexpectedly in `test_local_quote_map`. Is that something that could be platform-dependent?

```
.[2019-05-11 17:00:45:760] [DEBUG] BEGIN TEST: test_local_quote_map
result (ListLink
  (ConceptNode "B") ; [3464090702952180973][1]
  (ConceptNode "A") ; [1447400368074742694][1]
) ; [14267999396490119281][1]
expected (ListLink
  (ConceptNode "A") ; [1447400368074742694][1]
  (ConceptNode "B") ; [3464090702952180973][1]
) ; [12038113228570855210][1]


In MapLinkUTest::test_local_quote_map:
/Users/maparent/OpenSource/atomspace_maparent/build/tests/atoms/../../../tests/atoms/MapLinkUTest.cxxtest:317: Error: Assertion failed: result == expected
[2019-05-11 17:00:45:762] [DEBUG] END TEST: test_local_quote_map
```

opencog/guile/SchemePrimitive.h
linas: ah, but we still ask for c++11 not c++ 14.  ... I'm going to change that now...

maparent: I need this for clang right now because I'm telling clang to use C++11. Do you want me to try with C++14?

maparent: I can confirm that I can build with c++14 with clang, with the same set of errors as c++11.

maparent: That said... I noticed errors before I also recompiled cogutils with c++14. So other people could have mix-and-match issues with libraries using different c++ versions. (Sigh.) So I would tend to favour keeping c++11 compatibility while possible. You call, of course.

linas: Hmm. atomspace and opencog have almost no dependencies at all, just guile, python and boost, and there's no C++ in those first two. The ABI is supposed to guarantee that things compiled with c++11 can link & run correctly with things compiled with c++14.... Anyway, the compiler flags are in cogutils, anyway, so there is no way to compile the atomspace with different compiler flags, unless you hack them in.

My goal was to get rid of the integer-sequence code.

maparent:
Ok. My last modification is a bit more radical; I had to redefine `Unify::TypedSubstitution` as a pair, since the original `::value_type` definition has an implicit `const` in the key, which means that it cannot do copy assignment, which means that it cannot be used in `AndBIT::HandleBITNodeMap`. Open to discuss other ways to do it. But all tests now pass on clang, except `test_local_quote_map` in `MapLinkUTest` as described above, which I suspect is an overly rigid assumption about ordering.

linas:
You've made many different changes here' I'm thinking it might be easier (for Nil & I) if this was busted up into multiple pull reqs. Most of the changes are no controversial; I few of them. I'd like to understand better (e.g. the python change, the de-constification of things.

maparent:
I made it a point to break down the changes in meaningful commits; I could break it down further if desired. (Please confirm.) The de-constification is because any structure with any const field cannot have an `operator=`. And without an `operator=`, it cannot be part of a map, set, etc. Clang seems to be more rigid about enforcing that than gcc, but it's actually correct. I get fairly obscure error messages about not finding an appropriate `operator=` for this or that, but it boils down to killing the const.
The python change: First the technique of chopping off /usr and pre-pending CMAKE_INSTALL_PREFIX failed on mac. The path I got from sysconfig was directly usable.
However, I would also like to understand better what you were trying to achieve with that chop/prepend, maybe it can be repaired rather than suppressed.
I realize now that sysconfig is not as useful on unix, and that the distutils.sysconfig may have been a better bet; I'll revise that. (I thought it was wrong on mac, but I realize I got confused by the other issue.)
(I'll also have to find a way to discover the right location for guile modules, it is elsewhere on mac.)
Still interested in discussing some of those issues synchronously somewhere.

maparent:
Not mac-specific, but I had some issues with guile loading, so I looked into libsmob.so installation. It's put in `lib/opencog`, and the `LTDL_LIBRARY_PATH` is manipulated to include that. Though it does work, I would argue for instead using the result of 
`pkg-config guile-2.2 --variable=extensiondir`
as an install location. If you agree, I'll make that a separate PR.

maparent:
I altered the python commit, back to `distutils.sysconfig`. We still need to discuss why you thought it necessary to prepend `CMAKE_INSTALL_PREFIX `.

ngeiswei:
@linas said
> @ngeiswei three people posted on the IRC channel in the last week. (no one stick around long enough to get an answer, though)

Oh, it's still up indeed

http://irc.netsplit.de/channels/details.php?room=%23opencog&net=freenode

I wonder why I can't access it... I'll put it back on the wiki then...

linas:
>  const field cannot have an operator=
Yes, that is the point -- we want to delete that operator, so that those things are not assignable.  It helps prevent bugs.  

> without an operator=, it cannot be part of a map, set, 
I don't understand what that means ... again, the whole point is that when it gets placed into the map, set, etc. It should be const -- and not assignable.  That's why we want operator= to get blanked out --   I'm guessing that you are tripping over some kind of clang bugs; in the long term, though, we really do want these to be const, and to use const as often as possible, in as many places in the code as possible.

> The python change
We have a very long history of hard-to-debug python bugs. No one wants to be the python maintainer. So I am just being conservative here, figuring that some change, any change to python will probably break something somewhere for someone, and since no one wants to fix those bugs .. yeeech.  If this change was in a distinct pull req, then, in principle, it would be easier to find/check/revert that one change only, instead of trying to figure out what all the other changes in the pull req are.

linas:
> pkg-config guile-2.2 --variable=extensiondir

That sounds like a good idea. 

Is there anything similar to that for python? 

>  why you thought it necessary to prepend CMAKE_INSTALL_PREFIX .

I don't think that at all. I think that is just some historical piece of cruft dating back to the dawn of time.  The last two people to monkey with the python bindings were @noskill  and @vsbogd and maybe they should take a look at that, see if/how this change affects them. They actively use python; I don't.

maparent:
const: There are two interrelated issues. Mac OS is using clang and libc++ vs gcc and libstdc++. I do not know if it is possible to use clang with libstdc++. But it turns out that, in libc++, sets and trees actually use the assignment operator when they rebalance, and that runs afoul const. So it's not a clang bug, clang is actually being very accurate; you could well call it a libc++ design flaw.
If you hunt around stackoverflow and reddit, you see people advising against const member variables. They're taking a stance. You believe in the value of const. I see the value of both stances.
I would propose a pragmatic compromise: Let's design a "maybeconst" macro that resolves to const or nothing, depending on libstdc++ vs libc++.
Would you be open to that?

Python: I will move that change in a separate PR. I'm very comfortable with python.
May do all this tomorrow.

linas:
> sysconfig vs.  distutils.sysconfig

Again - a python thing. I have no clue. I am vaguely aware that ubuntu and debian install python modules in some other directory, than the one that python itself uses, and that this has historically caused an endless string of confusion and bug reports, affecting not just opencog, but other projects as well. I don't know what the best solution for this is.  Again, @noskill and @vsbogd are the primary python users.

linas:
> I would propose a pragmatic compromise: 

You know, I actually don't care all that much, in the end. For code readability, just removing const is a lot easier to read/understand/remember than having a `maybeconst` macro. And I tend to prefer code that is easier to read/understand. I do not beleive that there is much of (if any) performance impact, so just removing const is fine.

linas:
p.s. when you are ready, just please remove the WIP from the title; I think this can be merged.

maparent:
Readability matters, yes. Another offer: private variable and public getter? (Only works for one of the cases...) but if you think it's mergeable... I'll take the python out and rebase.

maparent:
python taken out.

maparent:
Note that there remains the issue of the failing test because of `test_local_quote_map` ordering.

ngeiswei:
I'm fine with removing the consts.

Let me have a look at `test_local_quote_map`...

Issue #2159, opened 2019-05-07
AtomSpace too early dtored by guile


ngeiswei:
## Overview

While using the pattern miner, guile prematurely dtors a temporary atomspace holding KB + results.

## How to reproduce

In the following example https://github.com/opencog/opencog/tree/master/examples/learning/miner/sumo here's what seems to happened (best use commit
https://github.com/opencog/opencog/commit/6474595b631348d7f895eed0b52a05eec6f6fe62 to guaranty reproducibility):

1. Current atomspace is `parent-as`
2. Create `tmp-as` as child of `parent-as` https://github.com/opencog/opencog/blob/6474595b631348d7f895eed0b52a05eec6f6fe62/opencog/learning/miner/miner-utils.scm#L492
3. Set `tmp-as` as current atomspace https://github.com/opencog/opencog/blob/6474595b631348d7f895eed0b52a05eec6f6fe62/opencog/learning/miner/miner-utils.scm#L493
4. Create a list of atoms as corpus to mine in `tmp-as`
5. Miner `tmp-as` and put results in `tmp-as`
6. Set `parent-as` as current atomspace https://github.com/opencog/opencog/blob/6474595b631348d7f895eed0b52a05eec6f6fe62/opencog/learning/miner/miner-utils.scm#L533
7. Guile considers `tmp-as` is no longer useful and deletes it, as a result most results turn into invalid atoms.

If https://github.com/opencog/atomspace/blob/af9395ab81b57c2a03a836bc0cbeecb5fdcc07a3/opencog/guile/SchemeSmobGC.cc#L63 is commented out the bug goes away.

ngeiswei:
Certainly there must be a way to simply the test case to reproduce the bug, but I'm leaving that for later as it is not current the highest task on my list.

linas:
Umm, "working as designed" aka "not a bug"? Since the symbol `tmp-as` is never referenced again, it is impossible to access anything in the `tmp-as` -- that atomspace is dead to the world, and so can be safely deleted. 

I presume you made C++ lists/sets/arrays of the atoms that you found, and/or scheme lists. As long as   you have C++ Handle's to the atoms, those Handles will continue to point at valid atoms; but those atoms will not belong to any atomspace (the atomsapce pointer will be null).  If you have scheme lists holding atoms, then those atoms should still be there. If you have neither, then the atoms will go "poof" and be garbage-collected, since nothing can ever possibly access them, from either C++ or guile.

[...]

PR #2162, opened 2019-05-10
Add warning in cog-set-atomspace! about deleting previous atomspace

ngeiswei:
Add warning in `cog-set-atomspace!` about previous atomspace being deleted if not referenced.


ngeiswei:
Relates to #2159

PR #2164, opened 2019-05-11
Attempt to force explicit crashes on dereferences atomspaces

linas:
This pertains to bug https://github.com/opencog/atomspace/issues/2159


linas:
@ngeiswei this is a patch intended to explicitly force any bugs in the pattern miner to trigger immediately, instead of getting delayed.  Could you try it out, and see if it helps isolate the root cause of yor crashes?

ngeiswei:
@linas, actually I have already modified the user API of the pattern miner so I don't have to worry about that problem anymore.

vsbogd:
Is it still actual?

PR #2165, opened 2019-05-14
PYTHON_DEST starts with /usr/local vs /usr on mac

maparent:
The path wrangling makes assumptions that are not valid on mac. Deleting it works on my mac and on an ubuntu machine (and in the test environment). Some of the discussion happened in https://github.com/opencog/atomspace/pull/2161 . Can @noskill and/or @vsbogd please take a look?


CMakeLists.txt
```diff
@@ -266,8 +266,6 @@ IF(HAVE_PY_INTERP)
 			OUTPUT_VARIABLE PYTHON_DEST
 		)
 
-		# Chop off leading /usr
-		STRING(REPLACE "/usr" ${CMAKE_INSTALL_PREFIX} PYTHON_DEST "${PYTHON_DEST}")
```
vsbogd: Prefixing `PYTHON_DEST` by `CMAKE_INSTALL_PREFIX` makes sense for a case when one would like to have separate folder to install building results. As make system user I expect if I specify prefix then all files will be copied under this prefix folder, so I would keep prefixing. 

However instead of replacing `/usr` prefix could be specified as a `prefix` argument of the `get_python_lib()` call in line 265: https://github.com/opencog/atomspace/blob/b84086239471008db68d8ee9572643d2fd207a9d/CMakeLists.txt#L265

Separate issue is that standard Python modules looking path doesn't include `/usr/local/...` folder.

maparent: Excellent idea. Modified. (Also rebased on macos branch to make it easier for me to test, but tested both on mac and ubuntu.)

vsbogd:
Looks good to me, but I didn't look at changes outside of `PYTHON_DEST` scope. Are they discussed in separate PR?

maparent:
Yes, #2161.

vsbogd:
May be then it makes sense to merge this change to PR #2161 ? Or just merge them one after one.

maparent:
They were separated so you could look at this specifically. We'll merge them in sequence.

linas:
I asked Marc-Antoine to separate the python changes from the rest.

linas:
So, uh, this pull req appears to be identical to #2161 except that the title is different.  ???

maparent:
I rebased it after 2161, to allow testing, but it has one extra commit (the python path), so that commit could be looked at independently.

maparent:
and now rebased on master.

linas:
ah OK

PR #2166, opened 2019-05-15
Relax MapLinkUTest::test_quote_arg_map as suggested by Mr Parent

ngeiswei:
As per https://github.com/opencog/atomspace/pull/2161#issuecomment-491564340


maparent:
I can confirm that all tests pass on mac now. Thank you!

Issue #2167, opened 2019-05-16
Fail to ground a glob that appears in more than one clause


leungmanhin:
Haven't had time to look into this yet, it can be reproduced by:

```scheme
(use-modules (opencog) (opencog exec))

(EvaluationLink
    (PredicateNode "pred-1")
    (ListLink
        (ConceptNode "blah")
    )
)
(EvaluationLink
    (PredicateNode "pred-2")
    (ListLink
        (ConceptNode "blah")
    )
)
```

This fails, expecting `(ConceptNode "blah")` but it gives nothing...
```scheme
(define g1
    (GetLink
        (TypedVariableLink
            (GlobNode "$G")
            (TypeSetLink
                (TypeNode "ConceptNode")
                (IntervalLink
                    (NumberNode 1)
                    (NumberNode -1)
                )
            )
        )
        (AndLink
            (EvaluationLink
                (PredicateNode "pred-1")
                (ListLink
                    (GlobNode "$G")
                )
            )
            (EvaluationLink
                (PredicateNode "pred-2")
                (ListLink
                    (GlobNode "$G")
                )
            )
        )
    )
)

(cog-execute! g1)
```

This works, after replacing `GlobNode` with a `VariableNode`:

```scheme
(define g2
    (GetLink
        (TypedVariableLink
            (VariableNode "$G")
            (TypeNode "ConceptNode")
        )
        (AndLink
            (EvaluationLink
                (PredicateNode "pred-1")
                (ListLink
                    (VariableNode "$G")
                )
            )
            (EvaluationLink
                (PredicateNode "pred-2")
                (ListLink
                    (VariableNode "$G")
                )
            )
        )
    )
)

(cog-execute! g2)
```

Also works if the pattern contains only one of the above `EvaluationLink`.

linas:
I'm looking at this, and so far, all I see is this: when it tries to match the second clause, it hits `PatternMatchEngine.cc` circa line 762ff which notices that glob already has a grounding; it assume that grounding is wrong, forces backtracking.  Somehow, it did not know that it has started search a brand-new clause.  If I try to disable this or related code, it goes into an infinite loop.  I don't understand why there's an infinite loop; it makes me think there is also a second bug?  Looking at the blame-log, this code was last modified two years ago, as part of a cleanup from you @leungmanhin I'll continue debugging this till I find a solution or until I get bored.

linas:
Ignore above comment; I think I found the actual bug. unclear  how to fix it.

[...]

PR #2169, opened 2019-05-17
Pivot correctly around GlobNodes

linas:
This fixes bug #2167


linas:
Merging, as this fixes the original bug report.  However, this is not yet the final version, as more complex tests still fail.

PR #2173, opened 2019-05-20
Simplify and clean up the pattern compiler.

linas:
Remove dead code; simplify analysis of structure.


linas:
SurprisingnessUTest is failing, but this fails about half the time, anyway. The changes in this pull-req shouldn't afftect the pattern matcher, and shouldn't affect suprisingness in any way.

PR #2176, opened 2019-05-21
Pull StateLink management out of the atomspace 

linas:
Simplify atomspace implementation by handling StateLink entirely with the StateLink class.


linas:
Merging. As usual, SurprisingnessUTest keeps failing.

Issue #1576, opened 2018-02-12
Strange behavior when doing pattern matching with ImplicationLink


leungmanhin:
To reproduce it:

```scm
(use-modules (opencog) (opencog exec))

;  It may not make sense but it's used to demonstrate the issue
(define impl
  (ImplicationLink
    (AndLink
      (EvaluationLink
        (PredicateNode "some-pred")
        (ListLink
          (VariableNode "$x")
        )
      )
      (TrueLink)
    )
    (ConceptNode "goal")
  )
)

; Link with an anchor
(StateLink (AnchorNode "A") impl)

; PM query
(define sat
  (SatisfactionLink
    (TypedVariable
      (VariableNode "$y")
      (TypeNode "ImplicationLink")
    )
    (AndLink
      (StateLink (AnchorNode "A") (VariableNode "$y"))
    )
  )
)

(cog-evaluate! sat)
```

I would expect the evaluation to be true but it turns out to be false. On the other hand if I replace the above `StateLinks` with `ListLinks`, it returns true...

leungmanhin:
Also if I replace `(VariableNode "$x")` with any type of node that is not a variable, or type that variable, it will return true as well. So is a free variable causing this problem?

ngeiswei:
Can you try to replace `ImplicationLink` by `ImplicationScopeLink`?

leungmanhin:
Yes `ImplicationScopeLink` works. I'm wondering why does the scope matter in this case?

ngeiswei:
I suppose what happens is that `SatisfactionLink` finds the match (the implication), runs it, finds a variable in it and include it to the satisfaction problem. The `ImplicationScope` hides the variable, x, from the `SatisfactionLink` so the problem doesn't occur.

Regardless of whether it is or not a wanted behavior of `SatisfactionLink`, it `ImplicationScope` is really what you want to use, if you carefully read https://wiki.opencog.org/w/ImplicationScopeLink you'll probably understand why, please let me not otherwise.

leungmanhin:
It seems to be the case... though why would it be included in the satisfaction problem, I think `ImplicationLink` is declarative?

ngeiswei:
I would agree [that it's an unwanted behavior], maybe @linas can shine a light on it.

linas:
I was going to look at this today, but I didn't. Maybe tomorrow? Is this urgent?

leungmanhin:
It's not urgent, thanks @linas!

[...]

PR #2177, opened 2019-05-22
Verify that StateLink+Get works correctly.

linas:
There is a distinction between free variables and bound variables,
when used with StateLinks and the pattner matcher. This is exposed
in issue #1576. This test tightens testing, to make sure that things
stay that way.

### DIFF
--- a/tests/atoms/StateLinkUTest.cxxtest
+++ b/tests/atoms/StateLinkUTest.cxxtest
@@ -49,6 +49,7 @@ public:
 	void test_getting();
 	void test_putting();
 	void test_list();
+	void test_scope();
 };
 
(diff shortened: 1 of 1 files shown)


linas:
Just like always, SurprisingnessUTest fails. Merging anyway.

ngeiswei:
@linas said
> Just like always, SurprisingnessUTest fails

You mean locally, right? Cause it has passed on circleci. If it fails locally please paste somewhere the stdout so I can fix it.

linas:
@ngeiswei SurprisingnessUTest fails in circle-ci about half the time. I keep hitting the "retest" button until it passes.

PR #2180, opened 2019-05-23
Miscellaneous commentry cleanup

linas:
Some long-ago-lost commits had assorted minor fixes and commentary changes.  
This includes a significant re-org of the types-definition file, which is entirely cosmetic.


ngeiswei:
BTW, @linas, I wonder what it would take to make atom types user-definable at run-time (name, how inherit from other types, and even how to execute, etc). Do you think it's worth thinking about it, writing a github issue, etc? Or is it a waste of time as it's impossible/difficult for some reason?

linas:
@ngeiswei yes, run-time defined types have been a long-time desirable feature.  The pros and cons:

* So far, no one has really needed them. In the name of minimalism, one should not implement features that no one wants.
* Obviously, the C++-backed atoms need to be defined at compile time. Currently, maybe half, or more are defined at compile time.
* There is a performance hit: AtomTable holds a `std::vector` of length `NUM_TYPES` - That would need to be a hash table, and it could slow down atom-space references by 2x maybe.  Aotomspace access is already too slow...
* Restoring atoms from SQL whose definitions have been deleted is .. troublesome
* Python and scheme bindings are troublesome...

Open a github issue. I'm surprised we don't have one already.

Mostly, the above boils down to three points:
* performance performance performance
* Its doable, but there's a pretty big can of worms
* So what neat things could we do with run-time defined types, and are there other ways, better ways of solving the same problem?

linas:
@ngeiswei incoming sets are organized by type, and performance there is delicate, since graph traversal is controlled by this.

PR #2181, opened 2019-05-24
Streamline atomtable insertion code

linas:
Mostly, this just eliminates a block of duplicated code and some branches that go with it.


linas:
Merging despite having opencog fail consistently:
```
	 17 - AnaphoraTest (Failed)
```
presumably the above failures are due to recent python3 work

PR #2182, opened 2019-05-25
streamline atomspace clearing

linas:
There is no particular reason to always perform slow atomspace clearing. Sure, it does some consistency checks, but they always pass, anyway.


linas:
As usual, getting test failures:
```
The following tests FAILED:
	 17 - AnaphoraTest (Failed)
```
presumably due to recent python3 changes to opencog.  I'm ignoring these test failures, for now.

PR #2183, opened 2019-05-25
Remove a pointless, un-needed class 

linas:
Pointless complexity makes the system hard to understand .. and slow. Get rid of it.


linas:
As usual, opencog unit tests are failing:
```
The following tests FAILED:
	 17 - AnaphoraTest (Failed)
```
Merging anyway.

PR #2184, opened 2019-05-25
Move class-server intialization to class-server directory

linas:
Delete a pointless file in the process.


linas:
unit test crashes see - opencog/opencog#3503 and opencog/opencog#3509

ngeiswei:
> Move class-server intialization to class-server directory

I don't see any class-server directory, is it normal?

linas:
> I don't see any class-server directory, 

I meant -- the directory that the ClassServer actually lives in.  When ClassServer.cc got moved to a new location, this other file should have been too, but it was forgotten.

PR #2185, opened 2019-05-25
Remove unused functions, dead code

linas:
re-arrange location of a lock so it actually protects stuff.


linas:
unit test crashes see - opencog/opencog#3503 and opencog/opencog#3509

Issue #2190, opened 2019-05-28
Create atom type at run-time, via scheme, python or haskell.


ngeiswei:
## Overview

Currently the only way to create new atom types is at compile time via C++. It could be useful to allow the creation of new types at run-time via the binding languages, Scheme, Python and Haskell. It should be possible to define their inheritance relationship with the other types, as well as their associated executions if any.

According to @linas here is a list of cons (extracted from https://github.com/opencog/atomspace/pull/2180#issuecomment-495770974):

* So far, no one has really needed them. In the name of minimalism, one should not implement features that no one wants.
* Obviously, the C++-backed atoms need to be defined at compile time. Currently, maybe half, or more are defined at compile time.
* There is a performance hit: AtomTable holds a `std::vector` of length `NUM_TYPES` - That would need to be a hash table, and it could slow down atom-space references by 2x maybe.  Aotomspace access is already too slow...
* Restoring atoms from SQL whose definitions have been deleted is .. troublesome
* Python and scheme bindings are troublesome...

Mostly, the above boils down to three points:
* performance performance performance
* Its doable, but there's a pretty big can of worms
* So what neat things could we do with run-time defined types, and are there other ways, better ways of solving the same problem?

ngeiswei:
@linas, regarding
> There is a performance hit: AtomTable holds a std::vector of length NUM_TYPES - That would need to be a hash table, and it could slow down atom-space references by 2x maybe. Aotomspace access is already too slow...

Why would you need a hash table, can't the new types be simply pushed on that vector?

Hmm, or maybe it would require that their number ID be content generated (given their names and inheritance relationship with the other types) which indeed would require a hash...

vsbogd:
> There is a performance hit: AtomTable holds a std::vector of length NUM_TYPES - That would need to be a hash table, and it could slow down atom-space references by 2x maybe. Aotomspace access is already too slow...

@linas, do you mean this vector: https://github.com/opencog/atomspace/blob/7b060552a0d7bf2ba04988aaca929380b576dbdc/opencog/atomspace/AtomTable.h#L88-L89

It looks like it is accessed in very complex methods `add` and `extract` and most affected method should be `get_handles_by_type` which is used in `InitiateSearcCB.h`. So it is not clear if impact is serious.

vsbogd:
I would add that adding additional C++ Atom types in external libraries without modifying atomspace code would also be useful.

[...]

PR #2191, opened 2019-05-29
Serialize addition of new atom types.

linas:
Only one module at a time can add new atom types.


linas:
I realized this was needed while thinking about #2190

Issue #2193, opened 2019-05-29
Recursion with DefinedSchema crashes


buj:
Inspiration from the example with SequentialAndLink and printing stuff to console: [https://github.com/opencog/atomspace/blob/master/examples/atomspace/recursive-loop.scm](https://github.com/opencog/atomspace/blob/master/examples/atomspace/recursive-loop.scm)
The above mentioned uses evaluatable links (`DefinedPredicate` and `SequentialAnd`) to achieve recursion. So I thought of what would happen with `DefinedSchema` instead:
```
(Define
   (DefinedSchema "factorial")
   (Lambda
      (TypedVariable
         (Variable "$n")
         (Type "NumberNode")
      )
      (Cond
         (GreaterThan
            (Variable "$n")
            (Number 0)
         )
         (Times
            (Variable "$n")
            (ExecutionOutput
               (DefinedSchema "factorial")
               (Minus
                  (Variable "$n")
                  (Number 1)
               )
            )
         )
         (Number 1)
      )
   )
)

(cog-execute! (ExecutionOutput (DefinedSchema "factorial") (Number 5)))
```
It crashes; tried the same in python, also crashes.

Other than that, is there some link type corresponding to: `begin {statement1; statement2; statement3; ...} end`? This could be used in a similar way to `SequentialAndLink` in the abovementioned example---execute many statements, the result is the result of the last one.

linas:
Well, this is certainly an unexpected use of the atomspace.  But I guess its valid. 

It appears that execution causes infinite recursion, where the definition is substituted for repeatedly, until the stack is blown out, and then it crashes. I'm looking to see how it might be fixed now.

The reason I said that this is an unexpected use of the AtomSpace is that the AtomSpace is intended to be a knowledge-store, and for it to be effective for that, algorithms have to be able to crawl over the contents of the atomspace, and perform transformations on it.  By writing executable code in atomese,  your creating data structures that no ordinary algo can understand and process.  There's nothing terribly wrong with that, its just .. pushing the limits.  Computing factorials  in Atomese is going to be a thousand times slower (literally) than any other programming language... and worse, Atomese is not a human-friendly language.  Its not meant for human programmers.  Just saying ...

linas:
So, this is fixed in pull rq #2194

ngeiswei:
Maybe this example could be added in the example folder. I know it is not a typical use of Atomese but it is such a classical example. I could see MOSES learning that kind of programs.

linas:
[...]

PR #2194, opened 2019-05-30
Fix for bug #2193

linas:
Well, calling it a "fix" is kind of wrong -- there are maybe 8 or 9 fixes in here -- this is a big block of work that has been deferred for years, sitting on the back burner, unfinished, undone.  Well, no its a bit more done than before.

All unit tests pass; however, over in opencog-land, `MOSESPLNSynergyUTest` and `PLNRulesUTest` fail, which suggests that PLN is depending on undocumented, untested function to operate correctly. I'll try to look at that, but I fear that debugging PLN unit tests is hard, so I'm scared of having to go into that code.

### DIFF
--- a/opencog/atoms/base/Handle.cc
+++ b/opencog/atoms/base/Handle.cc
@@ -269,6 +269,16 @@ std::string oc_to_string(const HandleMapSeq& hms, const std::string& indent)
 	return ss.str();
 }
 
+std::string oc_to_string(const HandleMapSeqSeq& hmss, const std::string& indent)
+{
+	std::stringstream ss;
+	ss << indent << "size = " << hmss.size() << std::endl;
+	for (unsigned i = 0; i < hmss.size(); i++)
+		ss << indent << "--- maps[" << i << "] ---" << std::endl
+		   << oc_to_string(hmss[i], indent + OC_TO_STRING_INDENT);
+	return ss.str();	
+}
+
 std::string oc_to_string(const HandleMapSet& hms, const std::string& indent)
 {
 	std::stringstream ss;
--- a/opencog/atoms/base/Handle.h
+++ b/opencog/atoms/base/Handle.h
@@ -199,6 +199,9 @@ typedef std::map<Handle, HandleSet> HandleMultimap;
 //! a sequence of ordered handle maps
 typedef std::vector<HandleMap> HandleMapSeq;
 
+//! a sequence of sequences of ordered handle maps
+typedef std::vector<HandleMapSeq> HandleMapSeqSeq;
+
 //! a set of ordered handle maps
 typedef std::set<HandleMap> HandleMapSet;
 
@@ -311,6 +314,8 @@ std::string oc_to_string(const HandleMap::value_type& hmv,
 std::string oc_to_string(const HandleMultimap& hmm,
                          const std::string& indent=empty_string);
 std::string oc_to_string(const HandleMapSeq& hms,
+                         const std::string& indent=empty_string);
+std::string oc_to_string(const HandleMapSeqSeq& hmss,
                          const std::string& indent=empty_string);
 std::string oc_to_string(const HandleMapSet& hms,
                          const std::string& indent=empty_string);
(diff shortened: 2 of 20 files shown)


linas:
Ahh, all three failures in `PLNRulesUTest` is because quotes are not being stripped out. Maybe that will be easy to fix.

ngeiswei:
@linas before merging can you let me have a look it, just want to make sure it will have no collateral damages beside the ones caught by unit tests.

linas:
OK, I got `PLNRulesUTest` to pass; it requires eager evaluation of arguments, whereas I switched most of the code to lazy evaluation.   However, `MOSESPLNSynergyUTest` is still failing, and I have no effective way to debug why.  @ngeiswei could I convince you to take a look? I kind of want to merge this pull req, but can't do that without knowing why it's failing.

BTW, I'm thinking that the switch to lazy evaluation means that some of the quotes might no longer be needed.  Maybe. I can't really tell.

ngeiswei:
I'm on it.

linas:
Oh, hang on ... `MOSESPLNSynergyUTest` fails when I run it by hand, but it passes when I run it with `make test` ... that's not nice ...

ngeiswei:
`MOSESPLNSynergyUTest` passes for me, both with the other test and as standalone. But I still want to have a deeper look at your PR. I'll give you the go when you can merge (or I'll merge myself if you give me the go conditioned on my review).

linas:
OK. I'm done. I just pushed a unit test; it looks like everything passes now (except for the usual failures of bug opengcog/opencog#3509)  Merge it whenever you see fit.

ngeiswei:
@linas there is bug that I have highlighted in that unit test https://github.com/ngeiswei/atomspace/commit/4853619929ba4e5f29acc0e48e8e475e93414c9f

ngeiswei:
I think I know what is going on. `(Variable "$B")` is used as declaration of the constructed lambda, however since `(Variable "$B")` also maps to the body, then somewhere somehow the variable of the lambda gets substituted by the body again, wrongly.

ngeiswei:
I found the problem, the bug comes from the `PLN_NEEDS_UNQUOTING` block, by walking the tree once more `(Variable "$B")` in the vardecl of the to-be constructed lambda is replaced by the body.

I think the correct fix is to consume quotations rather than re-walk. Also that way it doesn't have to be a case specific PLN hack but rather something that broadly applies, I need to think a bit more about it though...

ngeiswei:
@linas, I'll resume work on it tomorrow, if you wish to work on it you may want to merge my branch

https://github.com/ngeiswei/atomspace/tree/in

into your branch

https://github.com/linas/atomspace/tree/in

as it contains the failing unit tests + some oc_to_string overloading to make debugging easier.

linas:
So @ngeiswei, I merged and provided a fix for your issue, here: c27559a777e42a42673a1aa2683dcdde3c73ea6e The concept of the fix is simple: one should either reduce, or walk_tree, but not both. 

The old code, prior to this pull req, had the idea of "eager execution", done by `walk_tree` which would force substitution of variables, then execute the arguments, then plug the arguments in, and execute again.  That is, for `f(x,y,z)` it would force-execute `x,y,z` first, and then plug them in, and run `f`.  Of course, this leads to an inf loop if any of `x,y,z` contains `f` itself.  The discovery of that inf loop was bug #2193 

I knew this was a problem, so I worked on "lazy execution", on and off, for years.  The problem with lazy execution was that I could not get all of the unit tests to always pass, and there was always some blocker (e.g. circular shared lib dependencies) Over time, the other blockers got fixed, so now, lazy execution becomes possible, finally.  The main mechanism for it is to call `beta_reduce` instead of `walk_tree`.

What `beta_reduce` does is a very minimal substitution of search results into the implicand pattern, and nothing more (in particular no dequotation). Then it executes `f(x,y,z)` as-is, so that `f` can decide whether or not to further expand or execute any of it's arguments.  So, for example, the inf-loop in `factorial` terminates, once the `Cond` link is hit. 

Besides fixing these kinds of issues, I think that `beta_reduce` is much faster than `walk_tree`, because it does a lot less. (I have not actually measured - again, this is where we need benchmarks, but don't really have any).

Overall, the code in `Instantiator.cc` seems overly complex and convoluted to me, so a project for the next few years is to reduce the number of lines of code in there. But, for now, it works, so .. whatever. Good enough for now.

linas:
I'm gonna merge for now; I want to fix PutLink, but that should be a distinct pull req, and not tangled into this one.

PR #2195, opened 2019-05-31
PutLink - change API to match the generic execute() API

linas:
This also removes the "eager execution" flag from `Instantiator.cc` 
This is ongoing cleanup inspired by reviewing #2194


linas:
Merged; the only test failures are the usual ones from opencog

PR #2207, opened 2019-06-07
Refactor Cython module macro to shared cmake include

ferrouswheel:
Based on comment here: https://github.com/opencog/opencog/pull/3518/files#r291023552

Will link related OpenCog PR soon


linas:
OK, thanks. Note - the renaming of of OpenCogFunctions ->OpenCogGuile will probably affect as-moses and agi-bio, too.  FYI  @Dagiopia @ngeiswei  @eddiemonroe this will affect you ...

Issue #2214, opened 2019-06-10
Problem evaluating DefinedPredicate


buj:
I'm getting some counterintuitive results with DefinedPredicates. Basically, I set up a small genealogy database (consisting of a few `InheritanceLinks`). Now I want to query it and find out if some `A` is a parent of `B`, that is, if there is an `InheritanceLink` with strength > 0.5 and confidence > 0.5. I devise three different methods, which should merely differ on the syntactic level; but they return different results (one is obviously correct and the other not so much).

The code shall explain the rest:
```
; Populate the atomspace with facts.
(define anna (Concept "anna"))
(define mama (Concept "mama"))
(Inheritance (stv 1. 1.) anna mama)


; Auxiliary methods for debugging.
(define (prt-atom x)
  (display x)
  (stv 1. 1.)
)
(define (prt-mean x)
  (display "mean is ")
  (display (cog-mean x))
  (newline)
  (stv 1. 1.)
)
(define (prt-confidence x)
  (display "confidence is ")
  (display (cog-confidence x))
  (newline)
  (stv 1. 1.)
)

; Something is "true-enough" if both its strength and confidence are
; above 0.5. This also prints debugging info.
(Define
  (DefinedPredicate "true-enough")
  (Lambda
    (SequentialAnd
      (True (ExecutionOutput
        (GroundedSchema "scm:prt-atom")
        (Variable "$x")
      ))
      (True (ExecutionOutput
        (GroundedSchema "scm:prt-mean")
        (Variable "$x")
      ))
      (True (ExecutionOutput
        (GroundedSchema "scm:prt-confidence")
        (Variable "$x")
      ))
      (GreaterThan
        (StrengthOf (Variable "$x"))
        (Number 0.5)
      )
      (GreaterThan
        (ConfidenceOf (Variable "$x"))
        (Number 0.5)
      )
      (True (ExecutionOutput
        (GroundedSchema "scm:prt-atom")
        (Concept "success!")
      ))
    )
  )
)
; Convenience method.
(define (true-enough x)
  (cog-evaluate!
    (Evaluation
      (DefinedPredicate "true-enough")
      x
    )
  )
)

; Essentially the same as "true-enough", except that it immediately
; puts the two variables into an inheritance link (thus saving some
; typing).
(Define
  (DefinedPredicate "aux")
  (Lambda
    (VariableList
      (Variable "$x")
      (Variable "$y")
    )
    (Evaluation
      (DefinedPredicate "true-enough")
[...]

Issue #1782, opened 2018-06-26
Python module does not load ...


linas:
This is a followup to issue opencog/opencog#3226 -- 
Starting the cogserver results in the following message being printed:

```
[2018-06-26 20:39:40:204] [INFO] Starting CogServer loop.
[2018-06-26 20:39:55:415] [INFO] Using config file found at: /home/ubuntu/run/./dj-en.conf

[2018-06-26 20:39:55:428] [INFO] [global_python_initialize] Start
[2018-06-26 20:39:55:439] [INFO] [global_python_initialize] Adding OpenCog sys.path directories
[2018-06-26 20:39:55:440] [WARN] PythonEval::try_to_load_modules Failed to load the opencog.atomspace module
[2018-06-26 20:39:55:440] [INFO] [global_python_initialize] Finish
[2018-06-26 20:39:55:440] [INFO] [global_python_initialize] Start
[2018-06-26 20:39:55:440] [WARN] PythonEval::atomspace_py_object Failed to load theopencog.atomspace module
[2018-06-26 20:39:55:440] [INFO] PythonEval::initialize_python_objects_and_imports Finished initialising python evaluator.
```

linas:
Above is after a clean install:
```
rm -r /usr/local/include/opencog
rm -r /usr/local/lib/opencog
rm -r /usr/local/share/opencog
```
and then rebuilding/installing cogutils, atomspace, opencog.

The python module gets installed here, on my system:
```
/usr/local/lib/python3.5/dist-packages/opencog/atomspace.so
```
and `ldd -r` shows that its fully linked, so that is not the issue ...

linas:
My current guess is that python is not looking in the `/usr/local` directory, and/or is looking it `site-packages` instead of `dist-packages`.   This is a very old generic python issue: ubuntu uses `site-packages` (or is it Debian??) but the mainstream python maintainers do not.  Plus, this policy is different between python 2.7 and 3.5 so its very confusing to figure out the correct install location.

ChehST:
Hi, is there "$ python -m site" your path ?
> /usr/local/lib/python3.5/dist-packages/opencog/atomspace.so

linas:
```
$ python -m site
sys.path = [
    '/home/linas',
    '/usr/lib/python2.7',
    '/usr/lib/python2.7/plat-x86_64-linux-gnu',
    '/usr/lib/python2.7/lib-tk',
    '/usr/lib/python2.7/lib-old',
[...]

PR #2216, opened 2019-06-11
It's not possible to get a TV unless the atom is in an atomspace

linas:
If an atom is not in an atomspace, we don't really know what TV it is supposed to have.
This fixes issue #2214  - the TruthValueOf, StrengthOf and ConfidenceOf links weren't
working correctly.

### DIFF
--- a/tests/cython/README.md
+++ b/tests/cython/README.md
@@ -13,7 +13,7 @@ apt-get install python-nose
 You may need to set up the PYTHON path:
 export PYTHONPATH=${PROJECT_BINARY_DIR}/opencog/cython
 or, if installed:
-export PYTHONPATH=/usr/local/lib/python3.5/dist-packages/opencog:${PYTHON}
+export PYTHONPATH=/usr/local/lib/python3/dist-packages/opencog:${PYTHON}
 
 For example:
 export PYTHONPATH=build/opencog/cython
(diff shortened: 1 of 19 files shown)


tests/cython/README.md
```diff
@@ -13,7 +13,7 @@ apt-get install python-nose
 You may need to set up the PYTHON path:
 export PYTHONPATH=${PROJECT_BINARY_DIR}/opencog/cython
 or, if installed:
-export PYTHONPATH=/usr/local/lib/python3.5/dist-packages/opencog:${PYTHON}
+export PYTHONPATH=/usr/local/lib/python3/dist-packages/opencog:${PYTHON}
```
ferrouswheel: I have my opencog install in `/usr/local/lib/python{3.5,3.6}/site-packages/opencog` and not specifying a minor version will cause all sorts of breakages.

(which isn't to say one of our cmake files installs stuff to /usr/local/lib/python3/dist-packages/opencog - but if that's the case I think we should fix it to avoid pain in future)

I saw your comment [here](https://github.com/opencog/atomspace/issues/1782#issuecomment-500492476), so looks like it shouldn't be too hard to get the correct path despite the distutils bug.

ferrouswheel: Ok - the reason I was getting things installed to '/usr/local/lib/python{3.5,3.6}/site-packages' instead of '/usr/local/lib/python3/dist-packages' like everyone else is that I was running the build in a virtualenv (while testing cython compatibility).

linas:
Feel free to modify the README to specify both.  I personally use that README as a cheat-sheet, I just cut and paste from it.  For a brief period, (a few years?), one did not need to specify a PYTHONPATH at all -- things just worked. As of the last few weeks or month or so, one must once-again manually specify the PYTHONPATH. This is because of Amen's change to the CMakefile to use distutils to discover the install path.

Issue #2225, opened 2019-06-14
Bug with rewriting ImplicationScope wrapped in ExecutionOutputLink


ngeiswei:
The following

```scheme
(define (dummy x) x)

(Evaluation
   (Predicate "P")
   (Concept "A"))

(define bl
(Bind
   (Present
      (Evaluation
         (Predicate "P")
         (Variable "$X")))
   (ExecutionOutput
      (GroundedSchema "scm: dummy")
      (ImplicationScope
         (Evaluation
            (Predicate "P")
            (Variable "$X"))
         (Evaluation
            (Predicate "Q")
            (Variable "$X"))))))

(cog-execute! bl)
```

wrongly produces

```scheme
      (ImplicationScope
         (Evaluation
            (Predicate "P")
            (Concept "A"))
         (Evaluation
            (Predicate "Q")
            (Variable "$X")))
```

while it should produce

```scheme
      (ImplicationScope
         (Evaluation
            (Predicate "P")
            (Variable "$X"))
         (Evaluation
            (Predicate "Q")
            (Variable "$X")))
```

that is leave untouched the implication scope since `$X` is bound to it.

ngeiswei:
The following branch https://github.com/ngeiswei/atomspace/commits/rewrite-utest contains a unit test for it.

linas:
Yeah, looks like a bug ... in `Instantiator.cc` and/or maybe I guess `Context.cc` ... I guess you are working on this.

ngeiswei:
I found the problem. `FreeVariables::substitute_scoped` does not properly hide the variables in the scope in that instance. The thing is that in that `FreeVariables::substitute_scoped` is called with variables that aren't as `beta_reduce` in `Instantiator.cc` construct a `FreeVariables` including variables and non variables in vmap, see

https://github.com/opencog/atomspace/blob/fea73cf3582d0aa993032f1e0dcd06a893623dbf/opencog/atoms/execution/Instantiator.cc#L55-L56

and so `FreeVariables::substitute_scoped` fails to recognize that

```scheme
(Evaluation
  (Predicate "P")
  (Variable "$X"))
```

should not be substituted by

```scheme
(Evaluation
  (Predicate "P")
  (Concept "A"))
```

I see 2 ways to fix that
1. Fix `beta_reduce` so that `crub` only considers variables.
2. Fix `FreeVariables::substitute_scoped` so that it robustly supports non-variables.

[...]

PR #2229, opened 2019-06-19
Fix rewriting ImplicationScope wrapped in ExecutionOutputLink

ngeiswei:
Fix issue #2225

### DIFF
--- a/opencog/atoms/core/Variables.cc
+++ b/opencog/atoms/core/Variables.cc
@@ -278,6 +278,25 @@ Handle FreeVariables::substitute_scoped(const Handle& term,
 		{
 			// Make a copy... this is what's computationally expensive.
 			IndexMap hidden_map = index_map;
+
+			// Remove from hidden_map all non variables in case they
+			// contain hidden variables.
+			for (auto it = hidden_map.begin(); it != hidden_map.end();)
+			{
+				if (it->first->get_type() != VARIABLE_NODE)
+				{
+					// TODO: we could additionally check whether the
+					// non-variable contains a variable from vees, thus
+					// allowing more subtree subtitution to take place
+					// which might be faster.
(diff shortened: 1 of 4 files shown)


ngeiswei:
@linas, I'm confident the fix is correct, however it could perhaps be optimized (by allowing non-variable substitutions that do not contain variables of `vees`, though it's not obvious to me whether it would be faster as it does involve extra checking).

opencog/atoms/core/Variables.cc
```diff
@@ -278,6 +278,25 @@ Handle FreeVariables::substitute_scoped(const Handle& term,
 		{
 			// Make a copy... this is what's computationally expensive.
 			IndexMap hidden_map = index_map;
+
+			// Remove from hidden_map all non variables in case they
+			// contain hidden variables.
+			for (auto it = hidden_map.begin(); it != hidden_map.end();)
+			{
+				if (it->first->get_type() != VARIABLE_NODE)
```
linas: Should almost surely include GLOB_NODE.

linas: I'm confused by your fix. It is immediately above a block of code that says "Remove the alpha-hidden variables.", which also loops .. and removes variables. ... from `hidden_map`. So you added a second loop that also does "the same thing" as the first loop.  Except, clearly, it's not quite exactly "the same thing", because it fixes your bug.   .. which means that the original loop was somehow buggy. Or something.     This is not quite right. There should not be two loops that do the same thing.  (see lines 301-308 in the new file).

linas: oh.. I see what is going on. OK, disregard previous commment.

linas:
Yes, this is the correct fix. I will add GlobNode shortly, and change the comments slightly.

PR #2230, opened 2019-06-19
Re-organize the hidden variable renaming.

linas:
This reorganizes the loops in pull req #2229 and adds additional
comments.


linas:
@ngeiswei this makes the loop that you added be the second loop. That way, I can describe it with the words "also do this..."  It also adds a glob-node check.

ngeiswei:
OK, BTW, the reason I had set it as first loop is because I figured in would in some case make the second faster, but it's probably negligible anyway.

linas:
> the reason I had set it as first loop is because I figured in would in some case make the second loop faster, 

I was wondering if that iwas why you did that ....

> but it's probably negligible anyway.

And yes, you'd be correct on both counts ... it would be faster, and negligable. I decided that it was more important for readability to remove the variables first, and then the terms.    It seemed easier to explain it that way.

Issue #2231, opened 2019-06-20
Bug PresentLink when combined with virtual link in some way


ngeiswei:
The following 

```scheme
;; KB

(EvaluationLink
  (PredicateNode "P")
  (LambdaLink
    (ImplicationLink
      (VariableNode "$Z")
      (InheritanceLink
        (VariableNode "$X")
        (VariableNode "$Y")
      )
    )
  )
)

(EvaluationLink
  (PredicateNode "P")
  (LambdaLink
    (ImplicationLink
      (VariableNode "$X")
      (VariableNode "$Y")
    )
  )
)

;; Query

(define (dummy x) (stv 1 1))

(define query
  (BindLink
    (AndLink
      (EvaluationLink
        (GroundedPredicateNode "scm: dummy")
        (EvaluationLink
          (PredicateNode "P")
          (LambdaLink
            (ImplicationLink
              (InheritanceLink
                (VariableNode "$X")
                (VariableNode "$Y")
              )
              (VariableNode "$Z")
            )
          )
        )
      )
      (PresentLink
        (EvaluationLink
          (PredicateNode "P")
          (LambdaLink
            (ImplicationLink
              (InheritanceLink
                (VariableNode "$X")
                (VariableNode "$Y")
              )
              (VariableNode "$Z")
            )
          )
        )
      )
    )
    (Concept "OK")
  )
)

(cog-execute! query)
```

raises an exception

```scheme
ERROR: In procedure opencog-extension:
Throw to key `C++-EXCEPTION' with args `("cog-execute!" "Not implemented! Need to implement a stack, here. (/home/nilg/OpenCog/atomspace/opencog/query/DefaultPatternMatchCB.cc:265)
Function args:
...
```

If `Present` is removed from the query, then it works.

ngeiswei:
I have created a unit test for it https://github.com/ngeiswei/atomspace/commits/fix-present-bug and I am investing the problem.

linas:
That error message might be misleading.  The BindLink has zero variables in it.  The PresentLink is wrapping something that is effectively a constant: there are zero free variables inside that PresentLink, so that term will always be in the atomspace, by definition, so the PresentLink is trivially satisfied.  The pattern matcher should have printed a warning about constant clauses, and removed that clause.

That leaves you with just the GPN, which is also wrapping a constant. The GPN should print the constant, and be done.

[...]

PR #2233, opened 2019-06-24
Fix present bug (issue #2231)

ngeiswei:
Fix issue #2231

The fix is in commit https://github.com/opencog/atomspace/commit/867f7b9d242770147ab04e452207ebaa900d6c26

It was caused because `_pat_bound_vars` would be set in `DefaultPatternMatchCB::link_match` but never reset in `DefaultPatternMatchCB::post_link_mismatch` as due to the lack of implementation of `PMCGroundings::post_link_mismatch` the default `PatternMatchCallback::post_link_mismatch` was called instead.

### DIFF
--- a/opencog/atoms/base/Handle.h
+++ b/opencog/atoms/base/Handle.h
@@ -287,9 +287,6 @@ static inline std::string operator+ (const std::string &lhs, Handle h)
 
 // Debugging helpers see
 // http://wiki.opencog.org/w/Development_standards#Print_OpenCog_Objects
-// The reason indent is not an optional argument with default is
-// because gdb doesn't support that, see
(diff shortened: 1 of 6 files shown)


ngeiswei:
@linas, although there's more to it than this fix, for instance, as you pointed out, the present link was grounded and shouldn't be considered to begin with, but this is was a bug nevertheless and I'm confident the fix is correct thus I'm merging.

linas:
Yes, the fix looks good.

linas:
This must have been hard to find. good work.

ngeiswei:
> This must have been hard to find. good work.

Man, when it clicked you should have seen my face, I was in trance! :-D

Issue #2213, opened 2019-06-10
PythonEvalUTest reinitialises embedded Python interpreter in an unsupported way


ferrouswheel:
This is essentially a continuation of #1888 - but opening a new issue as suggested by the last comment there.

The issue `no attribute __reduce_cython__` appears to still happen with Python 3.5/3.6 and cython 0.26 when running PythonEvalUTest. This causes the test to fail.

The last comment on https://github.com/cython/cython/issues/1953 suggests upgrading to cython 0.28 may get something working, although also suggests resetting the interpreter in this way with extension modules is not actually supported by Python.

Sticking with Python version 3.5 to reduce variables involved, I upgraded to cython 0.28.6 which results in a segfault in `PythonEvalUTest` and `PythonUtilitiesUTest`.

Upgrading to the latest cython 0.29.10, results in:
```
[2019-06-09 23:57:49:728] [ERROR] Python error: Interpreter change detected - this module can only be loaded into one interpreter per process.. (/home/joel/work/opencog/atomspace/opencog/cython/PythonEval.cc:694)
```

and subsequent errors.

It feels like we are skirting unsupported territory by re-initialising the python interpreter multiple times in a process, and just got lucky with cython 0.23 and Python 3.5.

ferrouswheel:
Tested Cython 0.23 and Python 3.6.7 - this combination works and passes tests.

For anyone else wanting to test version compatibility, my work flow using [virtualenvwrapper](https://virtualenvwrapper.readthedocs.io/en/latest/) is:

```
mkvirtualenv opencog --python=/usr/bin/python3.5 # or any other version you have installed
# the above puts you in the virtualenv automatically,
# to enable it later, use "workon opencog"
# You can change the version spec to get a particular minor version, this will get the highest
# version in the 0.23.* series
pip install --force cython>=0.23,<0.24
cd atomspace/build
# make sure to forget the old cmake vars related to python
cmake -U CYTHON_EXECUTABLE -U PYTHON_* ..
# i found the change of cython executable wasn't detected, so need to clean the build
make clean
make && make test
deactivate
rmvirtualenv opencog
```

This avoids messing with your system environment, just don't `make install`, or things might get weird.

AmeBel:
opencog/opencog-deps has been update to ubuntu:18.04 with cython version 0.23

ferrouswheel:
@AmeBel Unless you've also implemented something to fix PythonEvalUTest, your PR will break the CI checks.

[...]

PR #2238, opened 2019-06-28
Remove LTDL_LIBARY_PATH and fix unit test confusion

ferrouswheel:
Needs matching PRs for OpenCog and any other repos that depend on this use of `LTDL_LIBRARY_PATH`

- Partially fixes opencog/opencog#3529
- This fixes the unit tests for Ubuntu 18.04 - which use RUNPATH by default instead of RPATH. See discussion [here](https://github.com/opencog/opencog/issues/3529#issuecomment-505270649). However PythonEvalUTest still fails due to #2213.

The main benefit is that it should make it much harder to accidentally test against your system path instead of the build dir.

If this looks acceptable, I will implement a corresponding PR for OpenCog.


opencog/scm/CMakeLists.txt
```diff
@@ -33,6 +33,18 @@ ADD_GUILE_MODULE (FILES
 	MODULE_DESTINATION "${GUILE_SITE_DIR}/opencog"
 )
 
+# Create three copies of (opencog as-config) module:
+SET(SCM_IN_BUILD_DIR "#t")
+# One for build dir
+CONFIGURE_FILE(opencog/as-config_TEMPLATE.scm ${GUILE_BIN_DIR}/opencog/as-config.scm)
+# and one for in-place testing of source scm modules
+CONFIGURE_FILE(opencog/as-config_TEMPLATE.scm ${CMAKE_SOURCE_DIR}/opencog/scm/opencog/as-config.scm)
```
ferrouswheel: Ideally we'd remove this, and change all the unit tests to use the build dir.

But this would mean that you couldn't just edit a scm file and immediately rerun a test executable... you'd have to run make to copy your scm files to the build dir first.

I'm not sure if that functionality should be preserved.

.gitignore
```diff
@@ -122,3 +122,6 @@ Makefile
 
 # Vim #
 *.swp
+
+# Config for source dir testing
+opencog/scm/opencog/as-config.scm
```
linas: missing newline char.

opencog/guile/SchemeSmob.cc
```diff
@@ -224,40 +224,12 @@ void SchemeSmob::module_init(void*)
...
-	// https://github.com/opencog/atomspace/issues/705 for details.
-	scm_c_eval_string("(add-to-load-path \"" PROJECT_SOURCE_DIR "/opencog/scm\")");
-	scm_c_eval_string("(add-to-load-path \"" PROJECT_BINARY_DIR "\")");
-#endif
+	// For testing - to avoid requiring manual change of guile load path.
+	// Alternatives to always adding this path would be:
+	// - make ctest manage the environment variable GUILE_LOAD_PATH during testing
+	// - set GUILE_LOAD_PATH env var in each test, before initialising Scheme
+	// - pass an optional load path through SchemeEval/SchemeSmob
+	scm_c_eval_string("(add-to-load-path \"" PROJECT_BINARY_DIR "/opencog/scm\")");
```
linas: well, this is still hackalicious, and should be marked as such ... 

The first alternative seems best.   Or even the second -- about half the unit tests already contain lines similar to `scm_c_eval_string("(add-to-load-path \"../../..\")")` trying to manually fiddle their way back to the PROJECT_SOURCE_DIR and all of those hacks have been fragile and fiddly, because they would break at the drop of a hat (err, at the drop of some new theory about how unit tests should work, imposed by external forces I have no control over.   It would be nice to strip out all of that crud .. or make external forces pay the accumulated code debt from their forced ultimatums.)  Whatever. This is a step in the right direction, but its still a hack, and the comments should clearly state its a hack.  Like `XXX FIXME` hack.

opencog/scm/CMakeLists.txt
```diff
@@ -33,6 +33,18 @@ ADD_GUILE_MODULE (FILES
 	MODULE_DESTINATION "${GUILE_SITE_DIR}/opencog"
 )
 
+# Create three copies of (opencog as-config) module:
+SET(SCM_IN_BUILD_DIR "#t")
+# One for build dir
+CONFIGURE_FILE(opencog/as-config_TEMPLATE.scm ${GUILE_BIN_DIR}/opencog/as-config.scm)
```
linas: The industry-standard naming convention is to call this file `as-config.scm.in`  and not `as-config_TEMPLATE.scm` -- I prefer the industry-standard naming convention, as that makes it easier to understand at a glance.

opencog/scm/CMakeLists.txt
linas: The cmake ALREADY copies about half of the scm files into the build dir.   The reason it does not copy all of them is because, at some point in history, as people hacked on this, something changed, and the tests started passing without the copy being necessary.    It would be cleanest to either remove the copying (since its currently a big confusing no-op that litters copies of files throughout the build tree) or make it into a requirement.

opencog/scm/CMakeLists.txt
```diff
@@ -33,6 +33,18 @@ ADD_GUILE_MODULE (FILES
...
+# and one for in-place testing of source scm modules
+CONFIGURE_FILE(opencog/as-config_TEMPLATE.scm ${CMAKE_SOURCE_DIR}/opencog/scm/opencog/as-config.scm)
+
+SET(SCM_IN_BUILD_DIR "#f")
+# One for installing
+CONFIGURE_FILE(opencog/as-config_TEMPLATE.scm ${GUILE_BIN_DIR}/opencog/as-config-installable.scm)
+INSTALL(FILES ${GUILE_BIN_DIR}/opencog/as-config-installable.scm DESTINATION ${GUILE_SITE_DIR}/opencog RENAME as-config.scm)
```
linas: Why are you renaming during install ?  Why not just set the name in the first go?

opencog/scm/opencog/as-config_TEMPLATE.scm
```diff
@@ -0,0 +1,78 @@
...
+; CMAKE uses this tag generate different files for in-build 
+(define-public atomspace-in-build-dir @SCM_IN_BUILD_DIR@)
+(define install-location "${CMAKE_INSTALL_PREFIX}/lib/opencog/")
+
+; If you create a new binary library for scheme, define it here
+; along with it's destination path in the build directory.
+; You also need to add the symbol to the install dir section below.
+
+(define-public opencog-ext-path-smob "@CMAKE_BINARY_DIR@/opencog/guile/")
+(define-public opencog-ext-path-exec "@CMAKE_BINARY_DIR@/opencog/atoms/execution/")
```
linas: This is OK-ish, but we "already" have a "more elegant" solution to this.   There is a CMakefile called `cmake/OpenCogGuile.cmake` which computes and sets these paths -- it does so generically, as a module is defined for the fist time, precisely so that no such special hand-curated lists of library paths need to be maintained. 

As I think about it, I kind-odislike this -- it risks becoming out-of-sync with the paths that `OpenCogGuile.cmake` is setting up.  If those paths get out of sync, there will be debugging confusion, since one of these two files effectively over-writes what the other one is setting up.   So it would be best to gut most of this file, and use the default paths provided by `OpenCogGuile.cmake`  instead.

(The bonus is that as-moses, agi-bio, etc. already use `OpenCogGuile.cmake` and so that way, you don't have to try to track those projects, too.)

opencog/scm/CMakeLists.txt
ferrouswheel: One reason for copying to the build dir, is to make sure all the scm files are laid out in a consistent module hierarchy. This is reflective of what the install dir layout will be, and so this layout is ideally what the tests should use.

An alternative to ensure people can still test scm files without explicitly rerunning make, would be to create a symbolic link from the build dir to the relevent file in the source dir. Then there would be no reason to avoid tests only using the build dir.

opencog/scm/CMakeLists.txt
ferrouswheel: The installable config and build-dir config are different in terms of locating the shared libraries. Only `as-config-installable.scm` is actually installed. They both need to be configured and if they used the same name, one would clobber the other.

opencog/scm/opencog/as-config_TEMPLATE.scm
ferrouswheel: My reading of `cmake/OpenCogGuile.cmake` is that it does nothing with shared libraries. It only does some copying of scm files. They all just assume that their shared library will be available. There are no paths to get out of sync, because they are currently never configured! We cross our fingers and hope the library is somewhere in the RPATH chain, or in LTDL_LIBRARY_PATH (the former doesn't work with RUNPATH, the latter mixes up install and build dirs).

Having said that, it makes sense to put the configuration of shared library dependencies as an optional argument to `ADD_GUILE_MODULE`. And would simplify updating downstream projects.

I'm not a cmake expert and it'll take me a while to figure out how to run `CONFIGURE_FILE` *after* all the scm modules have been defined and stored their paths. I'll see how I go.

.gitignore
ferrouswheel: fixed

opencog/guile/SchemeSmob.cc
ferrouswheel: I clarified it's still a hack

opencog/scm/CMakeLists.txt
ferrouswheel: Makes sense - I've renamed it. Though, sadly, various editors/tools don't seem to detect it's a scheme file when `.in` is the extension.

opencog/scm/CMakeLists.txt
linas: Hmm OK.

linas:
I'll take a closer look tomorrow. In the meanwhile: would it be possible to define a cmake macro, say, `DEFINE_GUILE_LIB(libfoo)` that declares `libfoo` to be that kind of library that needs to get loaded during a scheme extension, and that therefore does the needed path magic to make this work?  Perhaps this could avoid the need for `as-config.scm`, or at least avoid having to list individual libraries in it?  This would be nice, because it allows lib paths to always be managed locally, without needing a centralized location.

Its not quite obvious how `DEFINE_GUILE_LIB(libfoo)` would work. Perhaps it could auto-define a `libfoo.scm` with the needed path magic in it; maybe it should hold all of the `(load-extension "libfoo" ...)` business. Guile modules would then just include `libfoo.scm` as needed.

I know this is asking you to do more work :-) But better get it set up right the first time, than come back to it next year...  ponder this a bit; I'll look at the rest tomorrow.

ferrouswheel:
I made an attempt at defining some CMake functions which would hopefully be general enough for use later.

Making a wrapper scm per extension also sounds nice and possibly would feel less hacky vs what I've done.

linas:
OK seems reasonable; I guess this is mergable; are you done?

ferrouswheel:
Before this merged I should probably make equivalent changes to OpenCog. I
can do that today.

Should we maybe also notify maintainers for other downstream extensions
that were relying on LTDL_LIBRARY_PATH?

linas:
The only down-streams at this point are as-moses and agi-bio and I think they are light-weight users of this mechanism, so might be unaffected.

Yes, we should notify, as we have a habit of making changes that break @mjsduncan which then leaves him high & dry for a while.  Also @eddiemonroe  @tanksha @habush

opencog/guile/SchemeSmob.cc
```diff
@@ -224,14 +224,7 @@ void SchemeSmob::module_init(void*)
...
 	register_procs();
-	// XXX FIXME This is a hack for testing.
-	// It avoids requiring a manual change of guile load path.
-	// Alternatives to always adding this path would be:
-	// - make ctest manage the environment variable GUILE_LOAD_PATH during testing
-	// - set GUILE_LOAD_PATH env var in each test, before initialising Scheme
-	// - pass an optional load path through SchemeEval/SchemeSmob
-	//scm_c_eval_string("(add-to-load-path \"" PROJECT_BINARY_DIR "/opencog/scm\")");
-
+
```
ferrouswheel: Apparently the tests only passed locally due to having an installed version of atomspace.

Meanwhile, leaving the hack in place would make my opencog build point at the atomspace build dir when testing (instead of the system atomspace).

Thus I will do the work to make the AtomSpace tests pass without this.

opencog/scm/CMakeLists.txt
```diff
@@ -21,6 +21,11 @@ ADD_GUILE_MODULE (FILES
 
 ADD_GUILE_MODULE (FILES
 	opencog/exec.scm
+	MODULE_DESTINATION "${GUILE_SITE_DIR}/opencog"
+)
+
+ADD_GUILE_MODULE (FILES
+
```
AmeBel: @ferrouswheel why is the split required?

ferrouswheel: It shouldn't be required. Must be a remnant of when I was experimenting. Good catch!

ferrouswheel:
This should be ready for merging, along with opencog/opencog#3556

As mentioned on the OpenCog PR, the testing strategy that should be adhered to is: the `PROJECT_SOURCE_DIR` should only be explicitly added to the guile load path by tests if they  to access supporting scm files (e.g. that set up specific atomspace contents).

For any scm functions that are eventually installed to the system path, tests should rely on the `GUILE_LOAD_PATH` environment variable. This is automatically set up when calling ADD_CXX_TEST and ADD_GUILE_TEST by opencog/cogutil#179 and dbe9b78. For Python tests and others that call `ADD_TEST`, it is up to the developer to configure the environment if they need access to the opencog scm modules ([example here](https://github.com/opencog/atomspace/pull/2238/commits/0e27496f164fc97d9f53475c09df236bd91b02d2#diff-e60cc1eb0be5fcc2dcbdcdad5df3570fR85))

ferrouswheel:
Well, it will be ready once I fix the test failures that I'm not checking locally: PythonEvalUTest (doesn't work for Python 3.6.8) and Gearman (I don't have this installed)

linas:
> should only be explicitly added to the guile load path by tests if they to access supporting scm files

If the approx 110 or so unit tests for guile, 109 of them have supporting scm files.   I'm pretty sure that all of them use `(add-to-load-path "../../..")` instead of `PROJECT_SOURCE_DIR`.  I suppose that might be a distinct pull req.

cmake/OpenCogGuile.cmake
```diff
@@ -175,6 +175,121 @@ FUNCTION(ADD_GUILE_TEST TEST_NAME FILE_NAME)
...
+    get_target_property(MODULE_NAME ${CONFIG_TARGET} MODULE_NAME)
+    get_target_property(TEST_ENV_VAR ${CONFIG_TARGET} TEST_ENV_VAR)
+    foreach(PATH_PAIR ${SYMBOL_PATH_LIST})
+        string(REPLACE "|" ";" SYMBOL_AND_PATH ${PATH_PAIR})
+        list(GET SYMBOL_AND_PATH 0 SYMBOL)
+        list(GET SYMBOL_AND_PATH 1 LIBPATH)
+        
+        set(SCM_PATHS "${SCM_PATHS}(define-public ${SYMBOL} \"${LIBPATH}/\")")
+        if(WIN32)
+            set(SCM_PATHS "${SCM_PATHS}\r\n")
```
linas: ins Win32 still a thing? Did it not die a graceful death some while ago?

cmake/OpenCogGuile.cmake
```diff
@@ -175,6 +175,121 @@ FUNCTION(ADD_GUILE_TEST TEST_NAME FILE_NAME)
...
+        ; When installed, all the modules have the same path
+${INSTALL_PATHS}
+    
+    )
+)
+")
+    
+    FILE(WRITE "${OUTPUT_FILE}" "${CONFIG_SRC}")
+    
+ENDFUNCTION(WRITE_GUILE_CONFIG)
```
linas: newline!  Check your editor settings.

opencog/guile/SchemeSmob.cc
linas: Yay!

linas:
looks really clean, ow. Thanks!

cmake/OpenCogGuile.cmake
ferrouswheel: Have now updated the settings on my editor :sweat_smile:

ferrouswheel:
@ngeiswei Tests appeared to pass without it, so I've removed it in #2259 - thanks for noticing that!

PR #2241, opened 2019-07-01
silence cython language level warnings

ferrouswheel:
Silences this FutureWarning from cython:

`/home/joel/.virtualenvs/opencog/lib/python3.6/site-packages/Cython/Compiler/Main.py:367: FutureWarning: Cython directive 'language_level' not set, using 2 for now (Py2). This will change in a later release!`

### DIFF
--- a/opencog/cython/opencog/CMakeLists.txt
+++ b/opencog/cython/opencog/CMakeLists.txt
@@ -18,7 +18,7 @@ INCLUDE_DIRECTORIES(
 	${CMAKE_CURRENT_BINARY_DIR}
 )
 
-SET(CYTHON_FLAGS "-f")
+SET(CYTHON_FLAGS "-2" "-f")
 
 # Use this as a guide:
 # https://github.com/OpenKinect/libfreenect/blob/master/wrappers/python/CMakeLists.txt


linas:
MinerUTest failed; I assume this is sporadic; I just it the "retest" button.

linas:
MinerUTest failed a second time, seg fault. third time's a charm?

linas:
MinerUTest failed a third time in a row ... segfaulting. I don't understand why. It doesn't use python ... it passes on my machine ... it passes in other runs, that come before the pull req  .. so what is the root cause of the crash?  Screwed up shlib linkage?

In the past, I've built gdb wrappers for the unit tests, so that gdb would print a stack trace (in circleci) when a unit test failed. I don't recall how, off the top of my head.

@ngeiswei any reason why MinerUTest might fail, desipte a seemingly harmless python tweak?

ngeiswei:
@linas @ferrouswheel you probably just need to rebase the branch to have `MinerUTest` pass.

ferrouswheel:
@ngeiswei @linas I rebased, MinerUTest still times out and the time out is now on master since #2244

ngeiswei:
Sorry about that @ferrouswheel, rebase once again should fix the problem.

PR #2246, opened 2019-07-03
 Create a generic value increment function 

linas:
Inspired by discussion at singnet/language-learning#217
```
 cog-inc-value! ATOM KEY CNT REF -- Increment value on ATOM by CNT.
  The REF location of the FloatValue at KEY is incremented by CNT.
  CNT may be any floating-point number (positive or negative).
  The rest of the FloatValue vector is left untouched.
```


glicerico:
Nice!

PR #2247, opened 2019-07-04
Minor indentation fix

ngeiswei:


ngeiswei:
I'm perplex, all unit tests pass on circle-ci, yet some, like `PLNRulesUTest` use `cog-bind-af` which is not defined anywhere (which fails locally BTW).

ngeiswei:
Actually, it is defined in https://github.com/opencog/opencog/blob/ef3b780f12c5ed3fffbaa09047bcf19c3ca2d8ea/opencog/attentionbank/bank/AttentionSCM.cc#L88

So, somehow this doesn't seem to get installed on my system.

linas:
I think this is a side-effect of Joel Pitt's recent recent search-path reogranization, which needs a new cogutils and a new atomspace.  I'm double-checking now.

ngeiswei:
All unit tests now pass on my system, maybe because I restarted my machine (for lack of other explanation).

linas:
!? Tests pass on my system. atomspace circleci passes, (which runs both atomspace & opencog tests), passes  but opencog circleci fails. How can opencog circleci fail, when the atomspace ci is running the same tests and passing!?   I dunno. Some kind of version skew, I bet it will clear tomorrow.

ngeiswei:
Maybe the Ghost utests randomly fail.

PR #2248, opened 2019-07-04
Propely handle variable scheme in ExecutionOutput

stellarspot:
The following BindLink does not work if ExecutionOutputLink uses variable for schema.
For example:
```
(use-modules
 (opencog)
 (opencog exec))

(Define
 (DefinedSchema "mul")
 (Lambda
  (VariableList
   (Variable "x")
   (Variable "y"))
  (Times
   (Variable "x")
   (Variable "y"))))

(Define
 (DefinedSchema "sum")
 (Lambda
  (VariableList
   (Variable "x")
   (Variable "y"))
  (Plus
   (Variable "x")
   (Variable "y"))))

(Inheritance
 (DefinedSchema "mul")
 (Concept "current-schema"))

(define call-curent-schema
 (Bind
  (TypedVariable (Variable "$SCHEMA") (Type "DefinedSchemaNode"))
  (Inheritance
   (Variable "$SCHEMA")
   (Concept "current-schema"))
  (ExecutionOutput
   (Variable "$SCHEMA")
   (List
    (Number "6")
    (Number "7"))))
)

(display
 (cog-execute! call-curent-schema))
```
The expected result is: `(SetLink  (NumberNode "42.000000"))` - "mul" DefinedSchema is found.
The current result is `Segmentation fault (core dumped)`.

The fix calls `beta_reduce(...)` for ExecutionOutputLink schema in `Instantiator::reduce_exout(...)` method.

### DIFF
--- a/opencog/atoms/execution/Instantiator.cc
+++ b/opencog/atoms/execution/Instantiator.cc
@@ -116,6 +116,9 @@ Handle Instantiator::reduce_exout(const Handle& expr, bool silent)
 	Handle sn(eolp->get_schema());
 	Handle args(eolp->get_args());
 
+        if (not _vmap->empty())
+		sn = beta_reduce(sn, *_vmap);
+
 	// If its a DSN, obtain the correct body for it.
 	if (DEFINED_SCHEMA_NODE == sn->get_type())
 		sn = DefineLink::get_definition(sn);
--- a/tests/query/ExecutionOutputUTest.cxxtest
+++ b/tests/query/ExecutionOutputUTest.cxxtest
@@ -70,6 +70,7 @@ public:
 	void test_recursion(void);
 	void test_lambda(void);
 	void test_implication_scope(void);
+	void test_variable_schema(void);
 };
 
 #define an as.add_node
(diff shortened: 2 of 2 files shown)


linas:
Looks reasonable to me. 

By the way, instead of writing
```
(Inheritance
 (DefinedSchema "mul")
 (Concept "current-schema"))
```
you can also use
```
(State (Concept "current-schema") (DefinedSchema "mul"))
```
The nice thing about `StateLink` is that it guarantees that there is one and only one state, ever, globally, free of race conditions.  That is, it's thread safe, so even if one thread is constantly changing it, all other threads will see only one copy, no matter what.

linas:
And instead of writing `(Concept "current-schema")`  you can write `(Anchor "current-schema")`  The `AnchorNode` does NOTHING at all special, its completely plain. However, it does let the human reading the code to know that "this is a special place where things are put".

Once upon a time, long long ago, there was an idea that `AnchorNode` would also be a thread-safe pick-up/drop-off point, so that anyone (multiple threads) could attach things to it (with an `InheritanceLink`, for example) but one and only one thread could pick something up (by deleting the `InheritanceLink`). That is `AnchorNode` would be a kind-of thread-safe queue.  It never happened.  A newer idea is to have a publish-subscribe system, instead.

PR #2254, opened 2019-07-08
Fix execution include files destination

ngeiswei:
And add LibraryManager as the URE relies on it for creating ASCII representations of inference trees.

### DIFF
--- a/opencog/atoms/execution/CMakeLists.txt
+++ b/opencog/atoms/execution/CMakeLists.txt
@@ -56,5 +56,6 @@ ENDIF (HAVE_GUILE)
 # are fixed, first.
 INSTALL (FILES
 	MapLink.h
-	DESTINATION "include/opencog/atoms/core"
+	LibraryManager.h
+	DESTINATION "include/opencog/atoms/execution"
 )


opencog/atoms/execution/CMakeLists.txt
```diff
@@ -56,5 +56,6 @@ ENDIF (HAVE_GUILE)
 # are fixed, first.
 INSTALL (FILES
 	MapLink.h
-	DESTINATION "include/opencog/atoms/core"
+	LibraryManager.h
```
linas: I think I had previously removed `LibraryManager.h` from the install list, because it had struck me as a horrid, badly-designed API and I did not want to encourage it's use. Do you actually need it for something?

ngeiswei: The URE needs it to extract the formula name of a rule https://github.com/opencog/atomspace/blob/4abc384dd366fd36288da6ff84d28f59c1426da6/opencog/ure/backwardchainer/BIT.cc#L707

It's not crucial but it's convenient for pretty printing inference trees.

ngeiswei:
I'm merging because we don't have anything to replace it as of today.

linas:
> I'm merging because we don't have anything to replace it as of today.

OK. If you're feeling creative, see if you can create better, more descriptive names for the functions. (Yes naming is hard. That's why it's important.)

ngeiswei:
OK, yeah I always found that function name pretty lame (I came up with it :-/). I'll fix it in a subsequent PR.

linas:
Oh, sorry, I guess I was critiquing your code. I did not realize that. I thought this was written by .. someone else.

ngeiswei:
Critiques are welcome. Except I don't know how to make that function name better without elongating it, like `parse_language_library_function` or something, which might still be better though.

linas:
its' not urgent.

ngeiswei:
`parse_schema` or `parse_schema_name` would be adequate though.

PR #2256, opened 2019-07-08
Move the URE to its own repository

ngeiswei:
This is the first step of what was suggested over half a year ago https://github.com/opencog/opencog/issues/3391

What prompts me to do it now is because I believe it will make the opencog vs singnet situation easier to manage, letting aside the fact that it's the right thing to do from an engineering standpoint of course.

I should note that, beside the opencog repo, that directly links to the C++ URE API for testing, which could be avoid BTW, it is a mere convenience, this change should make no difference to the regular user since the URE is typically accessed via its Scheme or Python interface, the user just needs to build and install the URE, that's all.

The new URE repo is ready

https://github.com/opencog/ure

The sister PR for the opencog repo is ready

https://github.com/opencog/opencog/pull/3567

All I need is your green light. :-)

P.S: for those curious, I followed the following tutorial to make the split https://confluence.atlassian.com/bitbucket/split-a-repository-in-two-313464964.html


ngeiswei:
Obviously let me know if you find anything else that needs to be updated, I'm happy to do it. For instance if `ocpkg` is used by anyone I will update it, etc.

ngeiswei:
Four utests, all related to persistent storage are failing
```
The following tests FAILED:
	 49 - ValueSaveUTest (SEGFAULT)
	 50 - PersistUTest (Failed)
	 51 - DeleteUTest (Failed)
	 52 - MultiPersistUTest (Failed)
```

The cause of failure seems to be associated with these errors
```
In DeleteUTest::test_pq_recurse:
/ws/atomspace/tests/persist/sql/multi-driver/DeleteUTest.cxxtest:119: Error: Test failed: Failed to execute SQL command!
PQresult message: ERROR:  relation "uuid_pool" does not exist
LINE 1: SELECT increment_by FROM uuid_pool;
```

I have no idea why.

linas:
> what was suggested #3391

Well, the atomspace repo and URE weren't listed, because they were not a source of problems. It's the other stuff that was always half-broken, that I kept having to maintain, because no one else maintained it, that made me unhappy.

But, OK, the URE and the atomspace have drifted apart, and are almost unrelated projects, so sure.

ngeiswei:
The URE is mentioned in the comments... I do intend to split the Miner and PLN from the opencog repo next.

linas:
> atlassian

I'm not convinced those instructions are correct.  Compare to https://help.github.com/en/articles/splitting-a-subfolder-out-into-a-new-repository  One reason is that the new URE repo still shows me as a major contributor, when that clearly cannot be the case -- almost all of my contributions should have been filtered out. Or so I would think .. which makes me think that the commit history was not correctly filtered.

linas:
> The following tests FAILED:

Yeah, circleCI needs to be configured to not run those tests in parallel.  In parallel, they step on each other. Or we need to change them to use four different sql db's

ngeiswei:
Yeah, I was also surprised the filtering wasn't more aggressive, will look into that...

linas:
> surprised the filtering wasn't more aggressive,

Try the github instructions, and see if it comes out different.

ngeiswei:
> Yeah, circleCI needs to be configured to not run those tests in parallel. In parallel, they step on each other. Or we need to change them to use four different sql db's

But how comes that it used to work and now it doesn't?

linas:
>> not run those tests in parallel.
> But how comes that it used to work and now it doesn't?

It's random timing. Sometimes, the earlier tests run fractions of a second longer, and so the final run-order is different. Hitting "rerun failed tests" button until they pass is my solution.

ngeiswei:
Oh, indeed, only `MultiUserUTest` fails this time, `circleci` just needs to be configured to run on one core. I'll take care of that tomorrow. I'm not merging that now anyway, I'd like to look some more into the filtering issue.

linas:
Merge whenever you are ready.

ngeiswei:
`ShellUTest` broke, I didn't know we still had this problem. Re-trying...

AmeBel:
@ngeiswei I missed a dependency when updating to ubuntu:18.04, that is why it broke. Rebuilding the image now

ngeiswei:
Cool, looks like `ShellUtest` has been fixed, however there's still a remaining error in the last step `Persisting to Workspace`

```
The specified paths did not match any files in /ws
```

AmeBel:
It seems to be a circle-ci failure

ferrouswheel:
If the ure is being split off, this is the perfect time to put it in it's own python package `ure`.

Unlike guile, python only allows one package called `opencog` and it will depend on the python path as to which package is found first.

To be consistent, I would also suggest making ure have it's own top level guile module. However I'm less familiar with guile conventions around projects sharing root module names.

linas:
ure depends on atomspace as a pre-req; its an add-on that adds extra features; its not a stand-alone. The current guile naming convention is that the atomspace module is called `(opencog)` and submodules are `(opencog sub)` and you can next them arbitrarily deep, e.g. there is `(opencog nlp chatbot)` and etc. I'm pretty sure python also allows submodules ... I'm prety sure I've used `from foo.bar import baz;` in python; I assume that meant `bar` was a submodule for `foo`.

But if we are gonna rename things, I'd like to call the atomspace `atomspace` instead of calling it `opencog`....

Or we could call it (in scheme) `(opencog atomspace)`  but that would leave module `(opencog)` effectively empty... which is kind-of weird.

ferrouswheel:
Dependencies between python packages are defined in a `setup.py` and then, once published to https://pypi.org/, can be managed by pip. If you say `pip install ure`, it will, for example, work out you also need to install atomspace.

Python allows submodules, but when looking for a package, it will search `sys.path` sequentially. The first location the package is found will be the one that is used. It won't search other packages for a specific submodule. If it's not in the first package it finds, it will say it doesn't exist.

So if ure has a `opencog` package in it's build dir, and then adds the `ure` submodule. It will never see the opencog package installed by the atomspace repo.

However, after writing the above, it confused me that the opencog repo tests somehow see the atomspaces `opencog` package. The reason that works is due to some magic [here](https://github.com/opencog/opencog/blob/master/opencog/cython/opencog/__init__.py#L2). Despite using python for over a decade I had never run into this, but it explains why things work at the moment. Feels a bit like trouble to not have clear package boundaries though.

ngeiswei:
I am not a Python user so I won't comment on that. Re. Scheme modules what @linas says seems to make sense.

ngeiswei:
I'm merging, it's pretty clear that the circleci problem is independent of that PR.

ferrouswheel:
@ngeiswei It'd probably be worth updating the OpenCog README to list the URE as an optional prerequisite.

ngeiswei:
Good idea, thanks.

Dagiopia:
@ngeiswei shouldn't the guile module be be removed from here?
https://github.com/opencog/atomspace/blob/master/opencog/scm/opencog/ure.scm

ngeiswei:
Indeed, thanks!

PR #2259, opened 2019-07-09
Remove in src config for guile extensions and win32 stanza

ferrouswheel:
make minor chances suggested by review of #2238


linas:
thanks

ngeiswei:
Does that mean we drop win32 support? I think that's perfectly acceptable. I'm asking because there's a bunch of other places with "win32 stanza" awaiting to be removed if that's the case.

ferrouswheel:
Not sure. In this case it's not really necessary since it just makes the
generated config file more readable (but makes the cmake config less
readable!)

linas:
It never worked on win32. It sometimes almost compiled. It sometimes almost ran. It was a project that was never completed. Drop it, it's pointless, now.

ngeiswei:
I don't know if people still use WIN32 for MOSES though?

linas:
How much useful machine learning can you do in less than 4GB of RAM?

linas:
I mean, Rasberry-Pi's are more powerful than any Win32 box ever was.

ferrouswheel:
WIN32 is sometimes conflated with the Windows platform generally.

https://stackoverflow.com/questions/6679396/should-i-define-both-win32-and-win64-in-64bit-build

Pretty sure my 64GB of memory windows box with 1080 ti and 12 cores can
kick the ass of even the newest rpi4. ;-)

linas:
I'm 100% sure your windows box is not running win32.

ferrouswheel:
Here's another link to explain what `WIN32` means in cmake https://cmake.org/cmake/help/v3.0/variable/WIN32.html

> True on windows systems, including win64.
> Set to true when the target system is Windows.

linas:
Holy cow, Joel, you are showing your credulity. Win32 is an ancient programming API, dating back to 1995 or so. It was absolutely horrible, it demonstrated a deep, utter incompetence of the Windows system programmers.  I wrote a website using Wn32, and it sucked.  It is the primary, direct reason I was driven into the arms of Linux.  Linux would be an unknown, forgotten bit computing history trivia, if Win32 hadn't massively boosted its popularity.   There are factual, historical reasons why things happened the way they happened,  and trying to pretend that Microsoft of that era was technically incompetent is revisionist history. Stop that.  Win32 sucked. Get over it.

ferrouswheel:
This conversation started due to a WIN32 stanza in cmake. I don't care about your opinion of the windows platform, because it has nothing to do with what the symbol `WIN32` means in cmake.

`WIN32` is true on Windows 64bit. Get over it.

This is all a complete aside to whether OpenCog should support windows. However you are the one that started talking about 4GB memory limits and 32-bit windows out of the blue. I was just pointing out that's irrelevent due to what the `WIN32` define actually means.

linas:
It's not my opinion; it is a factual statement. You are too young to have experienced what Win32 was like. Cuppa-coffee ever ten minutes while you waited for your system to reboot. You can use google and read about it and educate yourself. Otherwise, this is a childish convrsation.

At any rate, Win32 defines an API that can access no more than 4GBytes of RAM -- that's what the 32 stands for -- a 32-bit address space, with 32-bit pointers. Given the fact that it's old, its broken, and its inapplicable to modern operating systems, even Windows operating systems, it strikes me as utterly pointless to carry around this zombie.  Let it rest in peace.

ngeiswei:
It had beautiful blue screens thought.

PR #2261, opened 2019-07-10
Replace opencog job circleci config by opencog repo's

ngeiswei:


ngeiswei:
Please don't merge yet, just trying to understand why opencog circleci fails.

PR #2262, opened 2019-07-10
Experiments with multi-threading in the matrix code.

linas:
This pull-req adds some scaffolding for multi-threaded operation in the matrix code.

guile-2.9.2 code gives perfect linear (24x speedup on 24 cores) when running the pattern matcher in parallel.  However, it gives no speedup at all, when running the matrix code multi-threaded. I don't know why. It's a bug.

### DIFF
--- a/opencog/matrix/bin-count.scm
+++ b/opencog/matrix/bin-count.scm
@@ -1,8 +1,8 @@
 ;
 ; bin-count.scm
 ;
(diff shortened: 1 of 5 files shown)


linas:
opencog unit tests are passing, but circle-CI is failing with `The specified paths did not match any files in /ws` after the unit tests complete. Merging.

PR #2272, opened 2019-07-16
Place build paths before the install path in Python sys.path

ferrouswheel:
This will likely be causing people confusion and weird failures while testing.

The way it worked before this PR, is it used installed modules for testing in preference to the build dir. This is clearly wrong.


ferrouswheel:
Actually this needs to be changed to iterate through the list in reverse (or increment the insert position).

PR #2276, opened 2019-07-17
Execute PutLink arguments

stellarspot:
The sample below does not work as expected because arguments of the PutLink first are used in LambdaLink and then executed:
```
(Inheritance (Concept "B") (Concept "Parent"))
(Inheritance (Concept "C") (Concept "Parent"))

(define put-link
 (PutLink
  (LambdaLink
   (VariableNode "x")
   (EvaluationLink
    (PredicateNode "relatives")
    (ListLink
     (VariableNode "x")
     (VariableNode "A"))))
  (GetLink
   (Inheritance (Variable "$X") (Concept "Parent")))))

(display
 (cog-execute! put-link))
```

Output is:
```
(SetLink
)
```

Expected result is:
```
(SetLink
   (EvaluationLink
      (PredicateNode "relatives")
      (ListLink
         (ConceptNode "B")
         (VariableNode "A")
      )
   )
   (EvaluationLink
      (PredicateNode "relatives")
      (ListLink
         (ConceptNode "C")
         (VariableNode "A")
      )
   )
)
```

The pull request #2195 removed the eager execution of PutLink where arguments were executed before the beta-reduction.
The `EvaluationLink::do_eval_scratch()`  https://github.com/opencog/atomspace/blob/23d6f03635c169110cf90883890ea89cfb9190cb/opencog/atoms/execution/EvaluationLink.cc#L873  method has comments:
```
		// Evalating a PutLink requires three steps:
		// (1) execute the arguments, first,
		// (2) beta reduce (put arguments into body)
		// (3) evaluate the resulting body.
```

The proposed fix returns the code to `Instantiator::walk_tree()` which executes PutLink arguments before beta-reduce.

### DIFF
--- a/tests/atoms/put-execute-arguments.scm
+++ b/tests/atoms/put-execute-arguments.scm
@@ -1,0 +1,27 @@
+(Inheritance (Concept "B") (Concept "Parent"))
+(Inheritance (Concept "C") (Concept "Parent"))
+
+(define put-execute-arguments
+ (PutLink
+  (LambdaLink
+   (VariableNode "x")
+   (EvaluationLink
+    (PredicateNode "relation")
+    (ListLink
+     (VariableNode "x")
+     (VariableNode "A"))))
+  (GetLink
+   (Inheritance (Variable "$X") (Concept "Parent")))))
+
+(define expected-put-execute-arguments
+ (SetLink
+  (EvaluationLink
+   (PredicateNode "relation")
+   (ListLink
+    (ConceptNode "B")
+    (VariableNode "A")))
+  (EvaluationLink
+   (PredicateNode "relation")
+   (ListLink
+    (ConceptNode "C")
+    (VariableNode "A")))))
--- a/opencog/atoms/core/PutLink.cc
+++ b/opencog/atoms/core/PutLink.cc
@@ -294,6 +294,12 @@ Handle PutLink::do_reduce(void) const
 	Handle bods(_body);
 	Variables vars(_varlist);
 	PrenexLinkPtr subs(PrenexLinkCast(get_handle()));
+	Handle args(_arguments);
+
+	if (args->is_executable())
+	{
+		args = HandleCast(args->execute());
+	}
 
 	// Resolve the body, if needed. That is, if the body is
 	// given in a defintion, get that defintion.
@@ -330,7 +336,7 @@ Handle PutLink::do_reduce(void) const
 	}
 
 	// Now get the arguments that we will plug into the body.
-	Type vtype = _arguments->get_type();
+	Type vtype = args->get_type();
 	size_t nvars = vars.varseq.size();
 
 	// FunctionLinks behave like pointless lambdas; that is, one can
@@ -348,7 +354,7 @@ Handle PutLink::do_reduce(void) const
 		if (LIST_LINK == vtype)
 		{
 			HandleSeq oset(bods->getOutgoingSet());
-			const HandleSeq& rest = _arguments->getOutgoingSet();
+			const HandleSeq& rest = args->getOutgoingSet();
 			oset.insert(oset.end(), rest.begin(), rest.end());
 			return createLink(oset, btype);
 		}
@@ -356,13 +362,13 @@ Handle PutLink::do_reduce(void) const
 		if (SET_LINK != vtype)
 		{
 			HandleSeq oset(bods->getOutgoingSet());
-			oset.emplace_back(_arguments);
+			oset.emplace_back(args);
 			return createLink(oset, btype);
 		}
 
 		// If the arguments are given in a set, then iterate over the set...
 		HandleSeq bset;
-		for (const Handle& h : _arguments->getOutgoingSet())
+		for (const Handle& h : args->getOutgoingSet())
 		{
 			if (LIST_LINK == h->get_type())
 			{
@@ -386,12 +392,12 @@ Handle PutLink::do_reduce(void) const
 	{
 		if (SET_LINK != vtype)
 		{
-			return reddy(subs, {_arguments});
+			return reddy(subs, {args});
 		}
 
 		// If the arguments are given in a set, then iterate over the set...
 		HandleSeq bset;
-		for (const Handle& h : _arguments->getOutgoingSet())
+		for (const Handle& h : args->getOutgoingSet())
 		{
 			HandleSeq oset;
 			oset.emplace_back(h);
@@ -418,7 +424,7 @@ Handle PutLink::do_reduce(void) const
 		// But what if `h` is not itself executable, but one of it's
 		// deeper elements is? What then? Oy, this is a mess.
 		HandleSeq oset;
-		for (const Handle& h: _arguments->getOutgoingSet())
+		for (const Handle& h: args->getOutgoingSet())
 		{
 			if (PUT_LINK == h->get_type())
 				oset.push_back(HandleCast(h->execute()));
@@ -434,7 +440,7 @@ Handle PutLink::do_reduce(void) const
 	if (LAMBDA_LINK == vtype)
 	{
 		HandleSeq oset;
-		oset.emplace_back(_arguments);
+		oset.emplace_back(args);
 		return reddy(subs, oset);
 	}
 
@@ -450,7 +456,7 @@ Handle PutLink::do_reduce(void) const
 	}
 
 	HandleSeq bset;
-	for (const Handle& h : _arguments->getOutgoingSet())
+	for (const Handle& h : args->getOutgoingSet())
 	{
 		const HandleSeq& oset = h->getOutgoingSet();
 		try
--- a/opencog/atoms/execution/Instantiator.cc
+++ b/opencog/atoms/execution/Instantiator.cc
@@ -310,18 +310,6 @@ Handle Instantiator::walk_tree(const Handle& expr, bool silent)
 		Handle hexpr(beta_reduce(expr, *_vmap));
 		PutLinkPtr ppp(PutLinkCast(hexpr));
 
-		Handle pargs = ppp->get_arguments();
-		Handle gargs = walk_tree(pargs, silent);
-		if (gargs != pargs)
-		{
-			HandleSeq groset;
-			if (ppp->get_vardecl())
-				groset.emplace_back(ppp->get_vardecl());
-			groset.emplace_back(ppp->get_body());
-			groset.emplace_back(gargs);
-			ppp = createPutLink(groset);
-		}
-
 		// Step two: beta-reduce.
 		Handle red(HandleCast(ppp->execute(_as, silent)));
 
--- a/opencog/atoms/execution/Instantiator.cc
+++ b/opencog/atoms/execution/Instantiator.cc
@@ -310,6 +310,18 @@ Handle Instantiator::walk_tree(const Handle& expr, bool silent)
 		Handle hexpr(beta_reduce(expr, *_vmap));
 		PutLinkPtr ppp(PutLinkCast(hexpr));
 
+		Handle pargs = ppp->get_arguments();
+		Handle gargs = walk_tree(pargs, silent);
+		if (gargs != pargs)
+		{
+			HandleSeq groset;
+			if (ppp->get_vardecl())
+				groset.emplace_back(ppp->get_vardecl());
+			groset.emplace_back(ppp->get_body());
+			groset.emplace_back(gargs);
+			ppp = createPutLink(groset);
+		}
+
 		// Step two: beta-reduce.
 		Handle red(HandleCast(ppp->execute(_as, silent)));
 
--- a/tests/atoms/PutLinkUTest.cxxtest
+++ b/tests/atoms/PutLinkUTest.cxxtest
@@ -96,6 +96,7 @@ public:
 	void test_quoted_eval();
 	void test_quoted_lambda();
 	void test_arguments_execution();
+	void test_arguments_execution_with_lambda();
 };
 
 #define N _as.add_node
@@ -2106,6 +2107,58 @@ void PutLinkUTest::test_arguments_execution()
 	logger().info("BEGIN TEST: %s", __FUNCTION__);
 
 	std::string rs = _eval.eval("(load-from-path \"tests/atoms/put-execute-arguments.scm\")");
+	logger().debug() << "rs = " << rs;
+
+	Instantiator inst(&_as);
+	Handle put = _eval.eval_h("put-execute-arguments"),
+		result = HandleCast(inst.execute(put)),
+		expected = _eval.eval_h("expected-put-execute-arguments");
+
+	logger().debug() << "result = " << oc_to_string(result);
+	logger().debug() << "expected = " << oc_to_string(expected);
+
+	TS_ASSERT_EQUALS(result, expected);
+
+	logger().info("END TEST: %s", __FUNCTION__);
+}
+
+/**
+ *
+ * Test
+ *
+ * (Inheritance (Concept "B") (Concept "Parent"))
+ * (Inheritance (Concept "C") (Concept "Parent"))
+ *
+ *  (PutLink
+ *  (LambdaLink
+ *   (VariableNode "x")
+ *   (EvaluationLink
+ *    (PredicateNode "relation")
+ *    (ListLink
+ *    (VariableNode "x")
+ *     (VariableNode "A"))))
+ *  (GetLink
+ *   (Inheritance (Variable "$X") (Concept "Parent"))))
+ *
+ * which should return
+ *
+ *  (SetLink
+ *  (EvaluationLink
+ *   (PredicateNode "relation")
+ *   (ListLink
+ *   (ConceptNode "B")
+ *  (VariableNode "A")))
+ *  (EvaluationLink
+ *   (PredicateNode "relation")
+ *   (ListLink
+ *    (ConceptNode "C")
+ *    (VariableNode "A"))))
+*/
+void PutLinkUTest::test_arguments_execution_with_lambda()
+{
+	logger().info("BEGIN TEST: %s", __FUNCTION__);
+
+	std::string rs = _eval.eval("(load-from-path \"tests/atoms/put-execute-arguments-with-lambda.scm\")");
 	logger().debug() << "rs = " << rs;
 
 	Instantiator inst(&_as);
--- a/tests/atoms/PutLinkUTest.cxxtest
+++ b/tests/atoms/PutLinkUTest.cxxtest
@@ -95,6 +95,7 @@ public:
 	void test_quoted_exec();
 	void test_quoted_eval();
 	void test_quoted_lambda();
+	void test_arguments_execution();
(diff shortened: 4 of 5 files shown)


tests/atoms/put-execute-arguments.scm
```diff
@@ -0,0 +1,27 @@
...
+  (EvaluationLink
+   (PredicateNode "relation")
+   (ListLink
+    (ConceptNode "B")
+    (VariableNode "A")))
+  (EvaluationLink
+   (PredicateNode "relation")
+   (ListLink
+    (ConceptNode "C")
+    (VariableNode "A")))))
```
linas: your editor seems confused about what the end of a file should look like; maybe you are running it in binary-edit mode, instead of text mode?

linas:
Good catch, reasonable fix. The original idea of lazy evaluation vs. eager evaluation is to avoid evaluating the args if the PutLink threw them away.  That is, avoid the call `walk_tree` unless it is actually needed. I kind-of don't want to go back to eager evaluation, if possible.  So I'm thinking the bug is actually somewhere else.  

I'm thinking that the bug is actually in `PutLink.cc` itself, somewhere between lines 293 and 463.   I'm thinking that line 423 needs to be changed from `if (PUT_LINK == h->get_type())` to `if(h->is_executable())` just like the comments in 413-419 suggest.  In addition to this change, maybe a better fix is to change line 388 to look like this:
```
--- a/opencog/atoms/core/PutLink.cc
+++ b/opencog/atoms/core/PutLink.cc
@@ -386,7 +386,12 @@ Handle PutLink::do_reduce(void) const
        {
                if (SET_LINK != vtype)
                {
+if (_arguments->is_executable()) {
+  Handle result(_arguments->execute());
+  return reddy(subs, {result});
+} else{
                        return reddy(subs, {_arguments});
+}
                }
 
                // If the arguments are given in a set, then iterate over the set...
```

Would that also fix the bug? If so, then I think it is a better fix in the long-term.

linas:
And an even better fix might be do to this only if the body is a lambda, so that only lambda get eager-evaluated, everything else stays lazy.
```
--- a/opencog/atoms/core/PutLink.cc
+++ b/opencog/atoms/core/PutLink.cc
@@ -293,6 +293,7 @@ Handle PutLink::do_reduce(void) const
 {
        Handle bods(_body);
        Variables vars(_varlist);
+Handle args(_arguments);
        PrenexLinkPtr subs(PrenexLinkCast(get_handle()));
 
        // Resolve the body, if needed. That is, if the body is
@@ -327,10 +328,13 @@ Handle PutLink::do_reduce(void) const
                vars = lam->get_variables();
                btype = bods->get_type();
                subs = lam;
+if (args->is_executable()) {
+args=args->execute();
+}
        }
 
        // Now get the arguments that we will plug into the body.
-       Type vtype = _arguments->get_type();
+       Type vtype = args->get_type();
        size_t nvars = vars.varseq.size();
 
        // FunctionLinks behave like pointless lambdas; that is, one can
@@ -348,7 +352,7 @@ Handle PutLink::do_reduce(void) const
                if (LIST_LINK == vtype)
                {
                        HandleSeq oset(bods->getOutgoingSet());
-                       const HandleSeq& rest = _arguments->getOutgoingSet();
+                       const HandleSeq& rest = args->getOutgoingSet();
                        oset.insert(oset.end(), rest.begin(), rest.end());
                        return createLink(oset, btype);
                }
@@ -356,13 +360,13 @@ Handle PutLink::do_reduce(void) const
                if (SET_LINK != vtype)
                {
                        HandleSeq oset(bods->getOutgoingSet());
-                       oset.emplace_back(_arguments);
+                       oset.emplace_back(args);
                        return createLink(oset, btype);
                }
 
                // If the arguments are given in a set, then iterate over the set...
                HandleSeq bset;
-               for (const Handle& h : _arguments->getOutgoingSet())
+               for (const Handle& h : args->getOutgoingSet())
                {
                        if (LIST_LINK == h->get_type())
                        {
@@ -386,12 +390,12 @@ Handle PutLink::do_reduce(void) const
        {
                if (SET_LINK != vtype)
                {
-                       return reddy(subs, {_arguments});
+                       return reddy(subs, {args});
                }
 
                // If the arguments are given in a set, then iterate over the set...
                HandleSeq bset;
-               for (const Handle& h : _arguments->getOutgoingSet())
+               for (const Handle& h : args->getOutgoingSet())
                {
                        HandleSeq oset;
                        oset.emplace_back(h);
@@ -418,7 +422,7 @@ Handle PutLink::do_reduce(void) const
                // But what if `h` is not itself executable, but one of it's
                // deeper elements is? What then? Oy, this is a mess.
                HandleSeq oset;
-               for (const Handle& h: _arguments->getOutgoingSet())
+               for (const Handle& h: args->getOutgoingSet())
                {
                        if (PUT_LINK == h->get_type())
                                oset.push_back(HandleCast(h->execute()));
@@ -434,7 +438,7 @@ Handle PutLink::do_reduce(void) const
        if (LAMBDA_LINK == vtype)
        {
                HandleSeq oset;
-               oset.emplace_back(_arguments);
+               oset.emplace_back(args);
                return reddy(subs, oset);
        }
 
@@ -450,7 +454,7 @@ Handle PutLink::do_reduce(void) const
        }
 
        HandleSeq bset;
-       for (const Handle& h : _arguments->getOutgoingSet())
+       for (const Handle& h : args->getOutgoingSet())
        {
                const HandleSeq& oset = h->getOutgoingSet();
                try
```

linas:
Could you try out one of these fixes (I guess the second fix) and see if you can make it work?  You may need to tweak something to make it work, my code above is just pseudocode, it might not compile as written...

ngeiswei:
@linas said
> And an even better fix might be do to this only if the body is a lambda, so that only lambda get eager-evaluated, everything else stays lazy.

Why? isn't
```
Put
  <vardecl>
  <body>
  <args>
```
sugar syntax for
```
Put
  Lambda
    <vardecl>
    <body>
  <args>
```
?

stellarspot:
The lazy version of PutLink arguments execution does not work because the problem is that it is not know is the result of the execution SetLink or not.
Because if it is a SetLink then PutLink needs to be applied to SetLink outgoing set.

Here is an example which does not use LambdaLink:
```
(Inheritance (Concept "B") (Concept "Parent"))
(Inheritance (Concept "C") (Concept "Parent"))

(define put-link
 (PutLink
   (EvaluationLink
    (PredicateNode "relatives")
    (ListLink
     (VariableNode "$Z")
     (ConceptNode "A")))
  (GetLink
   (Inheritance (Variable "$X") (Concept "Parent")))))

(display
 (cog-execute! put-link))
```

Version with arguments lazy execution first substitute GetLink into EvaluationLink and then execute the GetLink. The result is:
```
(EvaluationLink
   (PredicateNode "relatives")
   (ListLink
      (SetLink
         (ConceptNode "B")
         (ConceptNode "C")
      )
      (ConceptNode "A")
   )
)
```

The expected result is:
```
(SetLink
   (EvaluationLink
      (PredicateNode "relatives")
      (ListLink
         (ConceptNode "B")
         (ConceptNode "A")
      )
   )
   (EvaluationLink
      (PredicateNode "relatives")
      (ListLink
         (ConceptNode "C")
         (ConceptNode "A")
      )
   )
)
```

To use the lazy execution it needs to know if the result of execution link is SetLink or not.

stellarspot:
I updated the fix:
- Moved PutLink argument execution into PutLink.do_reduce() method
- Added a test with PutLink where LambdaLink is not used

linas:
> Why? Isn't `Put` sugar syntax for ...

No. it was always meant to be in prenex form. Lambdas may or may not be in prenex form; I think that's kind of the whole point of having ReWriteLink and PrenexLink -- to solve the core issues of trying to figure out "what does lambda even mean?" for a graph-rewriting system.

linas:
Thank you @stellarspot !

The short answer to your questions is that the problem with `walk_tree` is that it is not very modular, its not "OO-style".  I'm attempting to move more and more code and branching-logic into the the `execute()` method of individual C++ classes, instead of having a master-class like `walk_tree` be responsible for the logic flow. 

The reason that `walk_tree` exists is because it was written a long time ago, when there were only three c++ classes: `class Node` and `class Link` and their base  `class Atom`.  It took a while to understand that other derived classes were possible, and, after that, it took a while to understand how to use them  (the fourth class was `class BindLink`, and all that it did was a static analysis for the pattern it contained (it "compiled" the pattern) ... it did not actually execute the search.  Moving execution into the c++ `execute()` method came much later...)

The need for `lazy evaluation` first became clear when I wrote my first tail-recursive loop in Atomese.  It was a `DefinedPredicate` which called itself ... a classic infinite loop. With eager evaluation, it would have to be expanded an infinite number of times, before being executed... obviously impossible. With lazy evaluation, it is expanded only once per loop iteration, to determine if the loop should halt or continue.  Once you think about that, you realize lots of things could be or should be lazy-evaluated ... (see `examples/atomspace/recursive-loop.scm`)

The experimentation is not over. There is a slow process of discovering good ideas, and a slow process to converting the code to actually use the good ideas.  The conversion takes years... The next major conversion needs to be a gradual replacement of the use of `SetLink` by `MemberLink` -- but this is hard - it will be confusing, hard to understand, and will take a long time, but it will make the system better in the long run.

linas:
@stellarspot 

Can I get you to do one more thing?  Can you write a unit test for something like this:
```
(define put-link
 (PutLink
  (LambdaLink
   (VariableList (Variable "x") t (Variable "y")
   (EvaluationLink
    (PredicateNode "relatives")
    (ListLink
     (Variable "x") 
     (Variable "y")
     (Concept "A"))))
(ListLink
  (GetLink   (Inheritance (Variable "$X") (Concept "Father")))
  (GetLink   (Inheritance (Variable "$Y") (Concept "Mother"))))
```
 I suspect it will fail, because I suspect there also needs to be an `if (is_executable()) execute();` in the part that unwraps the `ListLink` arguments.

stellarspot:
The ListLink with GetLink inside return empty SetLink:
```
(Inheritance (Concept "Jim") (Concept "Father"))
(Inheritance (Concept "Jane") (Concept "Mother"))

(define put-link
 (PutLink
  (LambdaLink
   (VariableList (Variable "x") (Variable "y") )
   (EvaluationLink
    (PredicateNode "relatives")
    (ListLink
     (Variable "x")
     (Variable "y")
     (Concept "A"))))
  (ListLink
   (GetLink (Inheritance (Variable "$X") (Concept "Father")))
   (GetLink (Inheritance (Variable "$Y") (Concept "Mother"))))))

(display
 (cog-execute! put-link))
```

Output:
```
(SetLink
)
```

stellarspot:
The ListLink with GetLink inside return empty SetLink:
```
(Inheritance (Concept "Jim") (Concept "Father"))
(Inheritance (Concept "Jane") (Concept "Mother"))

(define put-link
 (PutLink
  (LambdaLink
   (VariableList (Variable "x") (Variable "y") )
   (EvaluationLink
    (PredicateNode "relatives")
    (ListLink
     (Variable "x")
     (Variable "y")
     (Concept "A"))))
  (ListLink
   (GetLink (Inheritance (Variable "$X") (Concept "Father")))
   (GetLink (Inheritance (Variable "$Y") (Concept "Mother"))))))

(display
 (cog-execute! put-link))
```

Output:
```
(SetLink
)
```

linas:
> The ListLink with GetLink inside return empty SetLink:

Yep. Any chance I could convince you to fix this one remaining bit? While the idea is still hot in your mind?

stellarspot:
ListLink is not executable link so it seems that it is necessary to explicitly check ListLink outgoing set if it is necessary to execute them.

It looks like that it is not only problem of the ListLink but any link which contains executable links inside.

It seems that the object oriented solution where each link knows should it be executed or not does not work in this case. It requires to walk over link outgoing tree and execute executable atoms.
If lazy initialization is required, may be it is better to explicitly provide it by LazyLink. The first execution just removes LazyLink and the second execution actually executes the atom.

linas:
@stellarspot you are over-thinking this. It is just not that hard.  Near the location of your original patch, about 5 or 10 or 20 lines away, there is an if-statement that says `if (LIST_LINK == atomtype) { ..}` all you need to do is to stick the execute into that block of code, too.   It's about 3-4 lines of code, total, in the same style as before, in a nearby area.

stellarspot:
What about if there is one more ListLink inside the ListLink, should the ListLink itself check that its values are executable?
```
(Inheritance (Concept "Jim") (Concept "Father"))
(Inheritance (Concept "Jane") (Concept "Mother"))

(define put-link
 (PutLink
  (LambdaLink
   (VariableList (Variable "x") (Variable "y") )
   (EvaluationLink
    (PredicateNode "relatives")
    (ListLink
     (Variable "x")
     (Variable "y")
     (Concept "A"))))
  (ListLink
   (ListLink
    (GetLink (Inheritance (Variable "$X") (Concept "Father"))))
   (ListLink
    (GetLink (Inheritance (Variable "$Y") (Concept "Mother")))))))

(display
 (cog-execute! put-link))
```

linas:
> one more ListLink inside the ListLink, should the ListLink itself check that its values are executable?

No. ListLink does not execute arguments. ListLink is a tool to pass arguments to functions. It is not meant to be anything fancy.

stellarspot:
The ListLink can contain other links and that links can contain executable links.
If pass such executable links by-lazy each of them can be passed in different places and be executed different times which can have performance impact.

It is better to explicitly note  which executable links should be executed before passing to PutLink body and which should be passed by lazy.

linas:
No. You are wildly over-thinking and over-engineering the problem.  This is supposed to be a very simple 3 or 4 lines of code fix.   The PutLink is a beta-redex. That's all that it is. You seem to be thinking of PutLink as some kind of `ApplyLink` which knows how to apply functions to arguments.   It is explicitly NOT an applier!  It is a single-step beta-reduction.

stellarspot:
> The PutLink is a beta-redex. That's all that it is.

The fix for the current pull request shows that arguments of PutLink needs to be executed (because it is not known a priory if result of the arguments execution is SetLink or not). [PutLink.cc#L301](https://github.com/stellarspot/atomspace/blob/a1c184ff3312db0e6de2efde1652950c82c8cf3f/opencog/atoms/core/PutLink.cc#L301)
If PutLink is just a beta-redex,  execution of the PutLink arguments would not be necessary at all.

If PutLink arguments are always executed even there are cases there it does not need to do there is a question is it possible to mention somehow explicitly that arguments need to passed by lazy.

One way to do it is to explicitly mention it in the PutLink variables declaration.
For example:
```
  (PutLink
     (LazyVariableLink        ;; lazy variable declaration
        (VariableNode "$X"))
  ...
  )
``` 
 
> This is supposed to be a very simple 3 or 4 lines of code fix.

I believe that the issue not that ListLink does not execute GetLinks in the PutLink.
Just substituting LambdaLink arguments with not executed GetLinks should work as well.

So the initial query:
```
(define put-link
 (PutLink
  (LambdaLink
   (VariableList (Variable "x") (Variable "y") )
   (EvaluationLink
    (PredicateNode "relatives")
    (ListLink
     (Variable "x")
     (Variable "y")
     (Concept "A"))))
  (ListLink
   (GetLink (Inheritance (Variable "$X") (Concept "Father")))
   (GetLink (Inheritance (Variable "$Y") (Concept "Mother"))))))
``` 

after LabbdaLink arguments substitution should look like:
```
(define put-link
 (PutLink
   (EvaluationLink
    (PredicateNode "relatives")
    (ListLink
     (GetLink (Inheritance (Variable "$X") (Concept "Father")))
     (GetLink (Inheritance (Variable "$Y") (Concept "Mother")))
     (Concept "A")))
``` 
And after execution it needs to return the expected result.

Fixing it in the way that PutLink arguments need to execute values in ListLink just hides the real issue and leads that some other links in PutLink arguments need to execute their outgoing atoms.

linas:
> after execution

No. There is nothing left to put, after that. Again Put is a beta-redex. https://en.wikipedia.org/wiki/Beta_normal_form

linas:
The problem is that if one inverts the order of execution, things become a real mess.  You can't arbitrarily redefine the order of execution without affecting whether a system is confluent or not.   https://en.wikipedia.org/wiki/Confluence_(abstract_rewriting) 

Read also the related articles:
https://en.wikipedia.org/wiki/Rewriting
https://en.wikipedia.org/wiki/Normal_form_(abstract_rewriting)
https://en.wikipedia.org/wiki/Canonical_form
https://en.wikipedia.org/wiki/Abstract_rewriting_system

Note that the atomspace is not a "term rewriting system", it is a "graph rewriting system". The two are almost identical to one-another, but there are subtle differences that seem to crop up in various places.

stellarspot:
> No. There is nothing left to put, after that. Again Put is a beta-redex. https://en.wikipedia.org/wiki/Beta_normal_form

> The problem is that if one inverts the order of execution, things become a real mess. You can't arbitrarily redefine the order of execution without affecting whether a system is confluent or not. https://en.wikipedia.org/wiki/Confluence_(abstract_rewriting)

I see some contradiction from what is stated about PutLink and what really PutLink does.

Here is a simple example where ExecutionOutputLink is an argument of PutLink with LambdaLink as body:
```
(Define
 (DefinedSchema "sqr")
 (Lambda
  (Variable "$X")
  (Times
   (Variable "$X")
   (Variable "$X"))))

(display
 (cog-execute!
  (Put
   (Lambda
    (Variable "$X")
    (Plus
     (Number "2")
     (Variable "$X")))
   (ExecutionOutput
    (DefinedSchema "sqr")
    (Number "2")))))
```

First I run it without the fix #2276 where PutLink executes its arguments.
The steps are the following:
1) The LambdaLink argument is substituted by  ExecutionOutputLink:
```
(PlusLink
  (NumberNode "2.000000")
  (ExecutionOutputLink
    (DefinedSchemaNode "sqr")
    (NumberNode "2.000000") ))
```

2) The Instantiator calls `walk_tree` method on the result of the PutLink execution:
https://github.com/opencog/atomspace/blob/1fa59c57b5b7833c5ab7d21bcfa55c0e21c27108/opencog/atoms/execution/Instantiator.cc#L327

The final result is `(NumberNode "6.000000")`

Second, I run the same code on the current version of AtomSpace where PutLink executes its arguments.
The steps are the following:
1) PutLink executes the ExecutionOutputLink and passes the result to LambdaLink:
```
(PlusLink
  (NumberNode "2.000000")
  (NumberNode "4.000000"))
```
2) The Instantiator calls `walk_tree` method on the result of the PutLink execution:
The result is `(NumberNode "6.000000")`

What I see that the execution of PutLink arguments is not a beta-redux and it can affect whether a system is confluent.

Fix #2276 executes PutLink arguments. It is not possible to avoid it because it is not know is result of execution SetLink or not.

But execution of PutLink arguments can affect another samples that behave differently if PutLink arguments are executed instead of using beta redex.

As a possible solution I can  suggest to add some kind of LazyLink which can be used when someone knows that a PutLink arguments must not be executed.
Without such ability PutLink arguments are always executed even they do not need to be executed.

Issue #1853, opened 2018-08-01
Should we add an ArgumentList?


rTreutlein:
I have been working on Distributional Values recently and am using these for joint probability distributions which are attached to Cartesian Product which can easily be represented by a ListLink. The issue comes up when a VariableNode is grounded by this ListLink and I want to pass it to an EvaluationLink. The EvaluationLink sees the ListLink and assumes that you want to provide it with multiple Arguments even though that is not the case.
To avoid this we could add an ArgumentList which acts like a ListLink would right now. To allow people to specify when they wish to have multiple Arguments and a when to treat the Product as a whole.

linas:
The obvious work-around is to have a ListLink that wraps the ListLink.

Maybe having an ArgumentList is a good idea. But maybe having a CartesianProductLink is a better idea.  When would you use one instead of the other?  When does the ListLink get used? How does existing code change to accommodate this?  Or maybe something completely different, like getting rid of the ListLink in an EvaluationLink (so that you have just `(Evaluation Pred arg arg arg)` instead of wrapping args in a list).

I've run around in circles, over and over, contemplating all of these different kinds of variations. Adding more link types just adds more complexity, more confusion, more unit tests, more exceptions, if-and-but's in the documentation, so the payoff has to be worth it.  So far, in developing atomese, I've tried to work through all of the different ways of doing things the different alternatives, and pick the alternatives that seem to fit together best.  But this is a global-optimization problem; its hard.  Usually, a gain, a win, here, means that something somewhere else loses or gets uglier.  Usually, it takes about a year to figure it out, after one or two failed implementations. (I'm not joking about it taking a year).

In my new job title, I won't have much time to ponder these tradeoffs. That means you or others have to.  That means you should describe the actual problem you're having above in much greater detail (cause I don't fully understand it)  Maybe there's some easy work-around? I can't tell.

ngeiswei:
@rTreutlein A more detailed description of the problem, with examples, would be welcome indeed.

vsbogd:
Hi @rTreutlein , as far as I understand you always need to designate GroundedSchemaNode arguments, so last part of the BindLink should be:
```
ExecutionOutPutLink
                GroundedSchemaNode "scm: modus-ponense-formula"
                ListLink
                    VariableNode "$B"
                    VariableNode "$A"
                    InheritanceLink
                        VariableNode "$A"
                        VariableNode "$B"
```

ngeiswei:
[...]

PR #2285, opened 2019-07-23
Distributional Values

rTreutlein:
This is an improved and cleaned up version of my previous work on Distributional Values. All the code is in opencog/atoms/distvalue though it might make sense to move certains parts out of there. For example the CoverTree implementation could be moved to cogutils.

Atm this more of a Review Request to see if anything needs to be changed/improved.


opencog/atoms/distvalue/CTHist.cc
```diff
@@ -0,0 +1,399 @@
...
+
+			//Add all descendants to the parent
+			for (int c_idx : desc)
+			{
+				_nodes[c_idx].children.clear();
+				insert_rec(c_idx,_nodes[parent],level+1);
+			}
+			_nodes[nearest_idx].children.clear();
+
+			//Readd our node
```
linas: It is conventional, in C++, to have a single blank space after the `//` delimiter.

opencog/atoms/distvalue/CTHist.cc
```diff
@@ -0,0 +1,399 @@
...
+using namespace opencog;
+
+template <typename val_t>
+void CTHist<val_t>::insert(const DVec & pos,const val_t & value)
+{
+	insert(CoverTreeNode<val_t>(pos,value));
+}
+
+template <typename val_t>
+void CTHist<val_t>::insert(const CoverTreeNode<val_t> & x)
```
linas: This and all of the other methods should be documented.  In addition, there should 
be a general description of what a cover-tree is, and why it is needed, and how it is used.
(I use histograms all the time in the language-learning code, and can't even begin to
imagine why a cover tree is needed)

opencog/atoms/distvalue/CTHist.h
```diff
@@ -0,0 +1,181 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program; if not, write to:
+ * Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#ifndef _OPENCOG_CTHIST_H
+#define _OPENCOG_CTHIST_H
+
+#include <opencog/atoms/distvalue/CoverTree.h>
```
linas: I did not read this carefully yet, but it appears that there is nothing in 
the cover-tree code that requires atoms or the atomspace.  As such, 
if it is really generic code, then the `cogutil` repo would be a better place for it.

opencog/atoms/distvalue/DistributionalValueSCM.h
```diff
@@ -0,0 +1,95 @@
...
+	SCM ss_cdv_get_conditions(SCM);
+	SCM ss_cdv_get_unconditonals(SCM);
+	SCM ss_cdv_get_unconditonal(SCM,SCM);
+	SCM ss_cdv_get_joint(SCM,SCM);
+	SCM ss_cdv_merge(SCM,SCM);
+	SCM ss_cdv_join(SCM,SCM);
+
+	SCM ss_dv_merge_hi_conf(SCM,SCM);
+	SCM ss_cdv_merge_hi_conf(SCM,SCM);
+	//SCM ss_cdv_cde(SCM,SCM);
```
linas: Do you actually need all of the above? This is a pretty sharp and direct violation 
of the AtomSpace and scheme binding policy; the AtomSpace is not a programming
API meant for for human programmers!  Scheme (or python or haskell...) bindings
cannot be introspected, and thus are utterly useless and opaque for any algorithm.
Only human programmers can use the scheme bindings, and we are trying to 
keep humans out of this layer.  So .. again .. do you really need these?  Why?
Can you design it so that fewer of these (ideally, zero of these) are needed?

opencog/atoms/distvalue/README.md
```diff
@@ -0,0 +1,44 @@
+Distributional Values
+=====================
+
+An implementation of Distributional Values that might eventually replace all current TruthValues.
+
+Issues and Goals
+----------------
+
+One of the problems this is trying to solve is the combination of fuzzy and propabalistc logic. As both are used within OpenCog it is important to not only be able to clearly represent and distinguise them but also use them together in a combined for. That way no matter if the alogrithm expects fuzzy or propabalistc logic values they can work on a DistributionalValue. Which improves interoperability.
+Another issue this improves upon is the accuracy and detail of TruthValues albeit at the cost of higher complexity.
```
linas: please wrap lines so that they fit in an 80-column terminal (i.e. wrap at the 72nd column)
Also, please run the spell-checker.

linas:
Oh, and one more thing: the new `PRODUCT_LINK` - I'd like to discourage that (and/or have it in a distinct pull req) Formally, the current `LIST_LINK` is meant to be the same thing as a product type, and I'd rather not introduce a new type for this, without some very strong reason for it.

opencog/atoms/distvalue/DistributionalValueSCM.h
rTreutlein: I do need those as they are used by some PLN rules that work with DVs instead of TVs. Which i created to check weather this is actualy usable. But maybe this could be split off and moved to opencog.

rTreutlein:
Regarding the PRODUCT_LINK when i was creating the PLN example using DVs I initially used LIST_LINK instead but there is some special behaviour associated with LIST_LINK that was causing issues which was discussed in #1853 . There might be better solutions for this but this was the easiest to implement at the time and as it was primarily for testing i felt it was good enough.

opencog/atoms/distvalue/CTHist.h
ngeiswei: @linas it's because his histograms can be multi-dimensional. I agree that cover-tree should be in `cogutil`, but I actually wonder if this should not go to its own repo, since it is experimental and should remain experimental for a while. @rTreutlein, what do you think?

ngeiswei: By "this" I mean the whole PR, not just covertree.

rTreutlein: Sounds reasonable.

ngeiswei: OK, I've created https://github.com/opencog/distributional-value you can re-create your PRs on that repo.

rTreutlein:
Closing this as I'll be moving all this into it's own repository as per discussion.

linas:
OK.

linas:
OK, except that in #1853, the conclusion was that ListLink is fine, and that you don't need the ProductLink.

Also -- can you at least informally explain, here, why cover-trees and nearest-neighbors are needed for distributions? I'm just curious, right now, I can't even begin to imagine what the relationship is...

linas:
> per discussion.

where?

linas:
>> per discussion.

> where?

Oh, OK, github wasn't refreshing my browser window. I see it now.

opencog/atoms/distvalue/DistributionalValueSCM.h
linas: @rTreutlein FYI, I am urging @ngeiswei to convert the PLN formulas to use `PREDICATE_FORMULA_LINK` which evaluates the formulas using atomese (PlusLink, TimesLink, etc.) There are two reasons for this conversion:

1) It makes formulas at least 10x if not 100x faster than calling `GroundedPredicate "scm:foobar"` because calling `scm:foobar` incurs a 50 microsecond overhead (20K calls/sec) Same is true for `python:whatever`

2) In some whizzy future, things like `asmoses` or whatever learning algo could auto-learn new formulas, place them directly in the atomspace, and just go; no need for jumping through hoops to jam it into scm and then call that scm.

I figure you should know this, as it may alter your thinking about such things.

linas: See `examples/atomspace/formulas.scm` for a working example

opencog/atoms/distvalue/CTHist.h
linas: what's multi-dimensionality got to do with nearest neighbors and cover-trees?

rTreutlein: Well the real issue are dynamic multi dimensional Histograms. Because we don't know beforehand in what range the data will be we can't generate the bins beforehand. So we need to somehow generate the bins on the fly. There aren't really any data structures that can store and update arbitery sized bins without requiring you to rebuild most he structure each time. Wich is why i decided to just store the centers of bins. There still not many structures that can store these points in ordered way that can be easily update but Covertrees can do that better then most as far as i could find. And of course since we only have the centers to work with now it's quite important to find the closest center to some point of interest wich is where nearest neighbor search comes in.
It took quite a while to find a data representation that fit all the requirements. But as the readme doesn't explain that well enough atm ill work on that somemore so it's understandable why covertrees are the right choice in this case.

linas: If you can't generate the bins beforehand, then how can you know where the center of the bin is?  But if you do know what the centers of the bins are, then the edges of the bins are just at the half-way-point between two bins!?  I assume that you are still working with rectangular bins, when you go multi-dimensional.

(I fear confusing the issue, but should point out: when the bins aren't rectangular, then you get something called a "Voronoi diagram" in 2D and a "Wigner–Seitz cell" in general N-dimensions. The bin boundaries (cell boundaries) are still given by the half-way points between bin centers.)

It sounds to me like perhaps you are doing, or are attempting to do something called "agglomerative clustering", of assigning new points to the "nearest cluster".  I'm not sure what the best agglomerative clustering algo is; for the langauge learning, I just rolled my own, because it was easy.  Never occurred to me that I needed to somehow optimize the nearest-neighbor search ... Hmm. I will have to ponder what that means for my algos.  (My data is super-high-dimensional, like millions of dimensions, and is also super-sparse, only one in a trillion bins are non-empty.)

rTreutlein: Ofcourse i also don't know the centers of the bins beforhand. I created them one at a time as the data points come in (each data point becomes a bin center untill i have reach the max number of bins specified). This i could ofcourse also do with normal bins but the issue arises when I have to modify those bins for example splitting a bin or combinign 2 bins. Which i need to do because the first couple of data points that lead to these bins will likely not have created a very good distribution of bins. Doing this is hard without needing to rebuild the whole data structure containing these bins.
Doing this with just centers is easier. And yes i am doing agglomerative clustering. Each data point that get's added to a bin moves the center of the bin a small amount. Which should make the bins converge to a nice distribution over time.
Since only working with centers i have stoped thinking about the bins having a clear boundary so they are certainly not rectangular. I guess i could consider them "Wigner-Seitz cells" but haven't given that much thought yet. What i do if i want to know the expected frequency of a given Value is check the bin center clostes to it and a bin center in the opposite direction to the closest one and then do an interpolation to get my result.
Regarding sparse data my Histograms should be able to deal with that well as they only create bins at locations where we have actually seen data. But haven't really thought about using it for super-high-dimensions or bigger data sets. As i don't expect PLN to generate/use joint distributions that use more then a handfull of dimensions, though this is just my opinion reality might differ.

linas: OK, two questions. first, a 1D example: what happens if you decide you want 50 bins max, and then send in points from 1 to 1000 in sequential order (not random order)? Roughly, where do the bin centers end up (guesstimating?)

Why do you need a cover tree to do what you described?

rTreutlein:
Well for your example the bins would be at 1 to 49 and one at 525
approximately as i am currently assuming that data to be in a somewhat
random order. To improve on that we'd need a algorithm that splits/merges
bins when some threshold is reached. I was thinking about these a while
back but never got to implementing one because the data structures i was
experimenting with weren't flexible enough to facilitate them in higher
dimensions. But even with such improvements sequential clustering
algorithms are always biased on the first couple samples they see.

In the 1D example the covertree doesn't bring any benefit it would be
easier to just use a binary tree. But there is no generalization of a
binary tree for n-dimensions as rebalancing becomes impossible. There are a
lot of other data structures / trees for storing spatial data, where you
build the tree from a complete data set and then have easy access into that
data. But they don't allow you to incrementally build up the data structure
because they also can't be rebalanced. Now covertrees don't care about
dimensions they only care about distances so building it incrementally is
no problem. And then later on if you need to move data points in the tree
that is not completely trivial but doable.

Now i am curious what kind of data structure would you choose?
Here the list of requirements:
- Size linear in the number of bins
- can be incrementally built
- supports n-dimensions
- allows quick access < O(n) to find the bin
- possible to update position of a bin

I have thought quite a long time about it. And maybe i just got stuck in
one mode of thinking but covertrees where the only thing i found which
could reasonably full fill all requirements. And most things i read about
they never had to deal with incrementally building their Histograms which
really complicates things.

Am Do., 25. Juli 2019 um 16:34 Uhr schrieb Linas Vepštas <
notifications@github.com>:

linas:
OK, so you answered my question indirectly... the answer I was fishing for, the way is should be explained in the README, should be something like this:

"In order to assign a new datapoint to an existing cluster, the simplest and most naive procedure is to calculate the distance between the new datapoint and the N existing clusters, and then take the min distance.  This requires N calculations. If the distance calculation is expensive, then one can reduce the number of needed comparisons to O(log N) by using the covertree structure."

What was not clear is that you felt you needed to improve clustering performance, and that covertree was the way to do this.  And before that, it was not clear that you were using clustering instead of histograming.  Which sort-of changes the nature of the whole thing.

Actually, it makes it a lot more interesting to me. I'm trying to solve a rather similar problem in the language-learning code. There are a lot of similarities, but also some differences.

1) I don't give a fig about distributional values.  I just need to assign things to clusters. Once I figure out which cluster it belongs to, I just create `(MemberLink (Concept "thing") (Concept "clustername")) and then typically increment one or more count-values by some float-point weight.  Counting is the easy part. Finding the right cluster is the hard part.

2) During clustering, I have a way of splitting up `(Concept "thing")` into `(Concept "thing-a")` and `(Concept "thing-b")` such that the distance from `(Concept "thing-a")`  to `(Concept "clustername")` is minimized, and the distance from `(Concept "thing-b")` to  `(Concept "clustername")` is maximized.  This is for "word-sense disambiguation", the classic example is "saw", part of which clusters into "cutting implements" and part of which clusters into "looking-verb".

3) I don't have a fixed number of clusters before-hand. I do have a pretty good idea of the distance scale, though: two things can be clustered only if they are close AND ALSO if the distance is less than X; otherwise, start a new cluster.

4) dimensions: I'm confused by what you mean when you say "dimension". In my case, I have hundreds, maybe thousands of bins at the end, and maybe 100K items to place into one of the bins. It is more-or-less irrelevant to the clustering algo, but my "things" live in a very high-dimensional non-euclidean "space" of about 100 million dimensions or maybe more. But this last "dimension" should be irrelevant for clustering.

So I think it would be very very cool if you could provide a clustering tool that had just enough callbacks and decision-point-checks to allow me to do the above.   I've been slowing writing in a roll-your-own algo that does the above. Anton's language learning group isn't as picky, and so they just used some off-the-shelf k-means algo from SciPy or wherever which obviously gives poor results, but that's not thier prime concern right now.

So, yeah -- having a cover-tree based clustering algo that assigned things to clusters with O(log n) compares instead of O(n) compares for n clusters .. that would be awesome. But that is quite totally independent of values, of distributions.  I don't need a distributional value, I need fast clustering.

linas:
Oh, to be clear: in my case, the "cluster centers" live in this 100-million dimensional space. But that's OK, since the atomspace is a tool par excellance for storing ultra-high-dimensional vectors. Viz, whatever the clustering software is, it needs to delegate "cluster center" to my callbacks, as otherwise it would choke.  Also, it needs to delegate "how far away is X and Y" to my code, because the distance is not euclidean, and not isotropic, but I think it is homogeneous.

rTreutlein:
@linas Sorry for answering so late but off your 2 comments my email program only showed the last one so i hadn't read the first one till now.

In regards to your needs first off all i have to note that the findNearestNeighbot algorithm is not O(log(n)) but O(c^6 log(n)) where c is the Expansion Constant (Doubling Constant) of the data. I don't fully understand the meaning of that Constant but i know that is likely to increase as the number of dimensions increase. So CoverTrees might not be optimal for use-case because of this.

1. Well atm, I store the count and position of a bin in the CoverTree. In my case the Position is just as Vector of Doubles and i use that as an identifier but i am guessing that's not the case in your code. But that can be changed to a name or whatever works in your case. 

2. Could you expand on how that would affect clustering? Would you just assign "thing-a" and "thing-b" independently to clusters?

3. That's also easily doable.

4. My focus on dimensions is likely a remnant off me fixating on dynamic histograms, as this where is started thinking about this all. And because they usually do not support more then 1 dimension I tried to find a better way of doing it. And without really realizing i ended up in the domain of clustering. But of-course you are correct for clustering the dimensionality has no direct impact other then that  CoverTrees are sensitive to the Expansion Constant as noted above.

If you want to take a closer look at CoverTrees i would recommend this paper: https://izbicki.me/public/papers/icml2015-faster-cover-trees.pdf

If you want me to I can implement the required changes to the code and provide it to you for experimentation.

linas:
I'm not sure I'm asking you to make changes; more that I am thinking out loud and trying to describe to you the things that I'm doing. Now, what I'm doing seem to be fairly generic, and so I would like to see a good generic solution. But doing that won't be particularly easy, and making good use of it is hard.  

So let me work through a specific example, just to show where I'm coming from. So natural language: One of the core structures is ... well, I need a good name for it. The current code calls it a `SectionLink`. It is like a vertex (of a graph) plus a set of "half edges". I call them "half edges" because they resemble connectors: the other things (other atoms) they could connect to, like snap-together connectors of a plastic toy or the tabs of jigsaw puzzle pieces.  Its a generic way of representing parts of a graph.   These for a "pair", the pair being (vertex, set-of-half-edges) or (v,d) for short. And I count these.  So actually I have a matrix N(v,d).  The rows and the columns are vectors. So this is clearly very "distributional", and this is why its similar to what you've been thinking about. 

The big difference is that N(v,d) is very sparse (for me). So typically, I might have a 10K rows, and 1 million columns.  If I actually stored one float for each entry, that would require 10K x 1M x 4bytes = 40GBytes RAM.  Which is nuts, and only gets worse if I store doubles instead of floats, or if/when I need several other numbers, or store pointers, or the dimensions are different (some of my matricies are actually 10M x 10M in size ...) So when you say "Position is just as Vector of Doubles" -- that's a non-starter for me. It will never-ever fit into available RAM.  In my vectors, almost all of the vector entries are zero. For example, for one of my rows -- which is 1 million long, it will have only a few hundred non-zero entries.  (Actually, my rows follow a Zipf distribution -- half the rows have only a handful of non-zero entries; a tenth of them have hundreds of non-zero entries, and only a 1% of them have more than a few thousand non-zero entries) So -- storing billions of zeros is not a viable strategy.

The good news for me is I have already solved the above problem, for myself: (v,d) is just a pair of Atoms, and N(v,d) is a CountTruthValue. And I have code to compute "marginals" -- row and column subtotals, and also lots of other stuff (e.g. convert the count N(v,d) into a probability p(v,d) normalized to lie between 0.0 and 1.0 -- etc. all this code is in the `opencog/matrix` directory.)

What I don't have (well, I do have, but its only a rough prototype) is a way of clustering these. Clustering is hard (and slow) for several reasons.  The giant-sized vectors is just one of them.  The other problem is that some of the vectors are actually "degenerate": they're a mash-up of vectors that should have been distinct:  So, consider N(v,d) where v is the word "saw" and d is everything that the word "saw" can go with, and N is how often the relation was seen. So: one can "saw wood", and "saw plastic", or one can "saw a star" or "saw a galaxy". So the vector for "saw" is actually a sum of vectors for "cutting-verb", and "past-tense-looking-verb" and "cutting tool", so that's three distinct clusters, and I have to add parts of that vector to each.   So naive clustering is not quite enough...  I think I have several algos that can solve this, and some prototype code ... but again .. this gives you a feel for the problems I'm facing.

So: I'm very very interested in clustering, and I'm very very interested in distributions, but everything for me is super-high-dimensional, and super-sparse.

BTW, besides natural language, there are other examples.  One is in biochem, where the graphs are protein-protein interactions, or regulatory mechanisms between genes, etc.  The other is motion-perception: If I have 3D time-line tracks of a dancer's ankles, wrists, waist and head, can I tell what motion it is? (e.g. is it the "tango cluster", or the "swan lake cluster") Or the 3D time tracks for players on a football field -- what are they about to do? (Or military - positions of soldiers on a battlefield, etc.) again -- these are high-dimensional vectors (dozens or hundreds of 3D positions for ankles, elbows) and very very very sparse (almost all relative movements of ankle-to-elbow are physically impossible).  (The "natural representation" as motion tracks is not sparse, but its also hard to work with, for clustering.)

Issue #2203, opened 2019-06-05
Eliminate MapLink


linas:
MapLink does not seem to serve any useful purpose.  This issue proposes eliminating it, and/or debating reaoson why it should stay (and be improved/changed) Some background comments about it.:

* MapLink was originally proposed as a way to "not search the whole atomspace" but this is misleading:  GetLink already doesn't search the whole atomspace; it only pretends that it does. So MapLink offrs no real perf benefit over GetLink.
* An early idea was that  MapLink potentially allows you to build pipelines ... but so does GetLink. So this is no advantage.
* Turns out SetLink is a bad idea .. see the wiki page for SetLink, and also #1502. I want to replace all users of SetLink by MemberLink.

ngeiswei:
@linas, how to make `GetLink` not search the whole atomspace?

linas:
GetLink already "doesn't search the whole atomspace". it only seems to, conceptually. In practice, it searches only a tiny portion of it. Some examples (copied from #2202) 
```
(MemberLink (Concept "a") (Concept "my set"))
(MemberLink (Concept "b") (Concept "my set"))

(cog-execute! (Get (And
   (PresentLink (MemberLink (Variable "x") (Concept "my set")))
   (Evaluation (GroundedPredicate "scm:filter") (Variable "x")))
)))
```
which searches ONLY the members of "my set". 

Even better:
```
(cog-execute! (Bind
   (And
      (PresentLink (MemberLink (Variable "x") (Concept "my set")))
      (Evaluation (GroundedPredicate "scm:filter") (Variable "x")))  )
   (MemberLink  (Variable "x") (Concept "the next stage set"))))
```
which is a processing pipeline: accepts "my-set" as input, and places results into "the next stage set".  BTW, I use AnchorNode for this (i.e. `(Anchor "my set")` instead of `(Concept "my set")`) - anchors are a cleaner more obvious drop-off/pick-up points.  I suggest everyone could use them for organizing stages of a processing pipeline.

Processing pipelines could be even further improved, e.g. #1750  and/or #1507

ngeiswei:
Ah, as long as it is efficient it's OK, which I'm not entirely sure of, but maybe it's another issue.

vsbogd:
It is question for both this issue and issue #1502. 

For example query "get all baskets which contain only red balls" can be written as:
```scm
(cog-execute! (Get
  (Variable "basket")
  (And
    (Inheritance (Variable "basket") (Concept "basket"))
[...]

PR #2288, opened 2019-07-24
Implement AlwaysLink (for-all link in pattern matcher)

linas:
Per discussions in #2203 -- The `PresentLink` in the pattern matcher is a 
kind of `ThereExistsLink` but custom tailored so it can be used with the 
patten matcher. So of course, we need something analogous to a `ForAllLink`. 
Since that name is already taken by PLN, for a kind of ScopeLink, it seemed 
easier to just introduce the `AlwaysLink`. It is not clear if a `NeverLink` is needed.
(because "not always" is "sometimes")

This is just the most basic, bare-bones implementation. More complex uses are
probably buggy.


linas:
Merging. `UnifyUTest` in the URE is failing, but I believe that is totally 
unrelated: the URE unifier does not use the pattern matcher.

Issue #2281, opened 2019-07-22
Circle-CI is not running the actual code that in in the pull req


linas:
Circle-ci appears to be testing an earlier version of the code, 
and NOT the code in the pull req!

This is easily verified: intentionally introduce a bug. Create a pull req.
The unit tests will pass. Merge the pull req.  Now create a second 
pull req, fixing that bug. The unit tests will fail!

It doesn't have to be as drastic as the above -- just find some print
statement, change what it prints.  Circle-ci will no print the changed printf,
it will print the old one.

Likewise for compiler warnings -- create some code that generates
a warning,  you won't see that warning in the current run, you'll see
it in the next run.

I've done all three of the above, unintentionally, over the last few weeks,
and was surprised by the results. But now the pattern is clear: circle-ci
is not testing the pull req; it is testing what has already been merged.

vsbogd:
Looks like it works so after https://github.com/opencog/atomspace/pull/2266

vsbogd:
@ngeiswei, what is the reason of replacing CircleCI `-checkout` by manual checkout?

ngeiswei:
It seems faster but maybe I'm mistaken.

vsbogd:
But it checkouts current `master` instead of proper branch. Do you think CircleCI checkouts history as well? Probably it is a git restriction, recently I found that it is possible to clone from branch, event from tag, but it seems it is not possible clone from commit.

AmeBel:
Yap #2266 explains it. circle-ci `checkout` keyword will clone the branch that triggers the build. By always cloning master, pull-requests are not tested.

ngeiswei:
OK, my bad.

ferrouswheel:
If you're going to mess with the CI,  please at least get it reviewed by someone before merging it. Getting it wrong messes up everyone's development flow.

linas:
So I think this has been fixed everywhere, now...

PR #2289, opened 2019-07-24
Use circleci checkout command instead of git clone

ngeiswei:
Otherwise it checks out the master not the current branch!


ngeiswei:
Fix for #2281

ngeiswei:
I'm merging as it's clearly the right thing to do.

PR #2297, opened 2019-07-29
use c++ api in cython atom.__hash__ method

noskill:
before this fix hash(atom) would return different hashes for the same atoms in different atomspaces.

### DIFF
--- a/opencog/cython/opencog/atom.pyx
+++ b/opencog/cython/opencog/atom.pyx
@@ -151,8 +151,4 @@ cdef class Atom(Value):
             return -1
 
     def __hash__(self):
-        cdef cAtom* atom_ptr = self.handle.atom_ptr()
-        if atom_ptr == NULL:
-            raise RuntimeError("hash called on null pointer")
-        cdef ContentHash h = deref(atom_ptr).get_hash()
-        return PyLong_FromLongLong(h)
+        return PyLong_FromLongLong(self.get_c_handle().get().get_hash())
--- a/opencog/cython/opencog/atom.pyx
+++ b/opencog/cython/opencog/atom.pyx
@@ -1,4 +1,4 @@
-from cpython cimport PyLong_FromSize_t
+from cpython cimport PyLong_FromLongLong
 
 # Atom wrapper object
 cdef class Atom(Value):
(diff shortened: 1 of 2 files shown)


opencog/cython/opencog/atom.pyx
```diff
@@ -148,8 +150,9 @@ cdef class Atom(Value):
...
-        # Use the address of the atom in memory as the hash.
-        # This should be globally unique, because the atomspace
-        # does not allow more than one, ever.
-        return hash(PyLong_FromVoidPtr(a1.handle.atom_ptr()))
+    def __hash__(self):
+        cdef cAtom* atom_ptr = self.handle.atom_ptr()
+        if atom_ptr == NULL:
+            raise RuntimeError("hash called on null pointer")
+        cdef ContentHash h = deref(atom_ptr).get_hash()
+        return PyLong_FromLongLong(h)
```
vsbogd: I would replace it by:
```cython
return PyLong_FromLongLong(self.get_c_value_ptr().get().get_hash())
```

opencog/cython/opencog/atom.pyx
```diff
@@ -1,3 +1,5 @@
+from cpython cimport PyLong_FromLongLong
```
vsbogd:
```suggestion
from cpython cimport PyLong_FromSize_t
```

noskill:
Reverted commit with PyLong_FromSize_t, since it fails with cython from our docker image

noskill:
On some hashes current solution overflows

linas:
Oh .. shouldn't it be `PyLongLong_FromSize_t` ? Both `size_t` and `long long` are 64-bit, but I think PyLong is only 32-bit. Or something like that ...

noskill:
It should be PyLong_FromUnsignedLongLong, but it fails somehow with OverflowError: Python int too large to convert to C ssize_t. Perhaps cython error. Cython calls PyLong_AsSsize_t for some reason after call to PyLong_FromUnsignedLongLong.

noskill:
Ok, documentation says that python uses signed size_t so overflow is expected

noskill:
there is no PyLongLong_FromSize_t by the way

PR #2313, opened 2019-08-01
wrap c++ comparision methods in cython

noskill:


opencog/cython/opencog/atom.pyx
```diff
@@ -122,31 +122,15 @@ cdef class Atom(Value):
...
-        if is_equal:
-            return 0
-        else:
-            return -1
+    def __lt__(self, other):
+        cdef cAtom* p = self.get_c_handle().get()
+        cdef cAtom* o = ((<Atom>other).get_c_handle()).get()
+        return deref(p) < deref(o)
+
+    def __eq__(self, other):
```
vsbogd: I would add check blow in order to correctly handle comparing of Value and Atom classes:
```py
if not isinstance(other, Atom):
    return False
```

opencog/cython/opencog/atomspace.pxd
```diff
@@ -124,6 +124,10 @@ cdef extern from "opencog/atoms/base/Atom.h" namespace "opencog":
         string get_name()
         vector[cHandle] getOutgoingSet()
 
+        bool operator=(cAtom&)
```
vsbogd: Am I right that `operator==` should be used instead?
```suggestion
        bool operator==(cAtom&)
```

opencog/cython/opencog/atom.pyx
```diff
@@ -122,31 +122,15 @@ cdef class Atom(Value):
...
-            return NotImplemented
-        cdef Atom a1 = a1_
-        cdef Atom a2 = a2_
-        is_equal = (a1.atomspace == a2.atomspace and
-                     deref(a1.handle) == deref(a2.handle))
-        if is_equal:
-            return 0
-        else:
-            return -1
+    def __lt__(self, other):
```
vsbogd: I think an assert should be added that other is instance of `Atom`:
```py
assert isinstance(other, Atom), "Only Atom instances are comparable with atoms"

Comment (noskill) [reply opencog/cython/opencog/atom.pyx]: test_linkvalue_containing_atoms.py passes value for comparison, but value is action atom..

File: opencog/cython/opencog/atom.pyx (hunk 1 above)
Comment (noskill) [reply opencog/cython/opencog/atom.pyx]: test_get_list_of_items_from_value test passes value for comparison, but value is actually atom..

File: opencog/cython/opencog/atomspace.pxd (hunk 2 above)
Comment (noskill) [reply opencog/cython/opencog/atomspace.pxd]: yes, thanks

File: opencog/cython/opencog/atom.pyx (hunk 3 above)
Comment (noskill) [reply opencog/cython/opencog/atom.pyx]: ok

File: opencog/cython/opencog/atom.pyx [hunk 4]
```diff
@@ -124,31 +118,33 @@ cdef class Atom(Value):
...
-            return 0
-        else:
-            return -1
+    def __richcmp__(self, other, int op):
+        if op == Py_LT:
+            return self.__lt(other)
+        if op == Py_EQ:
+            return self.__eq(other)
+        if op == Py_GT:
+            return not self.__lt(other)
```
vsbogd:
```suggestion
            return not self.__lt(other) and not self.__eq(other)
```

opencog/cython/opencog/nameserver.pyx
```diff
@@ -77,12 +77,12 @@ cdef create_python_value_from_c_value(cValuePtr& value, AtomSpace atomspace):
     cdef cValue *c_ptr = value.get()
     if clazz is not None:
         if c_ptr.is_atom():
-            return clazz(ptr_holder = ptr_holder, atomspace = atomspace)
+            return clazz(ptr_holder = ptr_holder)
```
vsbogd: Branches are equal, so `if: else:` can be replaced by:
```py
return clazz(ptr_holder = ptr_holder)
```

opencog/cython/opencog/atom.pyx
noskill: other < self

opencog/cython/opencog/nameserver.pyx
noskill: done

linas:
Looks reasonable, but unit tests are failing -- looks like the URE python wrapper wants to access the atomspace, and is failing, as it has now been removed ... there needs to be matching pull req for URE to get this working.

noskill:
Yes, there is matching pull request  https://github.com/opencog/ure/pull/32

linas:
The cython bindings won't compile for opencog -- when changing the API like this, you should verify that all the major components still work.  This is one reason why API changes are ... troublesome.

Please create a pull req for opencog as well.

PR #2314, opened 2019-08-09
Add get_keys to python Atom binding

AmeBel:


opencog/cython/opencog/nameserver.pyx
```diff
@@ -74,11 +74,15 @@ cdef create_python_value_from_c_value(cValuePtr& value):
 
     thismodule = sys.modules[__name__]
     clazz = getattr(thismodule, type_name, None)
-    cdef cValue *c_ptr = value.get()
     if clazz is not None:
         return clazz(ptr_holder = ptr_holder)
 
-    if c_ptr.is_atom():
+    # For handling the children types of TruthValue.
+    if is_a(get_type(type_name), get_type("TruthValue")):
```
vsbogd: Lines 76-78 should be enough to return TruthValue, and this `if` should be not necessary. Did you see opposite?

vsbogd: I mean that it should be safe to remove lines 80-82, this condition should be handled by lines 76-78

AmeBel: If you comment out 80-82 on this change you get the following
```python
Python 3.6.8 (default, Jan 14 2019, 11:02:34)
[GCC 8.0.1 20180414 (experimental) [trunk revision 259383]] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from opencog.atomspace import AtomSpace, TruthValue, types
>>> atomspace = AtomSpace()
>>> a = atomspace.add_node(types.ConceptNode, 'a', TruthValue(0.5, 0.5))
>>> a.get_value(a.get_keys()[0])
  File "<stdin>", line 1, in <module>
  File "opencog/atom.pyx", line 73, in opencog.atomspace.Atom.get_value (/host/opencog/atomspace/build/opencog/cython/opencog/atomspace.cpp:8382)                                           
  File "opencog/nameserver.pyx", line 88, in opencog.atomspace.create_python_value_from_c_value (/host/opencog/atomspace/build/opencog/cython/opencog/atomspace.cpp:2488)                   
TypeError: Python API for SimpleTruthValue is not implemented yet
>>> 
```

vsbogd: Ah, ok, so the reason is that we have value which is not wrapped by Python class. But its parent value is wrapped. Then this fix seems resonable as fast fix. To make a proper fix I believe one need to traverse type hierarchy using nameserver until implemented class is not found.

vsbogd:
```suggestion
    if is_a(get_type(type_name), types.TruthValue):
```

opencog/cython/opencog/nameserver.pyx
```diff
@@ -74,11 +74,15 @@ cdef create_python_value_from_c_value(cValuePtr& value):
...
+        return TruthValue(ptr_holder = ptr_holder)
+
+    # For handling the children types of Atom.
+    if is_a(get_type(type_name), get_type("Atom")):
```
vsbogd:
```suggestion
    if is_a(get_type(type_name), types.Atom):
```

opencog/cython/opencog/nameserver.pyx
vsbogd: This and below requires additional import:
```python
from opencog.atomspace import types
```

vsbogd:
Thanks!

Issue #1645, opened 2018-04-25
Parallelize the pattern matcher


linas:
See opencog/cogutil#108 for the discussion.

vsbogd:
See [email](https://www.mail-archive.com/opencog@googlegroups.com/msg02041.html) for the discussion.

I will continue discussion here as @linas suggested.

@linas, thank you for spotting NextLink issue https://github.com/opencog/atomspace/issues/1507. Copying PatternMatchCallback was my first attempt to implement parallel matching. But many code parts expect that PatternMatchCallback.grounding() method collects the result and return it synchronously. And there are dozen different grounding() implementations which do it slightly differently.

NextLink idea can resolve this issue. Next steps as I see them:
- add NextLink to collect PatternMatch results asynchronously
- enable parallel matching using copy of PME and PMCB only for queries via NextLink (use single thread version for old clients)
- migrate old clients client by client to asynchronous results collection

Does it makes sense?

linas:
OK. Working on NextLink is kind-of diving into the deep end of the pool; its not clear that you've learned how to swim, yet. For example, NextLink was not originally envisioned as a way to launch some heavy-weight, multi-threaded search; quite the opposite, I was thinking of "lazy evaluation", where it gives you just one answer at a time, only if and when you ask for it.

A better way of doing parallelization is called "scatter-gather", or  "map-reduce", and NextLink is a poor API for map-reduce, and I have not spent any time thinking of a good atomese API for map-reduce/scatter-gather type parallel searching.  Thus, the next step should really be "is there a good map-reduce-type API for pattern matching"?

ngeiswei:
@linas said
> It really might be better if you initially focused on fixing some of the existing bugs, rather than tackling fundamental design issues in a major subsystem component ...

That's a wise suggestion.

@vsbogd, I guess you could take some time to go through the github issues of the atomspace repo to see if you find something with the right level challenge for you (probably avoiding the ones labeled "quick task" as they are already too easy for you and best kept for newcomers). As well as obviously keeping improving the benchmark (like addressing https://github.com/opencog/benchmark/issues/8) which is extremely worthwhile.

linas:
See also #1750 as a better way of publishing changes coming out from various subsystems.

linas:
[...]

PR #2316, opened 2019-08-14
Parallel Get/Put links 

stellarspot:
This is the draft pull request which arises from discussion #1502 "Stop using SetLink for search results".
The used idea is that pattern matcher creates results in parallel and other links read them in parallel. 

The aim of this pull request is to show how this can work with introduced ParallelLink which puts its results into queue and PutLink reads them.

In the current design both ParallelLink and PutLink work on the same thread. Making pattern matcher parallel can be considered as the second step if the current design with queue looks appropriate.

Here is a simple sample how it works:
```scheme
(Inheritance (Concept "ball-1") (Concept "green"))
(Inheritance (Concept "ball-2") (Concept "red"))
(Inheritance (Concept "ball-3") (Concept "green"))
(Inheritance (Concept "ball-4") (Concept "red"))

(cog-execute!
 (ParallelGet
  (Variable "$BALL")
  (Inheritance
   (Variable "$BALL")
   (Concept "green"))
  (SetNode "S") ))

(cog-execute!
 (Put
  (Variable "$BALL")
  (Inheritance
   (Variable "$BALL")
   (Concept "selected-balls"))
  (SetNode "S")))
```

ParallelGetLink uses SetNode as input. Set node contains queue which used to exchange results between atoms.

ParallelGetLink puts its results to queue and generates MemberLinks:
```scheme
(MemberLink
   (ConceptNode "ball-1")
   (SetNode "S"))
(MemberLink
   (ConceptNode "ball-3")
   (SetNode "S"))
```

PutLink uses SetNode as argument and reads results form the queue and produces:
```scheme
   (InheritanceLink
      (ConceptNode "ball-1")
      (ConceptNode "selected-balls") )
   (InheritanceLink
      (ConceptNode "ball-3")
      (ConceptNode "selected-balls"))
```


linas:
Wow! OK!  I think I like this!

* Rather than creating a `ParallelGet`, it might be better to just re-use the existing `GetLink`, and allow it to have an optional 3rd argument. 
* The cog-utils repo has implementations of a thread-safe queue and and a thread-safe stack; it would be better to use that, instead of inventing something new. 
* It might be better to have `QueueValue` be a special case of `StreamValue`? That is because `StreamValue` already has some basic infrastructure for time-varying values; and `QueueValue` extends that.

Some more comments to follow for specific lines of code.

opencog/atoms/value/QueueValue.h
```diff
@@ -0,0 +1,144 @@
...
+	static const Handle QUEUE_VALUE_KEY;
+
+	QueueValue(const std::string& name)
+		: Value(QUEUE_VALUE),
+		_name(name),
+		_handle_queue(new HandleClosableQueue()) { }
+
+	virtual ~QueueValue() {}
+
+	HandleClosableQueuePtr get_queue() const { return _handle_queue; }
```
linas: Rather than doing this, maybe it would be better to just add push and pop directly to the API?

opencog/atoms/value/QueueValue.h
```diff
@@ -0,0 +1,144 @@
...
+typedef ClosableQueue<Handle> HandleClosableQueue;
+typedef std::shared_ptr<HandleClosableQueue> HandleClosableQueuePtr;
+
+/**
+ * ClosableQueueValue contains a queue that allows to pass values from
+ * a producer to consumer.
+ */
+
+class QueueValue
+	: public Value
```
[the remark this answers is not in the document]
linas: Oh, hang on, maybe it would be better to inherit from LinkValue? 

Or maybe StreamValue and LinkValue should be merged into one ? Or something like that? Yes, that would be a different, distinct discussion.  So, for now, maybe this is OK as it is.

linas:
One last remark: parallelizing the pattern matcher seems like it should be "easy" but might be quite difficult. There is a very obvious place to parallelize: the top-level search-starting-point loop.  This is in `InitiateSearch` and it finds all of the places from which the search can be started. Replacing the `foreach` with a `parallel_foreach` seems like it would be very easy, and then it should "just plain work". Well, I tried that once, and it didn't work.  I don't know why. It seemed like there were some snakes down in there, waiting to bite.

opencog/atoms/core/PutLink.cc
```diff
@@ -431,24 +443,38 @@ Handle PutLink::do_reduce(void) const
...
-		HandleSeq bset;
-		for (const Handle& h : args->getOutgoingSet())
+		if (SET_NODE == vtype)
 		{
-			HandleSeq oset;
-			oset.emplace_back(h);
-			try
+			// If the argument is SetNode, then process atoms from queue
+			// stored in SetNode value
+			AtomSpace* as = getAtomSpace();
```
vsbogd: Is atomspace required here? It is not used in code below

opencog/atoms/atom_types/atom_types.script
```diff
@@ -388,6 +389,7 @@ SATISFYING_LINK <- PATTERN_LINK
 GET_LINK <- SATISFYING_LINK        // Finds all groundings, returns them
 QUERY_LINK <- SATISFYING_LINK      // Finds all groundings, substitutes.
 BIND_LINK <- QUERY_LINK            // Finds all groundings, substitutes.
+PARALLEL_GET_LINK <- SATISFYING_LINK  // Finds all groundings, puts them to queue
```
vsbogd: Should it be inherited from PATTERN_LINK instead?

opencog/atoms/pattern/ParallelGetLink.cc
```diff
@@ -0,0 +1,94 @@
...
+	// could defer this indefinitely, until its really needed.
+
+	HandleSet handle_set = do_execute(as, silent);
+	for (auto h: handle_set)
+	{
+
+		HandleSeq handle_seq;
+		handle_seq.push_back(h);
+		handle_seq.push_back(_target);
+		Handle member_link = createLink(handle_seq, MEMBER_LINK);
```
vsbogd: Why don't add results to the atomspace in the `ParallelSatisfier` instead? Am I right that current implementation collects results as `SetLink` and then puts them into atomspace as `MemberLink`? It would be simpler to put them as `MemberLink` without using a `Set` as intermediate state.

linas:
Closing. interesting prototype.

If you just want plain-old parallel `GetLink`, I think you can achieve this by looking at https://github.com/opencog/atomspace/blob/812580d19773cdf4de533998eca5b5a5a9f9cc7a/opencog/query/InitiateSearchCB.cc#L402-L417 and converting that loop into a parallel-for-each loop.   That parallelization is distinct from the other idea(s) in this prototype, which should be pursued as a distinct, unrelated project.  (The other idea in this prototype being that search results can be queued into a `QueueValue`, making it easy to pipeline producers and consumers (the pattern matcher being the producer, and some PutLink-type thing being a consumer.)

linas:
As usual, issue #1645 tracks the idea of parallelizing the pattern matcher, and issue #1502 tracks an earlier, related idea for `QueueValue`

PR #2317, opened 2019-08-16
move is_in_atomspace from cython to c++

noskill:
1) move is_in_atomspace from cython to c++
2) expose add_atom from cython

### DIFF
--- a/opencog/atomspace/AtomSpace.cc
+++ b/opencog/atomspace/AtomSpace.cc
@@ -495,6 +495,22 @@ Handle AtomSpace::set_truthvalue(const Handle& h, const TruthValuePtr& tvp)
     return Handle::UNDEFINED;
 }
 
+bool AtomSpace::is_in_atomspace(Handle const & h){
+     Type t = h->get_type();
+     if (h->is_node()){
+         if (this->get_handle(t, h->get_name()))
+             return true;
+         return false;
+     }
+     std::vector<Handle> handle_vector = h->getOutgoingSet();
+     if (h->is_link()){
+         if (this->get_handle(t, handle_vector))
+             return true;
+         return false;
+     }
+     throw std::runtime_error("Argument is not link and not node");
+}
+
 std::string AtomSpace::to_string() const
 {
 	std::stringstream ss;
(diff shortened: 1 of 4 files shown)


opencog/atomspace/AtomSpace.cc
```diff
@@ -495,6 +495,22 @@ Handle AtomSpace::set_truthvalue(const Handle& h, const TruthValuePtr& tvp)
     return Handle::UNDEFINED;
 }
 
+bool AtomSpace::is_in_atomspace(Handle const & h){
```
linas: This has a trivial one-line implementation: `return nullptr != get_atom(h);` -- therefore, this new code is not needed.  You should just copy this one-liner to the python code.

opencog/cython/opencog/atomspace_details.pyx
```diff
@@ -89,6 +89,12 @@ cdef class AtomSpace:
             atom = self.add_link(t, out, tv)
         return atom
 
+    def add_atom(self, Atom atom):
+        cdef cHandle result = self.atomspace.add_atom(atom.get_c_handle())
+        if result == result.UNDEFINED:
```
linas: Rather than returning nullptr, the `AtomTable` code should almost surely be changed to throw an exception. That would allow you to skip this check.

linas:
For the python API, it is probably better to just remove the `is_in_atomspace()` function, and instead, provide a `get_atom()` function -- this will return the atom if it is in the atomspace, else it will return null, and the user can test for null, if needed.

opencog/cython/opencog/atomspace_details.pyx
noskill: What if atom is DeleteLink?

linas: Oh. Interesting. I'll have to think about that. It still might be the right thing to throw ... 

Not all exceptions in the atomspace are "errors". Many are simply "fast paths" that are used to indicate "unusual conditions" (e.g. the `SilentException`) thus avoiding a useless if-statement for 99.9% of the cases, and paying the exception overhead for the 0.1% of the cases that need it.  It seems that this might  be the case here, too, with most cases being "actual errors", while the `DeleteLink` maybe should be a `SilentException`.

This is not urgent, its just .. a general goal to move towards.

PR #2320, opened 2019-09-01
Add missing cmake include

xloem:
This was needed to build on my system, Termux on Android, cmake 3.15.2 .


linas:
Thanks. Somewhat unexpected, but presumably harmless.

PR #2335, opened 2019-10-31
Fix variable declaration inference

### LINKED ISSUE #2329: Canonical VariableList inference isn't consistent
Consider the following

```scheme
(define X (Variable "X"))
(define Y (Variable "Y"))

(define lxy (Lambda (Inheritance (And X Y) Y)))
(define lyx (Lambda (Inheritance (And Y X) X)))
```

`lxy` and `lyx` are, to the uninformed user, alpha-equivalent. However the atomspace does not detect them as such, as it can be seen

```scheme
scheme@(guile-user)> (cog-handle lxy)
$3 = 12252611381647628989
scheme@(guile-user)> (cog-handle lyx)
$4 = 13289207854925892108
```

The reason is because the variable list is constructed according to the prefix order of variable appearances in the body of the scope. However when links are unordered, that ordering ultimately depends on the names of the variables. In my view, it's a bug, not a feature.

I think it can be fixed by modifying `VarScraper::find_vars` in `Variables.cc` to account for the unordered links, instead of blindly following the prefix order. I haven't put my finger on it yet, but I'm working on it. Any relevant comment is welcome.

ngeiswei:
Fix #2329.

The idea is to sort unordered links as much as possible by their semantics (thus unaffected by variable names) while looking for free variables.


ngeiswei:
@linas, I'm confident that fix is correct, however you may want to have a look at it.

As I suspected I didn't need to update `ScopeLink::compute_hash()` because it takes `ScopeLink::_varlist` into consideration, which is the only thing that this fix changes.

ngeiswei:
More specifically, that piece of code https://github.com/opencog/atomspace/blob/51bb939de2a964b6b834fc2bebb285fc8090aa37/opencog/atoms/core/ScopeLink.cc#L294-L298 is why we don't need to worry about modifying `ScopeLink::compute_hash`.

ngeiswei:
Hmm, the URE fails, I need to look into that before merging.

ngeiswei:
OK, everything passes, I give you (@linas, since you're the author of that code) the green light to merge.

linas:
Thanks!

PR #2337, opened 2019-11-01
Copy an Move assignment ctors

linas:
Provide explicit copy and move assignment constructors.  

I'm assuming this might provide some mild performance benefits, but have not checked.

### DIFF
--- a/opencog/atoms/base/Atom.h
+++ b/opencog/atoms/base/Atom.h
@@ -149,6 +149,11 @@ protected:
         _atom_space(nullptr)
     {}
 
+    Atom& operator=(const Atom& other) // copy assignment operator
+        { return *this; }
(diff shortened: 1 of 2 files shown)


ngeiswei:
Interesting. Did you do that for optimizing, or because the compiler asked you to?

linas:
I did it because of stupidity and excitement .. I read https://en.wikipedia.org/wiki/Rule_of_three_(C%2B%2B_programming)#Rule_of_Five and got excited, and I wrote some code, and then realized I was wasting my time, but created the pull req anyway. 

It might be the case that using the move constructor `Handle(&&)` might avoid the overhead of copying/incrementing/decrementing the shared pointer. Maybe. But I can't guess how often this happens; it needs measurements to be precise and I don't feel like measuring today.

Issue #2338, opened 2019-11-06
Quoted PutLink crashes


ngeiswei:
The following
```scheme
(Quote (Put
    (Unquote (Variable "$vardecl"))
    (Unquote (Variable "$body"))
    (Unquote (Variable "$args"))))
```
produces
```
Segmentation fault (core dumped)
```

ngeiswei:
BTW, the following
```scheme
(Quote (Put
    (Unquote (Variable "$body"))
    (Unquote (Variable "$args"))))
```
does NOT produce a segfault.

ngeiswei:
I've added a (disabled) unit test to reproduce the problem

https://github.com/opencog/atomspace/blob/d8e3a92833e168c13633a54dbfdc22158185397f/tests/atoms/core/PutLinkUTest.cxxtest#L2244-L2259

linas:
Hmm sometimes github auto-closes issues, and sometimes it doesn't.  If your pull req had said "this fixes #2338" then this would have auto-closed when it got merged.

PR #2341, opened 2019-11-06
Assign _vardecl, _body in case ScopeLink::extract_variables didn't

ngeiswei:
which can happen when the variable declaration is unquoted.


ngeiswei:
Potential fix for issue #2338 please @linas have a look and merge if appropriate.

linas:
Yeah, looks god to me.

Issue #2346, opened 2019-11-08
Pattern matcher does not return all permutations when it should


ngeiswei:
Provided the following KB

```scheme
(List (Concept "A") (Set (Predicate "P") (Predicate "Q")))
```

and the following query

```scheme
(define query-ok
  (Bind
    (Present (Set (Variable "$X") (Variable "$Y")))
    (Implication (Variable "$X") (Variable "$Y"))))
```

the pattern matcher, with `(cog-execute! query-ok)`, returns

```scheme
(SetLink
   (ImplicationLink
      (PredicateNode "P")
      (PredicateNode "Q")
   )
   (ImplicationLink
      (PredicateNode "Q")
      (PredicateNode "P")
   )
)
```

as it should.

However, for the following query

```scheme
(define query-not-ok
  (Bind
    (Present (List (Variable "$Z") (Set (Variable "$X") (Variable "$Y"))))
    (Implication (Variable "$X") (Variable "$Y"))))
```

it returns, with `(cog-execute! query-not-ok)`

```scheme
(SetLink
   (ImplicationLink
      (PredicateNode "P")
      (PredicateNode "Q")
   )
)
```

as it should not.

ngeiswei:
Some more weirdness. The following query still does not return all permutations

```scheme
  (Bind
    (Present (List (Variable "$Z") (Set (Variable "$X") (Variable "$Y"))))
    (Ordered (Variable "$Z") (Implication (Variable "$X") (Variable "$Y")))))
```

however, that one (`Ordered` has been replaced by `List`), does!!!

```scheme
  (Bind
    (Present (List (Variable "$Z") (Set (Variable "$X") (Variable "$Y"))))
    (List (Variable "$Z") (Implication (Variable "$X") (Variable "$Y")))))
```

ngeiswei:
I've added the following (disabled and failing) utest to reproduce the bug

https://github.com/opencog/atomspace/blob/ad843e6e578dfbfc8955c3e2ce669dcbec564622/tests/query/PermutationsUTest.cxxtest#L82-L105

linas:
Yikes!  Who should look into this?

ngeiswei:
I'm looking into it right now, but if you've got some light-bulb idea I welcome it.

linas:
[...]

PR #2348, opened 2019-11-08

linas:
This fixes issue #2346 and passes all unit tests, for me.   There are, however, more complex cases which are not handled.


linas:
Nil, everything passes, except for `SurprisingnessUTest` which is timing out. This is surprising, since I thought the miner did not use the pattern matcher.  Any clue what this is about, or is this OK to merge?

ngeiswei:
The pattern miner heavily relies on the pattern matcher. Better not merge. Will take a deep look at it on Monday.

linas:
Closing, obsoleted by #2353 which does assorted cleanup and fixes distinct from this actual fix.

PR #2353, opened 2019-11-08

linas:
This includes multiple SetLinks encourntered during search.


linas:
Don't merge; I am exploring yet other fixes.

linas:
Closing. Obsoleted/replaced by #2356 which is exactly the same fix,
except that its applied to refactored code.

PR #2356, opened 2019-11-09

linas:
This includes multiple SetLinks encountered during search.


linas:
argh, no this is broken.

linas:
Pull req #2357 is the non-broken version of this.

PR #2357, opened 2019-11-09

linas:
This includes multiple SetLinks encourntered during search.

### DIFF
--- a/opencog/query/PatternMatchEngine.cc
+++ b/opencog/query/PatternMatchEngine.cc
@@ -1163,8 +1163,23 @@ bool PatternMatchEngine::explore_term_branches(const Handle& term,
 
 	for (const PatternTermPtr &ptm : pl->second)
 	{
+		DO_LOG({LAZY_LOG_FINE << "Begin exploring term: " << ptm->to_string();})
 		if (explore_glob_branches(ptm, hg, clause_root))
 			return true;
+
+		// If no solution was found, and there are unordered links, then
+		// there may be alternate permuations of the unordered link that
+		// might satisfy this clause. So try those, until exhausted.
+		// Note that these unordered links might be buried deeply;
+		// that is why we iterate over them here.
+		while (_have_more)
+		{
+			_have_more = false;
+			_take_step = true;
+
+			if (explore_glob_branches(ptm, hg, clause_root))
+				return true;
+		}
 	}
 	return false;
 }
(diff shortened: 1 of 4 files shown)


ngeiswei:
Some pattern matcher query generated by the pattern miner "freezes" the pattern matcher. I'm trying to a create minimal test out of it.

ngeiswei:
The following query reproduces the freeze. It's already too large for my taste but I can't seem to shrink any further.

```scheme
;; KBs
(ListLink
  (LambdaLink
    (VariableList
      (VariableNode "$Y")
      (VariableNode "$Z")
      (VariableNode "$W")
    )
    (PresentLink
      (InheritanceLink
        (VariableNode "$W")
        (VariableNode "$Y")
      )
      (InheritanceLink
        (VariableNode "$Z")
        (VariableNode "$Y")
      )
    )
  )
  (NumberNode "5.000000")
)

(AndLink
  (InheritanceLink
    (VariableNode "$W")
    (VariableNode "$Y")
  )
  (InheritanceLink
    (VariableNode "$Z")
    (VariableNode "$Y")
  )
)

;; Query
(define query
(GetLink
  (PresentLink
    (ListLink
      (QuoteLink
        (LambdaLink
          (UnquoteLink
            (VariableNode "$f-vardecl")
          )
          (PresentLink
            (UnquoteLink
              (VariableNode "$cnj-bodies-1")
            )
            (UnquoteLink
              (VariableNode "$cnj-bodies-0")
            )
          )
        )
      )
      (VariableNode "$ms-0")
    )
    (ListLink
      (QuoteLink
        (LambdaLink
          (UnquoteLink
            (VariableNode "$f-vardecl")
          )
          (PresentLink
            (UnquoteLink
              (VariableNode "$cnj-bodies-1")
            )
            (UnquoteLink
              (VariableNode "$cnj-bodies-0")
            )
          )
        )
      )
      (VariableNode "$ms-1")
    )
  )
)
)
```

ngeiswei:
I think there's something wrong with the way `PatternMatchEngine::_have_more` is set which triggers the following

https://github.com/linas/atomspace/blob/0e2dca7830ca2c06260e9547f407ff4a35163d18/opencog/query/PatternMatchEngine.cc#L1175-L1182

to loop forever. Not sure why though.

linas:
OK, looking at it.

linas:
I dislike the fact that if I change the LambdaLink to an OrderedLink and then git rid of all the quotes/unquotes, then it works.  Something about unquoting makes it break. It might be taking too many steps...

linas:
woo hoo!

linas:
I added the above crazy test as a unit test.  Its passing (no inf loop) but it is returning the wrong number of solutions: its returning 1, there should be two.  So #2359 fixes this, it is currently finding two.  But it still has other issues...

PR #2359, opened 2019-11-10
Odometer prototype

linas:
This implements an "odometer" approach to cycling through various permutations of 
multiple Unordered links. Basically, a full permutation of one is explored, before the 
next one ticks by one. Tricky parts include figuring out how many terms there are in
an odometer, and also handling nested unordered links.

This seems to work right now, but is only weakly tested.


linas:
So after fiddling with this the whole day, I've come to the conclusion that there is no simple run-time deduction of the odometer structure.  Before starting matching, there will need to be a distinct analysis step to find all the odometers, and mark them up, so that the matcher knows when to roll over and when to stop. That requires a bunch of new code and a day or two to write it.    Here are some concrete examples:

This is an odometer:
```
(Pattern (Set A ...) (Set B...))
```
So all permutations of A have to be explored, before taking one step of B.  The code in this pull req does this correctly. (see `test_cube()`)

This is NOT an odometer:
```
(Pattern (Set B (Set A...) ...))
```
The old, existing code handles this case correctly (in the unit tests, its called `exhaust-3` and `exhaust-4`).  The code in this pull req gets confused by this one and mistakes it for an odometer, and rolls over A too many times, and misses some A-B combinations (so `exhaust-3` and 4 fail).

To avoid confusion between these two, it seems like a pre-analysis is needed to tell apart these cases.

linas:
All current unit tests passing. A later pull req will add more demanding tests.

Issue #2282, opened 2019-07-22
python wrapper crashes with segmentation fault when set_value is called with wrong parameters


astroseger:
The following (wrong) code crashes with segmentation fault instead of failing gracefully.

```python
from opencog.type_constructors import *
from opencog.utilities import initialize_opencog
from opencog.bindlink import execute_atom
from opencog.atomspace import TruthValue

# Initialize AtomSpace
atomspace = AtomSpace()
initialize_opencog(atomspace)

string_node = ConceptNode("String")
string_node.set_value("bad key", StringValue("Hello, World!"))

```

linas:
Hi @astroseger we do not have a regular python maintainer; care to try to fix this yourself?

vsbogd:
@noskill, @astroseger , can we close it?

astroseger:
yes we can!

> @noskill, @astroseger , can we close it?

PR #2363, opened 2019-11-13
verify key type in atom.set_value

noskill:
Check that key in atom.set_value is an instance of value

fix for issue https://github.com/opencog/atomspace/issues/2282

### DIFF
--- a/opencog/cython/opencog/atom.pyx
+++ b/opencog/cython/opencog/atom.pyx
@@ -62,6 +62,8 @@ cdef class Atom(Value):
             atom_ptr.setTruthValue(deref((<TruthValue>truth_value)._tvptr()))
 
     def set_value(self, key, value):
+        if not isinstance(key, Value):
+            raise RuntimeError("key should be an instance of Value, got {0} instead".format(type(key)))
         self.get_c_handle().get().setValue(deref((<Atom>key).handle),
                                 (<Value>value).get_c_value_ptr())
 
--- a/opencog/cython/opencog/atom.pyx
+++ b/opencog/cython/opencog/atom.pyx
@@ -63,7 +63,7 @@ cdef class Atom(Value):
 
     def set_value(self, key, value):
         if not isinstance(key, Value):
-            raise RuntimeError("key should be an instance of Value, got {0} instead".format(type(key)))
+            raise TypeError("key should be an instance of Value, got {0} instead".format(type(key)))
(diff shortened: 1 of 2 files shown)


opencog/cython/opencog/atom.pyx
```diff
@@ -62,6 +62,8 @@ cdef class Atom(Value):
             atom_ptr.setTruthValue(deref((<TruthValue>truth_value)._tvptr()))
 
     def set_value(self, key, value):
+        if not isinstance(key, Value):
+            raise RuntimeError("key should be an instance of Value, got {0} instead".format(type(key)))
```
vsbogd: I would suggest using `TypeError` as it suits this kind of issue better (see https://docs.python.org/3/library/exceptions.html#TypeError)
```suggestion
            raise TypeError("key should be an instance of Value, got {0} instead".format(type(key)))
```

tests/cython/atomspace/test_atom.py
```diff
@@ -62,6 +62,13 @@ def test_get_input(self):
 
         self.assertEqual(set(incoming), set([a, b]))
 
+    def test_invalid_key(self):
+        string_node = ConceptNode("String")
+        error_str = "key should be an instance of Value, got {0} instead".format(str)
+        with self.assertRaisesRegex(RuntimeError, error_str):
```
vsbogd:
```suggestion
        with self.assertRaisesRegex(TypeError, error_str):
```

opencog/cython/opencog/atom.pyx
```diff
@@ -62,6 +62,8 @@ cdef class Atom(Value):
             atom_ptr.setTruthValue(deref((<TruthValue>truth_value)._tvptr()))
 
     def set_value(self, key, value):
+        if not isinstance(key, Value):
+            raise TypeError("key should be an instance of Value, got {0} instead".format(type(key)))
```
linas: In the C++ API, the keys are required to be Atoms.

tests/cython/atomspace/test_atom.py
```diff
@@ -62,6 +62,13 @@ def test_get_input(self):
 
         self.assertEqual(set(incoming), set([a, b]))
 
+    def test_invalid_key(self):
+        string_node = ConceptNode("String")
```
linas: object created but never used...

noskill: string_node used for setting a value

PR #2364, opened 2019-11-14
check for atom instead of value in atom.set_value

noskill:
check atom instead of value in atom.set_value


noskill:
based on @linas comments from https://github.com/opencog/atomspace/pull/2363

linas:
thx!

Issue #2283, opened 2019-07-22
PlusLink in python api gives incorrect result and do not fails (as it should) when string is passed instead of VariableNode 


astroseger:
The following code should fail instead of giving incorrect result

```python
from opencog.type_constructors import *
from opencog.utilities import initialize_opencog
from opencog.bindlink import execute_atom
from opencog.atomspace import TruthValue

# Initialize AtomSpace
atomspace = AtomSpace()
initialize_opencog(atomspace)


(DefineLink(
    DefinedSchemaNode('add'),
    LambdaLink(
        VariableList(
            VariableNode('$X'),
            VariableNode('$Y')),
        PlusLink(
            '$X',                     #!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! 
            VariableNode('$Y')))))

res = execute_atom(atomspace,
                   ExecutionOutputLink(
                       DefinedSchemaNode('add'),
                       ListLink(
                           NumberNode('3'),
                           NumberNode('4'))))
print(res)

```

linas:
Said to be fixed in #2365

Issue #2302, opened 2019-07-30
VariableList python binding accepts list, returns empty variable list


noskill:
VariableList([TypedVariableLink(VariableNode("X1"), TypeNode("VariableNode"))])

gives

(VariableList
) ; [9225060886715003552][6]

linas:
fixed in #2365

PR #2365, opened 2019-11-14
check arg type in python's atom factory method

noskill:
raise error on type of argument to atom constructor in python
fix for issue https://github.com/opencog/atomspace/issues/2302


noskill:
fixes https://github.com/opencog/atomspace/issues/2283 too

Issue #2325, opened 2019-10-15
Introduce VariableSet for commutative variable declarations in scopes


ngeiswei:
## Problem

In some scopes, the order of the variables in the variable declaration does not matter. I.e one can say that such scope are Variable Declation Permutation Invariant, or VDPI for short.

That is typically the case for bind links. For example

```
Bind
  VariableList
    Variable "$X"
    Variable "$Y"
  Present
    Inheritance
      Variable "$X"
      Variable "$Y"
  List
    Variable "$X"
    Variable "$Y"
```

and

```
Bind
  VariableList
    Variable "$Y"
    Variable "$X"
  Present
    Inheritance
      Variable "$X"
      Variable "$Y"
  List
    Variable "$X"
    Variable "$Y"
```

are behaviorally equivalent, yet are not alpha-equivalent.

On the contrary, for instance

```
Get
  VariableList
    Variable "$X"
    Variable "$Y"
  Present
    Inheritance
      Variable "$X"
      Variable "$Y"
```

and

```
Get
  VariableList
    Variable "$Y"
    Variable "$X"
  Present
    Inheritance
      Variable "$X"
      Variable "$Y"
```

are not behaviorally equivalent.

It would be nice if there was an easy way to take this into account as to avoid over-representing these scopes. In other words, pushing alpha-equivalence up to VDPI.

## Proposal

I'm not completely sure but we could introduce a `VariableSet` construct, analog to `VariableList`, expect that it would re-order the variables following predetermined order like `UnorderedLink`.

The reason I'm suggesting that instead of just taking care of scopes that we know are VDPI like BindLink, is because sometimes whether this is determined by usage rather than scope type, so letting the user the freedom to choose could be a plus.

Their might be better ways, any thought on the matter is welcome.

linas:
`VariableSet` seems perfectly fine.

For `BindLink`, I've occasionally thought about moving the variable declarations out of the header, and into the body itself.  I don't recall the syntax that would be needed for this, and I can't find it in my notes. This hasn't been done because it would add yet more complexity to an already-complex decoding of the body of the `BindLink`. (its already quite the mess, with fixed clauses, virtual ones, evaluatable ones, optional/must-be-absent clauses, logical connectives tying them together... Throwing variable decls into that risks further confusing the issue, at least, at this point in time, in the current code base).  I'm not sure why I even bring this up... other than to say its possible, but ill-advised.

ngeiswei:
This seems to be working so I'm closing.

PR #2366, opened 2019-11-14
Implement VariableSet

ngeiswei:
Addresses issue https://github.com/opencog/atomspace/issues/2325

I'm happy with the PR. However, there a slight change in the definition of alpha-equivalence of scopes to note. Precisely, assuming that `<body>` contains a single free variable `$X`, then the following equivalences hold

```
Scope
  <body>
```

is equivalent to

```
Scope
  Variable "$X"
  <body>
```

is equivalent to

```
Scope
  VariableSet
    Variable "$X"
  <body>
```

That is because a scope without variable declaration produces by default a `VariableSet` over the free variables of that scope. So for instance

```
Scope
  Inheritance
    Variable "$X"
    Variable "$Y"
```

is equivalent to

```
Scope
  VariableSet
    Variable "$X"
    Variable "$Y"
  Inheritance
    Variable "$X"
    Variable "$Y"
```

Such convention has been chosen on the ground that if no variable declaration is provided, then the user shouldn't care about the order since that one is arbitrary (though fixed according to the semantics of the variables in the body), so a `VariableSet` should be used, not a `VariableList`.

`VariableList` can of course be used when the variable declaration is provided.

You also should note that, as of that PR, the follow equivalence does NOT hold.

```
Scope
  VariableList
    Variable "$X"
  <body>
```

is NOT equivalent to

```
Scope
  VariableSet
    Variable "$X"
  <body>
```

while it could, since there is only one variable, order is irrelevant to the semantics of the scope.

This could be corrected in the future if we so desire.

Wikipage still need to be updated (which I'll likely do tomorrow). Meanwhile, your feedback is welcome.

### DIFF
--- a/opencog/atoms/core/TypeUtils.h
+++ b/opencog/atoms/core/TypeUtils.h
@@ -35,6 +35,8 @@ namespace opencog
  *  @{
  */
 
+// TODO: most of what is here should be moved to Variables
+
 /**
  * Type checker.  Returns true if `value` is of type `type_spec`.
  * More precisely, returns true if `value` will fit into the type
--- a/opencog/atoms/core/TypeUtils.h
+++ b/opencog/atoms/core/TypeUtils.h
@@ -35,7 +35,7 @@ namespace opencog
  *  @{
  */
 
-// TODO: most of what is here should be moved to Variables
+// TODO: some of what is here could be moved to Variables
 
 /**
  * Type checker.  Returns true if `value` is of type `type_spec`.
--- a/opencog/atoms/core/TypeUtils.h
+++ b/opencog/atoms/core/TypeUtils.h
@@ -27,6 +27,7 @@
 #include <opencog/atoms/base/Handle.h>
 #include <opencog/atoms/atom_types/types.h>
 #include <opencog/atoms/core/VariableList.h>
+#include <opencog/atoms/core/VariableSet.h>
 
 namespace opencog
 {
@@ -183,9 +184,9 @@ TypeSet type_intersection(const TypeSet& lhs,
                           const TypeSet& rhs);
 
 /**
- * Generate a VariableList of the free variables of a given atom h.
+ * Generate a VariableSet of the free variables of a given atom h.
  */
-VariableListPtr gen_varlist(const Handle& h);
+VariableSetPtr gen_variable_set(const Handle& h);
 
 /**
  * Generate a variable declaration of the free variables of a given atom h.
@@ -194,13 +195,8 @@ Handle gen_vardecl(const Handle& h);
 
 /**
  * Given an atom h and its variable declaration vardecl, turn the
- * vardecl into a VariableList if not already, and if undefined,
- * generate a VariableList of the free variables of h.
- */
-VariableListPtr gen_varlist(const Handle& h, const Handle& vardecl);
-
-/**
- * Like above but return Variables instead.
+ * vardecl into a Variables object, and if undefined, generate a
+ * Variables object from the free variables of h.
  */
 Variables gen_variables(const Handle& h, const Handle& vardecl);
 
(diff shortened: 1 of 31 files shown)


ngeiswei:
There are some issues to fix before this is mergeable.

linas:
I'll try to look at this tomorrow.

ngeiswei:
I did all the corrections I wanted.

The URE and Miner unit tests should pass after merging the following PRs

https://github.com/opencog/ure/pull/47
https://github.com/opencog/miner/pull/19

which I'll merge as soon as this PR is approved.

linas:
whoops, I got distracted. The unit tests are failing, though ...

linas:
oh, I see, failing in ure, waiting on this to merge ... OK

ngeiswei:
The thing is if I merge the ure and miner sister PR, then they will break until that one is merged, which is why I would rather have that PR approved before I merge them, or at the same time.

ngeiswei:
I've added a wiki page for it https://wiki.opencog.org/w/VariableSet

Please, @linas (and possibly @vsbogd if you want to) have a look at it and let me know if you find anything wrong. Then I'll merge all 3 PRs.

linas:
I'm reviewing this now. I haven't finished; but here's one question that might be an easy question .. or not.
Currently, the following BindLink is invalid, but I would like to make it valid:
```
(BindLink
   (AndLink ; beginning of body: an unordered set of predicates
      (Present (Inheritance (Variable "$X") ( Variable "$Y"))
      (TypedVariable (Variable "$X") (Type 'ConceptNode))  ; a very special kind of predicate
      (TypedVariable (Variable "$X") (Type 'ConceptNode))
   ) ; end of body
  (List (Variable "$X") (Variable "$Y"))  ;consequent
)
```
The above is meant to be equivalent to 
```
(BindLink
   (VariableSet
        (TypedVariable (Variable "$X") (Type 'ConceptNode)) 
         (TypedVariable (Variable "$X") (Type 'ConceptNode))
   )
   (AndLink ; beginning of body: an unordered set of predicates
      (Present (Inheritance (Variable "$X") ( Variable "$Y"))
   ) ; end of body
  (List (Variable "$X") (Variable "$Y"))  ;consequent
)
```
I've wanted to support the above for a while, but writing teh C++ code for that seemed .. yucky.  Does your pull req make this easy?  Could we easily convert the one into the other?

opencog/atoms/core/TypeUtils.h
```diff
@@ -27,13 +27,16 @@
...
 #include <opencog/atoms/core/VariableList.h>
+#include <opencog/atoms/core/VariableSet.h>
 
 namespace opencog
 {
 /** \addtogroup grp_atomspace
  *  @{
  */
 
+// TODO: most of what is here should be moved to Variables
```
linas: ... or not. The type utils can also be used without variables, for example to implement (finish implementing) https://wiki.opencog.org/w/TypedAtomLink

tests/atoms/core/PutLinkUTest.cxxtest
```diff
@@ -2,7 +2,7 @@
  * tests/atoms/PutLinkUTest.cxxtest
  *
  * Copyright (C) 2015,2017 Linas Vepstas
- * Copyright (C) 2016 Nil Geiswieller
+ * Copyright (C) 2016 Nil Geisweiller
```
linas: eh

linas:
You state that

> However, there a slight change in the definition of alpha-equivalence of scopes to note. 

I somehow failed to notice the spot where this alpha-equivalance is computed/performed. Do you actually have a unit test for this?  i don't see that it's urgent/important if this isn't done; I was just curious.

ngeiswei:
Thanks @linas! Every passes (besides unrelated failure of UnorderedUTest), so I'm merging.

ngeiswei:
Regarding the code dealing with the various equivalences, it's here

https://github.com/opencog/atomspace/blob/f7d0313436d1d066ac5bfbf531af485d9efd9bb1/opencog/atoms/core/Variables.cc#L852

https://github.com/opencog/atomspace/blob/f7d0313436d1d066ac5bfbf531af485d9efd9bb1/opencog/atoms/core/ScopeLink.cc#L265

linas:
OK, thanks.  BTW, I'm vaguely contemplating adding utilities for 160-bit and 256-bit hashes. Maybe. It's probably not wise to replace the existing 64-bit hash with a 160-bit one. Maybe. Just FYI. May never happen at current rate of progress.

PR #2372, opened 2019-11-19
Revert commit 981e35a09 as it seems to be needed indeed

### LINKED ISSUE #2371: UnorderedUTest::test_embed randomly fails
Since 981e35a09 `UnorderedUTest::test_embed()` randomly fails.

ngeiswei:
Fix #2371.

Warning: fix entirely based on finding the breaking commit, no other level of understanding was used.


ngeiswei:
@linas, please review and merge if you think it is adequate.

linas:
Hmm. It should be a giant no-op.

PR #2373, opened 2019-11-19
create empty variable list when replacing variables

noskill:
in substitute-var utility function return empty VariableList, since VariableList with replaced variables is not valid

### DIFF
--- a/opencog/scm/opencog/base/apply.scm
+++ b/opencog/scm/opencog/base/apply.scm
@@ -22,6 +22,7 @@
 (define-public (substitute-var bindings body)
   (let ((body_type (cog-type body)))
     (cond ; recursive case
+          ((cog-subtype? 'VariableList body_type) (VariableList))
           ((cog-subtype? 'Link body_type)
            (apply cog-new-link
                   (cons body_type


linas:
Looks reasonable to me, but its not my code.  Side comments:

1) Documentations should be provided, This can be done like so:
```
(define-public (foo bar)
"
  This foo bar documentation is printed when user does
  ,describe foo  or simply ,d foo at the command line.
"
 (cons foo bar))
```

2) There is now an `AlwaysLink` in the pattern matcher, which implements for-all semantics. It's like `PresentLink` or `AbsentLink`, except that the pattern matches only when every match, every variation, always has that clause satisfied.   See the [examples/pattern-matcher/always.scm]  example (baskets with colored balls  issue from half a year ago).

ngeiswei:
@noskill, maybe it should only remove outgoings of VariableList that are non variables. Anyway, that can be merged for now. BTW, I no longer use that code since I'm using meta-rules.

ngeiswei:
URE fails but that's probably because that PR is based on an older master, so I'm merging.

linas:
>  no longer use that code 

Maybe we should remove this? It's undocumented. I read it a bit more carefully, and it seems to be duplicating what `PutLink` does, but outside of the atomspace.

ngeiswei:
If nobody is using it I'm for removing it. @noskill, do you need that code?

noskill:
I used substitute-var to substitute variables with some atoms in a bindlink for debug purposes. Perhaps the same can be done with atomese, not sure how..

ngeiswei:
Maybe you can use `PutLink`. Otherwise no problem, can stay.

noskill:
PutLink doesn't work with BindLink

noskill:
ERROR: In procedure cog-new-link:
Throw to key `C++-EXCEPTION' with args `("cog-new-link" "PutLink was expecting a ListLink, SetLink or GetLink! (/home/noskill/projects/atomspace/opencog/atoms/core/PutLink.cc:223)

linas:
?

If you use PutLink all by itself, then that exception just says that you are using PutLink wrong; in this case RTFM.

I believe that one should be able to say stuff like
```
(PutLink  
    (VariableList (Variable "x") (Variable "y"))
    (BindLink 
         (VariableList (Variable "a") (Variable "b"))  
         (FooBarLink  (Variable "x") (Variable "y"))
         (BarFooLink (Variable "a")  (Variable "x")  (Variable "b")))  ; end of bind
   (List (Concept "p") (Concept "q")))
```
and the result of running that should be
```
 (BindLink 
         (VariableList (Variable "a") (Variable "b"))  
         (FooBarLink  (Concept "p") (Concept "q"))
         (BarFooLink (Variable "a")  (Concept "p")  (Variable "b")))
```
and then... Hmm. I guess `cog-execute!` would then keep going, and try to run the BindLink, too ...   I've never tried this before ... maybe its' buggy ...

noskill:
I read the manual e.g. wiki page. It says
```
 PutLink
    <clause containing multiple variables>
    ListLink
       <atoms that will replace the variables>
```

In my case <clause containing multiple variables>  is a bindlink. 
While you wrote 

```
 PutLink
   VariableList
    <clause containing multiple variables>
    ListLink
       <atoms that will replace the variables>
```

Anyway it doesn't work like expected:

```scheme
scheme@(guile-user)> (define b1 (BindLink
   (VariableList
      (TypedVariableLink
         (VariableNode "ob")
         (TypeNode "ConceptNode")
      )
   )
   (AndLink
      (InheritanceLink
         (VariableNode "ob")
         (ConceptNode "object")
      )
   )
   (VariableNode "ob")
))
```
```scheme
scheme@(guile-user) > (cog-execute! (PutLink (VariableList (VariableNode "ob") ) b1 (ListLink (ConceptNode "a"))))
$12 = (SetLink
   (ConceptNode "b")
   (ConceptNode "a")
)
```

linas:
OK, well, you are violating alpha-renaming rules with that put; the atomspace will alpha-hide the arguments to the bindlink.

PR #2377, opened 2019-11-21
Make it easier for the compiler to inline.

linas:
Move simple returns into the header file.

### DIFF
--- a/opencog/atoms/pattern/PatternTerm.cc
+++ b/opencog/atoms/pattern/PatternTerm.cc
@@ -5,21 +5,7 @@
  * All Rights Reserved
  *
  * Created by Nil Geisweiller Oct 2016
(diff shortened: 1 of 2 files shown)


ngeiswei:
Does this really makes a difference? Did you check the generated assembly code?

linas:
> Does this really make a difference? Did you check the generated assembly code?
 
I think it makesa difference, I did not check assembly code. I think it makes a difference because when one sets `-flto` in `CFLAGS` (LTO==link-time-optimization) the performance actually improves 1-2%  what LTO does is inline functions from across modules.  Currently LTO is disabled in `OpenCogGccOptions.cmake` because it makes build about 4x slower than otherwise -- unbearably slow during development.

PR #2380, opened 2019-11-21
Include attention and cogserver in circleci dependencies.

linas:
failure to include these explains some of the recent breakages.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -120,7 +120,8 @@ jobs:
           root: /ws/
           paths:
             - ure
-            - ccache
+            # See https://github.com/opencog/cogutil/commit/da53c4079c3972904857a365a7761ac5923e4e92
+            # - ccache
 
   miner:
     docker:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -177,6 +177,119 @@ jobs:
             - miner
             - ccache
 
+  cogserver:
+    docker:
+      - image: opencog/opencog-deps
+        user: root
+        environment:
+          CCACHE_DIR: /ws/ccache
+    working_directory: /ws/cogserver
+    steps:
+      - attach_workspace:
+          at: /ws
+      - run:
+          name: Set number of make jobs
+          command: echo "export MAKEFLAGS=-j2" >> $BASH_ENV
+      - run:
+          name: Install CogUtil
+          command: cd /ws/cogutil/build && make -j2 install && ldconfig
+      - restore_cache:
+          name: Restore GHC Cache
+          keys:
+            - ghc-{{checksum "/ws/atomspace/opencog/haskell/stack.yaml"}}-{{ arch }}
+      - restore_cache:
+          name: Restore Haskell Deps Cache
+          keys:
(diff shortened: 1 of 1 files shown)


vsbogd:
@linas , thanks for adding dependencies, but it is not enough, they should also be part of the workflow, which is at the end of the config: https://github.com/opencog/atomspace/blob/fd2957f4fbf3223665b890de9c7909c47422e7dc/.circleci/config.yml#L364-L383

vsbogd:
Another place is `opencog` job, `cogserver` and `attention` should be installed like `URE` before building `opencog` itself: https://github.com/opencog/atomspace/blob/fd2957f4fbf3223665b890de9c7909c47422e7dc/.circleci/config.yml#L317-L319

linas:
@vsbogd fixed. Please merge if you think this is the right fix. I just copied this from the cogutils config.yml.

vsbogd:
Ah, CI fails, we had already seen this issue with @ngeiswei, same component were built twice.

vsbogd:
The same fix should help: https://github.com/opencog/cogutil/commit/da53c4079c3972904857a365a7761ac5923e4e92#diff-1d37e48f9ceff6d8030570cd36286a61

linas:
Is it because of
```
      - opencog:
          requires:
            - atomspace
            - ure
            - cogserver
            - attention
```
and atomspace was previously required by both ure and cogserver ... and also cogserver was required by attention?

vsbogd:
Yes, it is strange but the reason is that `atomspace` is in two chains of builds.

linas:
Something wrong with the "two chains of builds" theory. I stubbed out the ccache in the *other* path, not the URE path, and it still failed. So this is specific to URE, somehow. Does URE maybe have copies of files that are also in the atomspace? @ngeiswei ?

vsbogd:
I believe to try second way you need to disable caching in CogServer and Attention at the same time.
Nil's experiments: https://github.com/opencog/cogutil/pull/201

linas:
Hmm current checks are failing in cogserver `ShellUTest` The test itself passes (prints `OK`) but then crashes during exit, with 
```
terminate called after throwing an instance of 'concurrent_stack<opencog::SchemeEval*>::Canceled'
  what():  std::exception
```
I have a really pretty good idea of where, how and why this is happening, and should ebe able to fix it quickly,  except that I cannot reproduce it, so am kind-f blind about exactly where.  (I ran 300 or 500 reps on my systems without any failures, and the the cat literally pressed the reset button on the box.)

A stack trace for this would be very useful.

linas:
I've run 30 repeats of ShellUTest with zero failures, on boost 1.62  I'm on Debian stable, `gcc --version gcc (Debian 8.3.0-6) 8.3.0`  and `ldd --version ldd (Debian GLIBC 2.28-10) 2.28`

There's something about the circleci environment that differs

linas:
OK, so, with boost-1.67, ShellUTest intermittently fails; it always passes for boost-1.62

linas:
I'm going to merge this (for now) and then disable the cogserver so that the ShellUTest doesn't interfere.

PR #2381, opened 2019-11-21
Include attention and cogserver in circleci dependencies. 

linas:
Variation on pull req #2380, searching for an alternative fix.


linas:
OK, so this did not work.

Issue #2382, opened 2019-11-22
Fiddling with auxiliary atomspace in peculiar way invalids a handle


ngeiswei:
This one is wild. By accessing an atomspace via an atom in it, and re-accessing that handle makes it invalid. Here's the shortest way to reproduce it, I'll try to squeeze if I can but that's all I've got so far. Read carefully the comment in the end, as the bug can be circumvented by accessing the auxiliary atomspace directly instead of via an atom, really weird.

```scheme
(use-modules (opencog logger))
(cog-logger-set-timestamp! #f)
(cog-logger-set-stdout! #t)

;; Save initial atomspace
(define initial-as (cog-atomspace))

;; Define an auxiliary atomspace
(define auxiliary-as (cog-new-atomspace))

;; Switch to the auxiliary atomspace
(define pre-pre-as (cog-set-atomspace! auxiliary-as))

;; Inside this auxiliary atomspace, create concept A
(define auxiliary-A (Concept "A"))

;; Switch the atomspace where A is (which is auxiliary-as)
(define pre-as (cog-set-atomspace! (cog-as auxiliary-A)))

;; Verify that the current atomspace, auxiliary-as and the prevous one
;; are all the same
(cog-logger-info "[1] current as = ~a" (cog-atomspace))
(cog-logger-info "[1] auxiliary-as = ~a" auxiliary-as)
(cog-logger-info "[1] pre-as = ~a" pre-as)

;; Come back to the previous atomspace (which should still be auxiliary-as)
(cog-set-atomspace! pre-as)

;; Verify that the current atomspace and auxiliary-as are the same
(cog-logger-info "[2] current as = ~a" (cog-atomspace))
(cog-logger-info "[2] auxiliary-as = ~a" auxiliary-as)

;; Come back to the previous of the previous atomspace (which should be the initial atomspace)
(cog-set-atomspace! pre-pre-as)

;; Verify that the current atomspace and initial-as are the same
(cog-logger-info "[3] current as = ~a" (cog-atomspace))
(cog-logger-info "[3] initial-as = ~a" initial-as)

;; Switch again to the auxiliary atomspace
(define pre-as-again (cog-set-atomspace! auxiliary-as))

;; Create again concept A in it
(define A (Concept "A"))

;; Switch back to initial atomspace
(cog-set-atomspace! pre-as-again)

;; Now A is an invalid handle.
;;
;; Very strangly if the line
;;
[...]

PR #2389, opened 2019-11-24
AtomSpace memory management fix

linas:
This fixes issue #2382 where atomspaces were being prematurely deleted.


ngeiswei:
Thank you so much! It was a hard one to crack, well done!

linas:
Welcome. This had been a problem area for years; I can't say why it was so confusing, but I'm pretty confident that this was the last bug.  😊 😊 😊

PR #2393, opened 2019-11-27
Misc glob supports

kasimebrahim:
Miscellaneous changes to support GlobNode for PutLink, Unification and ...

### DIFF
--- a/opencog/atoms/core/Variables.cc
+++ b/opencog/atoms/core/Variables.cc
@@ -539,12 +539,12 @@ Handle FreeVariables::substitute_scoped(const Handle& term,
 		if (GLOB_NODE == h->get_type())
 		{
 			Handle glst(substitute_scoped(h, args, silent, index_map, quotation));
-			if (glst->is_node())
-				return glst;
-
 			changed = true;
-			for (const Handle& gl : glst->getOutgoingSet())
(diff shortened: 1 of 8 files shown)


opencog/atoms/core/Variables.cc
```diff
@@ -539,12 +539,12 @@ Handle FreeVariables::substitute_scoped(const Handle& term,
 		if (GLOB_NODE == h->get_type())
 		{
 			Handle glst(substitute_scoped(h, args, silent, index_map, quotation));
-			if (glst->is_node())
-				return glst;
-
 			changed = true;
-			for (const Handle& gl : glst->getOutgoingSet())
-				oset.emplace_back(gl);
+			if (glst->get_type() == LIST_LINK)
```
linas: I believe that this is unwrapping the results from `wrap_glb_with_list()`, is that right? If so, can you add a comment saying this?

linas:
Wow. Looks absolutely perfect .. clean, crisp, minimal, organized, with unit tests. I really like it!

opencog/atoms/core/Variables.cc
linas: I will add this comment myself, in just a moment.

kasimebrahim:
Thanks!

PR #2395, opened 2019-11-27
Use C++ math library instead of C one.

vsbogd:
Fix compilation error "std:round() is not defined"


dagims:
@vsbogd the weird thing is that no math function is used in that source. Maybe it's a leftover include?

vsbogd:
Ah, sorry, looks like this code was refactored recently. This fix is actual for `singnet` repo.

ngeiswei:
BTW, @vsbogd there's been a lot of changes recently, VariableSet, etc. I haven't generated singnet pull requests because that big round is not totally over, almost though!

Issue #2391, opened 2019-11-25
split python initialization and setting default atomspace in python api


noskill:
Current python api is confusing. My idea is

1)  give user set_default_atomspace function to set current default atomspace - analogous to cog-set-atomspace in guile.  
2) perform python initialization automatically - it is possible to check if python is already initialized using Py_IsInitialized function.

vsbogd:
> perform python initialization automatically - it is possible to check if python is already initialized using Py_IsInitialized function.

@noskill, what will trigger Python initialization? Do you mean importing any opencog package will initialize Python interpreter?

noskill:
I think it can be delayed until PythonEval instance creation,  wherever it happens.

vsbogd:
I would not remove `initialize_opencog` completely as lazy initialization has disadvantages.
In case of lazy initialization software developer don't control point when initialization happens:
- it may lead to unexpected errors after some work is already done;
- performance could be poor at some critical step because initialization happens at this moment.

Taking it into account I would keep `initialize_opencog` for cases when software developer would like to control moment initialization happens. If one doesn't work to call it right now `PythonEval::instance()` already creates new instance if it is not created already.

I agree that split Python initalization and atomspace setting is a good idea. I looked at `PythonEval` code and I see that `_atomspace` is used to set global `ATOMSPACE` variable for each loaded module. But this global `ATOMSPACE` variable is not used in code.

linas:
It would be useful to clarify the different usage scenarios.  The ones I see are:

1) User starts python first, then does `import atomspace` then creates atoms, etc. with python.

2) User starts cogserver first, then runs python scripts inside cogserver.

For case 1) it seems pointless to have `initialize_opencog` .. everything should be initialized by the time that the `import` finishes.  Also, its pointless to ask the user to set a "default atomspace" it could be set automatically during the import. Changing it makes sense only if the user wants to work with more than one atomspace (which is an advanced thing to do; almost no users will ever do this.).  

There are two interesting subcases:

1a) Python user loads scheme file(s) e.g. by using PLN, where the "formulas" are in `GroundedPredicate "scm:some-formula"`  and scheme has to get the correct atomspace. I'm not sure if there's a unit test for this or not.

1b) Python user starts the cogserver.  In this case, the cogserver needs to use the same atomspace as python. I'm not even sure if the python API's allow you to start the cogserver.

[...]

PR #2398, opened 2019-11-28
Remove global atomspace variable from PythonEval

noskill:
Atomspace set by initialize_opencog wasn't used anywhere, except by tests.

This PR implements some changes from discussion in this issue https://github.com/opencog/atomspace/issues/2391

### DIFF
--- a/opencog/atoms/execution/EvaluationLink.cc
+++ b/opencog/atoms/execution/EvaluationLink.cc
@@ -740,7 +740,7 @@ TruthValuePtr do_eval_with_args(AtomSpace* as,
 
 		// Be sure to specify the atomspace in which to work!
 		PythonEval &applier = PythonEval::instance();
-		return applier.apply_tv(as, schema.substr(pos), args);
+		return applier.apply_tv(schema.substr(pos), args);
 #else
 		throw RuntimeException(TRACE_INFO,
 			"This binary does not have python support in it; "
--- a/opencog/cython/PythonSCM.cc
+++ b/opencog/cython/PythonSCM.cc
@@ -98,12 +98,17 @@ void* PythonSCM::init_in_guile(void* self)
 	scm_c_define_module("opencog python", init_in_module, self);
 	scm_c_use_module("opencog python");
 	PythonEval& pev = PythonEval::instance();
+  	// Make sure that guile and python are using the same atomspace.
+	// This will avoid assorted confusion.
+	AtomSpace* as = SchemeSmob::ss_get_env_as("python-eval");
 
 	// This feels hacky, I guess, but I cannot figure out any other
 	// way of telling python which atomspace it is supposed to use
 	// by default.
 	pev.eval("from opencog.atomspace import AtomSpace");
 	pev.eval("from opencog.type_constructors import set_type_ctor_atomspace");
+	pev.eval("set_type_ctor_atomspace(AtomSpace(" +
+            std::to_string((uint64_t) as) + "))\n");
 
 	return NULL;
 }
--- a/opencog/cython/PythonSCM.cc
+++ b/opencog/cython/PythonSCM.cc
@@ -97,21 +97,13 @@ void* PythonSCM::init_in_guile(void* self)
 {
 	scm_c_define_module("opencog python", init_in_module, self);
 	scm_c_use_module("opencog python");
-
-	// Make sure that guile and python are using the same atomspace.
-	// This will avoid assorted confusion.
-	AtomSpace* as = SchemeSmob::ss_get_env_as("python-eval");
-	PythonEval& pev = PythonEval::instance(as);
+	PythonEval& pev = PythonEval::instance();
 
 	// This feels hacky, I guess, but I cannot figure out any other
 	// way of telling python which atomspace it is supposed to use
 	// by default.
 	pev.eval("from opencog.atomspace import AtomSpace");
 	pev.eval("from opencog.type_constructors import set_type_ctor_atomspace");
-	char buf[301];
-	snprintf(buf, 300, "set_type_ctor_atomspace(AtomSpace(%ld))\n",
-		(uint64_t) as);
-	pev.eval(buf);
 
 	return NULL;
 }
--- a/tests/cython/PythonEvalUTest.cxxtest
+++ b/tests/cython/PythonEvalUTest.cxxtest
@@ -136,45 +136,27 @@ public:
     {
         logger().debug("[PythonEvalUTest] testPythonEvalSingletonCreationDeletion()");
 
-        AtomSpace* atomSpace = NULL;
 
         // Tell the python evaluator to create its singleton instance.
-        PythonEval::create_singleton_instance(NULL);
+        PythonEval::create_singleton_instance();
 
         // Delete the singleton instance of the PythonEval.
         PythonEval::delete_singleton_instance();
 
-
-        // Create an atomspace.
-        atomSpace = new AtomSpace();
-
-        // Tell the python evaluator to create its singleton instance
-        // with this atomspace.
-        PythonEval::create_singleton_instance(atomSpace);
+        PythonEval::create_singleton_instance();
 
         // Delete the singleton instance of the PythonEval.
         PythonEval::delete_singleton_instance();
 
-        // Delete the atomspace.
-        delete atomSpace;
-        atomSpace = NULL;
 
         // Initialize Python.
         global_python_initialize();
 
-        // Create an atomspace.
-        atomSpace = new AtomSpace();
-
-        // Tell the python evaluator to create its singleton instance
-        // with this atomspace.
-        PythonEval::create_singleton_instance(atomSpace);
+        // Repeat the process
+        PythonEval::create_singleton_instance();
 
         // Delete the singleton instance of the PythonEval.
         PythonEval::delete_singleton_instance();
-
-        // Delete the atomspace.
-        delete atomSpace;
-        atomSpace = NULL;
 
         logger().debug("[PythonEvalUTest] testPythonEvalSingletonCreationDeletion() DONE");
     }
@@ -182,35 +164,31 @@ public:
     void testPythonEvalEvalExpr()
     {
         logger().debug("[PythonEvalUTest] testPythonEvalEvalExpr()");
-
-        // Create an atomspace and singleton instance to use it.
-        AtomSpace* atomSpace = new AtomSpace();
-        PythonEval::create_singleton_instance(atomSpace);
+        PythonEval::create_singleton_instance();
         PythonEval* evaluator = &PythonEval::instance();
 
         // Check that our atom space global is visible and can add a node.
         evaluator->clear_pending();
         evaluator->eval_expr("from opencog.atomspace import *");
+        evaluator->eval_expr("ATOMSPACE = AtomSpace()");
         evaluator->eval_expr("ATOMSPACE.add_node(types.ConceptNode,'test')");
         evaluator->eval_expr("print ('Total atoms = ', ATOMSPACE.size(), '\\n')");
 
-        // Delete the singleton instance and atomspace.
+        // Delete the singleton instance.
         PythonEval::delete_singleton_instance();
-        delete atomSpace;
-        atomSpace = NULL;
         logger().debug("[PythonEvalUTest] testPythonEvalEvalExpr() DONE");
     }
 
     void testApplyAndApplyTV()
     {
-        AtomSpace *as = new AtomSpace();
-        PythonEval::create_singleton_instance(as);
+        PythonEval::create_singleton_instance();
         PythonEval* python = &PythonEval::instance();
 
         // Define python functions for use by scheme.
-        python->eval(
-            "from opencog.atomspace import types, Atom, TruthValue\n"
+        std::string eval_res = python->eval(
+            "from opencog.atomspace import vsfvkdfjnv Atom, TruthValue\n"
             "def add_link(atom1, atom2):\n"
+            "    ATOMSPACE = atom1.atomspace\n"
             "    link = ATOMSPACE.add_link(types.ListLink, [atom1, atom2])\n"
             "    print ('add_link(',atom1,',',atom2,') = ', link)\n"
             "    return link\n\n"
@@ -219,8 +197,10 @@ public:
             "    print ('return_truth TruthValue(0.75, 100.0)')\n"
             "    return TruthValue(0.75, 100.0)\n\n"
             );
+        std::cout << "eval_res:" << std::endl << eval_res << std::endl;
 
 #ifdef HAVE_GUILE
+        AtomSpace * as = new AtomSpace();
         // Create a scheme evaluator to trigger cog-evaluate and cog-execute.
         SchemeEval* scheme = new SchemeEval(as);
 
@@ -256,14 +236,14 @@ public:
                 );
         error = scheme->eval_error();
         scheme->clear_pending();
+        delete as;
         TSM_ASSERT("Failed cog-evaluate!", !error);
 #endif // HAVE_GUILE
     }
 
     void testCodeBlockWithNewline()
     {
-        AtomSpace *as = new AtomSpace();
-        PythonEval::create_singleton_instance(as);
+        PythonEval::create_singleton_instance();
         PythonEval* python = &PythonEval::instance();
 
         // Define python functions with newline within the code block
--- a/tests/cython/PythonEvalUTest.cxxtest
+++ b/tests/cython/PythonEvalUTest.cxxtest
@@ -184,9 +184,12 @@ public:
         PythonEval::create_singleton_instance();
         PythonEval* python = &PythonEval::instance();
 
+        TS_ASSERT_THROWS(std::string eval_res = python->eval(
+                    "from opencog.atomspace import vsfvkdfjnv\n"), RuntimeException);
+
         // Define python functions for use by scheme.
         std::string eval_res = python->eval(
-            "from opencog.atomspace import vsfvkdfjnv Atom, TruthValue\n"
+            "from opencog.atomspace import Atom, TruthValue\n"
             "def add_link(atom1, atom2):\n"
             "    ATOMSPACE = atom1.atomspace\n"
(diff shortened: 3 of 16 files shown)


tests/cython/PythonEvalUTest.cxxtest
```diff
@@ -136,81 +136,59 @@ public:
...
-        AtomSpace *as = new AtomSpace();
-        PythonEval::create_singleton_instance(as);
+        PythonEval::create_singleton_instance();
         PythonEval* python = &PythonEval::instance();
 
         // Define python functions for use by scheme.
-        python->eval(
-            "from opencog.atomspace import types, Atom, TruthValue\n"
+        std::string eval_res = python->eval(
+            "from opencog.atomspace import vsfvkdfjnv Atom, TruthValue\n"
```
vsbogd: Does this typo demonstrates that this tests never fails? :-)

noskill: Unintentionally, though it somehow passes

opencog/cython/PythonSCM.cc
```diff
@@ -97,21 +97,13 @@ void* PythonSCM::init_in_guile(void* self)
...
 
 	// This feels hacky, I guess, but I cannot figure out any other
 	// way of telling python which atomspace it is supposed to use
 	// by default.
 	pev.eval("from opencog.atomspace import AtomSpace");
 	pev.eval("from opencog.type_constructors import set_type_ctor_atomspace");
-	char buf[301];
-	snprintf(buf, 300, "set_type_ctor_atomspace(AtomSpace(%ld))\n",
-		(uint64_t) as);
-	pev.eval(buf);
```
vsbogd: Agree, that approach used in this code is error-prone, but removing this changes behavior. 
In scenario when Python code is being called from Scheme and Python is creating new atom using atom type constructor default atomspace is None and it should lead to errors. Nevertheless first example I found where Python is called from Scheme accurately uses atomspace in Python https://github.com/opencog/atomspace/blob/master/examples/atomspace/python.scm and isn't affected.

noskill: One can set default atomspace for python from scheme calling initialize_opencog from python. Though i'll split initialize_opencog into two methods in another PR.

vsbogd: After thinking about it I realized that as we have an API which implicitly sets an atomspace into atom constructors then initializing the atomspace when Scheme initializes Python is normal. This PR is about splitting Python API initialization and type constructor atomspace setter so these both steps should be used when initializing Python from Scheme.

The problem is that in the following two scenarios we should have the different behavior:
1. Scheme initializes Python 
2. Python initializes Scheme which initializes Python

In the first scenario we should set Scheme's atomspace as a type constructor atomspace for Python.
In the second scenario we should set Python's atomspace as a type constructor atomspace for Scheme and during the second initialization just do nothing. To do this we need to know whether PythonEval was initialized for the first time or not. If PythonEval is just initialized then we need to set type constructor atomspace.

noskill: user can set atomspace from scheme using:
```scheme
(python-call-with-as "set_type_ctor_atomspace" (cog-atomspace))
```

Though I can make common global pointer to default atomspace for both python and scheme

linas: This creates a bug for the user ... it is not reasonable for the user to know that sometimes, under certain special conditions, that python and scheme sometimes use two different atomspaces, and then try to guess when this might happen, and try to manually patch that up in their code.  It's a unified system, it should behave like a unified system.

linas: There's also a third scenario:

3. Cogserver initializes AtomSpace, and then initializes Scheme and Python

opencog/cython/PythonSCM.cc
```diff
@@ -97,21 +97,13 @@ void* PythonSCM::init_in_guile(void* self)
...
-	// This will avoid assorted confusion.
-	AtomSpace* as = SchemeSmob::ss_get_env_as("python-eval");
-	PythonEval& pev = PythonEval::instance(as);
+	PythonEval& pev = PythonEval::instance();
 
 	// This feels hacky, I guess, but I cannot figure out any other
 	// way of telling python which atomspace it is supposed to use
 	// by default.
 	pev.eval("from opencog.atomspace import AtomSpace");
 	pev.eval("from opencog.type_constructors import set_type_ctor_atomspace");
```
linas: If you are removing the code below, then please remove the imports, and the comments above, as they are no longer needed. (and teh comments are no longer correct)

tests/scm/scm-python-shared-atomspace.scm
```diff
@@ -19,9 +19,10 @@
 from opencog.atomspace import AtomSpace, TruthValue, types
 from opencog.type_constructors import atomspace
```
linas: If the atomspace is no longer global (below), then the import above is not needed. Same remark for `scm-python-arity.scm` test, above.

tests/cython/bindlink/test_functions.py
```diff
@@ -1,18 +1,20 @@
-# 
+#
 # Test Functions
 
 from opencog.atomspace import types, Atom, TruthValue
 from opencog.type_constructors import *
```
linas: This import also seems not to be needed.  Also, if I understand properly, I guess that `opencog.type_constructors` exports both `ATOMSPACE` and also `atomspace` both of which are set to the same value!???

examples/atomspace/python.scm
```diff
@@ -29,6 +29,7 @@
 (python-eval "
 from opencog.atomspace import AtomSpace, TruthValue
 from opencog.atomspace import types
+from opencog.utilities import set_type_ctor_atomspace
```
linas: In the unit tests at the bottom, we have `from opencog.type_constructors import atomspace` which imports the single global atomspace.   But here, we have to use `opencog.utilities` to set that single global atomspace. Shouldn't the setter and the getter live in the same place?

opencog/atoms/execution/EvaluationLink.cc
```diff
@@ -740,7 +740,7 @@ TruthValuePtr do_eval_with_args(AtomSpace* as,
 
 		// Be sure to specify the atomspace in which to work!
 		PythonEval &applier = PythonEval::instance();
-		return applier.apply_tv(as, schema.substr(pos), args);
+		return applier.apply_tv(schema.substr(pos), args);
```
linas: This is surely a bug ... during pattern matching, possible matches are evaluated in a temporary atomspace. If, in the end, the match is good, then the results are placed in the main atomspace. If there is no match, the results are discarded.  With the above change ... where are the temporary results written? How can you know which atomspace to use?

opencog/cython/PythonSCM.cc
noskill: If we implicitly share default atomspace between interpreters, then what should happen if one of interpreters change default atomspace?

opencog/atoms/execution/EvaluationLink.cc
linas: I'm pretty sure that none of the python unit tests test for this, since the pattern matcher is almost never used with python.  But if you go through the unit tests and look for `GroundedPredicate "scm:foo"`  (which calls foo in the temp atomspace) and then create a matching `GroundedPredicate "py:foo"`, you'll see the failure.

linas:
It's a big change, but I guess overall I like it.  The change to `apply_tv()` is a real bug though; evaluations always have to occur in the specified atomspace, and not some other random atomspace.  Only the caller knows what the correct atomspace is.

opencog/atoms/execution/EvaluationLink.cc
noskill: Which test should it port to python to expose bug?

opencog/cython/PythonSCM.cc
linas: Ahh, good question. Nothing, I think.  There's no "default" atomspace, not really; there's only the "initial atomspace created when system is started" and after that, its the "atomspace of the current context", the context being per-thread. There are four "common" contexts:

1. During pattern matching, partial results are evaluated in a temp atomspace; the correct atomspace to use is set using the arguments to the c++ `ExecutationOutputLink` and the c++ `EvaluationLink`. Sometimes, these can be two or three deep, because Gets/Puts can be nested arbitrarily. (Yes, this does happen).  Also, different threads can have different atomspaces (e.g. one thread is handling sensory input, e.g. from cameras; a different thread is doing PLN/mining/language/etc) 

2. During pattern mining ... there is also some kind of temp atomspace, I don't know what its for.  It's not one of the temps used by the matcher.

3. in agi-bio, the default atomspace is read-only and shared; its huge. All updates (TV changes, adds, deletes) occur in an overlay atomspace that is read-write.  (and pattern matching is then in a temp atomspace on top of this overlay) It's done this way to save disk space: the shared-read-only is huge and contains genetic data, everyone uses the same copy, and then everyone creates just minor patches to it, for different experiments.

4. in guile, at least, atomspaces are per-thread, so that two different threads can run at the same time, each thread using a different atomspace.  I'm not sure where this is used.

opencog/atoms/execution/EvaluationLink.cc
linas: Not sure, but here's how to find out: in `DefaultPatternMatchCB` you will find 
```
   _temp_aspace = grab_transient_atomspace(as);
   _instor = new Instantiator(_temp_aspace);
```
and then, below, in `DefaultPatternMatchCB::eval_term` you will see 
```
gvirt = HandleCast(_instor->instantiate(virt, gnds, true));
```
which uses the temp atomspace, and also two diferent uses of
```
EvaluationLink::do_eval_scratch(_as, gvirt, _temp_aspace, true);
```
put an assert in there, and see which unit tests trigger this.  All users of `EvaluationLink::do_eval_scratch()` will be using temp atomspaces.

linas: Or just put the assert into `EvaluationLink::do_eval_scratch()` and see which unit tests trigger that.

linas: So for example ... `do_eval_with_args` is *always* called with the temp atomspace, from what I can tell, so the `apply_tv()` will always specify the temp atomspace.

opencog/cython/PythonSCM.cc
noskill: Vitaly had great idea to use thread-local queue per thread.  I then we can set current default atomspace per thread instead of per interpreter. 

Somehow like this:
scheme or python call c++ method

```c++
push_atomspace_for_thread(std::shared_ptr<AtomSpace> a)
```

then scheme or python can call c++ factory for atom creation:
```c++
AtomPtr create_with_current_context(Type t, ...args)
```

where create_with_current_context takes atomspace from thread-local queue

opencog/atoms/execution/EvaluationLink.cc
noskill: I ported CondLinkUTest::test_grounded_cond. It passed, i still do no see why should it fail. Pattern matcher just checks TruthValue returned.. TruthValue might not be in any atomspace at all.

opencog/cython/PythonSCM.cc
noskill: Though with this design one would have to set atomspace upon thread creation, hmm..

opencog/atoms/execution/EvaluationLink.cc
vsbogd: From the code which Anatoly changed the only usage of this parameter was to set it to `_atomspace` field of the `PythonEval`. And only usage of `_atomspace` field was to set it as global `ATOMSPACE` variable in Python modules. And global `ATOMSPACE` variable is not used by Python code. So may be intention was as you described but in fact this code doesn't work by this way.

opencog/cython/PythonSCM.cc
vsbogd: To explain the idea further Anatoly and me discussed that working with two global variables on two platforms (Scheme and Python) is really difficult. Use cases described above suggest one need to set the global atomspace variable before calling Python from Scheme (or vice versa) and revert the value after returning from the call. And what if this global variable is changed by the code in the parallel thread.

We have concluded the following:
- each thread should have its own pointer to the current atomspace which is used by the code of the thread;
- each thread should have API to change its pointer before executing block of code and  revert this change after block is executed (something similar to `(cog-push-atomspace)` `(cog-pop-atomspace)` in Scheme);
- this pointer should be shared by Scheme, Python and C++ code.

Under these three conditions it doesn't matter how complex is a chain of the calls. C++ -> Python -> Scheme -> Python -> C++. The atomspace pointer is always same.

linas:
OK, this conversation is spiralling out of control; maybe we could have a simpler pull req that just removed the global all-uppercase `ATOMSPACE` which no one uses? I have no objection to that.  Also, maybe a second pull req that moves the set and get to the same module? And shortens the name from `set_type_ctor_atomspce` into something ... simpler and more meaningful. 

As to global variables, per-thread or not, I would rather avoid them. They're not needed for C++. The only reason they're used for scheme is to keep the bindings easy-to-read. But its "optional", because the following works:
```
(define as (cog-new-atomspace))
(define bs (cog-new-atomspace))
(Concept "aaa" as)   ; create this atom in as
(Concept "bbb" bs)    ;  create this atom in bs
; stupid cog-prt-atomspace doesn't take an argument; will fix this shortly
(cog-prt-atomspace)
(cog-set-atomspace! as)
(cog-prt-atomspace)
(cog-set-atomspace! bs)
(cog-prt-atomspace)
```

The scheme push-n-pop create "child" atomspaces; the temp/scratch atomspaces are "child atomspaces.  But the pattern miner creates unrelated, non-child atomspaces.

linas:
For a temp/scratch atomspace example, consider the following pseudocode:
```
(Get  (Variable "x")
   (And
       (Present (ListLink (Concept "foo") (Variable "x")))
       (Evaluation (GroundedPredicate "py:test_one")(List (Variable "x")))
       (Evaluation (GroundedPredicate "py:test_two")(List (Variable "x")))
    ))
```
with 
```
def test_one(v)
   thing = add_node(types.Concept, "barleycorn")
   thang = add_link(types.ListNode, thing, v, TruthValue(0.9, 0.75))
   thong = space_server.get(thang)
   return thong.getTruthValue()

def test_two(v):
    return TruthValue (0,1)   # false
```

In this example, the atoms `thing` `thang` and `thong` are supposed to be created in the temp atomspace, not the main atomspace.  The main atomspace is never polluted with these extra atoms. Doubly so, since the second test always fails, so the GetLink should always return the empty set.

That's the intent of the temp atomspace: to avoid polluting the main atomspace with any temporary atoms created during the search.

The `applier.apply_tv(as, schema.substr(pos), args);` needs to stay as-is, so that python can be told which atomspace will hold `thing`, `thang`, `thong`.  I don't really care how it works under the covers, or what API you want to have, in python.

vsbogd:
> As to global variables, per-thread or not, I would rather avoid them. They're not needed for C++. The only reason they're used for scheme is to keep the bindings easy-to-read.

I totally agree the best fix is using no global variables at all. But most of the Python and Scheme API code examples exploit global variables very intensive. And how can we guarantee the code written in this manner works correctly when it calls Python from Scheme or vice versa. The idea is to "merge" Python and Sceme global variables, add push/pop API to simplify nested context management and allow using them from C++ when it is needed.

Another option could be designing convenient Scheme and Python API without global variables.

> The scheme push-n-pop create "child" atomspaces; the temp/scratch atomspaces are "child" atomspaces. But the pattern miner creates unrelated, non-child atomspaces.

When I wrote about changing the atomspace pointer I didn't mean new atomspace is necessary a child of the current. It can be any unrelated atomspace.

linas:
The only "global" that I know of is the atomspace, and that's only in the scheme bindings; there aren't any globals in the C++ that I can think of.  (well, there are, but they are carefully hidden from the user, and are used to manage pools of various sorts).

The only reason the atomspace is a global is a historical accident: the cogserver was originally coded as a singleton. So the original bindings assumed only one atomspace. Later, when the idea of multiple atomspaces was added, backwards-compatibility meant that the scheme bindings continued to "pretend" there was just a single atomspace.  The API then set the "current" atomspace.  To make everything work lock-less in a multi-threaded env, the "current atomspace" had to be per-thread, thus avoiding race conditions, mutexes, etc.

The other way to avoid locks/mutexes is to pass the atomspace as an argument to everything that needs access to it.

The original python bindings arrived much later, and were multi-atomspace from the beginning, using an API of the form
```
as = AtomSpace()
bs = AtomSpace()
as.add_node(types.ConceptNode, "foo")
bs.add_node(types.ConceptNode, "bar")
```
At some point, someone (correctly) decided that this was a very clunky API, and changed it to
```
ConceptNode("foobar")
```
which is much more readable, but now requires a singleton global... it is/was the all-upper-case `ATOMSPACE` but then I guess someone changed that.  Or something. Then things got messy. If you make friends with Lake, he will tell you about the difficult times back then, when all these incompatibilities got created.  There was even an attempt to rewrite the entire atomspace, the pattern-matcher, everything - in python. (fishgram, buried deep in git) (And someone else, Boris Iordanov, an American-Russian Florida man started to rewrite it all in Java... Also both Haskel and Erlang were proposed ... some progress in haskell got made ... it was the era of language wars. )

opencog/cython/PythonSCM.cc
noskill: @linas I see that in PythonSCM::init_in_guile there is code that imports in python set_type_ctor_atomspace. Comment says it supposed to set default atomspace. It looks like it doesn't, where the default atomspace should come from?

tests/cython/bindlink/test_functions.py
noskill: It is needed to import type constructors

opencog/cython/PythonSCM.cc
noskill: Ok, i moved it back, seems there is no better solution in absence of c++ functions for the same purpose.

opencog/cython/PythonSCM.cc
noskill: I put this code back with some refactoring

linas:
I merged this, under the theory that it's better to to just get this in now, and fix the remaining issues later... I'll open new issues for each ...

PR #2406, opened 2019-12-02
Use OpenCogFindPython to find Python dependencies

vsbogd:
Use `OpenCogFindPython` macro to get Python dependencies uniformly.
Complementary PR to fix https://github.com/opencog/cogutil/issues/176 after merging https://github.com/opencog/cogutil/pull/211


linas:
thanks!

PR #2408, opened 2019-12-02
Inf&nan number support

kasimebrahim:
NOTE: commit c8fe151 replaces the existing std::to_string by boost::lexical_cast I had to change this because apparently std::to_string fails when creating an infinity NumberNode on at least some systems, which I couldn't find any pattern on. The problem doesn't rise when tried from the atomspace, but the same code i:e (NumberNode "-inf") would fail from asmoses.

If it is due to build configuration I couldn't find it. If any one has any idea on that or think boost::lexical_cast could rise a problem I didn't see, I will be glad to proceed to fix the configuration error rather than this PR.

### DIFF
--- a/opencog/atoms/core/NumberNode.h
+++ b/opencog/atoms/core/NumberNode.h
@@ -23,6 +23,7 @@
 #ifndef _OPENCOG_NUMBER_NODE_H
 #define _OPENCOG_NUMBER_NODE_H
 
+#include <boost/lexical_cast.hpp>
 #include <opencog/atoms/base/Node.h>
 #include <opencog/atoms/value/FloatValue.h>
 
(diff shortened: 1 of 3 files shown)


opencog/atoms/core/NumberNode.h
```diff
@@ -54,7 +55,7 @@ class NumberNode : public Node
 	// the European comma as a decimal separator blows up the code.
 	static std::string double_to_string(double x)
 	{
-		std::string vs(std::to_string(x));
+		std::string vs = boost::lexical_cast<std::string>(x);
 		std::size_t found = vs.find(',');
```
linas: well, if you are going to do this, then you can remove lines 44 through 63, since lexical_cast ignores the locale, and won't have the problem described in the comments. i.e. just do this:
```
static std::string double_to_string(double x){
    return boost::lexical_cast<std::string>(x); }
```
```

linas: fixed

linas:
Re: boost: some parts of boost are stable and work well, some parts are buggy. It's hard to know which until it's too late :-) For example, `boost:signals2` was a disaster. Currently, `boost:asio` is bug-ridden and a recurring source of breakage/failure. Other parts, that we use in existing code, seem to work OK.

Issue #2409, opened 2019-12-02
Python: type_ctor_atomspace is inconsistent.


linas:
The python bindings are inconsistent with regard to this global:
so: in ` opencog/cython/PythonSCM.cc` we have
```
"from opencog.type_constructors import set_type_ctor_atomspace")
```
but in `tests/scm/scm-python-shared-atomspace.scm`
```
from opencog.utilities import get_type_ctor_atomspace
```

The getter and setter should live in the same module, not in two different ones. That module should probably NOT be `utilities`. Also PLEASE change the name to something not-insane; for example `set_current_atomspace` would work well. This is a follow-on to pull req #2398

linas:
@noskill @vsbogd followup to #2398

vsbogd:
They are in fact in same module: https://github.com/opencog/atomspace/blob/2c99e42c9b7f6141da942853f92ebfb39c1e642f/opencog/cython/opencog/type_constructors.pyx#L14-L21
but Python importing/exporting specifics allows using them from another module which imports `type_constructors`

vsbogd:
Fixed by #2415

linas:
OK, so #2415 doesn't look like a fix to me ... it looks like you've pushed around some kind of hidden bug ... since `utilities` never exported that symbol, how did that code ever work?

Or maybe utilities still exports that symbol, in which case this is not fixed.

linas:
Closing; #2416 marks it deprecated, and #2420 changes the overall idea.  Thanks!

PR #2415, opened 2019-12-03
Fix package import

vsbogd:


vsbogd:
Fix for https://github.com/opencog/atomspace/issues/2409

tests/scm/scm-python-shared-atomspace.scm
```diff
@@ -17,7 +17,7 @@
 ; Define a python func returning a TV
 (python-eval "
 from opencog.atomspace import AtomSpace, TruthValue, types
-from opencog.utilities import get_type_ctor_atomspace
+from opencog.type_constructors import get_type_ctor_atomspace
```
linas: I don't get it ... if `utilities` didn't actually export `get_type_ctor_atomspace`, then is it really needed? How did this code work?

Or does python allow misc random modules to export symbols from other modules?  That seems very non-hygenic.

PR #2416, opened 2019-12-03
split PyEval initialization and setting default atomspace

noskill:
better names for setting and getting default atomspace:

set_default_atomspace and get_default_atomspace, also allow to call initialize_opencog without atomspace

fix for issue https://github.com/opencog/atomspace/issues/2391

### DIFF
--- a/opencog/cython/opencog/utilities.pyx
+++ b/opencog/cython/opencog/utilities.pyx
@@ -1,43 +1,54 @@
 from contextlib import contextmanager
 from opencog.atomspace cimport AtomSpace
 from opencog.atomspace import create_child_atomspace
-from opencog.type_constructors import get_type_ctor_atomspace, set_type_ctor_atomspace
+from opencog.type_constructors import get_default_atomspace, set_default_atomspace
+import warnings
+
 
 # Avoid recursive intialization
 is_initialized = False
 
-def initialize_opencog(AtomSpace atomspace):
 
+def initialize_opencog(AtomSpace atomspace=None):
+    """
+    Set default atomspace(deprecated feature) and
+    create python evaluator singleton object.
+
+    Calling this function should not be needed. Use set_default_atomspace to set default atomspace.
+    Python evaluator will be created on first evaluation.
+    """
+    if atomspace is not None:
+        warnings.warn("setting default atomspace with initialize_opencog is deprecated,\
+                use set_default_atomspace instead", DeprecationWarning)
+        set_default_atomspace(atomspace)
     # Avoid recursive intialization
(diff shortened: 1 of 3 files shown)


opencog/cython/opencog/utilities.pyx
```diff
@@ -1,43 +1,54 @@
...
     if is_initialized:
-        set_type_ctor_atomspace(atomspace)
         return
     is_initialized = True
-
     c_initialize_python()
-    set_type_ctor_atomspace(atomspace)
+
 
 def finalize_opencog():
```
vsbogd: Probably `finalize_opencog` should be marked as deprecated as well?

opencog/cython/opencog/utilities.pyx
```diff
@@ -1,43 +1,54 @@
...
+from opencog.type_constructors import get_default_atomspace, set_default_atomspace
+import warnings
+
 
 # Avoid recursive intialization
 is_initialized = False
 
-def initialize_opencog(AtomSpace atomspace):
 
+def initialize_opencog(AtomSpace atomspace=None):
```
vsbogd: Actually I am not sure if we should deprecate that. Sometimes it is better to make initialization manually to know exact moment when it is initialized instead of initializing it in some undefined moment.

noskill: initialize_opencog is not deprecated, but i want to deprecate passing an atomspace to initialize_opencog

linas: But `initialize_opencog` isn't needed any more: it can be done automatically, at the time that `import opencog` is done. 

The only reason it was exported is because it needed the AtomSpace argument, and the AtomSpace was unknown at the time of `import`. Now that it doesn't need that argument, it can be moved back to the `import`.

linas:
I'm going to merge this, as it looks reasonable, and defer the `initialize_opencog` issue as being a distinct issue.

opencog/cython/opencog/utilities.pyx
linas: The be clear: the call `c_initialize_python()` should be moved to the import -- If I recall correctly, it would be moved to `__init__.py`  and then get rid if the `global is_initialized` flag, its not needed.  All that's left is that `initialize_opencog()` is another deprecated synonym for `set_default_atomspace()`.  Meanwhile, `c_finalize_python()` would get called from the python finalizer (e.g. in `__init__.py`, you'd have some class with a `__del__` method on it, and it would call `c_finalize_python()` That way, when python exits, the class gets deleted, `__del__` gets called.

Issue #1728, opened 2018-06-08
Atom to_string and to_short_string should not end by a new line


ngeiswei:
This takes some control out of printing Atoms, for instance when logging an atom, since logging system itself adds a newline, it ends up with 2 new lines.

ngeiswei:
Of course to address that issue you do not only need to remove "\n" in https://github.com/opencog/atomspace/blob/master/opencog/atoms/base/Link.cc#L96 and so on, you also need to add it back in various places so that an atom does display nicely over multiple lines, so does `oc_to_string(const HandleSeq&)` over multiple lines, etc. You basically need to check all references where `Atom::to_string`, `Atom::to_short_string` and oc_to_string(const Handle&)` are used and fix them accordingly.

navaro1:
Hey, I will try to solve this issue.

linas:
As of about a month ago, `to_short_string()` no longer has a newline on it.   A unit test should be added for this, because the newlinefree version is a requirement for https://github.com/opencog/atomspace-dht

ngeiswei:
@linas, `Atom::to_short_string` of

```
List
  Concept "A"
  Concept "B"
```

returns

```
(ListLink   (Concept "A")  (Concept "B"))
```

that is the indentation before `(Concept "A")` and `(Concept "B")` is still present. I assume it's OK if that extraneous indentation is removed, let me know otherwise.

PR #2418, opened 2019-12-04
Fix Atom::to_string, Value::to_string and Atom::to_short_string

ngeiswei:
Fix issue #1728

1. The last line of a rendered atom or such does not end by a newline. It is up to the user to decide if thou so desires.
2. Remove extraneous whitespaces in `Atom::to_short_string()`.


ngeiswei:
All tests pass, I'm merging as I'm almost sure that the extraneous spaces in `to_short_string` are unneeded.

Issue #2410, opened 2019-12-02
Python temp atomspace usage


linas:
I believe that the python code, after pull rq #2498 no longer respects the temp atomspace.  Specifically, consider the following pseudocode:
```
(Get  (Variable "x")
   (And
       (Present (ListLink (Concept "foo") (Variable "x")))
       (Evaluation (GroundedPredicate "py:test_one")(List (Variable "x")))
       (Evaluation (GroundedPredicate "py:test_two")(List (Variable "x")))
    ))
```
with 
```
def test_one(v):
   thing = add_node(types.Concept, "barleycorn")
   thang = add_link(types.ListNode, thing, v, TruthValue(0.9, 0.75))
   thong = space_server.get(thang)
   return thong.getTruthValue()

def test_two(v):
    return TruthValue (0,1)   # false
```

In this example, the atoms `thing` `thang` and `thong` are supposed to be created in the temp atomspace, not the main atomspace.  The main atomspace is never polluted with these extra atoms. Doubly so, since the second test always fails, so the GetLink should always return the empty set.

That's the intent of the temp atomspace: to avoid polluting the main atomspace with any temporary atoms created during the search.

Before the pull req #2398, the `applier.apply_tv(as, schema.substr(pos), args);` API told python how to find the temp atomspace that will hold `thing`, `thang`, `thong`. Since that API is now gone, its not clear how this can work.

linas:
I think the pseudocode for a working version would be something like this:
```
TruthValuePtr apply_tv(AtomSpace*as, const std::string& func, Handle varargs)
{
    this->eval("from opencog.atomspace import AtomSpace");
    this->eval("from opencog.type_constructors import set_type_ctor_atomspace");
    this->eval("set_type_ctor_atomspace(AtomSpace(" + std::to_string((uint64_t) as) + "))\n");

   etc...
}
```
and likewise for `apply()`

linas:
@noskill @vsbogd follow-on to pull req #2398

vsbogd:
Test `https://github.com/opencog/atomspace/blob/master/tests/scm/scm-python-shared-atomspace.scm` which tests this passes.
See https://circleci.com/gh/opencog/atomspace/4106?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

Test log:
```
69/138 Testing: SCMPythonSharedAtomspace
69/138 Test: SCMPythonSharedAtomspace
[...]

PR #2420, opened 2019-12-05
Python threadlocal atomspace

noskill:
This PR implements thread-local c++ queue for current default atomspace to be used in python or other binding languages.

So now this code in python
```
ConceptNode("test")
```
calls c++ factory which takes atomspace to create ConceptNode from this thread-local queue.
PythonEval pushes atomspace to the queue before calling a callback function, and removes right after calling.

fix for issue https://github.com/opencog/atomspace/issues/2410


opencog/executioncontext/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+
+# The atom_types.h file is written to the build directory
+INCLUDE_DIRECTORIES( ${CMAKE_CURRENT_BINARY_DIR})
+
+
+ADD_LIBRARY (executioncontext
```
linas: I think it would be cleaner to just have this in the cython directory, instead of creating a new directory for it.

I know you're thinking that the scheme bindings could use this, but that's not quite the way scheme works: it uses the concept of a "fluid", which is kind of like thread-local-storage, except that it's its a full-fledged object in it's own right.  The concept of TLS works for C++/python, because in those languages, execution is hard-coded to always be in a thread (with the threads managed by the OS kernel); roughly speaking, this is possible whenever a language has an execution stack that holds transitions from caller to callee (the C++/python stack).  In functional languages, e.g. scheme, you have calls with continuations, so when you "return" or "continue", you may return to a different thread. (its vaguely like setjmp/longjmp, but guaranteed and not dangerous)  Thus, for compatibility, in order to emulate the C++/python-style concepts of threads/TLS, scheme has the concept of "fluids".  That is where the scheme bindings search for and find the "current" atomspace.   

... and that's why this code is unlikely to be useful to anyone, except for cython, which is why I suggest moving it to the cython directory.

tests/cython/bindlink/test_functions.py
```diff
@@ -61,3 +61,16 @@ def stop_go(atom):
...
+def func_one(v):
+    thing = ConceptNode("barleycorn")
+    thang = ListLink(thing, v)
+    thang.tv = TruthValue(0.9, 1)
+    return thang.tv
+
+
+func_two_result = TruthValue (0,1) # false
+def func_two(v):
+    return func_two_result
```
linas: This should be
```
def func_two(v):
    return ListLink(ConceptNode("barleycorn"), v).tv
```
which returns the TV that `func_one` had computed.  This proves that the atoms created by `func_one` are still accessible to `func_two`, i.e. it proves that func_one and func_two are sharing the same temporary atomspace.

tests/cython/bindlink/test_bindlink.py
```diff
@@ -166,5 +166,26 @@ def test_execute_atom_no_return_value(self):
...
+                   EvaluationLink(GroundedPredicateNode( "py: test_functions.func_two"),
+                   ListLink (VariableNode ("x")))))
+        result = execute_atom(self.atomspace, get)
+        self.assertFalse(result.out)
+        self.assertFalse('barleycorn' in str(self.atomspace.get_atoms_by_type(types.ConceptNode)))
+        test_functions.func_two_result = TruthValue(1,1)
+        result = execute_atom(self.atomspace, get)
+        self.assertTrue(result.out)
+        # still should not be in the current namespace
+        self.assertFalse('barleycorn' in str(self.atomspace.get_atoms_by_type(types.ConceptNode)))
```
linas: It would be useful to have a pair of functions:
```
def is_node_in_atomspace(atom_type, str): 
    return true if ... else return false 

def is_link_in_atomspace(atom_type, atom_lsit):
      return true if ... else return false 
```

I have these in scheme, hey are almost never needed, except .. sometimes they are. Typically, to get the truth value on something, but only if that something actually exists. 

in the ideal implementation, these call into C, so that it's fast, instead of trying to search a huge list (in the real world, `get_atom_by_type` might return millions of atoms)

opencog/executioncontext/CMakeLists.txt
noskill: I thought about that difference a bit before make the PR. Python 3.7 now also has something similar to fluids, contextvars..

Anyway If python callback will start a new thread then it won't be able to get atomspace set by PythonEval. I decided it is up to developer of the callback to decide what to do in such case..

tests/cython/bindlink/test_bindlink.py
noskill: I added new methods..

linas: Thanks!  Looks good.

opencog/executioncontext/CMakeLists.txt
linas: That sounds reasonable, I guess.

tests/cython/bindlink/test_functions.py
linas: fixed. OK

vsbogd:
>  In functional languages, e.g. scheme, you have calls with continuations, so when you "return" or "continue", you may return to a different thread. (its vaguely like setjmp/longjmp, but guaranteed and not dangerous)

@linas thank you very much for mentioning this. I didn't think about this from such angle. The original goal was to allow passing the atomspace between any pairs of APIs (from Scheme to Python, from Python to C++ and so on). But taking continuations into account I would say that the idea of the `ThreadLocal` atomspace loses most of its attractiveness. Even Python has `yield` operator which potentially allows creating a generator in one thread and call it from another one thus same generator will use different atomspaces.

As was mentioned in https://github.com/opencog/atomspace/pull/2398#issuecomment-559915526 the global atomspace is more or less a historical accident. And may be the best way is to stimulate Atomese users always using some specific atomspace to create new atoms. C++ and Python APIs can be designed in such a way to make it convenient. Recently I ported a large piece of Python rules in Atomese to C++ and I can share example of how it could be done in C++ code. 

Another thing which could made it more explicit: current atomspace (temporary atomspace for `BindLink` execution case) could be passed as a parameter of `GroundedSchemeNode` function. C++ `GroundedSchemeNode` are already implement in this way, but in Python and in Scheme the atomspace is passed via ThreadLocal and fluid respectively.

opencog/cython/opencog/utilities.pyx
```diff
@@ -52,3 +53,61 @@ def tmp_atomspace():
...
+    cdef vector[cHandle] handle_vector
+    for atom in outgoing:
+        if isinstance(atom, Atom):
+            handle_vector.push_back(deref((<Atom>(atom)).handle))
+        else:
+            raise TypeError("outgoing set should contain atoms, got {0} instead".format(type(atom)))
+    cdef cHandle result
+    result = c_add_link(t, handle_vector)
+    if result == result.UNDEFINED: return None
+    atom = Atom.createAtom(result);
```
vsbogd: `create_python_value_from_c_value` should be used instead, otherwise it returns incorrect Python wrapper

opencog/cython/opencog/utilities.pyx
```diff
@@ -52,3 +53,61 @@ def tmp_atomspace():
...
+
+def add_node(Type t, atom_name, TruthValue tv=None):
+    """
+    Add Node to the atomspace from the current context
+    """
+    cdef string name = atom_name.encode('UTF-8')
+    cdef cHandle result = c_add_node(t, name)
+
+    if result == result.UNDEFINED: return None
+    atom = Atom.createAtom(result);
```
vsbogd: create_python_value_from_c_value should be used instead, otherwise it returns incorrect Python wrapper

vsbogd:
I would also add unit tests on new functionality:
- new functions `get_context_atomspace`, `push_context_atomspace`, `pop_context_atomspace`, `clear_context`;
- in addition to `test_tmp_atomspace` I would add a test to check that when `execute_atom(atomspace, ExecutionOutputLink)` and `execute_atom(atomspace, EvaluationLink)` are called then result of the `ExecutionOutputLink` and `EvaluationLink` is present in `atomspace`.

vsbogd:
I published example of C++ API I used to port Python code here https://github.com/opencog/atomspace/issues/2428

noskill:
EvaluationLink is tested in test_evaluate_atom. ExecutionOutputLink is tested in test_execute_atom and in some other tests.. What was not covered by these tests?

opencog/cython/opencog/utilities.pyx
noskill: create_python_value_from_c_value doesn't accept cHandle

noskill:
sorry, i re-requested review..

vsbogd:
> EvaluationLink is tested in test_evaluate_atom. ExecutionOutputLink is tested in test_execute_atom and in some other tests.. What was not covered by these tests?

I mean checking that result is added into atomspace, not just returned.
`test_execute_atom` places expected `ListLink(ConceptNode("one"), ConceptNode("two"))` into atomspace before executing `BindLink` so it cannot be used to check whether `test_functions.add_link` really creates something in the atomspace. And `test_functions.add_link` actually uses atomspace from the first atom not the global one. 

I would suggest writing the following test:
```Python
def add_new_link():
    return InheritanceLink(ConceptNode("cat"), ConceptNode("animal"))

def test_add_atom_from_grounded_schema_node(self):
    test_as = AtomSpace();
    execute_atom(test_as,
            ExecutionOutputLink(
                GroundedSchemaNode("py: test_functions.add_new_link"),
                ListLink()
            )
        )
    self.assertTrue(test_as.is_link_in_atomspace(types.InheritanceLink,
        [test_as.add_node(types.ConceptNode, "cat"),
        test_as.add_node(types.ConceptNode, "animal")]))
```

And similar for `EvaluationLink`.

PR #2421, opened 2019-12-06
Fix BindLink alpha equivalence

ngeiswei:
And for that complete the canonical order of free variables.

Multiplicity of variables was not enough to disambiguate variable order by semantics, their paths from variable to root is necessary as well.

### DIFF
--- a/opencog/atoms/core/Variables.cc
+++ b/opencog/atoms/core/Variables.cc
@@ -43,43 +43,39 @@
 
 namespace opencog {
 
+/// The VarScraper struct provides the functions to find and sort free
+/// variables in a scope link that does not contain variable
+/// declaration or contains an unordered one (using VariableSet). The
+/// sorting of free variables only depends on the semantics of the
+/// variables, that is their positions in the scope, taking into
+/// account the commutativity property of unordered links.
 struct VarScraper
 {
 	/**
-	 * Mapping between variables and number of occurrences. This allows
-	 * to take into account the semantics of the variables, as to be
-	 * able to sort unordered links like
-	 *
-	 * (And
-	 *   (Inheritance (Variable "$X") (Variable "$Y"))
-	 *   (Inheritance (Variable "$Z") (Variable "$Z")))
-	 *
-	 * by their semantics (Z occurs twice, thus is sorted after X and
-	 * Y) as opposed to solely variable names, which is not robust to
-	 * alpha-conversion.
+	 * Mapping variables to their paths, where a path is a list of
+	 * pairs of atom type and outgoing index, with the particularity
+	 * that when the type is a subtype of unordered link, the index is
+	 * zero, because it is semantically meaningless.
 	 */
-	HandleUCounter _fvc;
+	HandlePathsMap _paths;
 
 	/**
 	 * Find variable declaration in such order that is consistent with
 	 * alpha-equivalence.
 	 */
-	void find_vars(HandleSeq& varseq, HandleSet& varset,
-	               const HandleSeq& outs, bool ordered_link);
+	HandleSeq operator()(const HandleSeq& outgoing, bool ordered_link);
 
 	/**
-	 * Return a mapping between free variables and the number of times
-	 * they occur in the given body.
-	 *
-	 * Such a mapping is then passed to the various ordering functions
-	 * so that semantics is considered prior to variable names.
+	 * Return a mapping between variables (free and non free) and their
+	 * paths. The reason we need the path of non free variables is
+	 * because they are subject to semantic comparison just like the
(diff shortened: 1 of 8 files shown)


opencog/atoms/core/Variables.h
```diff
@@ -38,6 +39,20 @@ namespace opencog
  *  @{
  */
 
+// Struct to build paths between variables and root. A path is a
```
linas: How is this different from `PatternTerm`?  That device was invented to record the unique path from some root to some atom under that root.

linas: I mean this is obviously much more lightweight than that; but I dislike having two mechanisms that are similar-but-different.

opencog/atoms/core/Variables.cc
```diff
@@ -283,7 +293,12 @@ bool VarScraper::is_variable(const Handle& h)
 
 bool VarScraper::is_ordered_link(const Handle& h)
 {
-	return nameserver().isA(h->get_type(), ORDERED_LINK);
+	return is_ordered_type(h->get_type());
+}
+
+bool VarScraper::is_ordered_type(Type t)
+{
+	return nameserver().isA(t, ORDERED_LINK);
```
linas: This is guaranteed to lead to bugs.  A better test would be to look for "not unordered" links.   There are rare cases where a user can say just "Link", and just-plain Link.s are ... ordered; whereas the above code assumes the opposite.  (Yes, I was already bitten by this in the pattern matcher)

opencog/atoms/core/Variables.cc
```diff
@@ -238,43 +250,41 @@ bool VarScraper::less_than(const Handle& lh, const Handle& rh,
...
-bool VarScraper::less_than_unordered(const HandleSeq& lhs, const HandleSeq& rhs,
-                                     const Context& ctx) const
+bool VarScraper::less_than_unordered_outgoing(const HandleSeq& lhs,
+                                              const HandleSeq& rhs) const
 {
-	return less_than_ordered(sorted(lhs, ctx), sorted(rhs, ctx), ctx);
+	return less_than_ordered_outgoing(sorted(lhs), sorted(rhs));
 }
 
+// TODO: maybe support context
```
linas: How is Context different from `PatternTerm`? So, `PatternTerm` already has quoting support...(added by you)

opencog/atoms/core/Variables.cc
```diff
@@ -238,43 +250,41 @@ bool VarScraper::less_than(const Handle& lh, const Handle& rh,
...
+	// Still a tie? These 2 variables are semantically equivalent. Sort
+	// by atom order then.
 	return lh < rh;
 }
 
-bool VarScraper::less_than_ordered(const HandleSeq& lhs, const HandleSeq& rhs,
-                                   const Context& ctx) const
+bool VarScraper::less_than_ordered_outgoing(const HandleSeq& lhs,
+                                            const HandleSeq& rhs) const
 {
```
linas: Now that `const Context& ctx` is gone, this seems to be a confusing duplicate of `handle_seq_less`  in `Handle.h` (Which, BTW, seems mis-declared, I think it should be declared as `namespace std { template <> struct less<opencog::HandleSeq>; }` or something like that ...

opencog/atoms/core/Variables.h
ngeiswei: Yes, it's very lightweight and abstracts away unordered outgoing indices.

I don't like redundancy either, but I think it's justified for now. I've added a note in that regard.

opencog/atoms/core/Variables.cc
ngeiswei: Thanks!

opencog/atoms/core/Variables.cc
ngeiswei: As it turns out context, quotations, etc, are not needed here, the only thing that matters is the semantic differences between variables, and this is entirely captured by `Path`.

opencog/atoms/core/Variables.cc
ngeiswei: See

https://github.com/opencog/atomspace/blob/e97da6f8d4d7b5a7ef921d99e11942895f1b208c/opencog/atoms/core/Variables.cc#L261-L264

to understand the fundamental difference.

linas: Yes, sure, but I wasn't talking about that function, I was pointing at `less_than_ordered_outgoing()`

linas: Ohhh, because its recursively embedded deep in there. OK.

linas:
OK, I marked them all as resolved.

Issue #2388, opened 2019-11-23
Multiple UnorderedUTest failures


linas:
At this time, there ~~nine~~ ~~six~~ ~~five~~ ~~two~~ ~~one~~ zero failing subtests in the UnorderedUTest:

* ~~`void xtest_big_jswiergo(void);`~~  fixed in #2401
* ~~`void xtest_odo_below(void);`~~  fixed in #2399
* ~~`void xtest_odo_distinct(void);`~~   fixed in #2403
* ~~`void xtest_odo_indistinct(void);`~~ fixed in #2425
* ~~`void xtest_odo_layer(void);`~~  fixed in #2403
* ~~`void xtest_odo_couple(void);`~~  fixed in #2397
* ~~`void xtest_odo_equ_pred(void);`~~ fixed in #2414
* ~~`void xtest_odo_equal(void);`~~ fixed in #2397
* ~~`void xtest_odo_couplayer(void);`~~ fixed in #2403

* ~~Also `SudokuUTest` is broken by #2401~~    fixed in #2424
* ~~See also #2396 for intermittent failures.~~ - fixed in #2401

All of these need to be fixed.  The first one is from issue #148

linas:
Closing. #2425 fixes the last remaining unit test. This was a huge amount of work, and much harder than I had hoped.  If there's an easier way, its hard to guess what that is.

linas:
(I thought this would take a day, two tops. It took three weeks (the actual work started with issue #2346 which lead to this issue.)

ngeiswei:
Same thing for issue #2329 (which ultimately led to that issue) I though it would take a day, it took over a month! The final solution is quite simple, but it took me a while to adopt the right approach and address (with your great help) all related bugs.

PR #2424, opened 2019-12-07
Fix clashing use of the permutation stepper stack.

linas:
There are two stacks for this, not one; they cannot be merged into one.  Phew.


linas:
See #2388 for general status

PR #2425, opened 2019-12-07
Implement a recursive odometer for unordered links

linas:
This fixes the last broken unordered link unit test. Unlike the
earlier attempt at an odometer, this one is properly recursive.


linas:
See #2388 for general status.

PR #2426, opened 2019-12-09
Remove `using opencog` from header.

vsbogd:
`using namespace opencog` in header exports all `opencog` namespace into global namespace.


linas:
thanks!

PR #2427, opened 2019-12-09
Remove out_of_range exceptions

vsbogd:
Replace catching exceptions by conditional statements as processing exceptions is slow operation.
This change improves forward chaining speed on ~10%.

### DIFF
--- a/opencog/atoms/core/Variables.cc
+++ b/opencog/atoms/core/Variables.cc
@@ -1272,23 +1272,22 @@ void Variables::extend(const Variables& vset)
 {
 	for (const Handle& h : vset.varseq)
 	{
-		try
+		auto index_it = index.find(h);
+		if (index_it != index.end())
 		{
-			index.at(h);
-
 			// Merge the two typemaps, if needed.
-			try
+			auto typemap_it = vset._simple_typemap.find(h);
+			if (typemap_it != vset._simple_typemap.end())
 			{
-				const TypeSet& tms = vset._simple_typemap.at(h);
+				const TypeSet& tms = typemap_it->second;
 				TypeSet mytypes =
 					type_intersection(_simple_typemap[h], tms);
 				_simple_typemap.erase(h);	 // is it safe to erase if
                                              // h not in already?
 				_simple_typemap.insert({h, mytypes});
 			}
-			catch(const std::out_of_range&) {}
 		}
-		catch(const std::out_of_range&)
+		else
 		{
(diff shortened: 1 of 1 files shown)


opencog/atoms/core/Variables.cc
```diff
@@ -1272,36 +1272,34 @@ void Variables::extend(const Variables& vset)
...
+				const TypeSet& tms = typemap_it->second;
 				TypeSet mytypes =
 					type_intersection(_simple_typemap[h], tms);
 				_simple_typemap.erase(h);	 // is it safe to erase if
                                              // h not in already?
 				_simple_typemap.insert({h, mytypes});
 			}
-			catch(const std::out_of_range&) {}
 		}
-		catch(const std::out_of_range&)
```
linas: well, but these changes make the code run slower.  In general, `try..catch` is always faster than `if(error) {}`

The point is that you always pay the code of `if(error)` even if there is no error -- and the cose is not just the if test, but a possible icache miss and/or itlb miss, which can be expensive. By contrast, `try..catch` always costs zero extra cycles/cache-misses, unless there is an error. Its faster, smalllerr, cheaper, easier and more fool-proof -- this is why exceptions were added to C++...

linas:
I'm confused by the comment that "this improves forward chaining speed on ~10%." This makes no sense at all ... how is that even possible? Are there lots and lots of errors? 

I'm assuming that in "normal" code, the try never-ever throws.  If it does throw, then of course, it will be expensive.  Is there some other bug somewhere else causing it to throw a lot?

linas:
Oh, OK, I had to read the original code; it was just insane.

vsbogd:
I found exceptions are thrown very frequently when I was looking at the ForwardChaining profile. Not sure if it is an unexpected behavior in the higher level of the ForwardChainer logic. At least one of the `catch` blocks is used like a condition statement, so I concluded it is a proper execution path. If some of the exceptions are really unexpected I can raise PR to re-throw them.

vsbogd:
I added benchmarks on this method https://github.com/opencog/benchmark/pull/25 All three catches are triggered in legal usage scenarios.

linas:
Yes, the original code, as written, was just ... crazy.  Exceptions are very expensive.

linas:
I mean, to be clear: thank you for fixing this.  The original code should not have used exceptions to handle what was an ordinary if-then-else operation, where both branches of the if-then-else occur frequently.

Issue #2428, opened 2019-12-09
Example of more compact C++ syntax to instantiate complex graphs in atomese


vsbogd:
I used the following API to port Python atomese code to C++. My goal was to make as small code changes as possible but at the same time get rid of using an implicit global atomspace.

This example of a C++ API implementation allows building complex atomese constructions mentioning the atomspace instance only once. `type_constructor` namespace members keep atom details until an atomspace is provided. Then each class recursively adds its subatoms and itself into atomspace provided. It is also possible use builder-like syntax instead of constructors with many parameters similarly to `tv()` method which sets `ThruthValue`.

Implementation details:
- I used namespace to eliminate conflict with atoms constructor from `opencog` namespace in this example; in real implementation either type constructors or atom constructors should have different names;
- I used `put_into()` method to not touch `AtomSpace` implementation, but it would be more simple to add `AtomSpace::add(type_constructors::Atom)` method.

Type constructors:
```c++
#ifndef TYPE_CONSTRUCTORS_H
#define TYPE_CONSTRUCTORS_H

#include <string>
#include <vector>

#include <opencog/atoms/base/Atom.h>
#include <opencog/atoms/truthvalue/TruthValue.h>
#include <opencog/atomspace/AtomSpace.h>

namespace type_constructors {

class Atom {
	private:
		opencog::TruthValuePtr truthValue;
	protected:
		opencog::Handle fillTruthValue(opencog::Handle handle) const {
			if (this->truthValue) {
				handle->setTruthValue(this->truthValue);
			}
			return handle;
		}
	public:
		virtual opencog::Handle put_into(opencog::AtomSpace& as) const = 0;
		Atom& tv(opencog::TruthValuePtr const& truthValue) {
			this->truthValue = truthValue;
			return *this;
		}
};

class Node : public Atom {
	protected:
		opencog::Type type;
		std::string name;

	public:
		Node(opencog::Type type, std::string const& name) : type(type), name(name) {}
		virtual opencog::Handle put_into(opencog::AtomSpace& as) const {
			return fillTruthValue(as.add_node(type, name));
		}
};

class Link : public Atom {
	private:
		opencog::Type type;
		std::vector<Atom const*> atoms;

	public:
		Link(opencog::Type type, std::vector<Atom const*> const& atoms) : type(type), atoms(atoms) {}
		virtual opencog::Handle put_into(opencog::AtomSpace& as) const {
			opencog::HandleSeq handles;
			for (auto const& atom : atoms) {
[...]

PR #2433, opened 2019-12-10
Add automated C++ function generation.

linas:
Proof of concept for #2428

This probably(?) should not be merged without unit tests and additional cleanup.  Your mileage may vary.


linas:
Things that are missing/broken in this pull req:

* untested.
* LINK_CTOR only defines functions for arity-2 links, it needs to also define arity 0,1,3,4,5 and a generic `HandleSeq` version.
* `CMAKE_BINARY_DIR` is broken .. it seems to be the source dir, instead. I don't know why. Specifically this line: `SET(CNAMES_FILE ${CMAKE_BINARY_DIR}/atom_names.h)` is broken.

linas:
@vsbogd coul;d I convince you to finish the work on this?  Tweak it as needed; this was only a proof-of-concept, it needs to be finished and polished

vsbogd:
@linas , np, I will try to finish it .

vsbogd:
I have started here https://github.com/opencog/atomspace/compare/master...vsbogd:type-constructors

linas:
Thanks @vsbogd I've merged your changes; if there's more, start with a new issue/pull-req.

vsbogd:
ok, thanks!

PR #2434, opened 2019-12-10
Cleanup hash printing

linas:
It's slightly more efficient to use stringstream,

### DIFF
--- a/opencog/atoms/base/Atom.cc
+++ b/opencog/atoms/base/Atom.cc
@@ -463,9 +463,12 @@ size_t Atom::getIncomingSetSizeByType(Type type) const
 
 std::string Atom::id_to_string() const
 {
-    return
-        std::string("[") + std::to_string(get_hash()) + "]" +
-        std::string("[") + (_atom_space? std::to_string(_atom_space->_atom_table.get_uuid()) : std::string("-1")) + "]";
+    std::stringstream ss;
+    ss << "[" << std::hex << get_hash() << "][";
+    if (_atom_space) ss << _atom_space->_atom_table.get_uuid();
+    else ss << "-1";
+    ss << "]";
+    return ss.str();
 }
 
 std::string oc_to_string(const IncomingSet& iset, const std::string& indent)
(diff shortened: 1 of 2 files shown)


opencog/atoms/base/Atom.cc
```diff
@@ -463,9 +463,12 @@ size_t Atom::getIncomingSetSizeByType(Type type) const
...
 {
-    return
-        std::string("[") + std::to_string(get_hash()) + "]" +
-        std::string("[") + (_atom_space? std::to_string(_atom_space->_atom_table.get_uuid()) : std::string("-1")) + "]";
+    std::stringstream ss;
+    ss << "[" << std::hex << get_hash() << "][";
+    if (_atom_space) ss << _atom_space->_atom_table.get_uuid();
+    else ss << "-1";
+    ss << "]";
+    return ss.str();
```
ngeiswei: Interesting, what is making it faster? `std::hex`? Or the fact that `stringstream` is faster than `string` (which I find surprising).

linas:
Oh dear, well, I did not actually check. But the old standard wisdom used to be that `string` always allocates contiguous memory, and has to reallocate a larger buffer every time you append.  Whereas `stringstream` simply created a linked-list as each new string chunk came in, to track that chunk,  and did not allocate a single contiguous buffer until you called the `str()` method, at which point the concatenation was actually performed. That was why it was supposed to be faster.  However, my understanding is dated, and this may no longer be the case.

What I should have said in the commit message is that the I think the hex numbers are more readable than the ints, and I figured no one else cared.  So it was mostly about using hex.

ngeiswei:
I like `std::hex` better too, as anything that reminds me my Atari ST days, might also be faster to convert to string and more compact.

linas:
FWIW, I just reviewed the docs for rvalue ctors (available since c++11) and get the impression that `std::string s = "[" + foo + "]"; ` would use one big "modifiable rvalue" and thus avoid copies. at lest according to one theory. And thus could be as fast (or faster??) than stringstream.

Issue #2423, opened 2019-12-06
Introduce Value::is_unordered_link


ngeiswei:
Based on that comment

https://github.com/opencog/atomspace/pull/2421#discussion_r354794102

I think it would be wise to introduce a

```C++
bool Value::is_unordered_link() const;
```

method, and use it instead of instances such as

https://github.com/opencog/atomspace/blob/2bca608896ff9c93492754842eb1dd8aad8b2380/opencog/atoms/core/FreeLink.cc#L60

there seems to be about a dozen instances of such (including one in the ure repo), which I think is large enough to justify it.

Let me know what you think.

linas:
Sure I think its harmless. Maybe even a tiny performance boost.

ngeiswei:
Why would it bring a (tiny) performance boost?

ngeiswei:
Oh, do you mean to implement it as

```c++
virtual bool UnorderedLink::is_unordered_link() const
{
	return true;
}
```

?

ngeiswei:
Is that what you have in mind

https://github.com/ngeiswei/atomspace/commit/1563b2704c28b84e2f7318091a32884e954f76e3

?

linas:
yes!

linas:
The `if(nameserver().isA(get_type()..` is mostly a lookup in the atom-types array.  I think the compiler inlines most of the function calls, so what's left is just that array lookup, and a possible dcache miss or dtlb miss from the array lookup. 

By contrast, the `virtual bool iswhatever()` is a function pointer call through a vtable that cannot be inlined, so its probably more instructions. But maybe less dcache misses, becuase no more array lookup. But then, maybe icache misses. But the vtable is smaller than the atom-types table, so .... 

I dunno its hard to guess what the compiler might do for these two cases, and its hard to guess how many cache misses might result.  Its a toss-up.

linas:
Actually, if the compiler is really smart, it can just place the return code from `is_unordere_link()` in the vtable, and then just look that up, without any function calls at all!  This would require only a small amount of vtable magic, so its possible that modern compilers already do this ...  I mean, if they can already do link-time-optimizations, this is surely an easier trick ...

linas:
BTW, if you are looking for performance optimizations, I just opened #2429 describing a deeply-rooted one that could make a fairly large difference.

vsbogd:
[...]

PR #2442, opened 2019-12-13
Implement Value::is_unordered_link

ngeiswei:
Addresses https://github.com/opencog/atomspace/issues/2423

### DIFF
--- a/opencog/atoms/core/FreeLink.cc
+++ b/opencog/atoms/core/FreeLink.cc
@@ -57,7 +57,7 @@ FreeLink::FreeLink(const Link& l)
 
 void FreeLink::unorder(void)
 {
-	if (not nameserver().isA(get_type(), UNORDERED_LINK)) return;
+	if (not is_unordered_link()) return;
 
 	// Place into arbitrary, but deterministic order.
 	// We have to do this here,  because some links,
@@ -72,7 +72,7 @@ void FreeLink::unorder(void)
 
 void FreeLink::init(void)
 {
-	bool ordered = not nameserver().isA(get_type(), UNORDERED_LINK);
+	bool ordered = not is_unordered_link();
 	_vars.find_variables(_outgoing, ordered);
 }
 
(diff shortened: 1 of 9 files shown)


ngeiswei:
The following unit test

https://github.com/opencog/atomspace/pull/2442/commits/efedf3f45d20054cd493c1537a5a39877cad0734

fails.

As far as I can tell, the factory for the `UnorderedLink` class is never called. Instead the factory for `EvaluationLink` is called because `EqualLink` inherits `VirtualLink` which (oddly) inherits `EvaluationLink`.

So `EqualLink` has multiple inheritances, `UnorderedLink` and `EvaluationLink`, aren't multiple factories supposed to be called?

ngeiswei:
I can confirm that the `UnorderedLink` ctor never gets called.

Looking `ClassServer::splice` it's clear `ClassServer::_atomFactory` only associates one factory per type. That seems like a deficiency of our factory system. Not sure where to go from there. Should we auto generate multi-inherited classes (such as `EqualLink`) when needed, as to "simulate" multi-factory?

linas:
Change to `VIRTUAL_LINK <- EVALUATABLE_LINK`, I think this would solve the problem.

Re; "Multiple factories", I want to avoid  c++ multiple inheritance .... at all costs, even. Down that path lies madness.   The Atom multiple inheritance is used primarily (only) to firm up the expected input and output types of the atoms, to firm up run-time type checking.

ngeiswei:
`VIRTUAL_LINK <- EVALUATABLE_LINK` works!

If we're dedicated to avoid multi-factory, then I can let the implementation of `is_unordered_link` as is it, otherwise, I would rather replace it with

```c++
virtual bool is_unordered_link() const
{
    return nameserver().isA(get_type(), UNORDERED_LINK);
}
```

which would be more resilient to factory f-up.

opencog/atoms/core/FreeLink.cc
```diff
@@ -57,7 +57,7 @@ FreeLink::FreeLink(const Link& l)
 
 void FreeLink::unorder(void)
 {
-	if (not nameserver().isA(get_type(), UNORDERED_LINK)) return;
+	if (not is_unordered_link()) return;
```
linas: !!?? But FreeLinks are always ordered! how dd we end up with this?

PR #2443, opened 2019-12-13
Assume FreeLinks are always ordered links.

linas:
This reverts commit e3c26feac825de54db8461484df446583a49dd77
which added this code, with minimal explanation as to what the
actual problem was.

The current problem is that the C++ code comment mentions
EqualLink and IdenticalLink, both of which *are* unordered,
but neither of which are FreeLinks.  Grepping through the
atom type definitions, I found no free, unordered link types.
(for the spacetime, attention and agi-bio).


linas:
Merging. I think this bug originally fixed some problem with a different
type hierarchy, and an earlier version of the type factory. That problem
would seem to be gone, now.

PR #2444, opened 2019-12-15
Simplify the complexity of the install path.

linas:
There was no apparent reason for the complexity; it was just
obfuscation of what's actually happening.


linas:
In fact this can be further simplified, but this should be enough for now.

PR #2447, opened 2019-12-17
Dependency fixes

linas:
Attempt to fix build-order dependency problem opencog/opencog-nix#54

This also cleans up header-file include dependencies


Radivarig:
Tried with `73ac503` commit but still the same output

PR #2449, opened 2019-12-18
call PythonEval from different threads

noskill:
add one more test to ensure PythonEval works correctly from different threads

### DIFF
--- a/tests/cython/PythonEvalUTest.cxxtest
+++ b/tests/cython/PythonEvalUTest.cxxtest
@@ -266,6 +266,42 @@ public:
         TS_ASSERT(true);
     }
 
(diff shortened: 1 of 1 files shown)


tests/cython/PythonEvalUTest.cxxtest
```diff
@@ -266,6 +266,42 @@ public:
         TS_ASSERT(true);
     }
 
+    void testCodeBlockWithNewlineThreaded()
+    {
+        PythonEval::create_singleton_instance();
+        PythonEval* python = &PythonEval::instance();
+	std::vector<std::thread> threads;
```
linas: indentation

PR #2450, opened 2019-12-18
Attempt to fix parallel-make dependency problems.

linas:
This is for opencog/opencog-nix#54

This time, the fix should really, really work, because the header
file will not show up until after it has been completely written.


linas:
See https://stackoverflow.com/questions/59420384/cmake-how-do-i-declare-a-dependency-on-a-generated-header-file for details

linas:
OK, I'm getting confused. Not only does this not fix things, but also #2448 didn't do the trick, and also #2447 didn't work, and in the meanwhile  #2420 showed up with incorrect build dependencies.

Is it possible that #2420 is at the root cause of the issue?

PR #2453, opened 2019-12-26
Localize use of PatternMatchEngine

linas:
Due to historical design reasons, the PatternMatchEngine had invasively invaded
too many corners of the code. As a result, efforts to parallelize the pattern matcher,
such as #2316 got complicated and off-track. This pull req is part one of a series to
isolate it down to one small area, which will make parallelization much easier to
accomplish.


linas:
Merging; this needs opencog/opencog#3625 to pass all unit tests.

linas:
Ugh. opencog actually requires #2455 to pass...

PR #2457, opened 2019-12-29
Mass removal of copy constructors

linas:
They seem to not be needed; I don't think they were ever needed.

This lays some groundwork for a later pull req that implements move constructors.


linas:
I' going to merge this now, because I think that it's unlikely that there's any 3rd-party code that will be affected. e.g. code that @vsbogd  might be running ...

PR #2458, opened 2019-12-29
Use move constructors for links and nodes

linas:
This should provide some minor performance improvements, but is potentially quite disruptive, depending on how you've built your system.  Requires the corresponding pull reqs:  opencog/attention#10 and opencog/opencog#3627


linas:
@vsbogd this might be disruptive to your code, so FYI.

linas:
@vsbogd Please take a look, and let me know if you have any objections to merging this.

vsbogd:
I would keep the possibility to call constructors with both lvalues and rvalues. If constructor allows only rvalues then caller needs manually copying lvalues to pass them as parameters. If caller doesn't need lvalue he can use `std::move` to pass it but if value is needed then copying is necessary.

Simple way to allow passing both lvalue and rvalue is replacing references in constructors by values and using `std::move` inside constructor. Then compiler will move value when rvalue is passed and copy it when lvalue is passed to constructor.  See example of class `C` implementation below:

```C++
#include <iostream>
#include <string>
#include <utility>

struct A {
	std::string str;

	A(std::string&& str) : str(std::move(str)) {
		std::cout << "A moving" << std::endl;
	}

};

struct B {

	int i;

	B(int i) : i(i) {
		std::cout << "B new" << std::endl;
	}

	B(B&& b) {
		std::cout << "B moving" << std::endl;
	}

	B(B const& b) {
		std::cout << "B copying" << std::endl;
	}

};

struct C {
	B b;

	C(B b) : b(std::move(b)) {}
};

int main()
{
	std::string str = "test";

	std::cout << "passing rvalue to A" << std::endl;
	// Next line doesn't compile
	// A a(str);
	// Next line doesn't compile as well
	// A a(std::string(str));
	A* a = new A(std::string(str));
	std::cout << "a.str = " << a->str << std::endl;

	std::cout << "passing lvalue to A" << std::endl;
	A aa(std::move(str));
	std::cout << "str = " << str << std::endl;
	std::cout << "aa.str = " << aa.str << std::endl;

	std::cout << "passing rvalue to C" << std::endl;
	C c(B(0));

	std::cout << "passing lvalue to C" << std::endl;
	B b(1);
	C cc(b);

	std::cout << "passing rvalue to C" << std::endl;
	C ccc(std::move(b));

	return 0;
}
```

linas:
OK, I'm redoing parts of this in slightly more user-friendly pieces-parts pull reqs.

vsbogd:
Output of the program above:
```
passing rvalue to A
A moving
a.str = test
passing lvalue to A
A moving
str = 
aa.str = test
passing rvalue to C
B new
B moving
passing lvalue to C
B new
B copying
B moving
passing rvalue to C
B moving
B moving
```

vsbogd:
I read this article before making experiments: http://cpptruths.blogspot.com/2012/03/rvalue-references-in-constructor-when.html

linas:
Yeah, that's actually a stronger design. Thanks!

PR #2466, opened 2020-01-07
Remove decade-old cruft.

linas:
Concepts like the AtomPtr date back to when it actually was Atom*.
This hasn't applied in a very long time, now, and mostly just added
complexity and  overhead to operations. Just remove it. Also remove
the `_atom_store` table; it just duplicated information already available.
Also remove the size-caches; no one except the unit tests use them.
Also fix bug in the size-computation that no one noticed before.


linas:
BTW, this improves addNode performance by 50% and addLink performance by 18% and remove performance by 10% ... which serves as an example of what can be accomplished by removing old cruft.

PR #2469, opened 2020-01-08
Use Handles, not LinkPtr for the incoming set

linas:
The  goal here is to save CPU overhead associated with casting
between LinkPtr and Handle. Based on actual measurements, it
seems that `getIncomingSet`is now 10% faster, which is not as 
dramatic as expected. Both `addNode` and `addLink` improve very
slightly (as this is where some of the casting occurs.)

### DIFF
--- a/opencog/atoms/base/Atom.cc
+++ b/opencog/atoms/base/Atom.cc
@@ -61,7 +61,7 @@ namespace std {
 opencog::Type
 hash<opencog::WinkPtr>::operator()(const opencog::WinkPtr& w) const noexcept
 {
-    opencog::LinkPtr h(w.lock());
+    opencog::Handle h(w.lock());
     if (nullptr == h) return 0;
     return h->get_type();
 }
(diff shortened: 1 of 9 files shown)


linas:
I'm going to merge this, despite miner failing, and provide a parallel patch for miner shortly.

linas:
This needs opencog/miner#23 and https://github.com/opencog/opencog/commit/e4e6bbdb4fd8016b67a70d6acb6dece4ecd792b0 and https://github.com/opencog/attention/commit/1618320c45432803077c5af1bc9542e742e7635f

ngeiswei:
Great change. Thanks for taking care of the miner.

linas:
Welcome. Please note also #2464 which improves pattern-search performance by a lot (10%-30% depending on your scenario) and also #2466 which improves atomspace insert-remove performance by 10% to 60% depending on use-case.  Since they came before this one, so the cumulative performance impact should be quite noticable.  Although I guess YMMV, because benchmarks are not representative of real life...

PR #2470, opened 2020-01-11
Cache clauses that are repeatedly grounded.

linas:
The gene annotation code (https://github.com/MOZI-AI/annotation-scheme/)
contains patterns that have endpoints that are repeatedly searched.  For
those cases, performance is improved by as much as 25% by caching the
repeated search results. 

The code here does come at a cost; other searches are penalized by a few 
percent. :-(

There's no unit test, because I cannot figure out how to unit test, as this does
not alter search results.

### DIFF
--- a/opencog/atoms/core/FindUtils.cc
+++ b/opencog/atoms/core/FindUtils.cc
@@ -329,6 +329,17 @@ unsigned int num_unquoted_unscoped_in_tree(const Handle& tree,
 	return count;
 }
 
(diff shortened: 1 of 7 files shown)


linas:
See https://github.com/MOZI-AI/annotation-scheme/issues/98 for details

linas:
Merging. Although, in retrospect, I am unhappy about the overhead this adds to regular searches.

PR #2473, opened 2020-01-14
Don't compile debug prints.

linas:
Disabling the fine-level printing improves performance by 1% or 2%.
Since fine-level printing is only needed during debugging (which is
rare), making this a compile-time option seems reasonable to me.

This is based on the query-link benchmark, am retesting other benches.

### DIFF
--- a/opencog/query/DefaultPatternMatchCB.cc
+++ b/opencog/query/DefaultPatternMatchCB.cc
@@ -33,7 +33,7 @@
 
 using namespace opencog;
 
-#define DEBUG 1
+// #define DEBUG 1
 #ifdef DEBUG
 #define DO_LOG(STUFF) STUFF
 #else
--- a/opencog/query/InitiateSearchCB.cc
+++ b/opencog/query/InitiateSearchCB.cc
@@ -39,7 +39,7 @@
 
 using namespace opencog;
 
-#define DEBUG 1
+// #define DEBUG 1
 #ifdef DEBUG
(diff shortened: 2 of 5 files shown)


vsbogd:
I always use release builds and rebuilding atomspace to get logs is not convenient from this perspective. Anyway it seems strange that removing code which is almost always should check value of variable (`FINE <= logLevel`) and jump to the next line eats 1% of execution time. May be we need to investigate it further and improve instead.

For example:
- we can check that methods like `Logger::is_fine_enabled()` are inlined
- check that calling `fine()` without `LAZY_LOG` is fast enough if not it can be replaced by macros with calling `is_fine_enabled()` first;
- replace level comparision in `is_fine_enabled` by checking flag.

linas:
Even the simplest pattern searches generate tens of thousands of debug prints ... when is the last time you read through these? During normal ops, you'll generate gigabyte-sized logfiles in 10 minutes. So I don't really see how logging fine-grained debug messages are ever useful, unless you are debugging that component...

That they take up a few percent is not surprising: I was surprised only that they took up so little; I was thinking they might eat 10% or more. My goal is to get the pattern matcher to run fast, and that means to minimize the CPU time between each branch-point. (there's a log message placed at each branch-point)

ngeiswei:
I don't mind removing some or most of these, BTW, I rarely use them.

linas:
I'm gonna merge, this is not hard to re-enable if needed.

Issue #2475, opened 2020-01-17
Add a type attribute to get the type of an atom in python binding


Habush:
Sometimes, we might want to inspect the type of atom at runtime in a python code. However, currently, there is no way to get type of a specific atom using an attribute such as `atom.type`.

Habush:
There is already a `type_name` attribute that can be used defined in the [Value class](https://github.com/opencog/atomspace/blob/97a2239d86ceb6e25d34a10e3900d6d51136d634/opencog/cython/opencog/value.pyx#L17). Closing this issue and #2476

PR #2476, opened 2020-01-17
Add a type attribute to atoms

Habush:
This is a fix for issue #2475


vsbogd:
Please check why does it break the build.

opencog/cython/opencog/atom.pyx
```diff
@@ -39,6 +39,15 @@ cdef class Atom(Value):
                 else:
                     self._name = ""
             return self._name
+    property type:
```
vsbogd: In C++ code the `type` is attribute of `Value` class, it is less surprising if this property is added into Python `Value` class as well.

Habush: Hi, please see my second comment to the issue

PR #2483, opened 2020-01-26
Allow multiple rewrites for a given search.

linas:
There is no particular reason for a rewrite to produce only one result, and considerable performance gains are possible by allowing multiple rewrites at once.  So remove that artificial restriction. (This is a rather simple change.)


linas:
Closing, use #2484 instead.

Issue #2490, opened 2020-01-31
Explicit representation of the bottom type


ngeiswei:
`Variables::extend` applied on incompatible types, such as

```scheme
(TypedVariable
  (Variable "$X")
  (Type "ConceptNode"))
```

and

```scheme
(TypedVariable
  (Variable "$X")
  (Type "PredicateNode"))
```

must return a bottom type.

As of today the code would return

```scheme
(TypedVariable
  (Variable "$X")
  (TypeChoice))
```

which is actually conceptually correct, because an empty disjunction is false (by virtue of being the neutral element), thus corresponds to the bottom type.

However, the following pattern matcher query

```scheme
(Inheritance (Concept "A") (Concept "A"))
(define X (Variable "$X"))
(cog-execute! (Get (TypedVariable X (TypeChoice)) (Inheritance X X)))
```

returns

```scheme
(SetLink
   (ConceptNode "A")
)
```

which is inconsistent with `(TypeChoice)` being the bottom type.

Also, as it turns out we already have a bottom type

https://github.com/opencog/atomspace/blob/f53b99a9544dedebfaab06e135d90d12f9d719b6/opencog/atoms/atom_types/atom_types.script#L15-L18

however, due to not inheriting from `NODE` or `LINK` (or even `ATOM`) it has no scheme binding.

So the questions are:

1. How to properly handle and represent the bottom type?
2. What role should `(Typechoice)` have and how to make it consistent with the pattern matcher?

ngeiswei:
My suggestion would be

1. Introduce

```
Top
``

and

```
Bottom
```

which would be the type equivalents of

```
TrueLink
```

and

```
FalseLink
```

2. Fix the pattern matcher so that `(TypeChoice)` is treated as `(Bottom)`.

**Comment by linas** [comment]
I like the idea of Bottom Type, this seems to be consistent, and sticking it into `Variables::extend` and the pattern matcher seems reasonable. 

Just to be clear: For this:
```
(cog-execute! (Get (TypedVariable X (TypeChoice)) (Inheritance X X)))
```
and also for this:
```
(cog-execute! (Get (TypedVariable X (TypeChoice 'NoType)) (Inheritance X X)))
```
the GetLink can be pre-compiled into an always-fail condition. Trivially-fail. Some flag somewhere (maybe in the c++ `class QueryLink`, that aborts the search before it even starts, because it can never be satisfied. 

[...]

PR #2494, opened 2020-02-05
Impl bottom type

kasimebrahim:
as per https://github.com/opencog/atomspace/issues/2490

Note that I didn't introduce a Bottom, I used an empty TypeChoice for now since it simply fixes the PatternMatcher.

### DIFF
--- a/opencog/atoms/atom_types/atom_types.script
+++ b/opencog/atoms/atom_types/atom_types.script
@@ -438,6 +438,8 @@ TYPE_CO_INH_NODE <- TYPE_NODE
 DEFINED_TYPE_NODE <- TYPE_NODE
 
 // Used for type polymorphism.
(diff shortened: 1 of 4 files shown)


opencog/atoms/atom_types/atom_types.script
```diff
@@ -438,6 +438,7 @@ TYPE_CO_INH_NODE <- TYPE_NODE
 DEFINED_TYPE_NODE <- TYPE_NODE
 
 // Used for type polymorphism.
+// Note that the empty type choice corresponds to the bottom type.
```
ngeiswei: It looks good to me @kasimebrahim , could you just add in the comment a link to that wiki page https://wiki.opencog.org/w/TypeChoice#Bottom_type that justifies why `TypeChoice` is equivalent to the bottom type. Thanks.

Issue #2485, opened 2020-01-28
Variables::extend bug/miss-behavior


ngeiswei:
`Variables::extend` does not work properly when the same variable has 2 different types. Consider for instance the following unit test

https://github.com/opencog/atomspace/blob/950762c2c55498050a2d2632da24d73065b4bfbb/tests/atoms/core/VariablesUTest.cxxtest#L83-L108

Since the `(TypedVariable <var> (Type <type-name>))` construct checks the exact type, it should not consider the type intersection according to the hierarchy of types but rather type equality.

The culprit code is

https://github.com/opencog/atomspace/blob/950762c2c55498050a2d2632da24d73065b4bfbb/opencog/atoms/core/Variables.cc#L1269-L1279

But ultimately a correct fix should take into account the cases where `TypeInhLink` is used, which should then consider type intersection according to the type hierarchy.

Also, it is to be noted that the default type of an undeclared variable is

```
TypeInhNode "Atom"
```

not

```
TypeNode "Atom"
```

linas:
seems like a correct analysis. Patch forthcoming?

ngeiswei:
I think @kasimebrahim is looking into it.

linas:
So is #2496 a complete fix for this issue? Can this be closed, or is there more?

ngeiswei:
I'm closing, @kasimebrahim feel free to re-open if we've forgotten something.

PR #2496, opened 2020-02-06
Bugfix variables extend

kasimebrahim:
as per issue https://github.com/opencog/atomspace/issues/2485, this is a fix for `TypeNode` `TypeInhNode` and `TypeCoInhNode` misbehaving in `Variables::extend`. 

Note that TypeInhNode used to record immediate children and now will record recursive children, same is true for TypeCoInhNode with Parents.

And note that type_intersection is removed and replaced with just set intersection of the TypeSets, as it has no usage.

### DIFF
--- a/opencog/atoms/core/Variables.cc
+++ b/opencog/atoms/core/Variables.cc
@@ -717,7 +717,7 @@ void Variables::get_vartype(const Handle& htypelink)
 		Type vt = TypeNodeCast(vartype)->get_kind();
 		TypeSet ts;
 		TypeSet::iterator it = ts.begin();
-		nameserver().getChildren(vt, std::inserter(ts, it));
+		nameserver().getParentsRecursive(vt, std::inserter(ts, it));
 		_simple_typemap.insert({varname, ts});
 	}
 	else if (TYPE_CHOICE == t)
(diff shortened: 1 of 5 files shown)


opencog/atoms/core/Variables.cc
```diff
@@ -737,6 +737,28 @@ void Variables::get_vartype(const Handle& htypelink)
...
+					TypeSet ts;
+					auto i_it = std::inserter(ts, ts.begin());
+					nameserver().getChildrenRecursive(vt, i_it);
+					typeset.insert(ts.begin(), ts.end());
+				}
+			}
+			else if (TYPE_CO_INH_NODE == var_type)
+			{
+				Type vt = TypeNodeCast(ht)->get_kind();
+				if (ATOM != vt)
```
linas: there should also be an `else` statement: else if vt is Atom, then bottom type, i.e. the match will always fail.

opencog/atoms/core/Variables.cc
```diff
@@ -709,15 +709,15 @@ void Variables::get_vartype(const Handle& htypelink)
 		Type vt = TypeNodeCast(vartype)->get_kind();
 		TypeSet ts;
 		TypeSet::iterator it = ts.begin();
-		nameserver().getChildren(vt, std::inserter(ts, it));
+		nameserver().getChildrenRecursive(vt, std::inserter(ts, it));
 		_simple_typemap.insert({varname, ts});
```
linas: OK, the above suggests that there should be a method
```
TypeSet NameServer::getChildrenRecursive(Type vt) {...}
```
and then you could just say
```
_simple_typemap.insert({varname, namserver(),getChildrenRecursive(vt)});
```
instead of all that ugly iterator.begin stuff, and the repeated cut-n-paste of this 4-line block of code...

Anyway, such a change is not for this pull req, but for some other one.

opencog/atoms/core/Variables.cc
linas: Never mind ... handling this correctly is actually hard to do. There is a very special case, here: `(TypeChoice (TypCoInh 'Atom))`

PR #2498, opened 2020-02-08
Assorted misc fixes to new code

linas:
Assorted fixes to minor issues pointed out in #2496 and #2497

### DIFF
--- a/opencog/atoms/atom_types/NameServer.h
+++ b/opencog/atoms/atom_types/NameServer.h
@@ -170,8 +170,8 @@ public:
     unsigned long getParentsRecursive(Type type, OutputIterator result) const
     {
         unsigned long n_parents = 0;
-        for (Type i = 0; i < nTypes; ++i) {
-            if (recursiveMap[i][type] and (type != i)) {
+        for (Type i = 0; i < type; ++i) {
+            if (recursiveMap[i][type]) {
                 *(result++) = i;
                 n_parents++;
             }
--- a/opencog/atoms/atom_types/NameServer.h
+++ b/opencog/atoms/atom_types/NameServer.h
@@ -159,6 +159,16 @@ public:
         }
         return n_children;
     }
+    TypeSet getChildrenRecursive(Type type) const
+    {
+        TypeSet ts;
+        for (Type i = 0; i < nTypes; ++i) {
+            if (recursiveMap[type][i] and (type != i)) {
+                ts.insert(i);
+            }
+        }
+        return ts;
+    }
 
     /**
      * Given the type `type`, get all of the parents. This is
@@ -177,6 +187,16 @@ public:
             }
         }
         return n_parents;
+    }
+    TypeSet getParentsRecursive(Type type) const
+    {
+        TypeSet ts;
+        for (Type i = 0; i < type; ++i) {
+            if (recursiveMap[i][type]) {
+                ts.insert(i);
+            }
+        }
+        return ts;
     }
 
     template <typename Function>
--- a/opencog/atoms/core/Variables.cc
+++ b/opencog/atoms/core/Variables.cc
@@ -844,9 +844,12 @@ void Variables::get_vartype(const Handle& htypelink)
 
 	if (0 < intervals.size())
 	{
-		_glob_intervalmap.insert({varname, std::make_pair(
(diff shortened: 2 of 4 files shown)


linas:
Merging. I do not believe the openpsi crash has nything at all to do with this pull req.

kasimebrahim:
@linas @ngeiswei may I ask a question, Isn't an empty interval a valid interval? why should the intersection between `{1, 2}` and `{3, 4}` be `{0, 0}`? 
and if we take `{0, inf}` and `{0, inf}` the intersection will be `{0, SIZE_MAX}` does that mean `(Number "inf")` and `(NumberNode "1.8446744073709552e+19")` are semantically similar?

ngeiswei:
I agree, `{0, 0}` is valid, since it is a closed interval `[0, 0]`, not an open one `[0, 0)`.  @linas why would you not use `{NAN, NAN}` as implemented by @kasimebrahim to represent invalid interval?

As for `{0, inf}` and `{0, SIZE_MAX}` I suppose it is practically equivalent, but @kasimebrahim do you need to worry about that in the code?

kasimebrahim:
> As for {0, inf} and {0, SIZE_MAX} I suppose it is practically equivalent, but @kasimebrahim do you need to worry about that in the code?

No no I just wanted to know. because for example 
<pre>
(EqualLink
   (NumberNode "inf")
   (NumberNode "1.8446744073709552e+19")
)
</pre>
should not return truth value of 1, and I am just curious if that would create an issue in the future.

ngeiswei:
Ah, good point, yes, since https://wiki.opencog.org/w/EqualLink represents *semantic equality* I suppose it should return true, maybe. We might also want to take that into account while inserting intervals to the atomspace, just like we detect alpha-equivalent scopes, etc. Though I suppose it should  cause little to no trouble for the time being.

linas:
I used `{0,0}` to denote the empty-set. Since it's the interval that matches nothing, it seemed like a good solution.

I switched to integers instead of floats, because these intervals are always for counting (specifically, counting atoms) and counting is always done with non-negative integers. 

Floats create bottomless pits of confusion and bugs and bad code. It's very easy to create 0.99999999997 when multiplying/dividing by 3 or 7, but C/C++ will do this:  `int n = (int)0.999999997; printf("its %d\n", n);` and it prints zero.  So you have to either add 0.5 to everything, unless its negative, in which case you subtrat. Or use `std::round`.  Then there are problems with Nan. and inf and -inf ... and float vs. double. The more you try to be precise, the more special cases, traps, pitfalls you discover. It is best to just avoid these completely. Just use counting integers (non-negative integers). Realisitically, we will never have GlobNodes matching more than a few dozen atoms: after that, its combinatorial explosion.

By the time you are wondering about `(EqualLink (NumberNode "inf") (NumberNode "1.8446744073709552e+19"))` it is clear that you are doing something deeply wrong, thinking about the problem the wrong way. If you really really need EqualLink to handle floating-point intervals in some "mathematically precise way", you should be using MPFR and not raw C++ double/float. The MPFR library was designed to give accurate, reproducible, well-defined behavior to floating-point intervals.

ngeiswei:
> I used {0,0} to denote the empty-set. Since it's the interval that matches nothing, it seemed like a good solution.

{0,0} is a rather useless interval, but it is valid.

For instance the following query

```
;; KB
(define (Concept "A"))
(List A)

;; Query
(define G (Glob "$G"))
(define N0 (Number 0))
(define Q (Get
  (TypedVariable
    G
    (Interval N0 N0))
  (List G A)))
(cog-execute! Q)
```

returns

```
(SetLink
   (ListLink
   )
)
```

which is correct, but that's in contradiction with the same query with an invalid interval, such as

```
(define N1 (Number 1))
(define Q (Get
  (TypedVariable
    G
    (Interval N1 N0))
  (List G A)))
(cog-execute! Q)
```

which rightfully returns

```
(SetLink
)
```

So I fear that using {0,0} as invalid introduces a bug in `Variables::extend`. Maybe we can use instead {1,0} or such, since it is actually invalid.

linas:
OK, yes, `{1,0}` or any `{m,n}` with `m>n` would be fine, as long as appropriate places in the code check for that. I'm opening a new issue right now so that this is not forgotten, but don't plan to fix it today...

PR #2499, opened 2020-02-08
Fix install path issues.

linas:
This is part of an ongoing attempt to fix install issues
for guile modules.


linas:
Merging. Based on my testing, this seems to finally resolve all the bugs/issues related to guile module install.

PR #2500, opened 2020-02-10
Post search results to an AnchorLink

linas:
This is a partial, initial-draft solution to issue #1502 - it allows results 
to be posted to an AnchorLink, instead of returning them in a SetLink.
The changes to enable this are small, and the example is easy.


linas:
This pull req was reverted in eee7a61ea766d6f94d9b1ef3a2caab649d793ba4

Basically, it just seemed like a bad idea, in the end.

FWIW, if this kind of thing was wanted, a better solution would
be a new kind of Value, that dequeued from a QueueValue, and
plopped the results onto an AnchorNode. ***This*** is the real
reason for stripping away this code: its not generic enough.

Issue #2504, opened 2020-02-17
Variables::substitute_nocheck does not alpha-convert when required


ngeiswei:
## Problem

`FreeVariables::substitute_nocheck` does not alpha-convert scope links in order to avoid name collision, whereas the comment

https://github.com/opencog/atomspace/blob/23ce42661444021e80973a06ff52baf184aa4e88/opencog/atoms/core/Variables.cc#L435-L438

indicates that it should.

## Reproduction

Enable the following unit test

https://github.com/opencog/atomspace/blob/d1d898dc6daa179982ddaec2860d1164acbdc048/tests/atoms/core/VariablesUTest.cxxtest#L853

## Solution

Fix `FreeVariables::substitute_scoped` to perform alpha-conversion when required.

linas:
Very odd, because there's a huge blob of code: 

https://github.com/opencog/atomspace/blob/23ce42661444021e80973a06ff52baf184aa4e88/opencog/atoms/core/Variables.cc#L465-L536

that appears to be trying to protect the bound variables. So this is surprising.

linas:
Oh, never mind; I see. The unit test is testing something very different from what the comments 
https://github.com/opencog/atomspace/blob/23ce42661444021e80973a06ff52baf184aa4e88/opencog/atoms/core/Variables.cc#L435-L438 
were intended to describe. And likewise, the big blob of code is handling the intended case.  The unit test is valid, but unexpected, so I'm not surprised that it failed.

The existing code is designed to handle this case:
```
(Inheritance (Variable X)  (Scope (Variable X) (... stuff ...)))
```
so that when substituting, only the first `X` is substituted; the second `X` is not. That is why the comments talk about "alpha-hiding" instead of "alpha-renaming" -- the second `X` is just hidden away during substitution. 

What you describe in the new unit test requires an actual renaming.

PR #2505, opened 2020-02-19
Fix FreeVariables::substitute_scoped name collision

ngeiswei:
Fix issue https://github.com/opencog/atomspace/issues/2504

### DIFF
--- a/opencog/atoms/core/Variables.cc
+++ b/opencog/atoms/core/Variables.cc
@@ -430,13 +430,16 @@ std::string FreeVariables::to_string(const std::string& indent) const
 }
 
 /// Perform beta-reduction on the term.  This is more-or-less a purely
-/// syntactic beta-reduction, except for two "tiny" semantic parts:
-/// The semantics of QuoteLink, UnquoteLink is honoured, so that quoted
-/// variables are not reduced, and the semantics of scoping
-/// (alpha-conversion) is honored, so that any bound variables with the
-/// same name as the free variables are alpha-hidden in the region where
-/// the bound variable has scope.
-Handle FreeVariables::substitute_scoped(const Handle& term,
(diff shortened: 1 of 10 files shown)


ngeiswei:
@linas, I'm confident this is correct but you might have a look if you want.

Note that I could have made `FreeVariables::must_alpha_hide` and `FreeVariables::alpha_hide` take `Variables` instead of `Handle` to avoid double scope casting, but thought the performance penalty is negligible and so went with what I thought is more elegant.

linas:
I believe that recent compilers improved the casting performance, but its still underwhelming. Its approx 6x or 8x slower than a subroutine call, i.e. its not yet "trivial", but its an acceptable, reasonable cost, otherwise.

opencog/atoms/core/Variables.cc
```diff
@@ -462,76 +465,33 @@ Handle FreeVariables::substitute_scoped(const Handle& term,
...
-			// using the alpha-renamed variable index map.
-			HandleSeq oset;
-			for (const Handle& h : term->getOutgoingSet())
-			{
-				oset.emplace_back(substitute_scoped(h, args, silent,
-				                                    hidden_map, quotation));
-			}
-			return createLink(std::move(oset), term->get_type());
+			// Otherwise the new map must be passed down below
+			index_map_ptr = &hidden_map;
```
linas: This risks a memory-access fault, or other corruption, because `hidden_map` goes out of scope, and it's destructor runs, with the closing brace below. I think it's some happy accident that it didn't die ... probably because its on the stack, and the dtor wont run until the stack is popped.

linas: I fixed this in f84c6d84cd619e8c11105bf2abd1d7500efff479 in master.

PR #2510, opened 2020-02-19
More mucking with rvalues

linas:
This seems to improve performance by a miniscule amount... maybe 1% or less

### DIFF
--- a/opencog/atomspace/AtomTable.cc
+++ b/opencog/atomspace/AtomTable.cc
@@ -139,15 +139,15 @@ void AtomTable::clear()
     clear_all_atoms();
 }
 
-Handle AtomTable::getHandle(Type t, const std::string& n) const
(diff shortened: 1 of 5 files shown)


opencog/atomspace/AtomTable.cc
```diff
@@ -139,15 +139,15 @@ void AtomTable::clear()
     clear_all_atoms();
 }
 
-Handle AtomTable::getHandle(Type t, const std::string& n) const
+Handle AtomTable::getHandle(Type t, const std::string&& n) const
```
vsbogd: As `n` is moved below `std::string&` can be replaced by plain `std::string` safely. Same for `getHandle(Type, HandleSeq)

linas: ?? There's no more `string&`; I'm changing everything to `string&&`.

Issue #2507, opened 2020-02-19
Alpha-equivalence of scopes with inner scopes fails


ngeiswei:
## Problem

`Scope::is_equal` fails over scopes with inner scopes such that variables in the outer and inner scopes are renamed.

## Example

The following 2 scopes

```scheme
(define sc1 (Scope
  (Variable "$X")
  (Scope
    (Variable "$Y")
    (Inheritance
      (Variable "$X")
      (Variable "$Y")))))
```

```scheme
(define sc2 (Scope
  (Variable "$Z") 
  (Scope
    (Variable "$W")
    (Inheritance
      (Variable "$Z")
      (Variable "$W")))))
```

although alpha-equivalent are given different hash keys.

```scheme
(cog-handle sc1)
$1 = 15474112605993223571
(cog-handle sc2)
$2 = 17125956293990831059
```

## Reproduction

Can also be reproduced by enabling the following utest

https://github.com/opencog/atomspace/blob/5d373a0f3ba9edd9a8a776cec07149aead1a05b1/tests/atoms/AlphaConvertUTest.cxxtest#L412-L432

PR #2511, opened 2020-02-20
Fix alpha equivalence with inner scopes

ngeiswei:
Fix issue https://github.com/opencog/atomspace/issues/2507

### DIFF
--- a/opencog/atoms/core/ScopeLink.cc
+++ b/opencog/atoms/core/ScopeLink.cc
@@ -337,8 +337,7 @@ ContentHash ScopeLink::compute_hash() const
 	hsh |= mask;
 
 	if (Handle::INVALID_HASH == hsh) hsh -= 1;
-	_content_hash = hsh;
-	return _content_hash;
+	return hsh;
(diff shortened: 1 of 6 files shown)


ngeiswei:
@linas, I'm confident it's correct, but just in case you may want to have a look at it.

opencog/atoms/core/ScopeLink.cc
```diff
@@ -329,65 +334,69 @@ ContentHash ScopeLink::compute_hash() const
...
-	// Quotation
-	quotation.update(t);
-
-	// Other embedded ScopeLinks might be hiding some of our variables...
-	bool issco = nameserver().isA(t, SCOPE_LINK);
-	UnorderedHandleSet bsave;
-	if (issco)
+	// If it is a scope, call specialized scope hash with the index
+	// completemented with the new variables of that scope, hiding the
+	// old ones if necessary.
```
linas: It would be nice if the comment mentioned that this fixes #2507

PR #2514, opened 2020-02-24
Added Support for nested VariableDecls

rTreutlein:
This is required for the MetaRule support in the URE


linas:
Can you explain what it means for a variable decl to be nested? Also, unit tests are failing ...

ngeiswei:
I think it's possible to avoid nested variable declarations now that globs are supported by the URE.

@rTreutlein, I'm closing your PR for now, and I've copied the branch on my account. Thanks.

PR #2533, opened 2020-04-02
Replace "~" to "_" too in CMAKE_CURRENT_SOURCE_DIR

mhatta:
Replace "~" to "_" too in CMAKE_CURRENT_SOURCE_DIR since CMake doesn't allow it in target names.

Currently, the build of attention fails because of this.   I need to use "~" in directory names for Debian packaging.


linas:
I will add a note that says ~ is for Debian packaging...

linas:
thanks

PR #2534, opened 2020-04-03
Start work on the SetTVLink

linas:
Add two new link types. Actual implementation to follow.


linas:
@ngeiswei please note: I had to fix `VariablesUTest.cxxtest` to make up for failures due to the indeterminate sort order of `std::set`.  As you wrote these tests, I'm concerned that you have something similar  in URE or PLN that might get affected. FYI...

ngeiswei:
I'm looking into it @linas , investigating whether the unit tests or the tested code is wrong. Please don't merge until I get back to you.

ngeiswei:
@linas, if you rebase that branch, ignoring your changes in VariablesUTest.cxxtest, it should now pass.

PR #2536, opened 2020-04-03
Fix persist cmake config

ngeiswei:
Avoid having a linking error if persist-sql isn't built.


ngeiswei:
I feel confident enough about it to merge, @linas .

Issue #2532, opened 2020-03-27
(PLN) Truth Value updates


linas:
This issue lists several ideas for altering how truth values are updated. It's aimed at PLN, one of the current primary consumers of truth values. The goals are (1) simpler, easier-to-use (2) improved performance.

A review of different ways of updating TV's.  

**GroundedSchema**
The first: the current PLN usage, which uses GroundedSchemaNodes as a key work-horse. A typical rule is explicitly declarative:
```
Bind
     Variables ...
      Body ...
      ExecutionOutput
           GroundedSchema "scm: ..."
           List
                 Target
                  Inputs ...
```
The GroundedSchema is some scheme snippet that grabs the TV's from the `Inputs...`, performs some simple arithmetic, and updates the TV on `Target`.  (Its not always "simple arithmetic", but many of the base-cases are.)

This works, and is relatively easy to use. The biggest disadvantage is speed: for simple arithmetic, the CPU-time spent entering/exiting scm is approx 10x greater than the CPU time spent computing the arithmetic formulas. But is this an actual real-world bottleneck for PLN, or is it a drop in the bucket? @ngeiswei  ?  Is it safe to assume it mostly doesn't matter, and most PLN overhead is lost on chaining, unification, etc.?

**PredicateFormula**
The `PredicateFormulaLink`, used with `DefinedPredicate` allows arithmetic formulas to be specified in Atomese.  See [examples](../tree/master/examples/atomspace/formulas.scm) for a working example. In short, this looks like
```
Evaluation
    DefinedPredicate  "some name"
    List
        Inputs ...
```
Where the `DefinePredicate` is attached to a formula written in Atomese, i.e. with `PlusLink`, `TimesLink`, etc. The advantages of this is that the CPU overhead of the GroundedSchema is avoided. The disadvantage is that there is no way to specify a "target": only the TV on the EvaluationLink is modified.

(There are a few "theoretical advantages" to this form: since the formulas are in Atomese, they could be "easily" imported and/or learned: e.g. from MOSES. Also, in principle, atomese formulas could be byte-compiled for performance)

**Enhanced ExecutionOutput** (proposal)
The goal is to get past the limitation of the previous example. This would alter `ExecutionOutput` to allow the use of a GPN and or other predicate.  The form would be:
```
ExecutionOutput
    GroundedPredicate "scm: some-formula"
    List
           Target      ;; The atom whose TV will be modified
           Inputs ...  ;; the atoms that will be passed to the formula
```
[...]

PR #2540, opened 2020-04-04
Implement a SetTVLink

linas:
This implements the "static" version of the ideas discussed in #2532 
It includes unit tests and examples.


linas:
Merging; I think this is self-contained, complete, and not disruptive of other subsystems.

PR #2545, opened 2020-04-13
In scheme, undefined atom should be #f instead of '()

linas:
Fix an error made a very long time ago, of returning an
empty list for the undefined atom. This was an OK decision
if this had been common lisp, but, for scheme, and for srfi-1
this was a bad decision, polluting all sorts of code with
pointless checks for the empty list.  So this attempts to
fix that age-old problem.

This might break existing code ... hopefully not, but no
obvious way of knowing...


linas:
Merging ... unit tests seem to pass ...

Issue #2582, opened 2020-05-08
Patternmatcher's unexpected behaviors with logical links


Eman22S:
There are few problems I have came across while using bindlink to instantiate a hypergraph of Orderedlink. 
1. It seems to get into an infinite loop when I have this rule 
```
(define and-identity
   (BindLink
     (Variable "$op")
     (AndLink
       (Variable "$op")
       (True))

     (Ordered
      (AndLink
         (Variable "$op")
         (False)))
 ))
```
run in an atomspace that contains 
```
(AndLink
  (False)
  (True))
```
using `cog-execute! `

2. It does not subsitute glob when we have the above rule with the globnode instead of variable node
```
(define and-identity
   (BindLink
     (Glob "$op")
     (AndLink
       (Glob "$op")
       (True))

     (Ordered
      (AndLink
         (Glob "$op")
         (False)))
 ))
 ```

Rather it produces this as an output
```
$2 = (SetLink
   (OrderedLink
      (AndLink
         (GlobNode "$op")
         (FalseLink
         )
      )
   )
)
```

**Comment by linas** [comment]
>  infinite loop 

Sure. Sounds plausible. It's not hard to write infinite loops. And it seems that's what you did, so not a surprise.  It is impossible to detect infinite loops -- this is known to be a problem not solvable by Turning machines. The pattern matcher can detect and halt a few of simpler the simpler inf loops, but can't do so in general.  Apparently, it fails to detect the one you wrote.

To avoid infinite loops, you should type your variables, and I am guessing that you meant to use  `PresentLink` 00 which is what you need to find something to be contained in the atomspace. Withtout the `PresentLink`, the TrueLink is merely true. rather than being searched for. That is,  your first query is equivalent to 
```
(define and-identity
   (BindLink
     (Variable "$op")
     (AndLink
       (Variable "$op"))
     (Ordered
      (AndLink
         (Variable "$op")
         (False)))
 ))
```
which is equivalent to
```
(define and-identity
   (BindLink
     (Variable "$op")
     (Present (Variable "$op"))
     (Ordered
      (AndLink
         (Variable "$op")
         (False)))
 ))
```
which is no obviously an inf loop.

> (Glob

That's odd. It should have been an inf loop also.  I'm not sure what it wasn't. I'll look into that &fix it..

Inf loops can be useful, when they are tail-recursive -- you can write a variety of different atomese inf loops, that are useful for handling events, atomspace updates, reactions to incoming stimulus, etc.

[...]

PR #2585, opened 2020-05-08
Dramatically lower the inf-loop counter bound.

linas:
Some simple test cases are combinatorially explosive, and
take minutes to get up to a count of 20. So lower the count.


linas:
motivated by #2582

PR #2604, opened 2020-05-12
fast load

noskill:
custom scm parser for faster load of atomese in s-expression format. 
Somewhat ~8 times faster than using guile


opencog/persist/load_scm/fast_load.cpp
```diff
@@ -0,0 +1,172 @@
...
+        if(s[l1] == '(') count++;
+        if(s[l1] == ')') count--;
+    } while(l1 <= r && count > 0);
+    if(count != 0) {
+        throw std::runtime_error("Invalid syntax #2 in " + s);
+    }
+    r = l1 - 1;
+}
+
+void get_next_token(const std::string& s, int& l, int& r)
```
linas, line 46: It would be nice to have a comment that explained what this function did.  I mean, it is not hard to tell by reading the source, but still, reading the source takes more work, more effort than reading comments.

opencog/persist/load_scm/fast_load.cpp
```diff
@@ -0,0 +1,172 @@
...
+#include <vector>
+#include <stdexcept>
+
+bool check = false;
+void get_next_expr(const std::string& s, int& l, int& r)
+{
+    int l1 = l;
+    while(s[l1] != '(' && l1 < r) {
+        if(s[l1] != ' ' && s[l1] != '\t' && s[l1] != '\n') {
+            throw std::runtime_error("Invalid syntax #1 in " + s.substr(l,r-l+1) + " at |" + s[l1] + "|");
```
linas, line 22: it would be more efficient (fewer compares) to write this as
```
while (s[l1] == ' ' or s[l1] == '\t' or s[l1] == '\n') l1++;
if (s[l1] != '(') throw ...
````

File: opencog/persist/CMakeLists.txt [hunk 3]
```diff
@@ -7,4 +7,6 @@ IF (GUILE_FOUND)
 	ADD_SUBDIRECTORY (guile)
 ENDIF (GUILE_FOUND)
 
+ADD_SUBDIRECTORY (load_scm)
+
```
linas, line 11: It would be nice to have a `README` in the `load_scm` directory that explained what the code is for, and why it was created.

linas:
Oh, and one more thing: there should be at least some kind of short unit test for this....  and another: ideally, there's also scheme bindings for this... since all the existing MOZI code is in scheme.

ngeiswei:
Wow! Thanks a lot @noskill, and @linas for carefully reviewing it.

Yeah, scheme bindings would be very welcome. Any volunteer? If not I might take care it of.

I wonder if `persist` is the right folder, I don't have a better suggestion, just wondering...

ngeiswei:
Minor detail, it'd be more consistent if the C++ source file suffix would be `cc` instead of `cpp`.

linas:
`persist` is the right folder for any kind of i/o

as to details, there are also some whitespace issues, but whatever. Don't want to torture noskill too much.

noskill:
Sure, i am working on some more tests and scheme binding

PR #2611, opened 2020-05-14
bugfix and more tests for opencog::load_file

noskill:
test now includes node names with random ascii symbols

### DIFF
--- a/opencog/persist/README
+++ b/opencog/persist/README
@@ -15,6 +15,8 @@ gearman    -- Experimental support for distributed operation, using
 
 sql        -- Works well for most uses -- with caveats. It's slow.
 
+load_scm   -- fast load of atomese from files in s-expression format.
+
 others     -- There used to be more. They have all bit-rotted.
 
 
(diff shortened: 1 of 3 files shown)


linas:
thanks!

linas:
p,s, you are doing `git commit` from an account that differs from your `git pull` account. The git commit account looks like its anonymous...I think (not sure? you have to go into github and tell github that both of these belong to you.)

linas:
I just noticed that you didn't unescape the escaped quotes. I just now pushed #2612 to do that. It also has a few more minor fixes:

-- unescape escaped quotes
-- declare private functions static
-- verify utf-8 support
-- don't load any atoms at all if there is a syntax error.

noskill:
@linas I don't get it. https://github.com/opencog/atomspace/commit/fdf5dbd564357be49ceee43b56b79d10400132db says  Anatoly Belikov committed XX hours ago. git blame also seems to be aware of who is author. How do i check that commit is anonymous or not?

linas:
@noskill  `git log` reports the `Author` which is configured locally with either `git-config` in the repo you are working in, or by hand-editing either the local `.git/config` or the global `~/.gitconfig`.  That is separate from the github side, which doesn't know who that is -- just shows a grey icon.  It is probably because you are using a new or different signing  key, and haven't told github about it.

PR #2613, opened 2020-05-15
Thread safe instantiator

linas:
Attempt to create a thread-safe Instantiator class.
This is required for a multi-threaded pattern matcher.
Along the way, a fair portion of crufty code was found in
Context.cc .. it is removed, as it was leading to bizarre
crashes with bizarre stack traces.  Specifically,
commit 423cbd8feca2379fb11ffe0926a20ff368e4bb6e
was crashing in MapLinkUTest


linas:
OK, so ... clearly that won't work.

PR #2614, opened 2020-05-15
scheme binding for opencog::load_file

noskill:


linas:
Could you redo this patch, but without the changes to `PersistSCM.cc .h` and  `persist.scm `, but with the new unit test? I'm thinking I would rather have a slightly different API, and rather than explaining it, its probably easier for me to just do it ...

linas:
So circle-ci stopped working after I clicked on "use the new circle-ci interface" and now I cannot merge anything because the reasons for circle-ci failing is now completely hidden from view.

opencog/persist/load_scm/fast_load.cc
```diff
@@ -89,10 +89,7 @@ static std::string get_next_token(const std::string& s, uint& l, uint& r)
         uint l1 = l;
         for(; l1 < r && (s[l1] != '"' or ((0 < l1) and (s[l1 - 1] == '\\'))); l1++)
         {
-            // Unescape esaped quotes.
-            // XXX is there anything else we want to unescape?
-            if (s[l1] != '\\' or s[l1+1] != '"')
-                token.push_back(s[l1]);
+            token.push_back(s[l1]);
```
linas, line 92: !? But this is gong to result with a `\"` in the atom name ... there is nothing in the atomspace that will string this out later, and so you will now get an atom with a broken name in it...

noskill: (ConceptNode "test\"") produces valid atom. In what way it is broken?

linas:
I can no longer view circle-ci, and it is failing. This explains how to get the old UI back:
https://support.circleci.com/hc/en-us/articles/360047237534-Switching-to-CircleCI-s-Old-UI-and-back-to-the-New-UI-
However, since I can't log in, I cannot get the old UI back. Argh...

opencog/persist/load_scm/fast_load.cc
linas: I don't understand ... if you pass `(ConceptNode "test"")` into your tokenizer, it will complain when it finds the third quote-mark, right? Because the second quote ends the token search, and the third quote is going to mess up the close-paren... maybe I am reading your code incorrectly?

noskill: it fails on cases with backslash before the last token: like that 
 ```"test\\"```

I need to add backlash counting or something to properly handle such cases, currently condition just checks that the symbol previous to " is backlash, which is not correct.

While "test\"" is ok.

noskill:
The new scm test failed, i don't know why, i test it with guile -l test-load-file.scm and it passes.

opencog/persist/load_scm/fast_load.cc
noskill: @linas i mean ```(ConceptNode "test\"") ```, not ```(ConceptNode "test"") ```  - which is not valid of course

linas: Yes, OK, I understand. In general, unescaping backslashes during file-load is generically necessary. But this is your code, in the end you need to maintain it and get it working correctly.  I tried and failed :-/

linas:
> The new scm test failed

It was missing `(test-end tname)` and was hanging. I'll fix that and merge.

Are you able to view circleci test results? circleci stopped working for me today, I cannot even view it's output any longer.

linas:
FYI, the python code has this file `opencog/cython/opencog/load-file.h` and there appears to be some bizarre collision with that code. Debugging now.  But basically, if you put a diagnostic fprintf in your code, it enters an infinite loop reading empty lines.  This only happens when running the cmake test harness, and not otherwise. Its .. bizarre...

linas:
Oh, duhh, if the file does not exist, you enter into an infinite loop. There is no `eof()` because the file never opened in the first place.

linas:
I patch to check to see if the file exists. Then I hacked the unit test to go to the correct directory.

You probably want to ass support for file search paths, as documented here: https://www.gnu.org/software/guile/manual/html_node/Load-Paths.html 

I'm am unclear on what the python equivalent would be.

linas:
And finally #2617 renames the file, moves it around, and create a new scheme module called `(opencog persist-file)` .. and I swear I am going to take a vacation from hacking on the atomspace starting now.

noskill:
circleci seems to working now, i also had a problem seeing the test results

PR #2615, opened 2020-05-15
Thread safe instantiator2

linas:
Attempt to create a thread-safe Instantiator class.
This is required for a multi-threaded pattern matcher.

### DIFF
--- a/opencog/atoms/execution/Instantiator.cc
+++ b/opencog/atoms/execution/Instantiator.cc
@@ -31,11 +31,8 @@
 
 using namespace opencog;
 
-Instantiator::Instantiator(AtomSpace* as)
-	: _as(as), _halt(false),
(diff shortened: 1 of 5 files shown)


linas:
Merging despite circle-ci failure.

I cannot figure out what is happening ... There is a "new circle-ci UI" and I made the mistake of clicking on the button to use it.  The circle-ci test results are now invisible/unviewable and failing.  Since things build and pass unit tests for me, I assume that this is a temporary circle-ci system outage...

linas:
This explains how to get the old UI back:

https://support.circleci.com/hc/en-us/articles/360047237534-Switching-to-CircleCI-s-Old-UI-and-back-to-the-New-UI-

However, since I can't log in, I cannot get the old UI back. Argh...

PR #2616, opened 2020-05-15
Threaded pattern engine

linas:
This mostly resolves issue #1645 - "Parallelize the pattern matcher"
All unit tests pass; but the code is disabled, because creating the 
multiple threads using OMP-for-each is very costly.

### DIFF
--- a/opencog/query/Implicator.cc
+++ b/opencog/query/Implicator.cc
@@ -46,6 +46,7 @@ Implicator::Implicator(AtomSpace* as)
 bool Implicator::grounding(const GroundingMap &var_soln,
                            const GroundingMap &term_soln)
 {
+	LOCK_PE_MUTEX;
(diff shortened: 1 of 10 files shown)


linas:
Merging; circle-ci stopped working due to their "new improved UI"
so I can no longer see what's going on. Unit tests pass here.

ngeiswei:
@linas to see circleci outputs you can click on the red cross next to the last commit. The ure fails apparently, no idea why ATM.

linas:
> you can click on the red cross next to the last commit. 

And it asks me to login, and then it fails and shows nothing. It started failing like this after I clicked on the "try the new beta" button, and now I can't reset it, because I cannot contact cicleci support without logging in .. which I cannot do.

ngeiswei:
> And it asks me to login, and then it fails and shows nothing.

I chose to login with my github account and it worked.

Issue #2599, opened 2020-05-11
Deprecate VariableList and VariableSet


linas:
Using the `VariableList` to wrap variable declarations .. seemed to make sense at the time it was introduced. But day-to-day experience suggests that it make little practical difference, and merely makes things verbose and hard to understand.  Thus, it's use should be made optional.  The problem is, of course, that lots of existing code assumes it's presence.  Fortunately, most of that code is in the class in `opencog/atoms/core/Variables.h` and so modifying that should be enough to take care of most use-cases.

linas:
close, fixed in #2620

PR #2620, opened 2020-05-18
Enable embedded variable declarations

linas:
This fixes a very old annoyance. Reported in #2599

### DIFF
--- a/opencog/atoms/core/ScopeLink.cc
+++ b/opencog/atoms/core/ScopeLink.cc
@@ -103,9 +103,12 @@ void ScopeLink::extract_variables(const HandleSeq& oset)
 		return;
 
 	// If the first atom is not explicitly a variable declaration, then
-	// there are no variable declarations. There are two cases that can
-	// apply here: either the body is a lambda, in which case, we copy
-	// the variables from the lambda; else we extract all free variables.
+	// there are no prenex-order variable declarations. (There might
+	// still be in-line variable declarations; these are allowed for
+	// some link types, e.g. the various query links.) Anyway, handle
+	// one of two cases: either the body is a lambda, in which case,
+	// we copy the variables from the lambda; else we extract all free
+	// variables.
 	if (VARIABLE_LIST != decls and VARIABLE_SET != decls and
 	    // A VariableNode could a be valid body, if it has no variable
 	    // declaration, that is if the Scope has only one argument.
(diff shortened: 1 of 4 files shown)


ngeiswei:
That's cool! @linas 

How hard do you think it would be to support variable declaration dynamically generated as being for instance the output of an `OutputExecutionLink`? My guess is that it is non trivial. I'm asking cause I suspect that it would be useful for enabling full support of meta-rules (rules producing rules) in the backward chainer.

linas:
>  dynamically generated variable declarations

Possible. Most of the mechanism is in place already; its called "jit_analyze()" in the code.

So normally,  variable declarations and clauses are extracted and processed at the time that the PatternLink is created. However, if the pattern contains DefinedPredicates or DefinedSchema, then that analysis is deferred until run-time. (Because it assumed that the definitions were not available earlier)  (this was probably a design mistake; the original idea was that such definitions might be recursive, and so could not be pre-processed. But such recursive definitions inside the pattern engine can't be done, so the jit idea wasn't needed for this...)

Hmm. OK, so I convinced myself that the `jit_analyze` code was wrong and needs to be removed.  

.. But that doesn't mean that we can't do something similar for variable declarations. Reading through the code, it almost looks like you could write `(Define (DefinedSchema "vardecl") (TypedVariable (Variable "x") (Type 'Concept)))` and then `(Bind ... (DefinedSchema "vardecl") ...)` and that would almost work. 

I'm guessing what you want is `(Define (DefinedSchema "vardecl") (ExecutionOutput ...))`

Of course, it doesn't have to be `DefinedSchema` .. The `jit_analyze` could also grab any `ExecutionOutput` in the body of the pattern, run it, and stick that in the pattern.  Again - not hard, almost everything is in place already.  "Not hard" for me, means "a long afternoon, a day". And, based on experience, that means "well, its midnight and its still not done, argh..." so two days. sometimes three. Argh.  Most of the work is creating and debugging unit tests.

So the real question is: do you want this at runtime, or at patternlink creation time? As always:
* If you plan to use the BindLink more than once, it is best to do setup once, when the link is created. 
* GroundedAnythingNodes are ugly and inefficient, and it would be best to find a way of not using them.

Whatever. If you are seriously interested in this, open a new issue.

ngeiswei:
> do you want this at runtime

Yes, in fact it would probably be something like the variable declaration would be the result of some rewrite (bindlink or something), it's not entire clear to me. But anyway, thanks for your input. It's not needed for now so I'll leave it there and keep a note for future reference.

Issue #2603, opened 2020-05-12
Reduce Arithmetic links cause Segmentation fualt. 


kasimebrahim:
Any Ideas on why executing the following atomese is causing segmentation fault. Sorry I should probably spent more time debugging it my self, but I am temporarily preoccupied with constructing atomese program representation for continues/numerical programs in as-moses. Any idea will be appreciated. 

<pre>
(PlusLink
   (TimesLink
      (SchemaNode "$x")
      (NumberNode "3")
   )
   (TimesLink
      (SchemaNode "$x")
      (NumberNode "2")
   )
   (NumberNode "5")
)
</pre>

linas:
(1) nothing should ever segfault, so that's a bug.

(2) its an infinite loop, for some reason. I'll look at ti tomorrow.

kasimebrahim:
Thanks @linas!

linas:
not fixed yet .. I found another almost-similar bug...

linas:
This:
```
(use-modules (opencog) (opencog exec))
(cog-execute!
(PlusLink
   (TimesLink (VariableNode "$x") (NumberNode "3"))
   (TimesLink (VariableNode "$y") (NumberNode "3"))
   (TimesLink (VariableNode "$x") (NumberNode "2"))
   (NumberNode "5")))

(use-modules (opencog) (opencog exec))
(cog-execute!
(PlusLink
   (TimesLink (VariableNode "$x") (NumberNode "3"))
   (TimesLink (VariableNode "$y") (NumberNode "3"))
   (TimesLink (VariableNode "$z") (NumberNode "2"))
   (NumberNode "5")))
```

linas:
Closing, now all the cases are fixed.

linas:
BTW, I said it before, but I'll say it again: rather than using `(SchemaNode "$x")` use `ValueOfLink` -- this will allow you to evaluate all the rows with just one single call. If you're not sure what this means, ping me... (this was mentioned somewhere else, recently)

kasimebrahim:
Oh I didn't forget that, I think it is a pretty good idea, The reason I am not wrapping the `Schema` and `Predicate` nodes with `ValueOfLink` right now is it requires changes in our fitness evaluation functions and converter(til we finish porting URE based reduct). And I am planing to do it after I finish porting the representation. And since the representation depends much on the arity rather than how we represent feature variables it is easy to keep it up to date with `ValueOfLink`.

PR #2621, opened 2020-05-18
Infinite Recursion on TimesLink

kasimebrahim:
This is remaining bugfix on reduct infinite recursion problem discussed #2603.
@linas this is just a copy of your fix on PlusLink. If there should be more to consider let me know I can work on it.


linas:
thanks!  good catch .. I should have spotted this during the first fix, but was distracted.

Issue #2546, opened 2020-04-16
Eager Execution in PutLink


kasimebrahim:
I am working on an atomese representation for asmoses. The representation I have is an atomese program when executed with knob settings as an argument it gives the candidate program.
Here is an atomese representation equivalent to the combo representation `[and !and]($a [$b !$b])`.

<pre>
  (DefineLink
    (DefinedSchema "representation")
    (LambdaLink
      (VariableList
        (Variable "$knob-1")
        (Variable "$knob-2"))
      (PutLink
        (GlobNode "$args")
        (CondLink
          (EqualLink
            (Variable "$knob-1")
            (NumberNode 1))
          (AndLink (GlobNode "$args"))
          (EqualLink
            (Variable "$knob-1")
            (NumberNode 1))
          (NotLink (AndLink (GlobNode "$args"))))
        (ListLink
          (PredicateNode "$a")
          (CondLink
            (EqualLink
              (Variable "$knob-2")
              (NumberNode 1))
            (PredicateNode "$b")
            (EqualLink
              (Variable "$knob-2")
              (NumberNode 2))
            (NotLink (PredicateNode "$b")))))))
</pre>

Now the problem is if I execute the above the program I get is not fully reduced.

<pre>
(AndLink
  (PredicateNode "$a") 
  (CondLink
    (EqualLink
      (NumberNode "2")
      (NumberNode "1")
    ) 
    (PredicateNode "$b")
    (EqualLink
      (NumberNode "2") 
      (NumberNode "2")
    ) 
    (NotLink
      (PredicateNode "$b")
    )
  )
)
</pre>
What is going on is `ExecutionOutPutLink` executes it til it doesn't get executable atom, which in this case is satisfied with the `AndLink`. So in order to get that CondLink executed we need to add eager execution to PutLink.
Adding something like this in [PutLink:do_reduce](https://github.com/opencog/atomspace/blob/8b28e68b05325acdbbe8d85fddf3bae6492e16ee/opencog/atoms/core/PutLink.cc#L352) should fix it.

<pre>
	if (nameserver().isA(args->get_type(), LIST_LINK))
	{
		HandleSeq seq;
		for (int j = 0; j < args->get_arity(); ++j) {
			Handle arg = args->getOutgoingAtom(j);
			if (nameserver().isA(arg->get_type(), COND_LINK) or
				nameserver().isA(arg->get_type(), PUT_LINK))
				seq.push_back(HandleCast(arg->execute()));
[...]

PR #2622, opened 2020-05-18
Add OR to exception in evaluatable checker

kasimebrahim:
I needed this in asmoses when creating representation.

### DIFF
--- a/opencog/atoms/core/Checkers.cc
+++ b/opencog/atoms/core/Checkers.cc
@@ -44,6 +44,7 @@ bool check_evaluatable(const Handle& bool_atom)
 	// Make an exception for AndLink, its used in pattern matcher
 	// in an unseemly way.
 	if (bool_atom->get_type() == AND_LINK) return true;
+	if (bool_atom->get_type() == OR_LINK) return true;
 
 	for (const Handle& h: bool_atom->getOutgoingSet())
 	{


linas:
Why?

AndLink is treated differentlty due to a historical mistake; I'm in the process of slowly erasing that mistake.  I do not want to spread that mistake to OrLink.

I'm guessing that perhaps you should be using `ChoiceLink` which is a non-boolean-logic Or useful for presenting a menu of choices?

kasimebrahim:
Yeah so do you remember the representation atomeses we were talking about on #2546?
some thing like 
<pre> 
(define the-expr
  (AndLink
    (PredicateNode "$a")
    (CondLink
      (EqualLink (Variable "$knob-2") (NumberNode 1))
      (PredicateNode "$b")
      (NotLink (PredicateNode "$b")))))
</pre>
I needed it to do the same when we have `OrLink` inplace of `AndLink`.

linas:
I think what you need is to add 
```
+++ b/opencog/atoms/core/Checkers.cc
@@ -56,6 +56,7 @@ bool check_evaluatable(const Handle& bool_atom)
                if (VARIABLE_NODE == t) continue;
                if (GLOB_NODE == t) continue;
                if (DEFINED_PREDICATE_NODE == t) continue;
+               if (COND_LINK == t) continue;
 
                // Allow conjunction, disjunction and negation of
                // predicates. Since it cannot inherit from EVALUATABLE_LINK
```

I haven't tested, but I think that's the actual issue you are facing

kasimebrahim:
One more thing the `CondLink` is not actually `CondLink`, it is a `KnobLink`[gets converted to `CondLink` when execution though] which contains more information than `CondLink` when constructing representation.
example of an actual rep will be.
<pre>
(OrLink
  (KnobLink
    (VariableNode "$knob-376fd81c")
    (PredicateNode "$2")
    (NumberNode "0 1 2")
    (FalseLink ) 
    (FalseLink )))
</pre>
But I could check if it is `FunctionLink`?

linas:
Test the patch above, see if that fixes it for you. Also, the checker infrastructure should be updated to print a more informative error message: e.g "found a non-evaluatable atom XYZ in a link that will be evaluated".

linas:
Ho about `if (nameserver().isA(t, COND_LINK))` ? I want to keep this as narrow/strict as possible, at least for now.  

There are several meta-issues that haven't been solved; one is that static type-checking is ... OK but imperfect, and that dynamic type-checking is equally hap-hazard. Both are kind of a kludge, but I want to avoid making them even kludgier by having too many exceptions.  The correct, long-term "perfect" solution is not yet clear.

linas:
I'm assuming KnobLink inherits from CondLink ...

linas:
Oh, well, you could just also make make KobLink inherit from EvaluatableLink, and then no changes to atomspace source are needed.  So if you have some `asmoses_atoms.script` file then just change
```
KNOB_LINK <- WHATEVER
```
to 
```
KNOB_LINK <- WHATEVER, EVALUATABLE_LINK
```
and it will all just work.

kasimebrahim:
Thanks that works.

PR #2623, opened 2020-05-19
Remove deprecated messages upon loading randgen library

ngeiswei:


ngeiswei:
@linas, I don't think you'd have much disagreement about that PR, but I'll you merge.

linas:
I just added those messages a few weeks ago, why are you removing them? This module should be nuked, its completely useless... the native guile random generator API is far superior.

linas:
Oh, I just saw this remark: https://github.com/opencog/atomspace/commit/4f12fd8ac9cc316ed63a0f4150ce12e9498342bc#r39278620

Why don't you change the documentation to describe what it actually does: i.e. that it is used to set the random seed of the opencog-utils rendgen utility. 

Anyway ... this is a really bad strategy, BTW, because it assumes that some unknown code block somewhere is actually using the random number generator from cogutil, instead of (for example) the ones that c++ provides in the standard library (or the ones that the C library provides, which are different, again, and use other seeds.).  A superior API would be to have the unknown code block provide it's own API to set the random seed for only that generator that it uses, and not for someone-else's generator.  That way, the user doesn't have to make assumptions about how the code might have been written.

ngeiswei:
I see... I'm only using opencog randgen as to have a central point of control over the random seed, makes it easy to achieve reproducibility, globally. Ideally I suppose you're right, this would offer more flexibility, either to achieve reproducibility of some component only, or allowing to use independent random generators, which might be desirable for performance reasons, I guess.

ngeiswei:
> Why don't you change the documentation to describe what it actually does: i.e. that it is used to set the random seed of the opencog-utils rendgen utility.

I think that's what I've said in the warnings. I'll try to reword that but I'm not too sure how to make it much better.

ngeiswei:
I've reworked the docstrings, I think it is pretty clear that is the intended purpose, so I'm merging. Should you have more feedback for correction, let me know.

Issue #2627, opened 2020-05-22
VoidValue proposal


vsbogd:
`GroundedSchemeNode` functions are required to return some result and they cannot return `Handle::INVALID` because execution code calls methods on returned value: https://github.com/opencog/atomspace/blob/36a98d775e1ca294eedb590135c61a47f1a945ef/opencog/atoms/execution/Instantiator.cc#L678-L680

I propose introducing `VoidValue` to solve the issue. The value which has only instance and separate VOID_TYPE

linas:
What is the actual problem that you are trying to solve?

vsbogd:
I am trying to execute GroundedSchemaNode and it fails with segfault in Instantiator.cc

linas:
The meta-issue is that having allowing `Handle::UNDEFINED` aka nullptr handles seems to have been a design mistake, so I am trying to get rid of every location where they might occur.  Do you have a stack trace?  (Most recently, `MapLink` was returning `Handle::UNDEFINED` but that was fixed just 3-4 days ago.)

vsbogd:
Following script fails:
```py
import unittest
from opencog.exec import execute_atom
from opencog.type_constructors import *

class VoidFuncTest(unittest.TestCase):

    atomspace = AtomSpace()

    def setUp(self):
        set_default_atomspace(self.atomspace)
        InheritanceLink(ConceptNode("Frog"),       ConceptNode("animal"))

    def test_void_func(self):
        bind = BindLink(VariableNode("x"),
                 AndLink(PresentLink(VariableNode("x"))),
                 ExecutionOutputLink(
                     GroundedSchemaNode("py: void_func"),
                     ListLink(VariableNode("x"))))
        result = execute_atom(self.atomspace, bind)

def void_func(atom):
    print("function called")
    pass

if __name__ == "__main__":
    unittest.main()
```

Failure description:
```
======================================================================
ERROR: test_void_func (__main__.VoidFuncTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "../atomspace/tests/cython/bindlink/test_void_func.py", line 19, in test_void_func
    result = execute_atom(self.atomspace, bind)
  File "exec.pyx", line 14, in opencog.exec.execute_atom
AttributeError: 'NoneType' object has no attribute 'handle_ptr'

----------------------------------------------------------------------
```

linas:
[...]

PR #2629, opened 2020-05-22
Test for bad return value

linas:
Add unit tests for GroundedSchema / GroundedPredicate's that return bad values.
 * Scheme code was already handling these properly.
 * Python was handling some of these properly, but not all.
   In some cases, it threw opaque, hard to understand errors.
   In the case of #2627 it was not noticing a bad return value.

Also, consolidate two different python execution paths into one.
They were cut-n-paste copies of each-other. Merge them back
together again. This results in an error being throw for #2627

### DIFF
--- a/opencog/cython/PythonEval.cc
+++ b/opencog/cython/PythonEval.cc
@@ -1061,6 +1061,17 @@ TruthValuePtr PythonEval::apply_tv(AtomSpace * as,
     // Grab the GIL.
     PyGILState_STATE gstate = PyGILState_Ensure();
 
+    // Did we actually get a TruthValue?
+    if (0 == PyObject_HasAttrString(pyTruthValue,
+                                    "truth_value_ptr_object"))
+    {
+        Py_DECREF(pyTruthValue);
+        PyGILState_Release(gstate);
+        throw RuntimeException(TRACE_INFO,
+            "Python function '%s' did not return TruthValue!",
+            func.c_str());
+    }
+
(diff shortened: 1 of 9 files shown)


opencog/cython/PythonEval.cc
```diff
@@ -985,114 +985,76 @@ PyObject* PythonEval::call_user_function(const std::string& moduleFunction,
...
         throw RuntimeException(TRACE_INFO,
-            "Python function '%s' did not return TruthValue!",
+            "Python function '%s' did not return Atomese!",
             func.c_str());
 
     // Grab the GIL.
     PyGILState_STATE gstate = PyGILState_Ensure();
 
+    // Did we actually get a Value?
+    if (0 == PyObject_HasAttrString(pyValue, "value_ptr"))
```
vsbogd, line 1015: Checking if `pyValue` is subclass of the `Value` would be more safe. This is few more lines of code though.

vsbogd: Other code looks good to me.

linas: pyValue isn't a subclass of Value, its a pointer to a python object holding a pointer to a value.

linas: Unless you mean that I should call `PyObject_IsInstance()` ...

vsbogd: Yes, I mean `PyObject_IsInstance`

PR #2634, opened 2020-05-24
Add support for SimpleTruthValue in fast-file reader.

linas:
Add support for SimpleTruthValue in fast-file reader.

See https://github.com/MOZI-AI/knowledge-import/issues/10


linas:
Merging. URE keeps failing, but URE doe not use this code, so ... ?

PR #2635, opened 2020-05-25
Remove cinit as init constructor is enough

vsbogd:
`__cinit__` is required to initialize class instance on C level but Value doesn't require such initialization so `__init__` is enough and more safe.

### DIFF
--- a/opencog/cython/opencog/value.pyx
+++ b/opencog/cython/opencog/value.pyx
@@ -17,9 +17,6 @@ cdef class PtrHolder:
         return ptr_holder
 
 cdef class Value:
-    """C++ Value object wrapper for Python clients"""
-    def __cinit__(self, PtrHolder ptr_holder, *args, **kwargs):
-        self.ptr_holder = ptr_holder
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


vsbogd:
@ntoxeg FYI

ntoxeg:
@vsbogd Is it really more safe? `__cinit__` is guaranteed to be invoked exactly once, including in all derived classes - sounds like C fields should be set there. Let's see if the checks pass, I am not tightly against the removal, if the code is simpler - fair enough.

That `__cinit__` is written generically so that derived class can add their own arguments but unless some of them need `__init__` that doesn't have a pointer as the first argument, that flexibility is not needed.

vsbogd:
@ntoxeg , original idea was to wrap `C++` pointer by `PtrHolder` and then use only `Python` constructors. My understanding `__cinit__` is not required here because `PtrHolder` instance should be valid `Python` object.

ntoxeg:
Ah, that is actually true - `ptr_holder` needs to be assigned to in `__init__` of course.

PR #2637, opened 2020-05-25
Test execution of a result of a query

vsbogd:


linas:
The URE forward-chainer unit test is intermittently failing, maybe 2 out of 3 times. I don't know why. Possibly @ngeiswei can enlighten us?

Because this particular unit test seems to pas, I'll merge. (Yes, it looks like a valid test that is working correctly.)

PR #2639, opened 2020-05-25
Fix segfault in Instantiator

vsbogd:


linas:
What's causing the segfault?

vsbogd:
@linas, please close it, if you think that it is a right way.

linas:
Will you be submitting a VoidValue pull req? Now that I understood what it was, it seems reasonable...

ngeiswei:
Should that be merged or closed?

Issue #2628, opened 2020-05-22
importing AtomSpace or types throws error (python 2)


einalex:
```
Traceback (most recent call last):
  File "/home/rasberry/catkin_ws/src/rasberry_hri/scripts/experimenter_node", line 7, in <module>
    from experiments.experimenter import Experimenter
  File "/home/rasberry/catkin_ws/src/rasberry_hri/src/experiments/experimenter.py", line 5, in <module>
    from experiment import Experiment
  File "/home/rasberry/catkin_ws/src/rasberry_hri/src/experiments/experiment.py", line 22, in <module>
    from utils import start_process
  File "/home/rasberry/catkin_ws/src/rasberry_hri/src/utils.py", line 12, in <module>
    from opencog.atomspace import AtomSpace, types, TruthValue
  File "opencog/nameserver.pyx", line 59, in init opencog.atomspace (/home/rasberry/git/atomspace/build/opencog/cython/opencog/atomspace.cpp:15789)
TypeError: type() argument 1 must be string, not unicode
```

linas:
Can you port your code to python3?

einalex:
Porting isn't the problem unfortunately, that'd be easy.
Our platform runs on ros kinetic and there are some more python 2 dependencies. So I can't move forward.

linas:
The cython maintainers/champions are  ... unresponsive... so I'm going to do some guess-work. Does the patch below fix it for you?
```
--- a/opencog/cython/opencog/nameserver.pyx
+++ b/opencog/cython/opencog/nameserver.pyx
@@ -56,7 +56,7 @@ cdef generate_type_module():
     types["NO_TYPE"] = NOTYPE
     return types
 
-types = type('atom_types', (), generate_type_module())
+types = type("atom_types", (), generate_type_module())
 
 #This function is for refreshing new types
 #ex. When you import a cython module which include non-core atom ty
```
I've got some other ideas, but try that first.

linas:
This might work:
```
--- a/opencog/cython/opencog/CMakeLists.txt
+++ b/opencog/cython/opencog/CMakeLists.txt
@@ -21,7 +21,7 @@ INCLUDE_DIRECTORIES(
        ${CMAKE_CURRENT_BINARY_DIR}
 )
 
-SET(CYTHON_FLAGS "-3" "-f")
+SET(CYTHON_FLAGS "-2" "-f")
 
 # Use this as a guide:
[...]

Issue #2642, opened 2020-05-26
exec is a reserved word in python 2


einalex:
The recent work on python 3 introduced a `exec.pyx/pxd` file in `atomspace/opencog/cython/opencog`
This leads to a syntax error:
```
Traceback (most recent call last):
  File "/home/rasberry/catkin_ws/src/rasberry_hri/scripts/scheduler_node", line 10, in <module>
    from bdi.knowledge_base import *
  File "/home/rasberry/catkin_ws/src/rasberry_hri/src/bdi/knowledge_base.py", line 12, in <module>
    from opencog.bindlink import execute_atom, evaluate_atom
  File "/home/rasberry/git/atomspace/build/opencog/cython/opencog/bindlink.py", line 3
    from opencog.exec import *
                    ^
SyntaxError: invalid syntax
```

`exec` is a reserved word in python: https://docs.python.org/3/library/functions.html#exec

I suggest we rename that file to something not reserved.

vsbogd:
`exec` is not in the list of the Python 3 keywords: https://docs.python.org/3/reference/lexical_analysis.html#keywords And one can define function `exec` and call it. 

But it is a Python 2 keyword https://docs.python.org/2/reference/lexical_analysis.html#keywords
Your issue is more likely related to Python 2/3 mixed environment.

einalex:
well, it's defined in the code, so when I compile it with python 2 and run it, it throws an error.
The only mix of environments here is that the code assumes it isn't run under python 2 and thus uses python 2 reserved words, thus breaking compatibility needlessly.

einalex:
Yes, sorry about that. Unfortunately for this project I'm stuck on py2 :/

linas:
`exec` used to be called `bindlink` and maybe we can keep it that way for python2

linas:
fixed in #2647

linas:
closing, re-open if still an issue

PR #2646, opened 2020-05-26
Support python2 for cython.

linas:
This should be enough to fix issue #2628.


einalex:
This also needs
https://github.com/opencog/atomspace/issues/2642

PR #2647, opened 2020-05-26
exec is a reserved keyword in python2

linas:
This fixes issue #2642


einalex:
Your branch fixes the issue in my example code, but leads to a crash in my main project. Unfortunately it doesn't produce any related log entries even with log level set to debug.
I'm probably misusing the logger, as there are hardly any log messages at all.

linas:
> but leads to a crash in my main project.

Make sure you've got a clean build directory. This should not be required, but when things go crazy, it's a sanity check.

You may need to explictly remove ` /usr/local/lib/python3.7/dist-packages/opencog/exec.so` (or wherever the old one got installed). Try removing `/usr/local/lib/opencog/libexec.so` before reinstalling.

A new backwards-compat `/usr/local/lib/python3.7/dist-packages/opencog/exec.py` is installed, and maybe this is causing an issue.

If build times take too long, but you've got enough RAM, then use `ccache` which provides a huge speedup for builds.  `sudo apt install ccache; rm CMakeCache.txt; PATH=/usr/lib/ccache:$PATH cmake ..; make -j`

einalex:
I've been starting my builds with `rm -rf *` since a few hours ago.
I'll try and hunt down the remnants of the old system.
Let's see if that helps.

einalex:
Hrm, still crashes.

linas:
arghh stack trace? Maybe its unrelated? Maybe I missed a spot?

einalex:
Was due to some other changes that happened in the meantime, that caused the atomspace repo to become out of alignment with the other repos. Updating and rebuilding all of them fixed it.

Issue #2641, opened 2020-05-26
Remove constant clause with pattern matcher, or not?


ngeiswei:
My understanding is that constant clauses are removed during initialization. This is corroborated by this

https://github.com/opencog/atomspace/blob/7bc74eb08b1ec41f4605d78c47633cfa77ab8aef/opencog/atoms/pattern/PatternLink.cc#L59

which is however in contradiction with this unit test

https://github.com/opencog/atomspace/blob/7bc74eb08b1ec41f4605d78c47633cfa77ab8aef/tests/query/ConstantClausesUTest.cxxtest#L405-L422

Digging deeper, the reason this unit test passes is because its clause

```scheme
(Inheritance (Concept "DA") (Concept "DB"))
```

is actually not considered a constant by

https://github.com/opencog/atomspace/blob/7bc74eb08b1ec41f4605d78c47633cfa77ab8aef/opencog/atoms/pattern/PatternUtils.cc#L95-L119

because, `InheritanceLink` inherits from `EvaluatableLink`.

Of course fixing it results in `ConstantClausesUTest::test_constant_pattern()` to fail.

So I'm asking, how should constants be treated?

I should mention that on my [URE] end, removing constants is no problem, even if, as tested by `test_constant_pattern` these constants are in a different atomspace, because the URE does the checking anyway (as it was agreed previously).

So what should be done?

1. Should `test_constant_pattern` be modified to expect `(Set (Concept "DB"))`? And therefore `is_constant` be fixed?
2. Should `test_constant_pattern` not be modified? But in that case `is_constant` is still not doing what it says it does, in which case the pattern matcher should be fixed elsewhere as well.

ngeiswei:
BTW, the intermittently failing ure utest is linked to that issue. I do not mind fixing it but I need to know what is the desirable behavior.

linas:
The root issue goes back to a very old debate: Should the pattern engine look at TV's, or not? 

Clearly for `Evaluation GPN` it *must* run the GPN.  The GPN returns a TV, now what?  The current ad-hoc solution is to look at the TV with `if (tv->strength > 0.5)` and completely ignore the confidence.

What about `Evaluation PN` ? There is nothing to run! Should we look at the TV? The current ad-hoc solution is "yes" -  to see if it is present, and if it is, then  to look at the TV with `if (tv->strength > 0.5)` and completely ignore the confidence. 

What about `InheritanceLink`? There is nothing to run! Should we look at the TV? 

[...]

PR #2649, opened 2020-05-27
Make MemberLink and a bunch other types NOT EvaluatableLink

ngeiswei:
As discussed in

https://github.com/opencog/atomspace/issues/2641#issuecomment-634168508

The complete list of atom types not anymore EvaluatableLink are:

MemberLink
InheritanceLink
SimilarityLink
SatisfyingSetLink
ImplicationScopeLink
ImplicationLink

As a result ConstantClausesUTest::test_constant_pattern() fails and
thus has been corrected accordingly.


tests/query/ConstantClausesUTest.cxxtest
```diff
@@ -403,8 +403,11 @@ void ConstantClausesUTest::test_complex()
 }
 
 /**
- * Constant clauses in a child atomspace are expected not to be accessible
- * from the parent atomspace (so the rewrite term should not be triggered).
+ * Constant clauses in a child atomspace would be expected not to be
+ * accessible from the parent atomspace, so the rewrite term would not
+ * be triggered, however constant clause are removed, regardless of
+ * whether they belong to the queried atomspace or not, thus, since
+ * empty is synonymous to true, the rewrite term is triggered.
```
linas, line 410: Actually, this is not correct. If nothing was found, then nothing was found, so nothing should be re-written and returned.

So that's a bug .... will note this in #2651

PR #2654, opened 2020-05-28
fix includes in persist/file/fast_load.h

noskill:
missing definitions in header for std::string etc.

### DIFF
--- a/opencog/persist/file/fast_load.h
+++ b/opencog/persist/file/fast_load.h
@@ -25,9 +25,11 @@
 
 #ifndef FAST_LOAD_H
 #define FAST_LOAD_H
+#include <opencog/atomspace/AtomSpace.h>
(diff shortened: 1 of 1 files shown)


opencog/persist/file/fast_load.h
```diff
@@ -25,9 +25,11 @@
 
 #ifndef FAST_LOAD_H
 #define FAST_LOAD_H
+#include <opencog/atomspace/AtomSpace.h>
+#include <string>
 
 namespace opencog {
-    void load_file(std::string fname, AtomSpace& as);
+    void load_file(std::string fname, opencog::AtomSpace& as);
```
linas, line 32: You don't actually need to say `opencog::` here. It's already the case, since one line above, you already declare the namespace.

Issue #2650, opened 2020-05-27
Inconsistent unification in atomese


noskill:
I tried to use EqualLink for unification of a variable with a pattern:

```scheme
(MemberLink 
     (EvaluationLink
            (PredicateNode "has_name")
            (ListLink (ConceptNode "node1") (ConceptNode "name1")))
     (ConceptNode "node2"))

(define (pred1 x x1)
                (display (list x x1)) (newline) 
                (stv 1 0))
```

For some reason this is ok:

```scheme
(cog-execute! (QueryLink
    (AndLink 
        (MemberLink 
            (EvaluationLink (VariableNode "X")  (VariableNode "Y"))
            (ConceptNode "node2"))
       (EqualLink (VariableNode "X") (PredicateNode "has_name"))
       (EvaluationLink (GroundedPredicate "scm:pred1") (VariableNode "Y"))) 
   (VariableNode "Y")
))
$1 = (QueueValue
    (ListLink
     (ConceptNode "node1") ; [480756faa9d577d8][1]
     (ConceptNode "name1") ; [e064383668383b3][1]
   ) ; [8057748bbf1ea089][1]
)
```

But this one causes "C++-EXCEPTION' with args `("cog-execute!" "Infinite Loop detected! Recursed 6 times!"
```scheme
(cog-execute! (QueryLink  
   (AndLink 
       (MemberLink 
             (EvaluationLink (PredicateNode "has_name") (VariableNode "Y"))
             (ConceptNode "node2"))
        (EqualLink (VariableNode "Y") 
              (ListLink (VariableNode "N") (ConceptNode "name1"))))
    (VariableNode "Y")))
```

Another inconsistency:
```scheme
(DefineLink
    (DefinedPredicate "pred1")
    (PredicateFormula
        (Number 1)
        (Times
            (ConfidenceOf (Variable "$X"))
            (ConfidenceOf (Variable "$Y"))))
)
```

ok
```
(cog-evaluate! (Evaluation (DefinedPredicate "pred1") (ListLink 
...          (ConceptNode "node1")
...          (ConceptNode "name1"))))
$3 = (stv 1.000000 0.000000)
```

not ok
```
(cog-execute! (QueryLink
  (AndLink
     (MemberLink
       (EvaluationLink (VariableNode "X") (VariableNode "Y"))
       (ConceptNode "node2"))
     (EqualLink (VariableNode "X") (PredicateNode "has_name"))
[...]

PR #2655, opened 2020-05-28
Fix broken PredicateFormula in searches

linas:
This fixes the second bug described in #2650 

It also deals with several other minor issues that cropped up along the way:
* A missing CountOfLink
* A mis-behvaing/incomplete ArityLink


linas:
Merging; one important part of the original bug report is still not fixed, but this is at least a coherent set of changes that address most of the problem.

PR #2656, opened 2020-05-29
Add support for parsing strings in fast_load

Habush:
This PR extends the existing fast load functionality to support string expression. This can be useful when we want to parse small Atomese expressions.


opencog/persist/file/fast_load.h
```diff
@@ -25,11 +25,26 @@
...
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+#include <opencog/atoms/atom_types/NameServer.h>
+#include <opencog/atoms/base/Link.h>
+#include <opencog/atoms/base/Node.h>
+#include <opencog/atoms/core/NumberNode.h>
+#include <opencog/atoms/truthvalue/SimpleTruthValue.h>
+#include <opencog/atomspace/AtomSpace.h>
```
linas, line 40: Can you move these header files back to the .cc file? There is nothing in this header file that needs any of these; they don't belong here ... they belong only in the c code that actually needs them.

tests/persist/file/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+LINK_LIBRARIES(atomspace load_scm)
+
+ADD_CXXTEST(FastLoadUTest)
```
linas, line 3: missing newline

tests/persist/file/FastLoadUTest.cxxtest
```diff
@@ -0,0 +1,39 @@
...
+    std::string in = R"((Evaluation (Predicate "is_a") (List (Concept "Earth") (Concept "Planet"))))";
+
+    Handle h = parseExpression(in, _as);
+
+    TS_ASSERT_EQUALS(5, _as.get_size());
+
+    TS_ASSERT_EQUALS(2, h->getOutgoingSet().size());
+
+    logger().info("END TEST: %s", __FUNCTION__);
+}
```
linas, line 39: missing newline

opencog/haskell/CMakeLists.txt
```diff
@@ -18,11 +18,11 @@ INSTALL (TARGETS atomspace-cwrapper
...
-    )
-ENDIF (HAVE_STACK)
+#IF (HAVE_STACK)
+#    ADD_CUSTOM_TARGET(haskell-atomspace-lib ALL
+#        DEPENDS atomspace-cwrapper
+#        COMMAND sh build.sh ${CMAKE_CURRENT_BINARY_DIR}
+#        WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+#        COMMENT "Building Haskell bindings."
+#    )
+#ENDIF (HAVE_STACK)
```
linas, line 28: this has nothing at all with file loading .. if this is needed, please submit a separate pull req for it.

opencog/persist/file/fast_load.cc
```diff
@@ -231,17 +213,37 @@ void opencog::load_file(std::string fname, AtomSpace& as)
...
 
             expr_cnt++;
-            as.add_atom(recursive_parse(expr, l, r, line_cnt));
-            expr = expr.substr(r+1);
+            h = as.add_atom(recursive_parse(expr, l, r, line_cnt));
+            expr = expr.substr(r + 1);
         }
     }
+
+    return h;
```
linas, line 230: This is .. kind of strange ... it returns only the last atom in the stream. Is this actually useful? 
Maybe you want to return a `std::vector<Handle>` of them?

opencog/persist/file/fast_load.cc
```diff
@@ -231,17 +213,37 @@ void opencog::load_file(std::string fname, AtomSpace& as)
...
+    if (not f.is_open())
+       throw std::runtime_error("Cannot find file >>" + fname + "<<");
+
+    parseStream(f, as);
+    
     f.close();
 }
+
+//Parse an Atomese string expression and return a Handle to the parsed atom
+Handle opencog::parseExpression(std::string expr, opencog::AtomSpace &as) {
```
linas, line 246: To avoid an un-needed string copy, change the signature to `const std::string&`

opencog/persist/file/fast_load.cc
```diff
@@ -203,21 +190,16 @@ static Handle recursive_parse(const std::string& s,
...
 }
 
-/// load_file -- load the given file into the given AtomSpace.
-void opencog::load_file(std::string fname, AtomSpace& as)
-{
-    std::ifstream f(fname);
-    if (not f.is_open())
-       throw std::runtime_error("Cannot find file >>" + fname + "<<");
-
+static Handle parseStream(std::istream& in, AtomSpace& as){
```
linas, line 193: Why static? Why not export this to the world? It seems useful enough...

opencog/persist/file/fast_load.cc
linas: Oh, I guess its useful if you want to parse only one atom at a time ...

Habush: > it returns only the last atom in the stream. Is this actually useful?

My initial use case was to return a single atom that wraps the other atoms as the addition to the atomspace is done recursively (using the `recursive_parse` function). But  your suggestion of returning `std::vector<Handle> ` makes more sense.

Habush: But I am wondering if we should use `std::set<Handle>` instead of a vector to avoid adding the same atom more than once as it can occur multiple times in an input. What do you think?

opencog/persist/file/fast_load.cc
```diff
@@ -203,21 +202,16 @@ static Handle recursive_parse(const std::string& s,
...
 }
 
-/// load_file -- load the given file into the given AtomSpace.
-void opencog::load_file(std::string fname, AtomSpace& as)
-{
-    std::ifstream f(fname);
-    if (not f.is_open())
-       throw std::runtime_error("Cannot find file >>" + fname + "<<");
-
+Handle parseStream(std::istream& in, AtomSpace& as){
```
linas, line 205: OK, so you made this not be `static` but you didn't put it in the header file ...

opencog/persist/file/fast_load.cc
linas: Well, don't over-engineer things. Right now, fast-file-load is fast because it spends about half its time parsing strings, and half it's time in `AtomSpace::add_atom()`. You want to avoid doing things that will slow it down for no good reason.  So:
* Creating giant `SetLink` of atoms is pointless and useless and slow. It takes a lot of CPU time to validate each, a lot of CPU time to set up the incoming sets, and then, if/when you delete the `SetLink`, you waste a lot of time changing the incoming sets back.  It's just a bad idea.
* Maintaining a HandleSeq or a HandleSet is useless if no one is going to look at it. It might not sound like a lot of CPU time, but compared to parsing strings, or compared to `AtomSpace::add_atom()`. it might be another 5% or 10% ... do you really need to slow down the code?
* The ***whole idea*** of the atomspace is that it is a "blackboard", a "place where you can find things". You shouldn't really need to know what was loaded, because you can always find it.  If you feel that you do need to know ... ask yourself why? is it really needed? or is it just a comfortable old habit that has no meaning, has no purpose? (old meaningless habits are great for culture and society, but not for programming.)  I suspect that you don't *really* need the list of loaded atoms.

Habush: > I suspect that you don't really need the list of loaded atoms.

I don't (at least for now). I was just suggesting solutions to your comment. But we can change this to return a container type when the need arises.

linas: Well, do you need the return value at all? Or are you planning for some uncertain future?

PR #2658, opened 2020-05-30
Rework how the pattern engine mixin callbacks work

linas:
Fixing the bug with the DefinePredicate allowed PatternLink code to 
move back to the diretory that the rest of PatternLink lives in. This 
in turn meant that the search algo no longer had to live with PatternLink
but could live with the pattern engine, where all of the rest of the search
happens.  This triggers a series of dominoes. Along the way, assorted
mixin classes were renamed to hopefully be more clear about what they
actually do.

This requires matching changes to AttentionBank, the Pattern Miner, 
and to OpenCog.


linas:
These have dependencies on this pull:
https://github.com/opencog/attention/pull/16
https://github.com/opencog/opencog/pull/3640
https://github.com/opencog/miner/pull/41

Issue #2644, opened 2020-05-26
OrLink only produces correct query results if defined with a single Link in its outgoing set


einalex:
As the title says, `OrLink` produces an empty `SetLink` if defined over more than one condition.

Output:
```
hungry people (OrLink (PresentLink ...))
expected: (SetLink (ConceptNode "me"))
(SetLink (ConceptNode "me"))

hungry or thirsty people (OrLink (PresentLink ...) (PresentLink ...))
expected: (SetLink (ConceptNode "me") (ConceptNode "you"))
(SetLink)
```

Test code:
```
#!/usr/bin/env python

from opencog.type_constructors import *
from opencog.bindlink import execute_atom, evaluate_atom
from opencog.utilities import initialize_opencog
from opencog.atomspace import AtomSpace, types

if __name__ == "__main__":
    atomspace = AtomSpace()
    initialize_opencog(atomspace)
    me = ConceptNode("me")
    you = ConceptNode("you")
    hungry = ConceptNode("hungry")
    thirsty = ConceptNode("thirsty")

    StateLink(me, hungry)
    StateLink(you, thirsty)

    query1 = GetLink(
                TypedVariableLink(VariableNode("someone"), TypeNode("ConceptNode")),
                OrLink(PresentLink(StateLink(VariableNode("someone"), hungry))))

    who_is_hungry1 = execute_atom(atomspace, query1)
    print("\nhungry people (OrLink (PresentLink ...))")
    print('expected: (SetLink (ConceptNode "me"))')
    print(who_is_hungry1)

    query2 = GetLink(
                TypedVariableLink(VariableNode("someone"), TypeNode("ConceptNode")),
                OrLink(PresentLink(StateLink(VariableNode("someone"), hungry)),
                            PresentLink(StateLink(VariableNode("someone"), thirsty))))

    who_is_hungry2 = execute_atom(atomspace, query2)
    print("\nhungry or thirsty people (OrLink (PresentLink ...) (PresentLink ...))")
    print('expected: (SetLink (ConceptNode "me") (ConceptNode "you"))')
    print(who_is_hungry2)

```

linas:
Can you use `ChoiceLink` instead? 

The long-term solution for this is to automatically convert (under the covers, inside the engine) or-link+present-link combinations into choice-links but I'm feeling lazy right now. Mostly because one then has to convert `not or present` into `and absent` which rapidly snowballs into conjunctive-normal-form pathologies, and I've been trying to avoid that ...

[...]

PR #2664, opened 2020-06-01
Choice link with compound terms

linas:
Allow ChoiceLinks where each choice alternative is a compound term
i.e. has multiple terms wrapped in a PresentLink.  This (partly) fixes issue #2644


linas:
This is a commit of work-in-progress. Issue #2644 is NOT fixed in this commit.

PR #2665, opened 2020-06-02
Export test-equal

ngeiswei:
Otherwise the tests cannot be run in standalone.


ngeiswei:
Please someone double check this. If I don't export `test-equal` then I cannot run some scheme unit tests in standalone.

linas:
I didn't double-check, but it seems reasonable.

PR #2667, opened 2020-06-02
Add VoidValue to return when no return vaulue is expected

vsbogd:
Implementation of #2627


linas:
thanks!

linas:
You forgot to provide VoidValueUTest!

PR #2669, opened 2020-06-03
Add VoidValue unit test

vsbogd:
Unit test from #2667


linas:
can you re-enable it in the CMakefile?

PR #2679, opened 2020-06-07
Move Grounded nodes to own directory

linas:
This does several things:

1) it untangles a nasty dependency loop with python and everything else
2) it simplifies ideas like GroundedObjectNode that singularity.net wants to have.


linas:
@noskill  and @vsbogd FYI

linas:
Unit tests will not pass until opencog/ure#91 is merged.

PR #2680, opened 2020-06-08
Grounded Python and Scheme Nodes

linas:
Split out the if-then-else in the GroundedPredicate/GroundedSchema  
into distinct python and scheme classes. This allows language-specific
optimizations to be performed for each different language, thus potentially
improving perforrmance, on a per-language level.


linas:
FYI @noskill and @vsbogd I believe that this makes GroundedObjectNode ***even simpler**, easier to deal with.  It should now be straightforward to stick customer per-language optimizations and special callouts and whatever else is needed into the corresponding "Runner" files. You can jam precompiled Python code in there, pointers to special python objects, whatever. All wrapped up in one easy-to-manage location.

PR #2684, opened 2020-06-09
Fix cog-get-atoms

ngeiswei:
That function did not follow its specification, as stated in its comment. That is the given type was not considered when its subtypes where to be considered.


ngeiswei:
I [pln and miner] am the only user of this function when `subtype` is `#t`, thus this change should have no impact on anybody but me, as of now. So I'm merging.

PR #2686, opened 2020-06-13
Debug printing cleanup

linas:
Mostly provides improved PatternTerm printing.


linas:
Merging. Since circle-ci has locked me out, I'm no longer able to determine the root cause. I assume its an intermittent failure; everything passes here...

linas:
This is staging for #2687

Issue #2687, opened 2020-06-13
Pattern analysis simplifications.


linas:
Pattern analysis can be simplified and generalized as follows:
* Remove the gathering up of the `pmandatory', `absents` and `always` clauses, and instead keeping them in one list.
* Remove `_fixed` and use the `ptm->hasAnyVirtual()` to determine component bridging.

This is a blocker for #2644

PR #2688, opened 2020-06-14
Simplify pattern analysis

linas:
Recent rework resulted in assorted cruft having accumulated in the 
pattern analysis code.  Remove it; this simplifies the overall flow.


linas:
This is staging for #2687

PR #2690, opened 2020-06-15
Run cogserver tests during CI build

vsbogd:


.circleci/config.yml
```diff
@@ -217,6 +217,9 @@ jobs:
       - run:
           name: Build tests
           command: cd build && make tests
+      - run:
+          name: Run tests
+          command: cd build && make test
 #
```
linas, line 223: Did you read lines 223 to 233, immediately below?

linas:
So does the `SET_PROPERTY` thing actually fix it for you? I'd like to understand why your system is different from my system... (which does not need the `SET_PROPERTY` thing...) What version of the linker do you use (ldd --version`) and gcc (`gcc --version`)

vsbogd:
You can see in CI log that it cannot build `atomspace` with `SET_PROPERTY` commit because of cyclic dependency:
https://circleci.com/gh/opencog/atomspace/8058?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

I didn't tried to build it locally, it fails in CI. The environment is `opencog/opencog-deps` docker.

linas:
OK. Hmm. Well, I'm running the shell script in issue #5 and have so far seen 2 failures out of 9 tests.  Given that both BackwardChainerUtest fails intermittently, and that the pattern miner SurprisingnessUTest fails intermittently,  I'm still not sure I want to re-enable these tests.  It's gotten hard to get circle-ci to pass unit tests, and this just further deteriorates the situation.  The cogserver is pretty low priority ...

vsbogd:
But the issue https://github.com/opencog/cogserver/issues/27 shows that `libtruthvalue.so` cannot be used as is. So it is an `opencog/atomspace` issue.

linas:
Sure, but you have two different things in one pull request. I don't mind merging the `SET_PROPERTY` thing. I do mind merging the re-enablement of the unit tests.  Maybe open a different pull req for SET_PROPERTY?

vsbogd:
At the moment PR is for testing only. And test shows that adding `SET_PROPERTY` doesn't compile.

linas:
Well, The `SET_PROPERTY` belongs in `lib/AtomSpaceConfig.cmake.in` as mentioned in opencog/cogserver#27

ngeiswei:
@linas, you still have ure or miner utest intermittent failures?

Can you point to the circleci runs that failed? Cause I can't reproduce that locally.

linas:
> circleci runs 

3-4 days ago, different compoents kept failing, I kept having to hit retest....

> intermittent

Yesterday, I ran SurprisingnessUTest in a loop. It failed on the very first try, then ran 113 more times without failures. So that beats me... 

I did not test  BackwardChainerUTest because it passes and then the double-free-crash, probably due to the temp atomspaces...?

linas:
> SurprisingnessUTest ... ailed on the very first try, then passed

@ngeiswei FYI, I am continuing to see this behavior. I make some significant change to the atomspace, SurprisingnessUTest fails, I panic, I run it again, it passes. But it only fails the first time, every subsequent attempt passes.... !?  I'm now paying attention to what I am doing (maybe a stale build ??) but thought I should mention it.

PR #2691, opened 2020-06-16
Assorted minor cleanup

linas:
Remove some pointless code, and some other assorted cleanup.


linas:
this is staging for #2687

PR #2694, opened 2020-06-16
Improve is_constant comment + add unit tests

ngeiswei:


ngeiswei:
I have added these unit tests to make sure that `is_constant` is doing what I though it did.

PR #2697, opened 2020-06-16
Absent present cleanup

linas:
Provide correct implementations for absent/present-link checks.


linas:
Never mind, too soon. Buggy.

PR #2698, opened 2020-06-17
Relax boolean link validator

ngeiswei:
Currently the validator associated to BOOLEAN_LINK is
check_evaluatable, the problem is boolean links can also be used on
atoms that are not evaluatable, such as implication links, according
to PLN.

These links cannot be made evaluatable without invalidating the
function is_constant in PatternUtils.h.  So I think the best thing for
now is to relax the validator of BOOLEAN_LINK.  I have relaxed
check_evaluatable, however I suppose the latter should either be
renamed check_boolean, or a new check_boolean validator should be
introduced that uses check_evaluatable + the additional code in this
commit.

### DIFF
--- a/opencog/atoms/core/Checkers.cc
+++ b/opencog/atoms/core/Checkers.cc
@@ -72,6 +72,18 @@ bool check_evaluatable(const Handle& bool_atom)
 		if (QUOTE_LINK == t) continue;
 		if (UNQUOTE_LINK == t) continue;
 
+		// Negation, conjunction or disjunction over Implication,
+		// Equivalence, Inheritance, Similarity and their subtypes is
+		// currently required by PLN for higher order reasoning. We may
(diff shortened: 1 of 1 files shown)


ngeiswei:
This fix allows to make the following PR to pass BTW https://github.com/opencog/pln/pull/22

ngeiswei:
@linas, I'm gonna merge this now, first because I'm reasonably confident about it, and second we need this rapidly for AGI-20 tutorials.

linas:
OK, but we need some kind of long-term plan to straighten this stuff out. Recall that ImplicationLink was marked "evaluatable" until a week or two ago ... 

Using the boolean links with things that are not evaluatable causes some fair bit of pain in the pattern engine, where there is an assorted collection of hacks trying to guess what the user wanted (Thus, PresentLink, etc and backwards compatibility)

PR #2703, opened 2020-06-18
Simplify struct Variables

linas:
Remove a lot of code. This code was used only by the unifier, and so now
it has been moved to live over there (see opencog/ure#99) This simplification
is a part of the ongoing effort to implement #2644.


linas:
Well, that's weird ... circleci was passing last night, failed this morning...

PR #2713, opened 2020-06-27
Return an empty scheme list if handle vector. is empty

Habush:
This fixes segfault error that happens when `scm_from` method is passed an empty handle vector. It checks whether the handle vector is empty and returns an empty scheme list if it is.


linas:
I'm going to push a different fix. The code as written is insane; it would be better to just fix it, instead of working around a bug. ... just a moment.

linas:
I think #2714 provides a better fix, so I am closing. Please verify that #2714 actually works for you!

Habush:
It works as expected. Thanks!

Issue #2465, opened 2020-01-05
Compilation error on Mac OS


Habush:
While compiling the atomspace locally on my Mac OS (10.14.6), I get the follwing error:

> /Users/xabush/opencog/atomspace/opencog/cython/PythonEval.cc:1389:14: error: use of
>       undeclared identifier 'pipe2'
>     int rc = pipe2(pipefd, 0);  // O_NONBLOCK);

It looks like `pipe2` isn't available in OS X. is there a way to fix this?

linas:
I think you need to install the unix-compatibility layers for macos. 

Anyway, you're the first person in maybe a decade who's tried opencog on macos, so if this is the only problem .. that's actually marvelous.  Submit pull reqs as needed for anything else you trip across.

vsbogd:
For this specific line:
```C++
int rc = pipe2(pipefd, 0);  // O_NONBLOCK);
```
replacing it by:
```C++
int rc = pipe(pipefd);
```
Should solve the compilation issue.

Habush:
@vsbogd suggestion fixes the issue

PR #2715, opened 2020-06-27
Use the correct system call for pipe creation depending on the platform

Habush:
This PR provides a fixe for #2465 by using the correct `pipe` system call depending on whether the platform is Mac OSX or not.


linas:
From the man-page:
```
       int pipe(int pipefd[2]);
       int pipe2(int pipefd[2], int flags);
       If  flags is 0, then pipe2() is the same as pipe().  
```
Just change it to use `pipe()` instead of `pipe2()` ....

linas:
The point being that if we actually needed to mark it as `O_NONBLOCK` then the Mac implementation would fail, anyway. But since it seems to work without  `O_NONBLOCK`, we may as well use just plain `pipe()`.

linas:
thanks

PR #2721, opened 2020-07-07
Fix haskell test

vsbogd:


vsbogd:
Closing it until we find proper solution here https://github.com/singnet/atomspace/pull/128

linas:
The haskell bindings are unmaintained, but as the fix seems vaguely plausible and unit tests pass ... merging.

PR #2730, opened 2020-07-23
Fix python examples + improve link ctor with tv

ngeiswei:
1. Fix Python examples (accommodate to new API + move to Python 3).
2. Improve link constructor to accept optional `tv` argument.

I've also updated https://wiki.opencog.org/w/Python accordingly, up to https://wiki.opencog.org/w/Python#Attention_Values which will be updated as attention gets looked into.


cmake/OpenCogAtomTypes.cmake
```diff
@@ -241,8 +241,8 @@ FOREACH (LINE ${TYPE_SCRIPT_CONTENTS})
                 FILE(APPEND "${PYTHON_FILE}" "    return add_node(types.${TYPE_NAME}, node_name, tv)\n")
             ENDIF (ISNODE STREQUAL "NODE")
             IF (ISLINK STREQUAL "LINK")
-                FILE(APPEND "${PYTHON_FILE}" "def ${TYPE_NAME}(*args):\n")
-                FILE(APPEND "${PYTHON_FILE}" "    return add_link(types.${TYPE_NAME}, args)\n")
+                FILE(APPEND "${PYTHON_FILE}" "def ${TYPE_NAME}(*args, tv=None):\n")
+                FILE(APPEND "${PYTHON_FILE}" "    return add_link(types.${TYPE_NAME}, args, tv)\n")
```
vsbogd, line 245: I would explicitly mention that `tv` is a value for `tv` argument, otherwise there is a chance that it be used as part of outgoing:
```suggestion
                FILE(APPEND "${PYTHON_FILE}" "    return add_link(types.${TYPE_NAME}, args, tv=tv)\n")
```

vsbogd:
@noskill , could you please look at it

cmake/OpenCogAtomTypes.cmake
ngeiswei: Sorry, I don't understand what that means, could you expand @vsbogd?

ngeiswei: Ah, OK, I get it, but I don't think there is any ambiguity because args is list (as far as I understand anyway), so it should consistently work, unless the API of `add_link` changes of course.

However your suggestion doesn't hurt so I might just as well apply it.

vsbogd: Ah agree, it is a problem only when arguments are unpacked.

noskill:
The changes look good to me.

ngeiswei:
Thanks for the feedback guys.

linas:
Unit tests are now failing with 
```
======================================================================
ERROR: test_add_link (test_atomspace.AtomSpaceTest)
----------------------------------------------------------------------
  File "/home/linas/src/novamente/src/atomspace-git/tests/cython/atomspace/test_atomspace.py", line 73, in test_add_link
    l4 = Link(n1, n4, tv=TruthValue(0.5, 0.8))
TypeError: Link() got an unexpected keyword argument 'tv'

----------------------------------------------------------------------
Ran 77 tests in 6.407s
```

and
```
The following tests FAILED:
	153 - CythonAtomSpace (Failed)
```

PR #2731, opened 2020-07-23
Fix decade-old cython irritation

linas:
Fix pre-historic naming failure.


linas:
@vsbogd I think this affects no one, but please take a look and please merge if it looks OK to you.

PR #2733, opened 2020-07-24
Type expression should be parsed both as symbol and string

Habush:
This adds on #2682 to support type expression declared using strings. So now both `(Type 'Concept)` and `(Type "Concept")` will be parsed correctly.


linas:
Thanks!

PR #2744, opened 2020-08-05
Stub out signals. No one uses them.

linas:
They damage performance, and accomplish nothing else.


linas:
... or not. Apparently, the async unit test uses these.

Measurements with benchmark reveals that removing the signals gains a 3% to 7% performance improvement, depending on the workload.

I'm going to punt for now, because that unit test is actually important. But I'm still pretty sure no one else uses them...

PR #2748, opened 2020-08-07
Relax check_evaluatable to allow subtypes of concept and predicate

ngeiswei:


linas:
Looking at #2814 and #2816 ... this change leads non-evaluatable evaluatables:
```
(use-modules (opencog))
(use-modules (opencog exec))
(cog-evaluate! (Concept "foo"))  ;crash!
(SequentialAnd  (Concept "foo"))  ; no error
(cog-evaluate!  (SequentialAnd  (Concept "foo")) ) ;crash!
```

Thus declaring `ConceptNode` to be evaluatable does not actually make it so.

PR #2765, opened 2020-09-06
Hypergraphs, metagraphs, RAM CPU

linas:
This started out as an explanation of how to do "indexing" in a "graph database". It unwittingly morphed into a strong defense of the AtomSpace design. It currently seems to be saying this: "if you are interested in graph databases, and you are not using the AtomSpace, not only are you missing out on something great, but worse, you are using something terrible and inefficient".

Ben, if you are reading this, it would be great if you could review this. You know 98% of what it says, so this should  be easy. The thing that surprised me (I really didn't really expect it, as I wrote it) that the hypergraph/.metagraph DB's are amazingly strong pieces of technology, and offer some large, huge advantages over graph DB's.  These are very strong claims, given how mature the graph DB industry is.

I think it would be great if this could be peer-reviewed, dinged on and banged on, and transformed into a real, publishable piece of literature. This is the strongest defense of the AtomSpace design I know of, and I think it could have the power to turn heads. .. if we got it out there, in those places where heads turn.


namedgraph:
Are you claiming AtomSpace is more performant than graph databases?

linas:
> Are you claiming AtomSpace is more performant than graph databases?

I seem to be. I made assumptions about what they "must surely need to do", and arrived at those statements. But if they don't do what I think they do, and do something else, well then .. I would be wrong. It's possible that I misunderstand their capabilities. If there are mistakes, they need to be corrected; if you're an expert and can spot the mistakes, point them out!

namedgraph:
Are you familiar with RDF triplestores and the SPARQL query language?
https://github.com/semantalytics/awesome-semantic-web#databases

linas:
> Are you familiar with RDF triplestores and the SPARQL query language?

Yes. Ten years ago, the atomspace was triple-store-like, and the original query language was sparql-like. It's evolved rather far away from that starting point. The inadequacies of triple-stores made themselves painfully clear when the second/third opencog chatbot came up. There's a detailed explanation of what's wrong with triple-stores somewhere in here:  https://github.com/opencog/opencog/tree/master/opencog/nlp/chatbot-old  

To explain it quickly: humans don't speak in triples for a reason. They use adjectives and prepositions, and once you start trying to jam adjectives and prepositions into a triple store, you loose all sense of orthogonality. I mean, sure you can jam anything into a triple-store, if you try hard enough, but its not natural. If you try to traverse it as a graph, its a disaster.  Look at gremlin/tinkerpop, for example.

linas:
To be clear: if the triple is `some_relation(thing_a, thing_b)` then the canonical opencog "triple" is `(Evaluation (Predicate "some relation") (List (Concept "thing a") (Concept "thing b")))`  The RelEx subsystem is a triple-extractor for natural language, it converts English-language sentences like "John threw a green ball" to `throw(John, ball) ; _adj(ball, green)` -- try it -- it works. The problem is that this triple markup is not adequate for performing reasoning.  The R2L subsystem (relex to logic) does have a collection of several hundred rewrite rules for taking input triples and extracting logical relationships from them, with the goal of being able to apply reasoning on these, e.g. via PLN.  The system creaks under its own weight.  There is a reason why the neural net people have had a long string of victories. I'm trying to create a theoretical foundation for *why* neural nets work, and how we can back-track them onto good old-fashioned AI symbolic techniques.

LifeIsStrange:
Hi Linas, sorry but I can't restrain myself from telling you how admirative I am of your work.

Also, did you study a bit [shacl](https://www.w3.org/TR/shacl/) ? It's one of the "recent" advances in symbolic logic, I wonder how some of its advantages compare with the atomspace/could be backported to the atomspace

PR #2766, opened 2020-09-08
Add function for wrapping existing loggers

ngeiswei:
Define `wrap_clogger` for wrapping existings loggers. Used by the ure for exposing the ure logger to Python.

### DIFF
--- a/opencog/cython/opencog/logger.pxd
+++ b/opencog/cython/opencog/logger.pxd
@@ -1,0 +1,45 @@
+from libcpp cimport bool
+
+# basic wrapping for std::string conversion
+cdef extern from "<string>" namespace "std":
+    cdef cppclass string:
+        string()
+        string(char *)
+        char * c_str()
+        int size()
+
+
+cdef extern from "opencog/util/Logger.h" namespace "opencog":
+    # Need to get around cython's lack of support for nested types
+    enum loglevel "opencog::Logger::Level":
+        NONE "opencog::Logger::NONE"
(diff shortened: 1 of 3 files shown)


ngeiswei:
Used by https://github.com/opencog/ure/pull/107

ngeiswei:
Review of any one familiar with Python bindings is welcome, otherwise I'll merge in 24h.

opencog/cython/opencog/logger.pxd
```diff
@@ -0,0 +1,45 @@
+from libcpp cimport bool
+
+# basic wrapping for std::string conversion
+cdef extern from "<string>" namespace "std":
```
vsbogd, line 4: I am not sure it is enough, but Cython has std::string wrapper:
```
from libcpp.string cimport string
```

vsbogd (approved):

opencog/cython/opencog/logger.pxd
ngeiswei: Thanks for your feedback @vsbogd, I tried to replace the cdef with `from libcpp cimport string` but I'm getting subsequent error messages like

```
        cLogger(string s)
               ^

/home/nilg/OpenCog/atomspace/opencog/cython/opencog/logger.pxd:25:16: 'string' is not a type identifier
```

Since I'm not very familiar with Cython and this code was there before that change (it has just been moved around, not added) I'm gonna keep it that way for now. If you know how to completely fix it please provide more information though, any code simplification is always welcome.

PR #2778, opened 2020-11-12
Add dependency on base types being generated to storage types generation

ntoxeg:
To prevent race conditions. Building storage types has to wait until base types are **generated**, conditioning on the library being built is not strict enough.


AmeBel:
Thanks

PR #2782, opened 2020-12-01
Add IsTrueLink and IsFalseLink

ngeiswei:
Add virtual link IsTrueLink (resp. IsFalseLink) to check if an atom has a TRUE_TV (resp. FALSE_TV).

Added wikipages as well

https://wiki.opencog.org/w/IsTrueLink
https://wiki.opencog.org/w/IsFalseLink


ngeiswei:
IsTrueLink can already be formalized with

```
And
  GreaterThan
    StrengthOf A
    Number 0.99
  GreaterThan
    ConfidenceOf A
    Number 0.99
```

but such addition makes much more compact something that is, IMO, casually useful.

PR #2794, opened 2021-03-01
Add atomspace-rocks to the unit tests

linas:
Stable and usable .. should be tested.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,6 +78,51 @@ jobs:
             - atomspace
             - ccache
 
+  atomspace-rocks:
+    docker:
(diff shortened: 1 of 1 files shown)


linas:
@amebel -- you understand circleci -- could you add `apt get librocksdb-dev` to wherever opencog circleci dependencies are installed?  I think this is the only one. There's no other configuration required.

vsbogd:
@linas , why don't add it to the `opencog-deps` docker?

AmeBel:
done

linas:
Thank you! BTW, atomspace-rocks works much better than postgres ever did; it's 4x faster and the code is simpler.  And if you need multi-user atomspace-cog works for that.

PR #2803, opened 2021-03-26
Bug-fix caching bug.

linas:
Fix strange and hard to reproduce bug involving caching.


linas:
See unit test in #2812

PR #2806, opened 2021-04-08
Add a new unit test - MultiDeleteUTest

linas:
This unit test provoked a bug in the RocksDB backend (now fixed). Provokes a bug here too.

After revising the unit test, it runs at about 2/3rds the speed of rocks. Turns ut rocks is 30x faster when writes interleve with similar writes, but is about the same as postgres when writes and deletes interleave.


linas:
Merging despite atomspace-rocks failure; I'm working to fix that now. Its unrelated.

PR #2808, opened 2021-04-13
remove atomspace creation in get_default_atomspace

noskill:
nullpointer was passing to AtomSpace_factory


linas:
I added @vsbogd to reviewers, just now. You guys are the only python users I am aware of at this point in time, so if this works for you & unit tests pass, then this should be merged.

ngeiswei:
I'm merging as @vsbogd has already reviewed it on the singnet side.

PR #2809, opened 2021-04-16
Derived atom types should inherit factories

linas:
This allows atom types defined in external shared libraries to
inherit the atom factories defined in the base system (as well
as any factories defined in yet other shared libs that they might
depend on.) This resolves an issue discussed in opencog/pln#40


ngeiswei:
Thanks @linas

linas:
> Thanks @linas

Welcome!  You should merge opencog/pln#42 which actually provides the nit test for this.  (Hmm. I guess I should design a unit test that lives in the atomspace)

PR #2810, opened 2021-04-17
fix MultiUserUTest regression

linas:
The pull req #2807 introduced a regression in the `MultiUserUTest`; 
it would fail one out of three runs.  This pull req fixes the failure.
Both unit tests can be run in excess of 100 times with no failures.


linas:
Merging despite failure i the attention repo; whatever that failure is, it's unrelated to these changes.

PR #2814, opened 2021-04-30
Add SetLink to the list

tanksha:
This will make creating patterns like (Not (Set A)) possible, in working with singletons.


ngeiswei:
@tanksha, I think it would be more appropriate to be with that code

https://github.com/opencog/atomspace/blob/7ae69840fd7bd8e051218df9bcb229d287553205/opencog/atoms/core/Checkers.cc#L65-L69

since `SetLink` can be interpreted as a concept.  Would be good to have a comment explaining that as well.

ngeiswei:
@tanksha, I think it would be more appropriate to be with that code

https://github.com/opencog/atomspace/blob/7ae69840fd7bd8e051218df9bcb229d287553205/opencog/atoms/core/Checkers.cc#L65-L69

since `SetLink` can be interpreted as a concept.  Would be good to have a comment explaining that as well.

linas:
Nil,

The meta-issue is that none of these link types are actually "evaluatable" in Atomese.  Here's the problem: in the early days of "Atomese", at Hanson Robotics, there was a need to `(cog-evaluate! (SequentialAnd (recognize-person) (Or (smile) (blink) (say-hello))))` which means that 'recognize-person` had to be an atom that returned a (crisp) truth-value - if it returned false, the evaluation would stop.  If one of the atoms was not actually evaluatable (e.g. was a ConceptNode) it would crash. That is because there is no C++ `ConceptNode::evaluate()` method.  So the checker was invented to catch these kinds of errors early, while the system was being booted, instead of late, when the robot was onstage.

So now you see what the problem is -- if you call `(cog-evaluate! (And (Concept "foo")))` or `(cog-evaluate! (Not (Set)))`  an exception will be thrown. 

There are 3-4 different solutions:
1) get rid of `check_evaluatable` and instead create a `robot-code-static-typechecker` module, and another called `pln-static-typechecker`

2) invent something that would allow pln to overload the C++ `ConceptNode::evaluatie()` method with custom code that does what PLN needs it to do. (I've got several ideas on how to do this)

3) remove large parts of atomese from the atomspace, and move it to "robot-atomese" and "pln atomese" so that what's left in the atomspace is "common atomese that everyone uses".

I'll open a new issue to continue this discussion, we shouldn't have it here...

linas:
So -- `UnionLink`, `IntersectLink` and `ComplementLink` -- one advantage would be to compute numerical values in C++ code, where its fast, instead of calling `(GroundedShema "scm:compute-union-tv")` which is slow, due to the need to bounce between C++ and scheme.... let me pick this up in #2816

linas:
Hmm. As I attempt to think about #2816, I realize that perhaps `UnionLink` etc is the simplest solution that makes all my vague and hard-to-describe confusions melt away and/or be deferred to some distant future. Let me write a more detailed proposal there. This could all be pretty simple.

tanksha:
@ngeiswei I have made the suggested changes, please check.

linas:
Where are we at with this? It seems that whatever needed this, it was not urgent. Can I close this?

ngeiswei:
I think @tanksha 's change is correct for the time being.  The better longer term change could be to introduce `UnionLink`, etc.  However I have an important, and possibly relevant to that issue, change to suggest before that, that I will write an issue about soon.  But in the meantime of all these more important changes I think we can accept that change.

However it's failing circleci.  @tanksha, could you rebase your branch (oh, it's your master branch, you should use another one) onto the opencog/master?

ngeiswei:
@tanksha, I think you need to close that PR and create a new one (adding a reference to that one cause it discusses important matters) with a new branch.

ngeiswei:
That other option is to just close that PR and forget about it, but I believe the bio-atomspace relies on that change.

linas:
Pull req #2843 adds UnionLink. I plan to merge it as soon as it passes unit tests.

I still think that experimenting with novel kinds of truth values is worthwhile: e.g. having a `IndirectEvidenceTruthValue` that knew how to "calculate itself" on demand. (see #2816 for a very short sketch) @ngeiswei I believe you currently use `GroundedPredicateNode "scm:foo"` for this. I'm suggesting that doing it in C++ would be faster, and doing it in a TV rather than a Node/Link would simplify formulas and simplify rules. Maybe - that's why I call it an "experiment".  (Simplifying anything tends to open new horizons)

linas:
> failing circleci. 

Just hit "retest" I don't think you have to rebase.

ngeiswei:
I believe we can safely close that issue now.

PR #2815, opened 2021-05-05
Revert "Proposed fix for a new circular-dependency bug"

linas:
Reverts opencog/atomspace#2813

Fails to fix anything, as per comments at opencog/pln#44


linas:
Thanks @ntoxeg

Issue #2817, opened 2021-05-06
Persistance API cannot load an Atom which has double quotes inside name


vsbogd:
`Sexpr::encode_atom` doesn't escape double quotes inside name of the atom. Thus `(ConceptNode "\"CC\"")` will be encoded as `(ConceptNode ""CC"")`. Such atom will break loading procedure because first pair of quotes will be decoded as empty name:
```
Traceback (most recent call last):
  File "./demo_using_db.py", line 48, in <module>
    load_atomspace(ac.atomspace)
  File "./demo_using_db.py", line 29, in load_atomspace
    ''')
  File "scheme.pyx", line 44, in opencog.scheme.scheme_eval
RuntimeError: Python-Scheme Wrapper: Failed to execute '
        (use-modules (opencog))
        (use-modules (opencog persist))
        (use-modules (opencog persist-rocks))
        (define sto (RocksStorageNode "rocks:///tmp/foo.rdb/"))
        (cog-open sto)
        (load-atomspace)
        (cog-close sto)
    '
Backtrace:
           6 (apply-smob/1 #<catch-closure face020>)
           5 (apply-smob/1 #<catch-closure fadee20>)
In ice-9/boot-9.scm:
   2312:4  4 (save-module-excursion #<procedure fad37c0 at ice-9/eva…>)
In ice-9/eval-string.scm:
     38:6  3 (read-and-eval #<input: string fa58620> #:lang _)
In unknown file:
           2 (opencog-extension dflt-load-atomspace ())
In ice-9/boot-9.scm:
   751:25  1 (dispatch-exception 0 C++-EXCEPTION ("dflt-load-atom…" …))
In unknown file:
           0 (apply-smob/1 #<catch-closure fadede0> C++-EXCEPTION "…" …)

ERROR: In procedure apply-smob/1:
In procedure dflt-load-atomspace: Syntax error at line 0 Unexpected text: >>CC"")<<
Function args:
()
ABORT: C++-EXCEPTION
 (/home/opencog/projects/opencog/atomspace/opencog/cython/opencog/PyScheme.cc:63)
```

linas:
Yep! Do you want to fix this or should I?

linas:
never mind, I have a fix. Pushing. Thanks for pointing this out!

vsbogd:
Thanks @linas, but I believe we also need to unescape it when reading.

linas:
Still working on it. It touches more code than expected. I pushed a new branch just now.

[...]

PR #2818, opened 2021-05-06
Escape quotes in Nodes and StringValues

linas:
Quotes embedded in node names were not being correctly escaped.

This should fix issue #2817 

This also fixes the same issue for StringValues, and for the command interpreter.


vsbogd:
Sorry @linas, I will be able to test it only at the evening.

linas:
No hurry. I think I did a pretty through job.

PR #2827, opened 2021-07-14
Add missing cfloat include for DBL_MAX

ngeiswei:
Otherwise gcc complains with

```
MinLink.cc: In member function ‘virtual opencog::ValuePtr opencog::MinLink::execute(opencog::AtomSpace*, bool)’:
MinLink.cc:49:44: error: ‘DBL_MAX’ was not declared in this scope
   49 |                         result.resize(len, DBL_MAX);
      |                                            ^~~~~~~
```


ngeiswei:
My version of gcc is 11.1.0.

Issue #2828, opened 2021-07-17
Python example not working


manisabri:
https://github.com/opencog/atomspace/blob/master/examples/atomspace/python.scm#L40
Above line throws an exception 
 __init__() takes exactly 1 positional argument (2 given)

linas:
Fixed in #2833 -- Apparently, someone changed the python API, but did not change the example demos to use the new API.  The fix was easy.

PR #2833, opened 2021-08-08
Python bugfixes

linas:
Apparently, someone changed the python API, but did not change the example demos to use the new API.

The fix was easy...


linas:
This fixes #2828

PR #2836, opened 2021-08-12
Use more contains + fix a bug

ngeiswei:
* Rename a few methods `is_in` by `contains`
* Fix bug by working around `Handle::operator==` being non-content based

### DIFF
--- a/opencog/atoms/base/Handle.cc
+++ b/opencog/atoms/base/Handle.cc
@@ -138,6 +138,14 @@ bool content_eq(const opencog::HandleSetSeq& lhs,
 	return true;
 }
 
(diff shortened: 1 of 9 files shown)


ngeiswei:
@linas, I'm fairly confident it's correct so I'm gonna merge now.

I have 2 comments to however

1. Let it be atomspace A and its child B.  Atom X is added in B, then a link L is added to X in B as well.  Then X is added to A.  I think L should be pointing to X in A, and X in B should be destroyed.
2. There is an inconsistency between `Handle::operator<` which is content based and `Handle::operator==` which is pointer based.  This creates problems downstream requiring the sort of fix in that PR.

PR #2841, opened 2021-09-06
Use iterators as input for content_contains instead of HandleSeq

Habush:
A `vector` is passed to `content_contains` in [FreeVariables.cc](https://github.com/opencog/atomspace/blob/5871e38ec680040a1e48a3c5a33af5408bcd1076/opencog/atoms/core/FreeVariables.cc#L226) and a `set` is passed in [Context.cc](https://github.com/opencog/atomspace/blob/5871e38ec680040a1e48a3c5a33af5408bcd1076/opencog/atoms/core/Context.cc#L84). Hence, it's better to use an `InputIterator` as an argument for it to generally work for containers.

### DIFF
--- a/opencog/atoms/core/Context.cc
+++ b/opencog/atoms/core/Context.cc
@@ -81,7 +81,7 @@ bool Context::is_free_variable(const Handle& h) const
 	Type t = h->get_type();
 	return (t == VARIABLE_NODE or t == GLOB_NODE)
 		and quotation.is_unquoted()
-		and not contains(shadow, h);
+		and not content_contains(HandleSeq(shadow.begin(), shadow.end()), h);
 }
 
 bool Context::operator==(const Context& other) const


linas:
Ugh. This would be one of the very few uses of iterators in the atomspace, ever. We've managed to avoid using iterators until now, and my knee-jerk reaction is to continue avoiding them.  An alternate and simpler fix would be just to copy the `continent_contains` code, and pass a different argument.

@ngeiswei -- opinions?

Also, why is Miner failing?

linas:
> what's the reason behind this?

They're slow, running at half the speed of `for(...)`. They're hard to read (need to pass more arguments, need to write `begin()` and `end()` explicitly, or is that `start()` and `finish()`?) They're not thread-safe (doesn't apply to this example, but is true generically) and they are not parallelizable (doesn't apply to this example, but true generically).  I think there are more reasons, too. Iterators are deeply problematic, as a generic comp-sci concept.

Once upon a time, not sure if it is still true, writing `for (auto it = thing.begin(); it != thing.end(); it++)` caused `thing.end()` to be called once every time through the loop.  Very surprising, if you thought you were writing efficient code.

ngeiswei:
@Habush, I agree with @linas, better to avoid iterators when possible, my main problem with them is its inelegance, and performance-wise it's still relatively cheap to construct a vector, you may for instance write

```C++
content_contains(HandleSeq(from, to), h);
```

If really you want the flexibility of iterators, I would think it's better to wait for C++20 Range (disclaimer, I've never used it, so I might be wrong here).

ngeiswei:
BTW, C++20 is almost completely supported in gcc  https://gcc.gnu.org/projects/cxx-status.html#cxx20 though is still tagged as experimental and probably require a version of gcc that's too high for most at this point, so let's wait a bit more and then switch to it.

Habush:
Okay. Reverted the method to use a `HandleSeq` and fixed the issue in Context.cc

linas:
Thank you.   ... are there any other uses of `contains`?

linas:
Very confusing. `contains` came from `util/algorithm.h` in cogutils....

linas:
I looked for other uses of `contains` and there do not seem to be any left.

Habush:
Btw, the recent [change](https://github.com/opencog/atomspace/commit/fe228df80fe811646174b44409ddaa88d6a861cd) to set thread names using the `prctl` syscall results in the atomspace  building only on Linux as the `prctl` syscall is Linux-specific (_this can be discussed in a separate issue_)

linas:
Open another issue. I was not aware that it built on anything but linux. I'll take any simple patch to stub it out, it was to make using gdb easier.

linas:
Simple patches to eliminate/reduce compiler warnings would be great too. As of right now, it builds on 32-bit linux with no compiler warnings, and passes all unit tests.

opencog/atoms/core/Context.cc
```diff
@@ -81,7 +81,7 @@ bool Context::is_free_variable(const Handle& h) const
 	Type t = h->get_type();
 	return (t == VARIABLE_NODE or t == GLOB_NODE)
 		and quotation.is_unquoted()
-		and not contains(shadow, h);
+		and not content_contains(HandleSeq(shadow.begin(), shadow.end()), h);
```
ngeiswei, line 84: Actually, this change should not be necessary.  That is because `opencog::contains` is specialized for sets, and `Handle::operator<` is already based on content.  Now if the compiler somehow chooses the generic `opencog::contains` instead of the specialized one for sets, then it would indeed introduce a bug that this change would fix.

@Habush, does the absence of that change breaks something on your end?  Or did you fix it solely based on code review?

 Of course the alternative is to have `Handle::operator==` being based on content, as I have repeated many times (maybe I'll just make it happen one day, the last time I tried it broke a bunch of utests, and so I gave up).

linas: > `Handle::operator==` ...  broke a bunch of utests

That's because the utests do some mighty weird things, comparing atoms that are in and not in atomspaces, deleting them, comparing deleted atoms, you name it, all-around craziness. They're also quite gnarly. I don't want to invest time in "fixing" them, and I think removing them would be a bad idea. Other than that, yes, I think I now finally agree that `Handle::operator==` could be overloaded. (It should do a pointer compare first, and then mark the rest of the body with `unlikely` so that the optimizer can do thee right thing.

ngeiswei: `unlikely`?  Sick (as in cool) https://en.cppreference.com/w/cpp/language/attributes/likely

Anyway, changing `Handle::operator==` to be content based is not urgent on my end either.

linas:
likely/unlikely have been in gcc for 20 years, and I'm pretty sure clang has them too (and so mac os has them) so if c++20 is not available, its not hard to write a portability wrapper.  

That web page does not explain it, but by using this hint, the compiler will typically inline the likely part, and leave the unlikely part either as the actual subroutine call, or it will move the code out-of-line, so that it has to take a jump to reach it.  The likely parts have no jumps, and thus are more easily optimized as part of the main flow. Moving the unlikely code out-of-line also reduces i-cache misses (because moving it out of the main line makes the main execution line smaller)

The biggest benefit for us would be anywhere where stuff gets called a lot: e.g. stuff in `Handle.h` I'm not sure where else.

PR #2844, opened 2021-09-10
Fix the incorrect inheritance of UnionLink, etc.

linas:
The pull req #2843 was not quite right.


ngeiswei (approved):
Yeah, I think it's a safer assumption for now.

ngeiswei:
I suppose we could build an abstract atom type, such as `SetConnectorLink` or such, that `UnionLink`, etc, could inherit from, but for now it's good enough I think.

linas:
I've used base types only when they were actually needed for something, e.g. in checkers.cc or if the base class in c++ had common methods in it.

linas:
See also #2845

Issue #2845, opened 2021-09-10
Introduce SetConnectorLink (or such) as parent type of UnionLink, etc


ngeiswei:
When matching classes of atoms, it could be convenient to be able to agglomerate all atom types representing concepts/sets.  So for instance

```
(Get (TypeInh X Type 'SetConnector) X)
```
would match `Set`, `Union`, `Intersection`, `Complement` and possibly more.

linas:
Sure. Except that `SetConnector` feels non-standard.  Perhaps `SetOperator` or `SetOperation` would be better. Maybe `SetFunction` (Saying `SetValuedFunction` seems like too long a name).

However, saying that `Set` inherits from `SetOperation` seems awkward, so perhaps `Collection` is a better name.  That is, `Set`, `Union`, etc. are all `Collection` s.  

So I nominate `Collection` - that would be my favorite.

linas:
This is a follow-up to #2844

PR #2846, opened 2021-09-11
CollectionLink as base class of SetLink

linas:
This satsifies issue #2845

### DIFF
--- a/opencog/atoms/atom_types/atom_types.script
+++ b/opencog/atoms/atom_types/atom_types.script
@@ -156,16 +156,22 @@ ALPHA_CONVERTIBLE_LINK <- LINK
 // ===========================================================
 // Basic boolean algebra and set-theory-inspired links.
 
-SET_LINK <- UNORDERED_LINK
+// Plain old list aka Cartesian product
 LIST_LINK <- ORDERED_LINK
-MEMBER_LINK <- ORDERED_LINK
+
+// Base class for sets and for operations that return sets.
+COLLECTION_LINK <- UNORDERED_LINK
+SET_LINK <- COLLECTION_LINK
 
 // Measure-theoretic (probabilistic) versions of AND_LINK, OR_LINK
 // These use probabilistic formulas to compute truth values.
 // See issue opencog/atomspace#2814 for a discussion.
-UNION_LINK <- UNORDERED_LINK
-INTERSECTION_LINK <- UNORDERED_LINK
-COMPLEMENT_LINK <- UNORDERED_LINK
+UNION_LINK <- COLLECTION_LINK
+INTERSECTION_LINK <- COLLECTION_LINK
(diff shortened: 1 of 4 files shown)


ngeiswei:
`CollectionLink` is nice, but do you mean as in https://en.wikipedia.org/wiki/Collection_(abstract_data_type) which typically includes list as well?  Or as in https://en.wikipedia.org/wiki/Set_(mathematics)?

Additionally I'm concerned that `CollectionLink` inheriting from `UnorderedLink` would not permit to introduce `SetDifferenceLink` which should be ordered even though it does represent an unordered collection.  Thus I think `CollectionLink` should rather inherits from `Link`, and have `SetLink`, `UnionLink`, etc, doubly inherit from `UnorderedLink` and `CollectionLink`.  I don't think such double inheritance would be problematic w.r.t. to factories since `CollectionLink` have no factory.   This also opens the possibility to have `ListLink` inheriting from `CollectionLink`, if we want to (I'm neutral about it, but I see it as feasible because a list can be seen as a set of indexed elements, and thus union, etc, between lists and sets is actually definable).

linas:
> do you mean as in https://en.wikipedia.org/wiki/Collection_(abstract_data_type)  

Not what I was originally thinking of, no. It was just an English word I could think of that was neutral. Could also try `GroupingLink` but that feels awkward in English.  I can't think of any other names right now.

>  doubly inherit from `UnorderedLink` and `CollectionLink`

Good idea. I'll make that change right now.

> SetDifferenceLink

What the heck. May as well add that now. clear the decks.

> opens the possibility to have `ListLink` inheriting from `CollectionLink`

Interesting idea. That means that all data-like things inherit from `Collection` .. interesting idea. I'll make that change too, and see if it harms anything. We can revert later if it looks bad.

linas:
Currently `SubsetLink` inherits from `InheritanceLink` which might make sense for PLN, but sounds weird if we are sticking to plain-old venn-diagram set-theory concepts.  I'd like to change it to
```
SUBSET_LINK <- ORDERED_LINK
```
and as a side effect, you might want to add back in
```
EXTENSIONAL_INHERITANCE_LINK. <- INHERITANCE_LINK
```
This is not a burning issue, but seems like "the right thing to do" from a set-theory-notation point of view.  Let me know what you think @ngeiswei

linas:
@ngeiswei any further comments, or should I merge this?

ngeiswei (approved):
It all looks good, @linas.

Regarding whether `Subset` should or not inherit from `Inheritance`, I'm somewhat confused by it at the moment, I would say probably not (in PLN `Inheritance` is the disjunction of extensional a.k.a.`Subset` and intensional inheritance, but constructing that disjunction supposes proper casting from these inheritance links to high order predicates, and I'm not sure that entirely well defined yet).  Anyway, it's not a pressing issue as you say.

PR #2849, opened 2021-09-14
Use platform.h api in cogutil for setting thread names

### LINKED ISSUE #2842: Setting thread names using prctl syscall
Setting thread names using [prctl](https://man7.org/linux/man-pages/man2/prctl.2.html) syscall results in the atomspace only building on Linux as the syscall is specific to Linux. This was added in #2832.

Habush:
Fixes #2842 and uses the changes in opencog/cogutil#257


linas:
Thank you!

PR #2851, opened 2021-09-22
Move assorted atom types to PLN

linas:
A variety of Atom Types are defined only for the benefit of PLN, and are not otherwise generally useful. These are moved out of the atomspace, and into PLN with a matching pull req. opencog/pln#53


ngeiswei:
I have no problems with these changes @linas. I worry a little that it might break someone else code, we should probably notify the list or such,  @vsbogd, what do you think?

PR #2856, opened 2021-09-24
Add spacetime and pln testing in circleci

ngeiswei:


ngeiswei:
SpaceTime (required by PLN) fails on circleci but passes locally, no idea why.

linas:
That's odd. I used to test every now and then, and It used to pass. I upgraded the OS recently, and now it fails for me locally.  Possible that the newer versions of octree have this assert in it. I have version 1.9.5 and circleci has version 1.9.3 -- what version do you have ? `apt search liboctomap-dev`

linas:
I just pushed a fix to spacetime. Now the unit tests pass, for me. 

Why does PLN depend on spacetime?

ngeiswei:
PLN now includes temporal reasoning which needs TimeNode and AtTimeLink that are defined in spacetime.

linas:
It might be a good idea to define distinct kinds of time nodes ... one kind that has the correct format for what the spacetime server wants/needs and a different one for PLN. There are technical problems for representing time --- seconds since 1970? nanoseconds? Arbitrary HH:MM:SS DDMMYYYY strings? Are they infinitely sharp or do they have an implicit fuzziness?   Time, relative to what?

Physics trivia factoid: time runs slower on the surface of the Earth, than in space, thanks to general relativity. The clocks in the GPS satellites are slowed down on purpose, to account for this; its actually in the GPS spec, page 72 or something like that. (I actually read the GPS spec, back in the day.) Now, the adjustment is tiny: parts per billion, but it adds up -- a few meters here or there of accuracy, and the military-grade GPS is accurate to a centimeter. To talk about time, you have to indicate where you are talking about time, and how fast you are moving, and the path through the gravitational field (twin paradox).  This is kind-of irrelevant for PLN, but the upshot is that PLN should not depend on the whims and vagaries of the spacetime server.

BTW, the spacetime server is kind-of mis-conceptualized. It uses an octree rep, which is great if you are out in the open, in unobstructed space. If you are indoors, in a maze of rooms and corridors, it's the wrong representation.  There's no way to say "oh hey, there is a wall in between here and there, and you can't go through the wall."  or even "you can't see that thing from where you are standing" or even "that doorway is too small for you to fit through". It's not useful for robotics, and its not useful for spatial reasoning. What is it useful for?

For this reason, I'd like to let the spacetime server die a slow death of obselescence. It would be better if PLN did not depend on it.

linas:
> PLN now includes temporal reasoning 

As I thought about this, it occurred to me that you want PLN to be able to do reasoning on any total order (any linear sequence of events, of which "common-sense" time is just one example.)  This effectively does not change any logic at all in PLN, just chenges the name of `TimeNode` and `AtTimeLink` to something like  `TotalOrderNode` and `AtOrderLink`.

If PLN can do partial orders, then there would be
```
PARTIAL_ORDER_NODE <- NODE
TOTAL_ORDER_NODE <- PARTIAL_ORDER_NODE 
```
and then if you really really want it:
```
TIME_NODE <- TOTAL_ORDER_NODE
```
FWIW, Einstein's famous results from special relativity is that time is not a total order everywhere: its only a total order in the past and future light-cones, and unordered anywhere else. For this reason, of no other reason, supporting partial orders would be cool.

linas:
And one last comment: once one starts thinking about partial orders, one of course can fall into a black hole. So, for example, sets and set-membership form a partial order.  So do sets and subsets -- same idea, but the actual partial order is different.

More generally, a partial order is anything where you can draw an arrow from one dot to another, and it's acyclic (its a DAG). We already have a `GreaterThanLink` which returns true/false, but it only works on `NumberNode`, it does not work on sets or subsets or set membership.

Note that both the outgoing set and the incoming set are both DAG's and thus partial orders. Thus, we have both `MeetLink` and `JoinLink`, but these work ONLY on incoming/outgoing sets.

A generic partial order system would need a `GreaterThanLink` that worked on any kind of ordered thing, and also a `MeetLink` and a `JoinLink` that would work on any kind of ordered thing.  I think it would be pretty cool of PLN could do reasoning on partial orders. Of course, its a bit of an intellectual challenge just to create the right kind of atom types to describe partial orders.  I'll try to think a a bit more about this over the coming few years, because its interesting and could be worthwhile.

PR #2868, opened 2021-10-23
Handle ChoiceLinks nested under IdenticalLinks

linas:
This continues the work of the previous pull req, correctly handling multiple choices that might appear under IndenticalLinks.


linas:
p.s. I think I'm done for a while.

PR #2878, opened 2021-11-01
Fix crash in recursive put

linas:
Sigh. Code hit a naive infinite loop. Stop that.


linas:
pln is failing unit tests due to opencog/pln#60 .. so I'm merging this.

PR #2880, opened 2021-11-03
Add debug printing

linas:
This modifies exceptions to change the error message to be more informative.

This requires the latest version of `cogutil`, as otherwise, it won't compile. Seems that `cogutil` did not anticipate this use.


linas:
pln TemporalReasoningUTest still timing out, per issue opencog/pln#60  merging this pull req.

PR #2883, opened 2021-11-04
Support implicit present terms

linas:
If part of a query uses a term, that term is implicitly present in the AtomSpace.  The user should not have to explicitly specify it. This was noticed during work on issue #2644


linas:
Merge despite failure of `TemporalUTest` see opencog/pln#60

PR #2884, opened 2021-11-05
Provide better fix for determining when dummies are needed

linas:
All variables must appear somewhere, in order for the pattern engine to find them. This provides a better check for that.


linas:
`TemporalReasoningUTest` failure per opencog/pln#60

PR #2889, opened 2021-11-05
Do not allow quotation level to go negative.

linas:


linas:
Closing; this is a fix, but not the best fix.

opencog/atoms/core/Quotation.cc
```diff
@@ -70,10 +70,11 @@ void Quotation::update(Type t)
 	_local_quote = is_unquoted and LOCAL_QUOTE_LINK == t;
 
 	// Increment or decrement quotation level if locally unquoted
-	if (is_locally_unquoted) {
+	if (is_locally_unquoted)
+	{
 		if (QUOTE_LINK == t) _quotation_level++;
-	    else if (UNQUOTE_LINK == t) _quotation_level--;
-    }
+		else if (UNQUOTE_LINK == t and 0 < _quotation_level) _quotation_level--;
```
ngeiswei, line 76: I'm a little worried that this is gonna silently bite us in the post some day.  I wonder if it wouldn't be better to have the user of that class know how to deal with negative quotation level (but I say that while having an incomplete understanding of what you've done, at this point).

linas: I did not merge this change.

Issue #2886, opened 2021-11-05
Non monotonic behavior of the pattern matcher with present clauses


ngeiswei:
# Problem description

I have a query, containing only present clauses, that correctly succeeds.  However the presence of an another atom in the atomspace makes it fails.

# How to reproduce

Run the following scheme code:

```scheme
(use-modules (opencog))
(use-modules (opencog exec))

;; Query
(define query
  (let* (
     ;; Constants
     (R (Predicate "R"))
     (A (Execution (Schema "A")))
     ;; Variables
     (P (Variable "$P"))
     (Q (Variable "$Q"))
     ;; Clauses
     (P↝Q (Quote
            (Implication
              (Unquote P)
              (Unquote Q))))
     (Q∧A (And Q A))
     (Q∧A↝R (Implication
              Q∧A
              R)))
    ;; Query
    (Get
      (VariableSet P Q)
      (Present P↝Q Q∧A↝R))))

;; KB

;; Culprit
(Implication
  (And
    (Predicate "Qbis") ; Culprit
    (Execution (Schema "A"))
  )
  (Predicate "R"))

;; Premises
(Implication
  (Predicate "P")
  (Predicate "Q"))
(Implication
  (And
    (Predicate "Q")
    (Execution (Schema "A"))
  )
  (Predicate "R"))

;; PM
(define results (cog-execute! query))
```

It, incorrectly, returns nothing.

Remove the culprit atom and then it works.

ngeiswei:
@linas, I'm working on it.  However, if you have some idea at the top of your head as to why it fails, that's welcome.

ngeiswei:
When I uncomment the following lines

https://github.com/opencog/atomspace/blob/7385233d703cca06f6b9c83cd580d810368ce47c/opencog/query/PatternMatchEngine.cc#L2608-L2611

it no longer fails.

ngeiswei:
@linas, since I have a workaround (disabling that negative cache check) and more urgent work to do I'm gonna leave that bug for now and will come back to it after a few days or weeks, unless you fix it by then.

linas:
Caches (on any kind of system) are notoriously hard to make work right. These caches made a huge difference on the bio-agi search patterns (something like 2x or 3x speedup on multi-minute searches).

I'll look. I've got something else I should be doing, but this kind of bug is very irritating.

linas:
BTW, Some off-topic comments but I know you will read them if I post them here.

[...]

PR #2890, opened 2021-11-05
Balanced quotes during pattern compilation

linas:
Make sure that quoted terms are treated properly during pattern compilation. This fixes issue #2886


linas:
Merging. There is a failure but its the opencog/pln#60 failure

PR #2893, opened 2021-11-08
Track down and fix more unbalanced QuoteLinks.

linas:
This fixes all of them, except for on that is intentionally unbalanced in a unit test.

@ngeiswei There are still many unbalanced QuoteLinks in the URE code base. These can be found by stopping the debugger at `STRICT_QUOTATION_LEVEL` in `Quotation.cc` I don't know that it matters very much. The unbalanced quotes might be hiding more serious bugs .. ore maybe not.  The unit tests do pass...


linas:
p.s. this is continuing fallout from #2866 .. which was fixed yesterday. This pull req just tracks down and fixes more cases that don't seem to actually affect anything.

PR #2920, opened 2022-03-07
Stop using weak pointers

linas:
They're overkill, and are not needed, and are an overall drag on performance.

The AtomSpace guarantees that the incoming set will be valid.


linas:
Reopen, merge the new code, but leave it disabled.

PR #2926, opened 2022-04-07
Support deletion in deeply-stacked atomspaces

linas:
Part two to pull req #2925: When there is a stack of atomspaces, and the atom is deleted at the top of the stack, it is marked as being invisible, so that it will remain present deeper in the stack.  That is, AtomSpaces with "deleted" atoms behave as if that atom really is gone, while still being present and accessible if shifting the viewpoint to other atomspaces.

FYI: This means that AtomSpaces now behave like "Context TV" and "Hypothetical TV" from the old, old, atomspace design from a decade ago. Different AtomSpace expose different TV's and atoms: they provide the "context" for some particular truth/world-view.

### DIFF
--- a/opencog/atoms/base/Atom.cc
+++ b/opencog/atoms/base/Atom.cc
@@ -237,6 +237,27 @@ bool Atom::setUnchecked(void)
     return _checked.exchange(false);
 }
 
+bool Atom::isAbsent() const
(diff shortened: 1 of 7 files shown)


ngeiswei:
That really nice, @linas

Issue #2945, opened 2022-06-02
Compile *.scm files and install the *.go files


linas:
Currently, the first time you start the atomspace, several minutes will be spend compiling the `*.scm` files for the first time. This makes the first start slow. It would be better if the compile happened during the CMake stage, and the resulting `*.go` files were installed.

linas:
The work needed for this has been mostly completed in pull reqs #2949 #2950 #2951 However, one issue remains:

If a module uses the `MODULE_DESTINATION` keyword, then the compiled module will be installed to the incorrect location. Some new keyword is needed, or else some extra-clever guessing.

linas:
Closing; all remaining work completed in #2951

PR #2946, opened 2022-06-02
Fix screwball guile module name processing.

linas:
Or at least, attempt to do this. The code is spaghetti code.


linas:
Failed attempt to lay groundwork for #2945

PR #2980, opened 2022-07-28
Boolean operators for working with Boolean vectors

linas:
These are intended for use in AS-MOSES scoring functions.

The idea is that, by actually using executable Atomsese in scoring functions, 
ASMOSES will be able to score any kinds of nodes (including audio, video)
and not just the handful of types it is aware of.


linas:
Merge despite MinerUTest fail -- the reason for the fail is clear, and now fixed.

Issue #2502, opened 2020-02-13
Glob interval empty set is wrong.


linas:
See discussion in #2498 -- the current glob-matching/set-intersection code is using `{0,0}` to represent "the empty set", but this is incorrect.  The actual fix is maybe a few lines of code in one or two places, and an appropriate unit test.

sikefield3:
hi,   
 i'd like to do this, if it's ok...
i'm able to build and run it in docker.

linas:
Hi,

I just spotted this ... sure, please! I guess you'd like some pointers for where to start? But if you think you can do it, then just create a pull request -- that's it...

linas:
A place to start would be to read the problem description in #2498 and then look at the patch in #2498 and look at GlobInterval -- that is the thing that holds the lower and upper bounds  for matching. Hunt around for places where GlobInterval is used, make sure the bounds checks are reasonable.

linas:
I mean - this is not glorious or exciting work, it just gets you familiar with the code base.

sikefield3:
Cool!   
I'm going to have a look at the code in the next few days and then come back with my suggestion/further questions.

sikefield3:
So, I did  a search for GlobInterval and now have some idea what to do...   
Imho, a function like `bool IsEmpty(GlobInterval)` would be handy, where the check according to 
[your suggestion](https://github.com/opencog/atomspace/pull/2498#issuecomment-585909835) is done.   

Anyway, I'm going to create my pull request soon...

linas:
Yes, something like `IsEmpty()` sounds reasonable!

> I'm going to create my pull request soon...

Cool. FWIW, I sometimes don't check email for days in a row, don't panic if there's no immediate reply.

sikefield3:
I'm getting an unresolved ref. for my function, so maybe I could make a WIP-pull-request and you could have a look, if I should change the function signature or location ?   
Apart from that the changes I wanted to make are in place.   
Unit tests can take while, since I'm not yet familiar with CxxTest.

sikefield3:
[fix](https://github.com/opencog/atomspace/compare/master...sikefield3:atomspace:wip_is2502_new_empty_interval) is now done.   
unit tests still missing.

sikefield3:
I was just wondering about the [extend](https://github.com/opencog/atomspace/blob/ef3c0a35b1814e20678d9f957d14d8b10669854d/opencog/atoms/core/TypeChoice.cc#L139-L146) -function:   
[...]

PR #3005, opened 2022-11-15
Issue #2502 New empty interval

linas:
Fix for issue #2502

### DIFF
--- a/opencog/atoms/core/TypeChoice.cc
+++ b/opencog/atoms/core/TypeChoice.cc
@@ -108,7 +108,7 @@ void TypeChoice::init(bool glob)
 	for (const Handle& h : _outgoing)
 		analyze(h);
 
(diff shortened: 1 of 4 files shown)


linas:
Code looks good to me; I'm ready to merge this; any reason I shouldn't?

linas:
Oh, except that there may be a bug here. According the wiki page https://wiki.opencog.org/w/GlobNode, an interval [N, -1] means match at least N times, up to infinity.  i.e. -1 was used to denote "no upper bound". This seems cleaner than asking users to write `[N, 4810683]` when they want unbounded-above.

sikefield3:
But since `GlobInterval` is using unsigned ints, I think the user can assign the upper bound to -1 here and  get an interval `[N, MAX_INT]`
`typedef std::pair<size_t, size_t> GlobInterval;`

PR #3017, opened 2022-12-08
Is2502 test empty interval

sikefield3:
Unit tests for #2502: intersection and union of interval with empty interval


linas:
Bingo! Thanks!

PR #3027, opened 2023-01-31
Made atomspace/examples/README.md use links for files referred to

rinslow:
Made `atomspace/examples/README.md` referred files be links, so it will be easier to study and navigate the codebase.


linas:
Thanks. Although now the plain-text readme lines don't line up nicely. Not eveyone looks at README's with a web browser, some people use a plain text editor.

Weird that the unit tests failed. I'm investigating that now.

PR #3029, opened 2023-01-31
Atom hashes independent of type hierarchy details.

linas:
Prior to this, the addition or removal of atoms to the type hierarchy alters the hashes, which makes debugging hard than it otherwise needs to be.


linas:
FYI this pull req exposed bugs in the as-moses unit tests, which erw fixed. It also exposed bugs in PLN, which were not fixed.

PR #3031, opened 2023-02-03
Inserting FloatValue, resolves #3026

### LINKED ISSUE #3026: Another FloatValue example
While going through the examples, I found another exception similar to #3022 in
https://github.com/opencog/atomspace/blob/b2603a5a17e9c737f9ac6788c0aec791ae427b68/examples/pattern-matcher/value-of.scm#L45-L57
which gives me:
```
ice-9/boot-9.scm:1685:16: In procedure raise-exception:
Atomspace C++ exception:
(cog-new-link Types fail to agree, invalid Atom syntax: (DivideLink
  (ValueOfLink
    (Variable "$X") ; [5e8448863b0b6c19][1]
    (Predicate "some key") ; [2dafe92e20a67a8f][1]
  ) ; [80bf787f0b876249][1]
  (Number "12") ; [6c6aa63c55cec01][1]
) ; [d1742eb156926b78][-1] (/home/opcwdir/repodock/atomspace/opencog/atoms/   
...
```
It worked, when I put FloatValueOf in there:
https://github.com/opencog/atomspace/blob/b2603a5a17e9c737f9ac6788c0aec791ae427b68/examples/pattern-matcher/value-of.scm#L54

sikefield3:
This PR relates to #3026.   
In order to avoid an exception, Value has to be replaced with FloatValue in two places.   
Apart from that, missing modules were included in two places.

### DIFF
--- a/examples/atomspace/recursive-loop.scm
+++ b/examples/atomspace/recursive-loop.scm
@@ -23,7 +23,7 @@
 
 ; Define a predicate that tests to see if the sum of three random
 ; values is less than 2.5 or not.
-(define uniform (ValueOfLink someplace key))
+(define uniform (FloatValueOfLink someplace key))
 (Define (DefinedPredicate "keep going?")
    (GreaterThan (Number 2.5) (Plus uniform uniform uniform)))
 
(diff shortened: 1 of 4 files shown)


linas:
Thanks!

linas:
The `(use-modules (opencog persist))` is actually wrong; in that case the `(cog-delete!)` at the end needed to be changed to `(cog-extract!)` ... I'm fixing that now.

PR #3033, opened 2023-02-18
Atomic grant of a link between two Atoms

linas:
Introduce a new GrantLink between source and target Atoms. This link can be created once and only once for a given source. Thread-safe, atomic, first-person wins. All subsequent attempts to create a GrantLink with the same source Atom will return the original target.

This is unlike UniqueLink, with throws an error, and unlike DefineLink, which can only be used with DefinedPredicate, DefinedSchema. It is unlike StateLink, which is similar in that only one exists for that source Atom, but StateLinks allow the target to be changed at any time. GrantLinks freeze the target to it's initial setting.

GrantLinks are needed for thread-race safe lookup tables between different atoms.


linas:
Merging despite unit test failures; the URE unit tests are timing out. This pull req has no impact on URE whatsoever, it only introduces a new Atom type, and that's all.

linas:
See https://wiki.opencog.org/w/GrantLink for documentation

PR #3034, opened 2023-02-18
The DeleteLink code seems bit-rotted. Fix it up.

linas:
Some old redesigns appear to have left it with a bug and with some unfinished work. Get it working again, as designed. Also clean up documentation.


linas:
Merging despite URE unit test timeout. URE does not use DeleteLinks, the test timeout is unrelated to this change. I'm guessing cython is bit-rotted.

PR #3036, opened 2023-04-16
Fix typo

benjamin-asdf:


linas:
Thank you!

benjamin-asdf:
Hey @linas 
`examples/atomspace/basic.scm`

```
; $ guile -L opencog/scm -L build
;
; where "build" is where-ever you built opencog.

; Another possibility: add stuff to your ~/.guile file, for example:
; (add-to-load-path "/home/yourname/opencog/build")
; (add-to-load-path "/home/yourname/opencog/opencog/scm")
;
```

what is is opencog? I guess this was split from the opencog repo and
now I don't get what it is refering

linas:
> what is opencog?

Way back in the day, the `opencog` git repo contained everything and the kitchen sink. It became unmanageable as-is, and got split into a several dozen directories. The main guile wrapper module for the AtomSpace was called `opencog`, and the name was not changed, to maintain backwards-compat. These days, it just loads the AtomSpace module, and nothing else; there are dozens of additional scheme modules for other things. (There are also python modules, but those are slowly bit-rotting.)

The `c++-guile` example had bit-rotted; I just now reviewed and overhauled it. There's an extended README, and a how-to for using the module. Do a `git pull` to get the latest.

I'll look at `basic.scm` next, it might be bit-rotted, too.

linas:
BTW, the "main" opencog git repo now contains only the abandoned code. All the good stuff has been split out.

linas:
I re-wrote and extended the `basic.scm` demo. It explains more, and doesn't leave you hanging at the end of it.

benjamin-asdf:
Do I not need the file `opencog/as-config.scm` ? I am trying to load `opencog.scm` right now and 
Unable to find file "opencog/as-config.scm" in load path.

linas:
> Unable to find file "opencog/as-config.scm" in load path.

You should not load `opencog.scm` by hand; instead, you should  `make install`  and then `(use-modules (opencog))`  This will set up all the search paths in the right way.

The `as-config.scm` file is a bit of auto-generated boilerplate that allows unit tests to run, without installing the system. Frankly, I hate it; it just adds pointless complexity. But whatever, I lost that political battle eons ago.

There's another reason to not load `opencog.scm` by hand: it  specifies another dozen additional files to load, and those also won't be found if the paths aren't just-so.  Guile has a module system, just use it and it will do "the right thing".

linas:
Oh, and if you really really don't want to install, but still want to run the demos, then you have two options. One is to run an lxc or a docker container. There are some docker containers in the `opencog/docker` repo.  But LXC is nice, too.

The other option is to try to run the examples from the build directory, i.e. in the same way that the unit tests are run. This might work, maybe, if you're careful. But this risks carpet burns. Not recommended at all.  If you're smart, you won't do that.

PR #3161, opened 2025-10-11
Convert Evaluator to perform crisp evaluation only

linas:
This is a large batch of changes that remove all non-crisp evaluation from the evaluator. This is a major step in the program to remove all TruthValues from the AtomSpace.

### DIFF
--- a/tests/scm/SCMExecutionOutputUTest.cxxtest
+++ b/tests/scm/SCMExecutionOutputUTest.cxxtest
@@ -456,8 +456,8 @@ void SCMExecutionOutputUTest::test_boolean(void)
 	CHKEV(eval);
 
 	// When evaluated, this one should be false.
-	tv = EvaluationLink::do_evaluate(as.get(), boolex);
-	TS_ASSERT_LESS_THAN_EQUALS(tv->get_mean(), 0.001);
+	bool result = EvaluationLink::crisp_eval_scratch(as.get(), boolex, as.get());
+	TS_ASSERT(not result);
 
 	logger().debug("END TEST: %s", __FUNCTION__);
 }


linas:
Jeez @Adam-Vandervorst -- camping on my high activity but unimportant commits? I have unfinished important conversations with you, which I will get back to "real soon now". As to this:

-- It should have been done 6-7 years ago.
-- it wasn't because it's boring and tedious refactoring
-- Claude Sonnet 4.5 allowed me to do the refactoring in a few days (this and like 6-7 other pull reqs) instead of a few weeks or month.
-- Nothing is actually lost; the FloatValues were added 7-8 years ago, They are vectors of floats and are a generalized replacement for the old "truth values". There's an entire generic, generalized infrastructure for working with these FloatValues; one of the on-again, off-again projects is to get these to migrate to GPU's.

This last bit: GPU-ifying the Atomese abstract syntax trees (ASTs), is one of the more interesting projects. The idea is that an AST is just an abstract description of an algorithm: more concrete than pseudo-code, less concrete than python, scala, c++.  Given an AST, an "agorithm", I should be able to say "hey go run yourself over there", and the GPU target seems like a reasonable target.

This would allow the "stupid computer trick" of, for example, writing a transformer in Atomese.

I call it "stupid" because kind of no one is actually interested in such a stunt. I mean, why bother? 99.99% of all programmers in the industry are going to ask Claude or ChatGPT write a blob of python code to do whatever, and if they need changes or add features, they just ask the LLM to do that, until that blob becomes an unmanagable mess. After six months of being stuck in molasses, some fraction of these coders will understand that "good architecture" is a "good idea", and maybe (??) by then, the LLM's will advance far enough to be able to actually architect coherent interfaces. Claude can almost do this now, on a good day. So who needs this stuff, when you can just ask an LLM to write code for you? ... and that is why this is a "stupid idea".

But I dunno, some days it feels like a worthwhile idea.

Adam-Vandervorst:
> camping on my high activity but unimportant commits

It felt anomalous -- but now I got it explained with "Claude Sonnet 4.5" (not the productivity, which I'm very used to coming from you, but the amount of yak shaving per experiment)

> Nothing is actually lost

Right, decreasing the redundancy with the more general mechanism... if it were a commitment to a crisp system, I was going to try and halt you in your tracks by saying I have a good "discrete only" engine.


See you real soon!

linas:
Oh. hah. "Real soon now" is a Texas expression meaning "it might be a long long time". And "any day now" means "it might never happen". Another one I like is, when asking for driving directions, "ain't no way to get there from here". Little bits of Texas culture that have been mostly erased.

PR #3174, opened 2026-09-22
Add free hypermultiset enumeration and a Plan 9 namespace variant of AtomSpace

danregima:
This PR now bundles two independent additions built in the same session/branch:

## 1. Free hypermultiset enumeration

C++ implementation of the recurrent enumeration formula from "A Recurrent Enumeration of Free Hypermultisets" (`opencog/atoms/base/FreeHypermultiset.h/.cc`), matching the published `A000081`-derived sequence, with unit tests, corrected reference MATLAB/markdown docs (`doc/free-hypermultisets/`), a PowerShell emulator (`scripts/free_hypermultiset.ps1`), and a Docker-based Linux dev environment (`docker/atomspace/`, `scripts/atomspace-docker.ps1`) since this workspace has no C++ toolchain, CMake, or Docker/WSL access.

## 2. Plan 9 namespace variant of AtomSpace (`opencog/atoms/plan9/`)

Exposes an AtomSpace as a synthetic filesystem served over the real **9P2000** wire protocol (as used by Plan 9 from Bell Labs and Linux's `v9fs`/`9pfuse`), with per-connection ("per-process") namespaces built from Plan 9's `bind`/`mount`/union-directory semantics, plus a pluggable **sheaf** extension point.

- `Nine9P.h`/`.cc`: from-scratch 9P2000 message codec (every message type `Tversion`..`Rwstat`), `Qid`/`Stat` encode/decode, frame-scoped decoding.
- `NineNode.h`/`.cc`: abstract synthetic-filesystem node interface.
- `NineNamespace.h`/`.cc`: per-process bind/mount/union-directory table implementing real Plan 9 namespace semantics (`MREPL`/`MBEFORE`/`MAFTER`), independent of AtomSpace's own frame/COW mechanism so the Plan 9 rules stay exact.
- `AtomFs.h`/`.cc`: maps every atom onto `/by-type/<Type>/<id>/` and `/by-handle/<id>/` directories (`name`, `type`, `to_string`, `tv`, `outgoing/`, `incoming/`, `values/`), plus a `/new` control file to create atoms by writing Atomese/Scheme. `outgoing/`/`incoming/` are true hard links (Qid-equal aliases). Since this AtomSpace version has no separate `TruthValue` class, `tv` reads/writes the atom's `truth_key()` `FloatValue` directly.
- `SheafHook.h`/`.cc`: the "sheaves" extension point. There's no maintained in-repo C++ sheaf implementation (it lives in the external `opencog/learn` project as Scheme), so this is a decoupled interface: `StructuralSheafHook` is a dependency-free fallback (a Link's outgoing set as its connectors), and `SchemeSheafHook` (Guile builds only) bridges to a user-named Scheme procedure so real sheaf/connector logic can be plugged in without a build-time dependency.
- `NineServer.h`/`.cc`: per-connection 9P dispatcher (fid table, `attach`/`walk`/`open`/`read`/`write`/`clunk`/`stat`/etc).
- `NineTcpServer.h`/`.cc` (POSIX only): minimal TCP listener so the tree can actually be mounted via `v9fs`/`9pfuse`.
- New CxxTests cover the wire codec, namespace bind/mount/union ordering and independence-after-fork, and the AtomFs mapping.

See `opencog/atoms/plan9/README.md` for the full filesystem layout, supported 9P2000 subset, and mount instructions.

## Note for reviewers

This development environment has no C++ compiler, CMake, or Docker/WSL access, so none of this code could be built or run here. Everything was hand-reviewed against the real header signatures of `AtomSpace`, `Atom`, `Handle`, `FloatValue`, `SchemeEval`, etc. Please build and run the new unit tests (`FreeHypermultisetUTest`, `NineProtocolUTest`, `NineNamespaceUTest`, `AtomFsUTest`) before merging.


linas:
Hi,
Instead of combining additinos, it would be better to split them up, so each could be reviewed and handled on its own.

For example, the opencog/docker git repo already provides a lot of different docker containers, I  don't want one in the core atomspace.

For the Plan9 stuff --  Several remarks on that:
* The opencog/cogserver already provides a generic tcp/ip server, ind it should be "easy" to add plan9 network support to it. I think. It would be a natural place for that.
* The cogserver also has codecs for several different protocols, so it would be a good place to do the plan 9 codec as well.

If you care using Claude code, point it at the cogserver and it will figure it out.  If you are not using Claude code, you should be.

Multiple small pull requests are strongly prefered to one giant one. I really really want to see clean code done the right way for the right reasons, and small pull reqs help make that happen.

